PR #1, opened 2020-05-26
Configure build & deploy using GitHub Actions

foolip:


foolip:
https://github.com/foolip/webdriver-bidi/runs/708969386?check_suite_focus=true#step:8:55 shows this working.

foolip:
https://github.com/jgraham/webdriver-bidi/runs/708976785?check_suite_focus=true also ran successfully and pushed to https://github.com/jgraham/webdriver-bidi/tree/gh-pages. https://jgraham.github.io/webdriver-bidi/ will take a while to appear, I guess, it usually does.

PR #7, opened 2020-05-28
Change status to Living Document

foolip:
A less blessed status, and matches https://w3c.github.io/webdriver/.


foolip:
This actually broke the build. I'll set up PR preview so that we notice if Bikeshed fails.

foolip:
I've asked in https://github.com/tabatkins/bikeshed/pull/1682 what the right thing would be.

PR #24, opened 2020-06-03
Start to specify the transport mechanism

jgraham:
This adds the initial steps to handle a WebDriver new session with the
capability bidi set, validating that the parameter is a boolean and
returning the URL to a websocket connection. It also adds some of the
scaffolding for interacting with that connection.


index.bs
```diff
@@ -6,16 +6,111 @@ Status: w3c/UD
...
+<dfn>websocket url</dfn>, which is a string. These are both initially
+<code>null</code>.
+
+When data is recieved on [=connection=] it must be processed according
+to the requirements of the [[!RFC6455|WebSockets]]
+specification. When a [[RFC6455#section-6.2|WebSocket Message has
+been Received]], an implementation must [=handle an incoming
+message=]. When [[RFC6455#section-7.1.3|The WebSocket Closing
+Handshake is Started]] an implementation must [=end the
+session=].
```
AutomatedTester, line 46: What happens if a socket prematurely dies? Do we need to care?

index.bs
```diff
@@ -6,16 +6,111 @@ Status: w3c/UD
...
+When required to <dfn>end the session</dfn>, run the following
+steps:
+
+1. Close the underlying network connection [=connection=].
+
+2. Set [=connection=] and [=WebSocket URL=] to <code>null</code>.
+
+Issue: This should also reset any internal state
+
+Note: This does not end any [[WEBDRIVER2#sessions|WebDriver
```
AutomatedTester, line 84: Is this a different `session` mentioned above?

index.bs
jgraham: I'm not sure we need to care; WebDriver 1 doesn't specify what happens if the server connection drops.

index.bs
jgraham: Yeah, maybe this is confusing. The point is that the bidi connection has a different lifetime compared to the WebDriver session. This is necessarily true, because the client can send a close frame which will cause the connection to close, without affecting the HTTP connection. And we could force the overall session to close, but that adds coupling between the specs in a way that's not really necessary.

index.bs
AutomatedTester: I don't think we needed to care due to the command response nature and we didn't for `keep-alive`. I guess just wondering if it drops over a network(Think BrowserStack) do intermerdiary nodes need to restart and assume we're all good or  ...?

I don't think this needs to block right now but at least we need to be aware.

index.bs
```diff
@@ -7,14 +7,110 @@ Group: browser-testing-tools
...
+
+2.  Set up a network connection that listens on an implementation-defined
+    hostname <var>host</var> and port <var>port</var>. Set the [=connection=] to
+    this network connection.
+
+3.  Let <var>WebSocket URL</var> be the result of constructing a [=WebSocket
+    ws-URI=] with host <var>host</var>, port <var>port</var>, and empty path an
+    query.
+
+4.  Set the [=websocket url=] to <var>WebSocket URL</var>.
```
foolip, line 72: I'm not sure I can predict how the "ensure" aspect of this algorithm will turn out to be useful. If it's not so clear, maybe we can drop it until later when it's needed.

index.bs
```diff
@@ -7,14 +7,110 @@ Group: browser-testing-tools
...
+
+WebDriver clients opt in to a bidirectional connection by requesting a
+capability with the name "<code>bidi</code>" and value
+<code>true</code>.
+
+This specification defines an
+[=additional webdriver capability=] with the [=capability name=] "<code>bidi</code>".
+
+The [=additional capability deserialization algorithm=] for the
+"<code>bidi</code> capability, with argument <var>value</var> is:
```
foolip, line 105:
```suggestion
"<code>bidi</code>" capability, with argument <var>value</var> is:
```

index.bs
```diff
@@ -7,14 +7,110 @@ Group: browser-testing-tools
...
+ 1. If <var>value</var> is not a boolean, return [=error=] with [=error
+    code|code=] [=invalid argument=].
+ 2. Return [=success=] with data <var>value</var>.
+
+The [=matched capability serialization algorithm=] for the `bidi` capability,
```
foolip, line 110:
```suggestion
The [=matched capability serialization algorithm=] for the "<code>bidi</code>" capability,
```

index.bs
```diff
@@ -7,14 +7,110 @@ Group: browser-testing-tools
...
+with argument <var>value</var> is:
+ 1. If value is <code>false</code>, return [=success=] with data
+    <code>null</code>.
+ 2. Let <var>websocket url</var> be the result of running the steps to
```
foolip, line 114: Seeing how this is called from https://w3c.github.io/webdriver/#dfn-matching-capabilities, the naming of the this concept, matched capability serialization algorithm, seems pretty odd. Would "additional capability matching algorithm" be an improvement?

index.bs
```diff
@@ -7,14 +7,110 @@ Group: browser-testing-tools
...
+from the user agent to the controlling software, better matching the
+evented nature of the browser DOM.
+
+Transport {#transport}
+======================
+
+Message transport is provided using a [[RFC6455|WebSocket]]
+connection.
+
+A [=remote end=] has a <dfn>connection</dfn>, which is a network connection, and
```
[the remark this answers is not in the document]
jgraham: Yeah, I guess that makes sense. I'm trying to avoid having too many explicit dependencies on WebDriver, but I suppose if you have multiple sessions in a single instance you'd also have multiple websockets. Although in practice the whole multiple-sessions thing is really about intermediary nodes rather than actual browsers. I don't really know what our feeling about intermediary nodes is in this case.

index.bs
jgraham: Unfortunately it's already needed in a theoretical way. The way that session matching works in WebDriver you can have something like

```
capabilities: {firstMatch:[{bidi: true, browserName: "Some other browser"}, {bidi: true}])
```

In that case you'd go through the capabilities matching steps twice, once for the first entry, which would then fail, and once for the second entry. So because we establish the connection during capabilities matching, there's a spec-theoretic possibility that we previously established a connection for some prior set of capabilities and want to return that. Now, in practice, I don't expect an implementation to actually do that. But it's easier to write the small fiction here than it is to add another hook to allow the capabilities set to be post-processed after selection.

index.bs
jgraham: I don't think I follow which thing you want to be renamed.

index.bs
jgraham: OK, so I realised that this doesn't work anyway if we want to tie the URL to a session. So I guess I have to fix it properly.

index.bs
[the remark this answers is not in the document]
foolip: A bit off topic, but aren't multiple sessions pretty normal even without intermediaries? Start multiple instances of the browser to runs tests in parallel, or even have them talk to each other as I know some WebRTC testing setups do.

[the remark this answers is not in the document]
jgraham: That's still one session per "endpoint node", which basically means "browser instance". Something like ChromeDriver could support multiple sessions of course, but each would be a different browser instance.

index.bs
foolip: I agree that we shouldn't (can't?) establish a ws connection during [new session](https://w3c.github.io/webdriver/#new-session), at least not if "establish" means making a network connection. Rather, we bind the port and are ready to accept new ws connections, like a ws server.

Having the ws URL be unguessable will probably preserve some flexibility, but we need to return the ws URL somewhere. It's not the capability itself, so I guess either a "websocketURL" capability, or as a sibling to the capabilities in the response. I'm not sure if this detail has already been discussed?

Aside on capabilties: per my testing of current chromedriver, geckodriver and safaridriver, neither `{"capabilities":{"firstMatch":[{"bidi":true},{"bidi":false}]}}` nor `{"capabilities":{"firstMatch":[{"bidi":true},{}]}}` will work, because they all return an error due to `bidi` not being a know capability. If there are clients that could work with and without bidi, they'd have to try connecting twice.

index.bs
foolip: My comment didn't end up where I intended. I mean "matched capability serialization algorithm" sounds odd. But maybe with the new extension point you're adding this will go away.

index.bs
jgraham: We can return the URL in the bidi field of the capability response. There isn't a problem here :)

You're right, the correct new session capabilities would have been `{capabilities: {firstMatch: [ {bidi: true}, {} ] } }`.

index.bs
jgraham: At the moment I don't think we have any way to restart the connection if it drops. Making that actually work seems hard; you miss a load of events and it's hard to know how a client would actually cope. Can you file a separate issue on that?

jgraham:
This now depends on https://github.com/w3c/webdriver/pull/1527

index.bs
bwalderman: As an aside: I don't think using the bidi WebDriver protocol end-to-end over a network is going to be the best way to take advantage of bidi. Users could avoid a lot of potential reliability and latency issues by using bidi to communicate with a browser on a single machine, and building higher-level protocols on top of bidi. For example, rather than running a test script on a local machine and communicating back-and-forth with a remote end, a user might just submit their entire test script to the remote end to be executed on the same machine as the browser where the connection is much more reliable.

index.bs
foolip: Note that `{"capabilities":{"firstMatch":[{"bidi":true},{}]}}` actually doesn't work, that's one of the forms I tested. I've filed https://github.com/w3c/webdriver/issues/1528 with a lot more details.

foolip: If we're going to use the same capability to request bidi and return the ws URL, I'd like to bikeshed the name and suggest "webSocketUrl", because it describes what you're getting and allows for another protocol in some distant future. The existing "proxyAutoconfigUrl" is similarly named, and it's also similar to the [`GET /json/version` endpoint in CDP](https://chromedevtools.github.io/devtools-protocol/#get-jsonversion).

index.bs
foolip: @bwalderman that sounds like a good approach, but for large test suites already built on WebDriver that want to add some bidi-only functionality, running the bidi connection over a reliable network is likely what will happen. I've filed https://github.com/w3c/webdriver-bidi/issues/25 to keep track of this.

index.bs
```diff
@@ -46,36 +51,41 @@ Transport {#transport}
...
-When data is recieved on [=connection=] it must be processed according to the
+When data is recieved on [=WebSocket connection=] it must be processed according to the
 requirements of the [[!RFC6455|WebSockets]] specification. When [=a WebSocket
 message has been received=], an implementation must [=handle an incoming
 message=]. When [=the WebSocket closing handshake is started=] an implementation
 must [=end the session=].
 
-When required to <dfn>ensure a connection is established</dfn>, run
-the following steps:
+When required to <dfn>Establish a WebSocket Connection</dfn>, with
```
foolip, line 64: I'm going to push a commit to lowercase this algorithm, and at the same time use the shorter "To X, run these steps" style.

index.bs
```diff
@@ -46,36 +51,41 @@ Transport {#transport}
...
+argument <var>session id</var>, run the following steps:
```
foolip: Can you pass in *session* rather than *session id* here? This way we can say "session's WebSocket URL" below, which is better.

index.bs
jgraham: I've removed "session" for now to avoid confusion.

index.bs
```diff
@@ -7,14 +7,110 @@ Group: browser-testing-tools
...
+A [=remote end=] has a <dfn>connection</dfn>, which is a network connection, and
+a <dfn>websocket url</dfn>, which is a string. These are both initially null.
+
+When data is recieved on [=connection=] it must be processed according to the
+requirements of the [[!RFC6455|WebSockets]] specification. When [=a WebSocket
+message has been received=], an implementation must [=handle an incoming
+message=]. When [=the WebSocket closing handshake is started=] an implementation
+must [=end the session=].
+
+When required to <dfn>ensure a connection is established</dfn>, run
```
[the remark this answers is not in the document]
jgraham: I went with "start listening for a WebSocket connection", which I think is closer to the truth. Although I suppose it's kind of wrong in detail since there's actually a two-step process here 1. start listening on a port for new connections and 2. actually accept the connection. I guess I ought to do some more work to make this right :/

index.bs
```diff
@@ -7,14 +7,160 @@ Group: browser-testing-tools
...
+"<code>bidi</code>" capability, with parameter |value| is:
+
+ 1. If |value| is not a boolean, return [=error=] with [=error code|code=]
+    [=invalid argument=].
+
+ 2. Return [=success=] with data |value|.
+
+</div>
+
+<div algorithm="bidi capability serialization algorithm">
```
foolip, line 136: Just to check my understanding of how things are structured: this hook exists only to propagate the capability to *matched capabilities* in https://w3c.github.io/webdriver/#dfn-matching-capabilities. One *could* serialize the actual URL here, but it's possible this won't be in the finally chosen *matched capabilities* in https://w3c.github.io/webdriver/#dfn-capabilities-processing, so we shouldn't start listening until our "bidi new session algorithm" hook.

I think I found a bug while staring at this: https://github.com/w3c/webdriver/issues/1530

index.bs
```diff
@@ -7,14 +7,160 @@ Group: browser-testing-tools
...
+with parameters |session| and |capabilities| is:
+
+ 1. Let |bidi| be the result of [=getting a property=] named "<code>bidi</code>"
+    from |capabilities|.
+
+ 2. If |bidi| is undefined, return.
+
+ 3. Let |session id| be the [=session ID=] of |session|.
+
+ 3. Otherwise |bidi| is <code>true</code>. Let |url| be the result of
```
foolip, line 158: https://infra.spec.whatwg.org/#assertions would work well here.

index.bs
```diff
@@ -7,14 +7,160 @@ Group: browser-testing-tools
...
+    calling the algorithm to [=start listening for a WebSocket
+    connection=] with argument |session id|.
+
+ 4. Set the [=WebSocket URL=] for |session| to |url|.
```
foolip, line 162:
```suggestion
 4. Set |session|'s [=WebSocket URL=] to |url|.
```

index.bs
```diff
@@ -7,14 +7,160 @@ Group: browser-testing-tools
...
+<div algorithm="bidi new session algorithm">
+The [=WebDriver new session algorithm=] defined by this specification,
+with parameters |session| and |capabilities| is:
+
+ 1. Let |bidi| be the result of [=getting a property=] named "<code>bidi</code>"
+    from |capabilities|.
+
+ 2. If |bidi| is undefined, return.
+
+ 3. Let |session id| be the [=session ID=] of |session|.
```
foolip, line 156:
```suggestion
 3. Let |session id| be the |session|'s [=session ID|ID=].
```

index.bs
```diff
@@ -86,19 +89,22 @@ argument |session id|, run these steps:
 Note: For an [=endpoint node=] the hostname in the above steps will be
 typically be "<code>localhost</code>".
 
-<div algorithm>
-To <dfn>handle an incoming message</dfn>, run thes steps:
+<div algorithm> To <dfn>handle an incoming message</dfn> with parameter
```
foolip, line 92:
```suggestion
<div algorithm> To <dfn>handle an incoming message</dfn> for a
```

foolip: Same suggestion below, but that spans lines so I can't suggest like this :)

index.bs
```diff
@@ -149,9 +155,9 @@ with arguments |session| and |capabilities| is:
 
  3. Let |session id| be the [=session ID=] of |session|.
 
- 3. Otherwise |bidi| is <code>true</code>. Let |url| be the result of calling
-    the algorithm to [=establish a WebSocket connection=] with argument
-    |session id|.
+ 3. Otherwise |bidi| is <code>true</code>. Let |url| be the result of
+    calling the algorithm to [=start listening for a WebSocket
```
foolip, line 159:
```suggestion
    calling [=start listening for a WebSocket
```

index.bs
```diff
@@ -55,15 +55,18 @@ A WebDriver [=session=] has a <dfn>WebSocket connection</dfn>, which
...
+WebSocket message has been received=], an implementation must [=handle
+an incoming message=] with argument <var>connection</var>. When [=the
+WebSocket closing handshake is started=] an implementation must
+[=close the WebSocket connection=] with argument
+<var>connection</var>.
 
 <div algorithm>
-To <dfn>establish a WebSocket connection</dfn>, with
-argument |session id|, run these steps:
+To <dfn>start listening for a WebSocket connection</dfn>, with
```
foolip, line 68:
```suggestion
To <dfn>start listening for a WebSocket connection</dfn>, for a |session|
```

foolip: Applying this suggestion by itself won't work, but we need that to set "|session|'s WebSocket URL" below.

index.bs
```diff
@@ -7,14 +7,154 @@ Group: browser-testing-tools
...
+requirements of the [[!RFC6455|WebSockets]] specification. When [=a WebSocket
+message has been received=], an implementation must [=handle an incoming
+message=]. When [=the WebSocket closing handshake is started=] an implementation
+must [=end the session=].
+
+<div algorithm>
+To <dfn>establish a WebSocket connection</dfn>, with
+argument |session id|, run these steps:
+
+ 1. Set up a network connection that listens on an implementation-defined
```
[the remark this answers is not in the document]
jgraham: Yep, as soon as I updated this I realised it was all wrong :) Working on this now.

index.bs
jgraham: Well we can't serialize the actual url here if it includes the session id since that isn't known yet. But basically, yes, that's correct.

index.bs
foolip: @jgraham has renamed it "webSocketUrl"

index.bs
foolip: The style is a bit funny as a result, but let's fix that later.

foolip:
@jgraham has reviewed my commits on this PR, I'll go ahead and merge now.

Issue #14, opened 2020-04-16
Specify a low-level transport format


jgraham:
The bi-directional protocol needs a low-level format for encoding messages that are sent over the wire.

The discussion so far has assumed that the format will be JSON-based; this has compatibility with the HTTP protocol and good human readability. It matches other existing protocols (although note that CDP now has a binary variant) and likely makes sense if we think that the message parsing overhead will be small compared to other latencies in the system.

@burg proposed that we adopt [JSON-RPC](https://www.jsonrpc.org/specification) as the specific flavour of JSON for the protocol. This meets several requirements:
* Provides a command type and a response type with an id field on match commands and responses
* Provides an event type that isn't generated by a command
* Provides an error type

However there are several details that might not match our requirements
* Requires a version number on every message, bloating the protocol for little gain (updating versions in WebDriver would be a difficult undertaking given the compatibility guarantees we provide).
* The protocol has a pipelining feature where an array of commands can be sent at once. This kind of feature seems useful to us, but the particular requirement that the implementation produce an array of responses doesn't seem like it matches use cases we might have (like running multiple concurrent commands and getting the response from each as soon as its ready).
* The spec has several SHOULD-level requirements; this is owrrying from the point of view of obtaining interop.

foolip:
@jgraham we can consider this fixed now, right?

jgraham:
Yes.

Issue #21, opened 2020-05-29
Machine readable definitions


foolip:
This is the BiDi sibling of issue https://github.com/w3c/webdriver/issues/1510, see that issue's description for the full background.

The solution for REST and BiDi likely won't be the same, and we might do one without the other.

For BiDi specifically, @bwalderman has already put together a [`openrpc.json`](https://github.com/w3c/webdriver-bidi/blob/1d245d0eb34f2a31c4c632aff01cbede22c85043/proposals/openrpc.json) proposal.

foolip:
@bwalderman was openrpc.json assembled by hand? How about the [API Reference](https://github.com/w3c/webdriver-bidi/blob/1d245d0eb34f2a31c4c632aff01cbede22c85043/proposals/api/index.html), is that generated from openrpc.json?

jgraham:
All of these api definition formats seem to use JSON Schema for the actual definitions. I'm not convineced that we really care about the value add of the additional layers on top of that; from a skim it looks like the additional features are about service discovery and licensing, which I don't think we particularly care about. In particular I see the following as use cases for machine-readable defintions in the spec:
* Reduce the spec-text boilerplate describing de(serilaization) of messages
* Make more of the spec constraints machine verifiable (e.g. ability to cross check that all errors are one of the accepted codes)
* Give browser authors and client authors definitions they can use directly implement the de(serialization) of messages
* Provide better documentation of the expected message format compared to having to reverse engineer the browser steps

I see the following as non-goals:
* Allowing generic RPC clients to connect to WebDriver endpoints and navigate them without specific understanding of the protocol semantics

So I don't think we want endpoints that produce schema documents to allow clients to introspect the API or anything; in practice all the WebDriver and CDP clients are providing significant value-add over the mechanical conversion of protocol endpoints into code, and in any case updates to the spec will be accompanied by updates to the published schema, so we don't also need to allow introspection.

Given that, I think we should just write json schema directly and not try to adopt any of the higher layer stuff like [Async|Open]API which afaict are mostly addressing needs we don't have.

bwalderman:
@foolip yes, openrpc.json was hand-written and the API reference was generated from it using https://github.com/open-rpc/schema-utils-js and some HTML templates.

foolip:
I see. I guess it’s not worth the effort now to put that build step into CI, but if we have a schema file later that’d make sense.

christian-bromann:
[...]

PR #26, opened 2020-06-04
Define the basic transport-agnostic terms of the protocol

foolip:
Following https://github.com/w3c/webdriver-bidi/issues/14#issuecomment-633664055

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/26.html" title="Last updated on Jun 8, 2020, 4:02 PM UTC (1053fb8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/26/580fd3d...1053fb8.html" title="Last updated on Jun 8, 2020, 4:02 PM UTC (1053fb8)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -92,11 +92,11 @@ failure of that command. An [=error=] consists of:
  - A <dfn export for=error>message</dfn> (string)
  - Optional <dfn export for=error>data</dfn> (any type)
 
-A <dfn export>notification</dfn> is sent by the [=remote end=] to the
+An <dfn export>event</dfn> is sent by the [=remote end=] to the
 [=local end=]. It is similar to a [=response=], but does not
-correspond to a specific [=command=]. A [=notification=] consists of:
- - A <dfn export for=notification>name</dfn> (string)
- - <dfn export for=notification>parameters</dfn> ([=map=] with string
+correspond to a specific [=command=]. A [=event=] consists of:
+ - A <dfn export for=event>name</dfn> (string)
+ - <dfn export for=event>parameters</dfn> ([=map=] with string
     <a for=map>keys</a> and any type as <a for=map>values</a>)
 
 Issue: Should the strings here be [=strings=] (a sequence of unsigned
--- a/index.bs
+++ b/index.bs
@@ -61,8 +61,7 @@ Model {#model}
 
 This section defines the basic concepts of the WebDriver BiDi
 protocol. These terms are distinct from their representation at the
-<a href=#transport>transport</a> layer, and can be thought of as their
-in-memory representation.
+<a href=#transport>transport</a> layer.
 
 Note: These definitions do not form any normative requirements, rather
 they are used elsewhere as part of such requirements.
--- a/index.bs
+++ b/index.bs
@@ -37,6 +37,7 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
     text: local end; url: dfn-local-ends
     text: matched capability serialization algorithm; url: dfn-matched-capability-serialization-algorithm
     text: remote end; url: dfn-remote-ends
+    text: remote end steps; url: dfn-remote-end-steps
     text: session; url: dfn-sessions
     text: session ID; url: dfn-session-id
     text: set a property; url: dfn-set-a-property
@@ -66,41 +67,34 @@ protocol. These terms are distinct from their representation at the
 Note: These definitions do not form any normative requirements, rather
 they are used elsewhere as part of such requirements.
 
-A <dfn export>command</dfn> is sent by the [=local end=] to the
-[=remote end=], requesting it to perform some action. A [=command=]
-consists of:
- - An <dfn export for=command>id</dfn> (integer) to be used in the
-    resulting [=response=] or [=error=].
- - A <dfn export for=command>name</dfn> (string)
- - <dfn export for=command>parameters</dfn> ([=map=] with string
-    <a for=map>keys</a> and any type as <a for=map>values</a>)
+## Commands ## {#commands}
 
-A <dfn export>response</dfn> is sent by the [=remote end=] to the
-[=local end=] in response to a specific [=command=], and signifies the
-success of that command. A [=response=] consists of:
- - An <dfn export for=response>id</dfn> (integer) equal to the
-    corresponding [=command=]'s <a for=command>id</a>.
- - A <dfn export for=response>result</dfn> (any type)
+A <dfn export>command</dfn> is an asynchronous operation, requested by
+the [=local end=] and run on the [=remote end=], resulting in either
+[=success=] (with a result) or [=error=] (with details) being returned
+to the [=local end=]. Multiples command can run at the same time, and
+commands can potentially be long-running. As a consequence, commands
+can finish out-of-order.
 
-An <dfn export>error</dfn> is sent by the [=remote end=] to the
-[=local end=] in response to a specific [=command=], and signifies the
-failure of that command. An [=error=] consists of:
- - An <dfn export for=error>id</dfn> (integer) equal to the
-    corresponding [=command=]'s <a for=command>id</a>.
- - A <dfn export for=error>code</dfn> (a WebDriver [=error code=]
-    string)
- - A <dfn export for=error>message</dfn> (string)
- - Optional <dfn export for=error>data</dfn> (any type)
+Each concrete [=command=] type is defined by:
+ - A string <dfn export for=command>name</dfn>, implicit in its
+   definition.
+ - A set of <dfn export for=command>parameters</dfn>, which are the
+   inputs to the command
+ - A set of [=remote end steps=].
+ - A <dfn export for=command>return type</dfn>, which may be null.
 
-An <dfn export>event</dfn> is sent by the [=remote end=] to the
-[=local end=]. It is similar to a [=response=], but does not
-correspond to a specific [=command=]. A [=event=] consists of:
- - A <dfn export for=event>name</dfn> (string)
- - <dfn export for=event>parameters</dfn> ([=map=] with string
-    <a for=map>keys</a> and any type as <a for=map>values</a>)
+## Events ## {#events}
 
-Issue: Should the strings here be [=strings=] (a sequence of unsigned
-16-bit integers) or [=scalar value strings=]?
+An <dfn export>event</dfn> is a notification, sent by the [=remote
+end=] to the [=local end=], signaling that something of interest has
+occurred on the [=remote end=].
+
+Each concrete [=event=] type is defined by:
+ - A string <dfn export for=command>name</dfn>, implicit in its
+   definition.
+ - A set of <dfn export for=command>parameters</dfn>, which are the
+   details of the event that has occurred.
 
 Transport {#transport}
 ======================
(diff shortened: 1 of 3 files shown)


index.bs
```diff
@@ -38,13 +38,64 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
...
+client in response to a specific [=command=], and signifies the
+success of that command. A [=response=] consists of:
+ - An <dfn export for=response>id</dfn> (integer) equal to the
+    corresponding [=command=]'s <a for=command>id</a>.
+ - A <dfn export for=response>result</dfn> (any type)
+
+An <dfn export>error</dfn> is sent by the [=remote end=] to the client
+in response to a specific [=command=], and signifies the failure of
+that command. An [=error=] consists of:
+ - An <dfn export for=error>id</dfn> (integer) equal to the
```
jgraham, line 81: Is it possible to have errors not in response to commands?

index.bs
```diff
@@ -38,13 +38,64 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
...
+    corresponding [=command=]'s <a for=command>id</a>.
+ - A <dfn export for=error>code</dfn> (a WebDriver [=error code=]
```
jgraham, line 83: I don't think we should reuse WebDriver error codes here, or at least I think we want to extend them.

index.bs
```diff
@@ -38,13 +38,64 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
...
+ - A <dfn export for=command>name</dfn> (string)
+ - <dfn export for=command>parameters</dfn> ([=map=] with string
+    <a for=map>keys</a> and any type as <a for=map>values</a>)
+
+A <dfn export>response</dfn> is sent by the [=remote end=] to the
+client in response to a specific [=command=], and signifies the
+success of that command. A [=response=] consists of:
+ - An <dfn export for=response>id</dfn> (integer) equal to the
+    corresponding [=command=]'s <a for=command>id</a>.
+ - A <dfn export for=response>result</dfn> (any type)
```
jgraham, line 76: When you say "type" here, what are you imagining? I initially thought "JSON type" (string, number, object, etc.), but I'd like to be able to define types in the spec ("RealmId", "ContextId", "JSValue", etc.), which might have complex serilizations. If we mean the latter I think I"m OK with this, but we'll need to define "type" later.

index.bs
```diff
@@ -38,13 +38,64 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
...
+
+A <dfn export>notification</dfn> is sent by the [=remote end=] to the
+client. It is similar to a [=response=], but does not correspond to a
+specific [=command=]. A [=notification=] consists of:
+ - A <dfn export for=notification>domain</dfn> string)
+ - A <dfn export for=notification>name</dfn> (string)
+ - <dfn export for=notification>parameters</dfn> ([=map=] with string
+    <a for=map>keys</a> and any type as <a for=map>values</a>)
+
+Issue: Should the strings here be [=strings=] (a sequence of unsigned
```
jgraham, line 96: In practice everything has to serialize as JSON and arbitary JS strings arguably don't (or at least aren't very portable). In practice all the strings defined here are specification-defined values so it effectively doesn't matter.

index.bs
```diff
@@ -38,13 +38,64 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
...
+
+Note: These definitions do not form any normative requirements, rather
+they are used elsewhere as part of such requirements.
+
+A <dfn export>command</dfn> is sent by the client to the [=remote
+end=], requesting it to perform some action. A [=command=] consists
+of:
+ - An <dfn export for=command>id</dfn> (integer) to be used in the
+    resulting [=response=] or [=error=].
+ - A <dfn export for=command>domain</dfn> (string)
```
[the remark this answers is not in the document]
foolip: I'm not entirely sure of this either, but my suspicion is that we'll want to parse out and refer to a domain as its own thing, and not just an opaque part of a string. That isn't incompatible with strings like `Logging.*` being the parameter value for something.

Happy to change this when we start using it, and I'm not committed to it staying like this.

index.bs
```diff
@@ -38,13 +38,64 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
...
 from the user agent to the controlling software, better matching the
 evented nature of the browser DOM.
 
+Model {#model}
+==============
+
+This section defines the basic concepts of the WebDriver BiDi
+protocol. These terms are distinct from their representation at the
+<a href=#transport>transport</a> layer, and can be thought of as their
+in-memory representation.
```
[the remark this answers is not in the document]
foolip: I'll just drop the "and can be thought of as their in-memory representation" bit. I think it will turn out inconvenient to not have the ID on the command concept, but please comment again if you'd like to start that way and add it when we find it's a problem.

index.bs
foolip: I mean any https://infra.spec.whatwg.org/#primitive-data-types or other type that we define. We will indeed need to define exactly how to turn a hopefully-JSON-byte-sequence into these constructs.

index.bs
foolip: I think we'll have to deal with this when we specify executing a script where that returns a JS string with lone surrogates. I have opinions, but would like to defer the decision for now :)

index.bs
```diff
@@ -38,13 +38,64 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
...
+
+Note: These definitions do not form any normative requirements, rather
+they are used elsewhere as part of such requirements.
+
+A <dfn export>command</dfn> is sent by the client to the [=remote
```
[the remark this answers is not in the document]
foolip: Done.

index.bs
```diff
@@ -38,13 +38,64 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
...
+in response to a specific [=command=], and signifies the failure of
+that command. An [=error=] consists of:
+ - An <dfn export for=error>id</dfn> (integer) equal to the
+    corresponding [=command=]'s <a for=command>id</a>.
+ - A <dfn export for=error>code</dfn> (a WebDriver [=error code=]
+    string)
+ - A <dfn export for=error>message</dfn> (string)
+ - Optional <dfn export for=error>data</dfn> (any type)
+
+A <dfn export>notification</dfn> is sent by the [=remote end=] to the
```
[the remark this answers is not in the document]
foolip: I think I'd prefer "event" as well, but matched the terminology in https://github.com/w3c/webdriver-bidi/blob/master/proposals/core.md to begin with.

@bwalderman do you have thoughts on this?

index.bs
jgraham: Yes, I strongly agree with the idea of working as much as possible at the highest level types.

index.bs
jgraham: Right, we clearly need to define it in general, but it's not strictly necessary now because most of the "strings" here are spec values that happen to be serialized as strings in the JSON representation rather than arbitrary string data.

index.bs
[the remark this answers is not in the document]
jgraham: So I think I'd like to go a bit further and define the underlying types separately from the messages. So I'd like to have something like:

"""
A `Command Type` has a string `command name` and a set of parameters which are defined by subtypes.

A `Command message` consists of an instance of a `Command Type` and an `id` which is an integer used as a connection-unique identifier for a specific message.
"""

index.bs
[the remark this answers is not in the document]
foolip: Alright, done!

index.bs
foolip: Would you like me to link one of https://infra.spec.whatwg.org/#string or https://infra.spec.whatwg.org/#scalar-value-string and revisit if necessary later, or just remove this inline issue?

index.bs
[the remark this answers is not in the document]
foolip: I think I agree with where you're going. While there will be some behavior between all commands, there will be no instance of the **command** type, it's an abstract type. Where I've said "any type" is likely to not end up useful, the parameters should as you say be defined by the concrete subtypes.

When we get to defining some concrete command, I think we can start with prose in the style of "`thingHappened` is an **event** in the `Something` domain with parameters **type** (an enum of ...) and **message** (a string)."

I'll give it a shot to move in that direction tomorrow.

Aside: When we've repeated that enough times we might be tempted to look at https://github.com/w3c/webdriver-bidi/issues/21, so I have in the back of my mind that some of this won't be defined with a bunch of prose, just like we don't say "`doThing` is a **method** for the `Something` interface with arguments ..."

index.bs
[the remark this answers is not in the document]
bwalderman: My vote is for "event" as well. I originally used "notification" from the JSON-RPC spec, but what we're describing here is commonly known as an event in most places, including the DOM and CDP.

index.bs
bwalderman: Maybe this spec should have its own definition of an Error Code. We could start with the existing set of WebDriver error codes and add any new standard error codes we may need for new commands we introduce.

I think it would also be helpful for extension commands to be able to send custom errors, but that sounds like a discussion for another PR. I've filed #28 issue to track.

index.bs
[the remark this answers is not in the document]
foolip: I've pushed a change to rename it to event, here and in the various other places, except where JSON-RPC was the topic.

index.bs
foolip: It's done.

index.bs
```diff
@@ -38,13 +38,64 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
...
+<a href=#transport>transport</a> layer, and can be thought of as their
+in-memory representation.
+
+Note: These definitions do not form any normative requirements, rather
+they are used elsewhere as part of such requirements.
+
+A <dfn export>command</dfn> is sent by the client to the [=remote
+end=], requesting it to perform some action. A [=command=] consists
+of:
+ - An <dfn export for=command>id</dfn> (integer) to be used in the
```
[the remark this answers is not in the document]
foolip: The id no longer appears in this section. It comes up again in https://github.com/w3c/webdriver-bidi/pull/34 though.

index.bs
```diff
@@ -47,13 +48,54 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
...
+<a href=#transport>transport</a> layer.
+
+Note: These definitions do not form any normative requirements, rather
+they are used elsewhere as part of such requirements.
+
+## Commands ## {#commands}
+
+A <dfn export>command</dfn> is an asynchronous operation, requested by
+the [=local end=] and run on the [=remote end=], resulting in either
+[=success=] (with a result) or [=error=] (with details) being returned
```
jgraham, line 74: This is a little confusing, because `success` and `error` are really spec-internal notions and you can't "send a success", you can just respond to the remote end steps returning a `success` by sending a response message over the wire and respond to them returning an `error` by sending an error message over the wire.

foolip: You're right, I've pushed a change to not link the spec-internal notions.

foolip:
FWIW, yes, I agree this likely to be refactored again. To figure out precisely what we *do* need, I think it'd probably help a lot to have a concrete command or two to spec, which should make it obvious what bits will be repeated.

PR #38, opened 2020-06-08
Update core proposal to reflect (tentative) choices for "webSocketUrl"

foolip:


proposals/core.md
```diff
@@ -116,40 +116,36 @@ To use the new bidirectional protocol, the client must first establish a WebSock
 
 >[RESOLUTION](https://www.w3.org/2019/09/20-webdriver-minutes.html#resolution01): Bi-di is always enabled. An optional capability, defaulting to true, indicating that bi-di is desired. When a new session is established, the return value of the new session contains the new top-level property of the bi-directional URL
```
foolip, line 117: Note that what we landed in https://github.com/w3c/webdriver-bidi/pull/24 doesn't default to true. Should we revisit that?

jgraham: Defaulting to true seems wrong to me. Once support is there clients that are using it can start to default to true.

proposals/core.md
```diff
@@ -116,40 +116,36 @@ To use the new bidirectional protocol, the client must first establish a WebSock
 
 >[RESOLUTION](https://www.w3.org/2019/09/20-webdriver-minutes.html#resolution01): Bi-di is always enabled. An optional capability, defaulting to true, indicating that bi-di is desired. When a new session is established, the return value of the new session contains the new top-level property of the bi-directional URL
 
-A bidirectional session starts as a traditional session created with a POST /session command. The client can upgrade their session to a bidirectional one by connecting to the WebSocket URL provided in the response. The client can also request the optional "protocolVersion" capability to explicitly enable or disable bidi:
+A bidirectional session starts as a traditional session created with a POST /session command. The client can request a bidirectional using the "webSocketUrl" capability:
```
bwalderman, line 119: Nit: "The client can request a bidirectional session using the "webSocketUrl" capability:"

foolip: Ah, missing "session".

proposals/core.md
foolip: It seems wrong to me as well. Reading the conversation leading up to https://www.w3.org/2019/09/20-webdriver-minutes.html#resolution01 I can't find any clues about whether this was deliberate or not, and I wasn't at these meetings. I guess I'll just leave the contradiction here and maybe ask about it quickly in a future VC.

Issue #17, opened 2020-04-17
Specify how to enable an event stream


jgraham:
For various use cases (e.g. logging) we want the ability to enable a stream of events from the browser to the client. But it isn't possible to have all events enabled all the time; we want a design that allows us to enable and disable particular events, either for the session as a whole or for specific browsing contexts. In some cases enabling a specific event stream might also enable additional commands that can make use of data from the events.

The explainer document currently proposes a mechanism where different event streams have enable/disable pairs with a counter that is incremented for each enable and decremented for each disable (clamped at 0), with the stream only being actually disabled when the count reaches 0. But there are concerns that this mechanism may be over-complex, providing a protocol-level solution to application-level concern around how to handle multiple parts of the code. 

The simplest solution here seems to be to divide events into various groups ("domains" in CDP) and have an enable and disable command for each group. Some additional property may be needed to determine the scope of the enabling (e.g. one browsing context / agent vs global).

bwalderman:
Here's a strawman proposal for enabling events:

Related events are grouped into domains. The client can subscribe to all events in a domain, or to specific events in a domain. Events can be enabled globally for all contexts, or for a subset of contexts. When events are enabled globally, those events are received from all current and future contexts.

Most events are disabled initially. When a new WebDriver BiDi session is created, the client should enable any event streams they are interested in before proceeding with their test. It might be a good idea to enable events related to context discovery by default; Something like CDP's `Target` domain.

Subscribing to events globally is useful for automation scenarios where you want to be aware of everything that is happening throughout the browser. Sometimes though, you may want to enable events for a particular tab, so we should make it possible to specify a top-level browsing context or service worker to enable events on. Enabling events for a particular nested frame or a particular worker doesn't seem as useful so we probably don't need to make this a priority. 

If we follow CDP's general approach, and add parameters for choosing specific events and contexts we might end up with something like this:

```js
// Enable these specific Network events globally
{
    method: "Network.enable",
    events: ["requestWillBeSent", "responseReceived"] 
}

// Enable all Network.* events for this specific context
{
    method: "Network.enable",
    contexts: ["<tab A>", "<service worker B>"]
}
```

The values in the `contexts` array must be top-level contexts: a tab, service worker, or shared worker. The requested events will be enabled for that context, and all nested contexts such as frames and workers.

It might also be useful to allow the client to subscribe to events for particular contexts even if they don't know those context's IDs yet. For example, to subscribe to event for any current or future tab matching a certain URL:

[...]

PR #41, opened 2020-06-30
Update Target Discovery & Message Routing Proposal

bwalderman:
This change updates the core proposal considering some feedback from active GitHub issues as well as the working group meeting. A quick summary of changes:

1. An updated Object Model that more closely reflects how browsing contexts and various worker types are specified. Also adds some new worker types that were not in the proposal initially.
2. A string ID format to address any of these objects
3. Added "to" and "from" fields in the low-level transport format to route commands to an object and know which object an event is coming from
4. A subset of events that are always enabled to assist with target discovery
5. Removed the ref-counted proposal for event subscription. Subscriptions are now simply enabled or disabled per-target. Using domains instead of individual event names.

### DIFF
--- a/proposals/core.md
+++ b/proposals/core.md
@@ -100,7 +100,8 @@ All event messages have a "from" property indicating which object in the BiDi mo
 {
     "from": "browsingContext/333",
     "method": "logging.entryAdded",
-    "params": { "type": "info", message: "Hello World!" }
+    "params": { "type": "info", "message": "Hello World!" }
+
 }
 ```
 
--- a/proposals/core.md
+++ b/proposals/core.md
@@ -47,7 +47,7 @@ The interface for the new protocol would be a set of client-to-server commands, 
 }
 ```
 
-Commands include a "method" name and optional "params" object containing named parameters. A "to" field indicates which actor in the BiDi model to route the request to.
+Commands include a "method" name and optional "params" object containing named parameters. A "to" field indicates which object in the BiDi model to route the request to.
 
 *Sample success response*
 
@@ -71,7 +71,9 @@ Responses from the server include the "id" of the command they are responding to
 
 #### Events
 
-Since events may generate a large amount of traffic over the WebSocket, and may have a runtime cost in the browser, these should be opt-in. Commands should be provided so that a client can subscribe and unsubscribe. Some events that are crucial to target discovery will be enabled by default. These are discussed in the Target Discovery section below.
+Since events may generate a large amount of traffic over the WebSocket, and may have a runtime cost in the browser, these should be opt-in. Commands should be provided so that a client can subscribe and unsubscribe.
+
+Some events that are crucial to target discovery will be enabled by default. These are discussed in the Target Discovery section below.
 
 The first step to receive events on the client side would be to send a "subscribe" command:
 
@@ -92,7 +94,7 @@ The client sends a matching "unsubscribe" command when they no longer want to re
 
 *Sample event*
 
-All event messages have a "from" property indicating which actor in the BiDi model sent the event. Event messages don't have an "id" property since they are not associated with a command request message.
+All event messages have a "from" property indicating which object in the BiDi model sent the event. Event messages don't have an "id" property since they are not associated with a command request message.
 
 ```json
 {
@@ -163,7 +165,7 @@ A session has one or more windows (top-level browsing contexts). Each window has
 
 ### Updated WebDriver/BiDi Model
 
-The object model below represents a somewhat simplified view of the modern user agent as defined in https://html.spec.whatwg.org/. The model includes the all of the actors represented in WebDriver/HTTP and adds new Worker and Worklet actors. It also separates the notion of a JS Realm from a browsing context and introduces a type for representing complex JS Values:
+The object model below represents a somewhat simplified view of the modern user agent as defined in https://html.spec.whatwg.org/. The model includes the all of the objects represented in WebDriver/HTTP and adds new Worker and Worklet objects. It also separates the notion of a JS Realm from a browsing context and introduces a type for representing complex JS Values:
 
 ```ts
 UserAgent
@@ -240,6 +242,8 @@ Value
     id: string?
     value: object?
 ```
+
+The BiDi remote ends will send events to the client whenever new objects become available, or if an objects's basic properties (listed above) change.
 
 ### Identifiers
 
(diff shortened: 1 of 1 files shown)


proposals/core.md
```diff
@@ -147,38 +150,127 @@ WebDriver has a notion of a current top-level browsing context, and current brow
...
+Every object in the model has an `id` property that is unique throughout the Session and indicates what type of object it is. An `id` is a human-readable type name, following by a string UUID. Examples:
 
-- WebDriver Session
-    - Target* (Formerly "Window". Can be a page or service worker)
-        - Browsing Context (aka "Frame")
-            - Element
-        - Script Context* (document, web worker, service worker, etc.)
+ - `"userAgent"` - The root user agent
+ - `"window/83d07169-4143-4ae8-947c-713f6949329f"` - A window target
```
jgraham, line 253: I"m not a fan of these complex string types that require parsing. Why not `{type: "window", "id": "83d07169-4143-4ae8-947c-713f6949329f"}` for example.

proposals/core.md
```diff
@@ -147,38 +150,127 @@ WebDriver has a notion of a current top-level browsing context, and current brow
...
 
-### Identifying elements
+#### Frames & Elements
 
-Web element references can continue to work as they do today, with one caveat; since element references are valid only within a particular browsing context, any commands that operate on elements and any events involving elements will need to specify which browsing context the element(s) belongs to. For example, a command like getElementText would need both a "browsingContextId" and an "elementId" parameter.
+WebDriver/HTTP uses Web Element References to represent HTML Elements and to represent Frames in the Switch To Frame command. However, Web Element References are valid only within a particular browsing context. To obtain a BiDi Element or Frame ID given a Web Element Reference, the client would first need to know which browsing context the Web Element Reference belongs to. The `BrowsingContext` can provide `findElementByHttpHandle` and `findBrowsingContextByHttpHandle` commands to convert Web Element Referenced into a BiDi Elements or Frames respectively.
```
jgraham, line 271: In the case of frames, I agree we need a new concept of a unique identifier (which will be just like the identifier for a tlbc), but it is also possible to simplify the compatibility story here by allowing the concept of a relative frame reference like `{type: "frame", "parent": <window id>, "ref": <webdriver/http reference>}`

proposals/core.md
```diff
@@ -41,12 +41,13 @@ The interface for the new protocol would be a set of client-to-server commands,
 ```json
 {
     "id": 0,
+    "to": "browsingContext/111",
```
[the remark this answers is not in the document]
foolip: I think it's debatable wether routing should be inside `params` or alongside it, and I can see the argument for alongside. It's not a input to the remote steps, and so there will probably be bits of spec and implementation that would have to treat it especially, and where it would have to be excluded from some handling if it is inside `params`.

proposals/core.md
```diff
@@ -64,40 +65,42 @@ Responses from the server include the "id" of the command they are responding to
...
+
+*Subscribing to events*
 
 ```json
 {
     "id": 0,
+    "to": "window/222",
     "method": "subscribe",
-    "params": { "event": "scriptContextCreated" }
+    "params": { "domains": ["logging", "network"] }
```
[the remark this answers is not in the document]
foolip: For anyone following along, this is the topic of https://github.com/w3c/webdriver-bidi/issues/17.

proposals/core.md
```diff
@@ -41,12 +41,13 @@ The interface for the new protocol would be a set of client-to-server commands,
...
     "id": 0,
+    "to": "browsingContext/111",
     "method": "getTitle",
-    "params": { "browsingContextId": "<ID>" }
+    "params": {}
 }
 ```
 
-Commands include a "method" name and optional "params". Positional parameters in an array are supported, but named parameters in an object are more descriptive and map more closely to how WebDriver commands currently work.
+Commands include a "method" name and optional "params" object containing named parameters. A "to" field indicates which object in the BiDi model to route the request to.
```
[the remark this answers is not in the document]
foolip: In other words, there's an implicit target which is the browser instance itself. Is that what you're thinking, @jgraham?

proposals/core.md
```diff
@@ -64,40 +65,42 @@ Responses from the server include the "id" of the command they are responding to
...
-Event messages don't have an "id" property since they are fire-and-forget.
+All event messages have a "from" property indicating which object in the BiDi model sent the event. Event messages don't have an "id" property since they are not associated with a command request message.
 
 ```json
 {
-    "method": "scriptContextCreated",
-    "params": { "scriptContextId": "<ID>" }
+    "from": "browsingContext/333",
+    "method": "logging.entryAdded",
+    "params": { "type": "info", message: "Hello World!" }
```
foolip, line 103:
```suggestion
    "params": { "type": "info", "message": "Hello World!" }

```

The syntax highlighting is messed up after this, so I'll commit this suggestion myself.

proposals/core.md
```diff
@@ -147,38 +151,127 @@ WebDriver has a notion of a current top-level browsing context, and current brow
...
+
+```ts
+UserAgent
+    // One per session, maintains a list of active targets
+    targets: Target[]
+
+Target
+    // A top-level browsing context or worker
+    id: string
+    type: "Window" | "SharedWorker" | "ServiceWorker"
```
foolip, line 179:
```suggestion
    type: "Window" | "SharedWorker" | "SharedWorker" | "ServiceWorker"

```

proposals/core.md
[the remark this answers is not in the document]
bwalderman: There's a UserAgent object in the proposed model that would serve as a browser-wide target for something like ending the session.

proposals/core.md
```diff
@@ -189,56 +282,82 @@ Below are some sample bidi commands that illustrate how message routing would wo
...
 ```
 
-#### Close a target
+#### Close a top-level window target
 
 ```json
 {
     "id": 0,
+    "to": "window/83c286bc-dc72-4678-a79f-c033b76498d7",
     "method": "closeTarget",
```
foolip, line 311: This isn't new in the proposal, but I wonder if we really want closing to be generic? The only two things that can be closed AFAIK are windows and tabs. The whole browser itself would be closed by ending the session, I suppose.

So right now, window is the only type of target that can be closed at all.

bwalderman:
@foolip, @jgraham, thanks for the feedback so far. I'm going to put this PR aside for the moment so we can move the discussion to the relevant issue threads.

PR #50, opened 2020-08-26
Refactor the way we use CDDL

jgraham:
Previously the spec only provided fragments of CDDL to validate
specific kinds of messages. But it makes sense from an implementation
point of view to have full schema that can be used directly.

In practice we require two seperate schema, one for validation of
messages going from the local end (i.e. client) to the remote
end (i.e. browser) and one for messages going in the opposite
direction. This change adds those schema to the spec.

The modelling adopted here is likely not the only approach, but the
basic idea is that each kind of message is modelled as a choice
between several variants, representing each possible message in the
protocol. Because the messages have unique names (e.g. in the `method`
field) we know that zero or one will match. However the names are not
represented directly.

In terms of organistaion, this currently defines a schema prelude with
each message type specified by a set of choices, one per module. Then
each module defines a choice per command (or other message) and
finally the messages themselves define the actual schema. I expect
this organisation to change somewhat over time. In particular using
the `//=` oeprator in CDDL seems like it might avoid the need to
define so many terms upfront, but it also seemed to be poorly
supported in the CDDL library I tried.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/50.html" title="Last updated on Sep 16, 2020, 11:46 AM UTC (a2e2a1a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/50/bb5d17e...a2e2a1a.html" title="Last updated on Sep 16, 2020, 11:46 AM UTC (a2e2a1a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -205,9 +205,15 @@ commands can run at the same time, and commands can potentially be
 long-running. As a consequence, commands can finish out-of-order.
 
 Each [=command=] is defined by:
-- A <dfn export for=command>command id</dfn>. This is a unique integer id
-  provided by the [=local end=] for each command.
- - A <dfn export for=command>command type</dfn> which is defined by a [=remote
+- A <dfn export for=command>command id</dfn>. This is an integer id
+  provided by the [=local end=] for each command. The value of the id is not
+  used by the [=remote end=] and can be regarded as opaque.
+
+  Note: In particular nothing guarantees that the [=command id=] is unique
+  for a given session. For example, a [=local end=] that sends only one command
+  at a time could give each the same command id.
+
+- A <dfn export for=command>command type</dfn> which is defined by a [=remote
    end schema=] fragment containing a group. Each such group has two fields:
     - <code>method</code> which is a string literal of the form <code>[module
       name].[method name]</code>. This is the <dfn export for=command>command
@@ -274,12 +280,12 @@ To <dfn>process a command</dfn> given |data|:
    1. Match |data| against the [=remote end schema=]. If this results in a
       match, run the following inner steps:
 
-      1. Let <var>result</var> be the result running the [=remote end steps=]
+      1. Let |result| be the result of running the [=remote end steps=]
          for the command with the [=command name=] equal to the
          <code>method</code> property of |data| and with the [=command
          parameters=] from the matched data.
 
-      1. Assert: if <var>result</var> is a [=success=] its |data| matches the
+      1. Assert: if |result| is a [=success=] its data matches the
          schema for the [=result type=] corresponding to the command with name
          [=command name=].
 
--- a/index.bs
+++ b/index.bs
@@ -73,6 +73,13 @@ command/response format of WebDriver, this permits events to stream
 from the user agent to the controlling software, better matching the
 evented nature of the browser DOM.
 
+Infrastructure {#infrastructure}
+==========================
+
+This specification depends on the Infra Standard. [[!INFRA]]
+
+Network protocol messages are defined using CDDL. [[!RFC8610]]
+
 Protocol {#protocol}
 ==============
 
@@ -80,26 +87,114 @@ This section defines the basic concepts of the WebDriver BiDi
 protocol. These terms are distinct from their representation at the
 <a href=#transport>transport</a> layer.
 
-This specification uses Concise Data Definition Language (CDDL)
-[[!RFC8610]] to describe the format of messages transmitted between the
-[=local end=] and [=remote end=].
+The protocol is defined using [[!RFC8610|CDDL]] schema. For the convenience of
+implementors two seperate schema are defined; the <dfn>remote end schema</dfn>
+which defines the format of messages produced on the [=local end=] and consumed
+on the [=remote end=], and the <dfn>local end schema</dfn> which defines the
+format of messages produced on the [=remote end=] and consumed on the [=local
+end=]
 
-Note: These definitions do not form any normative requirements, rather
-they are used elsewhere as part of such requirements.
+## Schema ## {#protocol-schema}
+
+Issue: Should this be an appendix?
+
+This section gives the initial contents of the [=remote end schema=] and
+[=local end schema=]. These are augmented by the schema fragments defined in
+the remainder of the specification.
+
+[=Remote end schema=]
+
+```
+Command = {
+  id: uint,
+  CommandData,
+}
+
+CommandData = (
+  SessionCommand
+)
+
+EmptyParams = { *text }
+```
+
+[=Local end schema=]
+
+```
+Message = (
+  CommandResponse //
+  Event
+)
+
+CommandResponse = (
+  id: uint,
+  ResponseData,
+)
+
+ResponseData = (
+  Error //
+  CommandResult
+)
+
+Error = (
+  id: uint,
+  error: (
+    "unknown error" /
+    "unknown method" /
+    "invalid argument"
+  ),
+  message: text,
+  stacktrace: text,
+)
+
+CommandResult = {
+  value: ResultData,
+}
+
+ResultData = (
+  EmptyResult //
+  SessionResult
+)
+
+EmptyResult = { *text }
+
+Event = ()
+```
+
+## Session ## {#session}
+
+WebDriver BiDi uses the same [=/session=] concept as WebDriver.
 
 ## Modules ## {#protocol-modules}
 
-The WebDriver BiDi protocol is organized into modules. A
-<dfn>module</dfn> is defined by:
- - A unique string <dfn export for=module>name</dfn>, implicit in its
-   definition.
- - A set of [=commands=].
- - A set of [=events=].
+The WebDriver BiDi protocol is organized into modules.
 
-Each [=module=] represents a collection of related [=commands=] and
-[=events=] pertaining to a certain aspect of the user agent. For
-example, a module might contain functionality for inspecting and
+Each <dfn export for=module>module</dfn> represents a collection of related
+[=commands=] and [=events=] pertaining to a certain aspect of the user
+agent. For example, a module might contain functionality for inspecting and
 manipulating the DOM, or for script execution.
+
+Each module has a <dfn>module name</dfn> which is a string literal. The
+[=command name=] and [=event name=] for commands and events defined in the
+module start with the [=module name=] followed by a period <code>"."</code>.
+
+Modules which contain [=commands=] define [=remote end schema=]
+fragments. These provide choices in the <code>CommandData</code> group for the
+module's [=commands=],and can also define additional schema properties. They
+can also define [=local end schema=] fragments that provide additional choices
+in the <code>ResultData</code> group for the results of commands in the module.
+
+Modules which contain events define [=local end schema=] fragments that are
+choices in the <code>Event</code> group for the module's [=events=].
+
+An implementation may define <dfn>extension modules</dfn>. These must have a
+[=module name=] that contains a single colon <code>":"</code> character. The
+part before the colon is the prefix; this is typically the same for all
+extension modules specific to a given implementation and should be unique for a
+given implementation. Such modules extend the [=local end schema=] and [=remote
+end schema=] providing additional groups as choices for the defined
+[=commands=] and [=events=].
+
+Issue: Can we put the extension modules into the schema in some generic way.
 
 ## Commands ## {#commands}
 
@@ -109,20 +204,35 @@ result or an error being returned to the [=local end=]. Multiple
 commands can run at the same time, and commands can potentially be
 long-running. As a consequence, commands can finish out-of-order.
 
-Each concrete [=command=] type is defined by:
- - A string <dfn export for=command>name</dfn>, implicit in its
-   definition.
- - A <dfn export for=command>parameter type</dfn>, which is a |map|
-   whose entries are the inputs to the command.
- - A set of [=remote end steps=].
- - A <dfn export for=command>return type</dfn>, which may be null.
+Each [=command=] is defined by:
+- A <dfn export for=command>command id</dfn>. This is a unique integer id
+  provided by the [=local end=] for each command.
+ - A <dfn export for=command>command type</dfn> which is defined by a [=remote
+   end schema=] fragment containing a group. Each such group has two fields:
+    - <code>method</code> which is a string literal of the form <code>[module
+      name].[method name]</code>. This is the <dfn export for=command>command
+      name</dfn>.
+    - <code>params</code> which defines a mapping containing data that to be passed into
+      the command. The populated value of this map is the
+      <dfn export for=command>command parameters</dfn>.
+- A <dfn export for=command>result type</dfn>, which is defined by a [=local
+  end schema=] fragment.
+- A set of [=remote end steps=] which define the actions to take for a command
+  given [=command parameters=] and return an instance of the command [=return
+  type=].
 
-Command [=command/parameter type|parameter types=] and [=command/return
-type|return types=] are defined in this specification using
-[[!RFC8610|CDDL]].
+The <dfn export for=command>set of all command names</dfn> is a set containing
+all the defined [=command names=], including any belonging to [=extension
+modules=].
 
-The following <dfn export for=command>table of commands</dfn> lists the
-available [=command|commands=] by module and name.
+### Table of Commands ### {#table-of-commands}
+
+<div class=non-normative>
+
+<em>This section is non-normative.</em>
+
+The following table of commands lists the available [=command|commands=] by
+module and name.
 
 <table class="simple">
    <tr>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -73,34 +73,125 @@ command/response format of WebDriver, this permits events to stream
...
+on the [=remote end=], and the <dfn>local end schema</dfn> which defines the
+format of messages produced on the [=remote end=] and consumed on the [=local
+end=]
+
+## Schema ## {#protocol-schema}
+
+Issue: Should this be an appendix?
+
+This section gives the initial contents of the [=remote end schema=] and
+[=local end schema=]. These are agumented by the schema fragments defined in
```
miketaylr, line 102: nit: s/agumented/augmented/

index.bs
```diff
@@ -137,55 +241,45 @@ available [=command|commands=] by module and name.
...
+      [=invalid argument=]. Otherwise let <var>command</var> be the value of
+      the <code>method</code> property of |data|.
 
-   1. [=Assert=]: |result| [=match a CDDL specification|matches the CDDL
-      specification=] given by |command|'s [=command/return type=].
+   1. If |command| is not in the [=set of all command names=], return an
+      [=Error=] with [=error code=] [=unknown command=].
 
-   1. Return [=success=] with data |result|.
+   1. Return an [=Error=] with [=error code=] [=invalid argument=].
```
miketaylr, line 282: Is this last step correct? (It could be a default "return an error", assuming there are other situations for returning an invalid args error different from step 2).

Also in step 2, how happens with `command`? Should that get returned? Or something else?

jgraham: I don't think I understand the question. The basic structure of this algorithm (with slightly different numbering) is:

1. Try to match the input against the schema.
2. If the match succeeds, then run the remote end steps for the command that matched and return that result.
3. [In all further steps it must be true that the match failed]
4. Check if we got something that looks like a map containing a method name at all and if not return invalid argument error
5. If we have a method name and it's not a known method return invalid command error
6. Otherwise we have a valid method name but the schema didn't match so return an invalid argument error.

miketaylr: > Otherwise we have a valid method name but the schema didn't match so return an invalid argument error.

Got it, I think step 6 is what I was trying to describe as "default return an error", so forgive the noise.

index.bs
```diff
@@ -137,55 +241,45 @@ available [=command|commands=] by module and name.
...
+      data.
 
-   1. Let |result| be the result of [=trying=] to run the [=remote end
-      steps=] for |command| with |parameters|.
+   1. Otherwise there is no match. If |data| isn't a map or is a map without a
+      property named </code>method</code> return an [=Error=] with error code
+      [=invalid argument=]. Otherwise let <var>command</var> be the value of
+      the <code>method</code> property of |data|.
 
-   1. [=Assert=]: |result| [=match a CDDL specification|matches the CDDL
```
bwalderman, line 185: Why not keep this assert?

jgraham: Added this back in a slightly different way.

index.bs
```diff
@@ -73,34 +73,125 @@ command/response format of WebDriver, this permits events to stream
...
 
-Note: These definitions do not form any normative requirements, rather
-they are used elsewhere as part of such requirements.
+CommandResponse = (
+  Error //
+  CommandResult
+)
+
+Error = (
+  id: unit,
```
christian-bromann, line 134:
```suggestion
  id: uint,
```

Also wouldn't it make sense to move this into the `CommandResponse` group as we would always expect an `id` to a response, no?

index.bs
```diff
@@ -109,20 +200,33 @@ result or an error being returned to the [=local end=]. Multiple
...
- - A <dfn export for=command>return type</dfn>, which may be null.
+Each [=command=] is defined by:
+ - A <dfn export for=command>command type</dfn> which is defined by a [=remote
+   end schema=] fragment containing a group. Each such group has two fields:
+    - <code>method</code> which is a string literal of the form <code>[module
+      name].[method name]</code>. This is the <dfn export for=command>command
+      name</dfn>.
+    - <code>params</code> which defines a mapping containing data that to be passed into
+      the command. The populated value of this map is the
+      <dfn export for=command>command parameters</dfn>.
```
christian-bromann, line 211: Does this also need an `id` so that the local end can connect a response message to it?

index.bs
jgraham: Good point.

index.bs
jgraham: Yes, it does. It occurs to me that there are some details to work out here; I'm going to create an issue.

index.bs
```diff
@@ -73,34 +73,129 @@ command/response format of WebDriver, this permits events to stream
...
+Modules which contain [=commands=] define [=remote end schema=]
+fragments. These provide choices in the <code>CommandData</code> group for the
+module's [=commands=],and can also define additional schema properties. They
+can also define [=local end schema=] fragments that provide additional choices
+in the <code>ResultData</code> group for the results of commands in the module.
+
+Modules which contain events define [=local end schema=] fragments that are
+choices in the <code>Event</code> group for the module's [=events=].
+
+An implementation may define <dfn>extension modules</dfn>. These must have a
```
foolip, line 189: How about extension commands? Does the setup we have allow for that? (I ask because of extension specs, not vendor-specific stuff.)

index.bs
```diff
@@ -137,55 +253,53 @@ available [=command|commands=] by module and name.
...
 An <dfn export>event</dfn> is a notification, sent by the [=remote
 end=] to the [=local end=], signaling that something of interest has
 occurred on the [=remote end=].
 
-Each concrete [=event=] type is defined by:
- - A string <dfn export for=event>name</dfn>, implicit in its
-   definition.
- - A <dfn export for=event>parameter type</dfn>, which is a |map| whose
-   entries are the details of the event that has occurred.
+ - A <dfn export for=event>event type</dfn> which is defined by a [=local
```
foolip, line 264:
```suggestion
 - An <dfn export for=event>event type</dfn> is defined by a [=local
```

index.bs
```diff
@@ -137,55 +253,53 @@ available [=command|commands=] by module and name.
...
 
-   1. Let |module name| be the bytes in |qualified command name|
-      preceeding the U+002E FULL STOP character.
+<div algorithm>
+To <dfn>process a command</dfn> given |data|:
 
-   1. Let |command name| be the bytes in |qualified command name|
-      following the U+002E FULL STOP character.
+   1. Match |data| against the [=remote end schema=]. If this results in a
+      match, run the following inner steps:
```
foolip, line 281:
```suggestion
      match:
```

https://infra.spec.whatwg.org/#split-on-commas is an example of this terser style. "If something, then:" is also common, if you'd like that better.

index.bs
```diff
@@ -137,55 +253,53 @@ available [=command|commands=] by module and name.
...
 
-   1. Let |command| be the command listed in the [=table of commands=]
-      whose module is |module name| and whose name is |command name|.
+      1. Let |result| be the result of running the [=remote end steps=]
+         for the command with the [=command name=] equal to the
```
foolip, line 284:
```suggestion
         for the command with [=command name=] equal to the
```

The "the" felt a bit off, feel free to reject if my English ist nicht gut.

index.bs
```diff
@@ -137,55 +253,53 @@ available [=command|commands=] by module and name.
...
-   1. If |parameters| does not [=match a CDDL specification|match the
-      CDDL specification=] given by |command|'s [=command/parameter
-      type=], then return an [=Error=] with [=error code=] [=invalid
-      argument=].
+      1. Return <var>result</var>.
 
-   1. Let |result| be the result of [=trying=] to run the [=remote end
-      steps=] for |command| with |parameters|.
+   1. Otherwise there is no match. If |data| isn't a map or is a map without a
+      property named </code>method</code> return an [=Error=] with error code
```
foolip, line 295: The wrong Error (https://w3c.github.io/webdriver/#dfn-errors) ends up linked here. Did you mean to refer to `Error` in the CDDL schema?

index.bs
```diff
@@ -137,55 +253,53 @@ available [=command|commands=] by module and name.
...
+      [=invalid argument=]. Otherwise let <var>command</var> be the value of
+      the <code>method</code> property of |data|.
 
-   1. [=Assert=]: |result| [=match a CDDL specification|matches the CDDL
-      specification=] given by |command|'s [=command/return type=].
+   1. If |command| is not in the [=set of all command names=], return an
+      [=Error=] with [=error code=] [=unknown command=].
 
-   1. Return [=success=] with data |result|.
+   1. Return an [=Error=] with [=error code=] [=invalid argument=].
```
foolip, line 302: This bit is not new, but where is the ID of a malformed command extracted and sent back? Doesn't need to be fixed in this PR, but if it is missing an issue for that would be good.

index.bs
```diff
@@ -73,34 +73,129 @@ command/response format of WebDriver, this permits events to stream
...
+
+Issue: Should this be an appendix?
+
+This section gives the initial contents of the [=remote end schema=] and
+[=local end schema=]. These are augmented by the schema fragments defined in
+the remainder of the specification.
+
+[=Remote end schema=]
+
+```
```
foolip, line 107:
```suggestion
```cddl
```

I suspect my suggestions won't work, but can we then go with `<pre class=cddl>` like we do for Web IDL, so we can give these boxes some styling?

foolip (commented):
Slowpoke is sorry, review is here.

index.bs
```diff
@@ -73,34 +73,129 @@ command/response format of WebDriver, this permits events to stream
...
 ==============
 
 This section defines the basic concepts of the WebDriver BiDi
 protocol. These terms are distinct from their representation at the
 <a href=#transport>transport</a> layer.
 
-This specification uses Concise Data Definition Language (CDDL)
-[[!RFC8610]] to describe the format of messages transmitted between the
-[=local end=] and [=remote end=].
+The protocol is defined using [[!RFC8610|CDDL]] schema. For the convenience of
```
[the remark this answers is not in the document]
jgraham: I guess we should use "definition" rather than "schema".

index.bs
jgraham: It doesn't work unless pygments supports the type. There are a couple of projects to add CDDL support to pygments, but nothing in the base package. If we found one we like we could perhaps get bikeshed to include it.

index.bs
```diff
@@ -73,34 +73,129 @@ command/response format of WebDriver, this permits events to stream
...
+[=Remote end schema=]
+
+```
+Command = {
+  id: uint,
+  CommandData,
+}
+
+CommandData = (
+  SessionCommand
```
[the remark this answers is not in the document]
jgraham: Well you can do something like 
```
CommandData //= (
  ExtensionCommand
)
```

to add additional choices. So I think this is at least somewhat possible.

index.bs
```diff
@@ -73,34 +73,129 @@ command/response format of WebDriver, this permits events to stream
...
+
+Error = (
+  id: uint,
+  error: (
+    "unknown error" /
+    "unknown method" /
+    "invalid argument"
+  ),
+  message: text,
+  stacktrace: text,
```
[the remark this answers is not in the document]
jgraham: I was basing it on what we provide in WebDriver 1.0 rather than on JSON-RPC.

index.bs
```diff
@@ -73,34 +73,129 @@ command/response format of WebDriver, this permits events to stream
...
-[=events=] pertaining to a certain aspect of the user agent. For
-example, a module might contain functionality for inspecting and
+Each <dfn export for=module>module</dfn> represents a collection of related
+[=commands=] and [=events=] pertaining to a certain aspect of the user
+agent. For example, a module might contain functionality for inspecting and
 manipulating the DOM, or for script execution.
 
+Each module has a <dfn>module name</dfn> which is a string literal. The
+[=command name=] and [=event name=] for commands and events defined in the
+module start with the [=module name=] followed by a period <code>"."</code>.
```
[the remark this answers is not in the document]
jgraham: We did the opposite for WebDriver 1, but I tend to agree with you.

index.bs
jgraham: The point is supposed to be that extension modules can contain extension commands.

index.bs
```diff
@@ -73,34 +73,129 @@ command/response format of WebDriver, this permits events to stream
...
+
+An implementation may define <dfn>extension modules</dfn>. These must have a
+[=module name=] that contains a single colon <code>":"</code> character. The
+part before the colon is the prefix; this is typically the same for all
+extension modules specific to a given implementation and should be unique for a
+given implementation. Such modules extend the [=local end schema=] and [=remote
+end schema=] providing additional groups as choices for the defined
+[=commands=] and [=events=].
+
+Issue: Can we put the extension modules into the schema in some generic way.
```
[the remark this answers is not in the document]
jgraham: I thinkin practice using the facility in CDDL for extending existing choices should be enough. When I wrote this I was wondering if there should be some option like `ExtensionCommmand = (*text: any)`, but I don't think it's useful.

index.bs
```diff
@@ -109,20 +204,41 @@ result or an error being returned to the [=local end=]. Multiple
...
- - A set of [=remote end steps=].
- - A <dfn export for=command>return type</dfn>, which may be null.
+Each [=command=] is defined by:
+- A <dfn export for=command>command id</dfn>. This is an integer id
+  provided by the [=local end=] for each command. The value of the id is not
+  used by the [=remote end=] and can be regarded as opaque.
+
+  Note: In particular nothing guarantees that the [=command id=] is unique
+  for a given session. For example, a [=local end=] that sends only one command
+  at a time could give each the same command id.
```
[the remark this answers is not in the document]
jgraham: There is a processing model for the server which doesn't use the command id (it's not passed in to the remote end steps). We historically haven't specified anything for clients and I think that's the right approach to continue (basically the server has to handle whatever).

I can update the example to make it more of an edge case if you think that will make things clearer.

index.bs
jgraham: No, this is the Error specification type defined in WebDriver rather than the `Error` CDDL definition.

index.bs
jgraham: It's handled in the caller i.e. `handle an incoming message`.

index.bs
```diff
@@ -73,34 +73,127 @@ command/response format of WebDriver, this permits events to stream
...
+  ResponseData,
+)
+
+ResponseData = (
+  Error //
+  CommandResult
+)
+
+Error = (
+  id: uint,
```
bwalderman, line 139: Nit: Looks like `id` was moved to CommandResponse but not removed from Error here.

index.bs
```diff
@@ -109,20 +202,41 @@ result or an error being returned to the [=local end=]. Multiple
...
-Each concrete [=command=] type is defined by:
- - A string <dfn export for=command>name</dfn>, implicit in its
-   definition.
- - A <dfn export for=command>parameter type</dfn>, which is a |map|
-   whose entries are the inputs to the command.
- - A set of [=remote end steps=].
- - A <dfn export for=command>return type</dfn>, which may be null.
+Each [=command=] is defined by:
+- A <dfn export for=command>command id</dfn>. This is an integer id
+  provided by the [=local end=] for each command. The value of the id is not
```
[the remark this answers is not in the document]
bwalderman: Thinking some more about this, I'm not sure the command ID should be mentioned here at all. It's a transport-layer concept that has no bearing on the definition of the command itself. Maybe the definition and requirements for the command ID should go in the transport section instead?

bwalderman (changes requested):

index.bs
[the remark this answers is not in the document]
foolip: That sounds good, yeah. Here we'd just say "using CDDL" and elsewhere we have "CDDL definition" of things.

index.bs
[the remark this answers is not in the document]
foolip: I see, looks like this is called a socket/plug mechanism in CDDL: https://tools.ietf.org/html/rfc8610#section-3.9

That's promising, let's hope it pans out!

index.bs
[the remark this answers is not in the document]
foolip: Sure, we won't be using JSON-RPC error objects, but this is doing more than refactoring. If you mention the addition in the eventual commit message, that's OK I think.

index.bs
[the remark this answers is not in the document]
foolip: As found in the other thread, hopefully https://tools.ietf.org/html/rfc8610#section-3.9 will solve our needs.

index.bs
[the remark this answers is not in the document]
foolip: I think making the example less realistic isn't necessarily an improvement. How about a `<!-- TODO -->` to link the processing model later?

index.bs
foolip: In other words, we're not allowing for extension commands or extension events in modules that are defined by this specification?

I don't particularly mind at this point because we're a long way from adding anything vendor-specific. It seems to me we'd regret it like all other prefixing. Can we maybe have an open issue and a link to in the spec, to steer the first implementer tempted to do something towards that issue, where we can then work it out with a real case to solve?

index.bs
foolip: Aha. In that case, shouldn't it be lowercased?

index.bs
foolip: Do you mean step 4 of https://w3c.github.io/webdriver-bidi/#handle-an-incoming-message? That doesn't use the ID if it was there. I'll file an issue.

foolip: https://github.com/w3c/webdriver-bidi/issues/55

index.bs
foolip: Would you mind `<pre class=cddl>` which is what we do for Web IDL? For it to be syntax highlighted we'd need to patch Bikeshed, but we could at least draw a box around it with just that.

index.bs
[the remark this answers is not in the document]
foolip: Let's continue this dicussion in https://github.com/w3c/webdriver-bidi/pull/50#pullrequestreview-489039572.

index.bs
```diff
@@ -73,34 +73,127 @@ command/response format of WebDriver, this permits events to stream
...
 ==============
 
 This section defines the basic concepts of the WebDriver BiDi
 protocol. These terms are distinct from their representation at the
 <a href=#transport>transport</a> layer.
 
-This specification uses Concise Data Definition Language (CDDL)
-[[!RFC8610]] to describe the format of messages transmitted between the
-[=local end=] and [=remote end=].
+The protocol is defined using a [[!RFC8610|CDDL]] definition. For the
```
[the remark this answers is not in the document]
foolip: This came out of https://github.com/w3c/webdriver-bidi/pull/50#discussion_r486981187.

There are a few instances of "CDDL definitions" in that RFC as well, does that give us cover? The downside of "CDDL specification" is that it's so easily interpreted as referring to https://tools.ietf.org/html/rfc8610 itself. Perhaps not a problem in practice, but if "definition" isn't exactly wrong, then I'd slightly prefer it.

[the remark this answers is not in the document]
jgraham: Yeah, I'd also prefer not to write "CDDL specification" for the same reason.

index.bs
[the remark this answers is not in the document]
jgraham: I don't understand the suggestion; we have a processing model and it doesn't use the command id. The point of the note is just to set reader expectations. Can we take additional improvements to the wording here, if any, to a followup PR.

index.bs
jgraham: Modules are mostly a spec-organisation convenience; I think the only way they really leak into the protocol is in the way we enable events. So I don't think there's a *lot* of user-facing difference between allowing a `chrome:runtime` module with a `doSomethingChromeish` command and having the `runtime` module be extended with a `chrome:doSomethingChromeish` command. But I think organisationally it's clearer that everything in that module is vendor specific and it means that you always need to explicitly opt-in to the vendor specfic things; you can just subscribe to all `runtime` events and start depending on the browser-specific events you get back.

For WebDriver/HTTP we already have vendor extensions and it's not totally unproblematic, but it's nowhere near as problematic as e.g. prefixes in CSS because mostly for authors it's abstracted away by libraries and there's only a few of those, and typically people writing automation control both ends of the pipeline i.e. they're writing the tests and picking the browsers they run in. Whilst I *certainly* think we want the base spec to be good enough for all common use cases to be cross-browser, I also think we're going to end up with some non-trivial amount of vendor-specific stuff (much of which might be a candidate for later standardisation).

index.bs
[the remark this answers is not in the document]
jgraham: The requirements are already there. I've changed things so that this section is only stating facts about the command id; I couldn't see anywhere in the transport section that would make it clearer.

index.bs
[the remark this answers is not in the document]
jgraham: This changed in the end anyway.

index.bs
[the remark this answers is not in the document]
bwalderman: Makes sense. I agree "specification" is easier to confuse with the spec itself so "definition" seems better.

index.bs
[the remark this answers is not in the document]
bwalderman: Thanks. The new description is more clear.

bwalderman (approved):
LGTM

PR #51, opened 2020-09-02
Add initial support for subscribing to events.

jgraham:
This adds session.subscribe and session.unsubscribe commands to
subscribe to specified events for given browsing contexts (realms are
not yet supported).

The current semantics are that if you subscribe to events for a
specified browsing context you also get all events from descendent
browsing contexts unless you specifically unsubscribe in those scopes.

This only allows a single list of events/contexts per message; the use
cases for allowing multiple lists per message with different events in
dfferent browsing contexts seems less clear and should probably be
solved with a general batching mecahnism.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/51.html" title="Last updated on Oct 2, 2020, 10:40 AM UTC (12eefe9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/51/db6776c...12eefe9.html" title="Last updated on Oct 2, 2020, 10:40 AM UTC (12eefe9)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -662,14 +662,14 @@ with parameters |session| and |capabilities| is:
 
 </div>
 
-Common Data Types {#data_types}
+Common Data Types {#data-types}
 ===============================
 
 This section defines data types which are common to many modules.
 
 ## Browsing Context ## {#data_types-browsing_context}
 
-[= remote end schema=] and [= local end schema=]
+[=remote end schema=] and [=local end schema=]
 ```
 BrowsingContext = text;
 ```
@@ -679,9 +679,9 @@ which is a string uniquely identifying that browsing context. For browsing
 contexts with an associated WebDriver [=window handle=] the [=/browsing context
 id=] must be the same as the [=window handle=].
 
-## Realm ## {#data_types-realm}
+## Realm ## {#data-types-realm}
 
-[= remote end schema=] and [= local end schema=]
+[=remote end schema=] and [=local end schema=]
 ```
 Realm = text;
 ```
--- a/index.bs
+++ b/index.bs
@@ -725,7 +725,7 @@ To <dfn lt="updating the event map">update the event map</dfn>, given
        the union of |event names| and the result of [=trying=] to [=obtain a set
        of event names=] with |name|.
 
-    1. If the |list of contexts| is null, run the following substeps:
+    1. If the |list of contexts| is null:
 
       1. If |enabled| is true, for each |event name| in |event names|,
          append |event name| to |global event set|. Otherwise for for each
@@ -736,7 +736,7 @@ To <dfn lt="updating the event map">update the event map</dfn>, given
 
     1. Let |targets| be an empty list.
 
-    1. For each entry |context id| in the |list of contexts|, run the following substeps:
+    1. For each entry |context id| in the |list of contexts|:
 
       1. If there is no browsing context with [=browsing context id=] |context id| return
          [=error=] with [=error code=] [=invalid argument=]
@@ -748,9 +748,9 @@ To <dfn lt="updating the event map">update the event map</dfn>, given
       1. Get the entry from the |event map| for |context| and append it to
          |targets|.
 
-    1. For each |target| in |targets|, run the following steps:
+    1. For each |target| in |targets|:
 
-      1. For each |event name| in |event names|, run the following steps:
+      1. For each |event name| in |event names|:
 
         1. Set |target|[|event name|] to |enabled|.
 
--- a/index.bs
+++ b/index.bs
@@ -37,6 +37,7 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
     text: additional capability deserialization algorithm; url: dfn-additional-capability-deserialization-algorithm
     text: additional WebDriver capability; url: dfn-additional-webdriver-capability
     text: capability name; url: dfn-capability-name
+    text: current session; url: dfn-current-session
     text: endpoint node; url: dfn-endpoint-node
     text: error code; url: dfn-error-code
     text: error; url: dfn-errors
@@ -273,8 +274,8 @@ occurred on the [=remote end=].
     - <code>params</code> which defines a mapping containing event data. The
       populated value of this map is the <dfn export for=command>event
       parameters</code>.
- - A [=remote end event trigger=] which define when the event is triggered
-   and steps to construct the [=event type=] data.
+ - A <dfn export>remote end event trigger</dfn> which define when the event is
+   triggered and steps to construct the [=event type=] data.
 
 A [=/session=] has a <dfn export for=event>global event set</dfn> which is a set
 containing containing the event names for events that are enabled for all
@@ -288,15 +289,16 @@ disabled for a given browsing context.
 
 <div algorithm>
 
-To determine if an <dfn export>event is enabled</dfn> given an |event name| and
-|browsing context|:
+To determine if an <dfn export>event is enabled</dfn> given |session|,
+|event name| and |browsing context|:
 
   1. While |browsing context| is not null:
 
-    1. If [=browsing context event map=] contains an entry for |browsing
-       context|, let |browsing context events| be [=browsing context event
-       map=][|browsing context|].  Otherwise let |browsing context events| be
-       null.
+    1. Let |event map| be the [=browsing context event map=] for |session|.
+
+    1. If |event map| contains an entry for |browsing context|, let |browsing
+       context events| be |event map|[|browsing context|].  Otherwise let
+       |browsing context events| be null.
 
     1. If |browsing context events| is not null, and |browsing context events|
        contains an entry for |event name| return |browsing context
@@ -305,7 +307,8 @@ To determine if an <dfn export>event is enabled</dfn> given an |event name| and
     1. Let |browsing context| be the [=parent browsing context=] of |browsing
        context|, if it has one, or null otherwise.
 
-  1. If the [=global event set=] contains |event name| return true.
+  1. If the [=global event set=] for |session| contains |event name| return
+     true.
 
   1. Return false.
 
@@ -709,8 +712,12 @@ SessionResult = (StatusResult)
 
 <div algorithm>
 
-To <dfn lt="updating the event map">update the event map</dfn>, given a |list of
-event names|, a |list of contexts|, and a |enabled|:
+To <dfn lt="updating the event map">update the event map</dfn>, given
+|session|, |list of event names|, |list of contexts|, and |enabled|:
+
+  1. Let |global event set| be the [=global event set=] for |session|.
+
+  1. Let |event map| be the [=browsing context event map=] for |session|.
 
   1. Let |event names| be an empty set.
 
@@ -721,9 +728,9 @@ event names|, a |list of contexts|, and a |enabled|:
     1. If the |list of contexts| is null, run the following substeps:
 
       1. If |enabled| is true, for each |event name| in |event names|,
-         append |event name| to the [=global event set=]. Otherwise for for each
-         |event name| in |event names|, if the [=global event set=] contains
-         |event name|, remote |event name| from the [=global event set=].
+         append |event name| to |global event set|. Otherwise for for each
+         |event name| in |event names|, if the |global event set| contains
+         |event name|, remote |event name| from the |global event set|.
 
       1. Return
 
@@ -735,11 +742,11 @@ event names|, a |list of contexts|, and a |enabled|:
          [=error=] with [=error code=] [=invalid argument=]
 
       1. Let |context| be the browsing context with id |context id|. If the
-         [=browsing context event map=] does not contain an entry for |context|,
+         |event map| does not contain an entry for |context|,
          set the value of the entry for |context| to a new empty map.
 
-      1. Get the entry from the [=browsing context event map=] for |context|
-         and append it to |targets|.
+      1. Get the entry from the |event map| for |context| and append it to
+         |targets|.
 
     1. For each |target| in |targets|, run the following steps:
 
@@ -833,8 +840,8 @@ The [=remote end steps=] with |command parameters| are:
     1. Let the |list of contexts| be the value of the <code>contexts</code>
        field of |command parameters| if it is present or null if it isn't.
 
-    1. Return the result of [=updating the event map=] with |list of event
-       names|, |list of contexts| and enabled true.
+    1. Return the result of [=updating the event map=] with [=current session=],
+       |list of event names|, |list of contexts| and enabled true.
 </div>
 
 ### Unsubscribe ### {#command-session-unsubscribe}
@@ -870,6 +877,6 @@ The [=remote end steps=] with |command parameters| are:
     1. Let the |list of contexts| be the value of the <code>contexts</code>
        field of |command parameters| if it is present or null if it isn't.
 
-    1. Return the result of [=updating the event map=] with |list of event
-       names|, |list of contexts| and enabled false.
+    1. Return the result of [=updating the event map=] with [=current session=],
+       |list of event names|, |list of contexts| and enabled false.
 </div>
--- a/index.bs
+++ b/index.bs
@@ -273,41 +273,39 @@ occurred on the [=remote end=].
     - <code>params</code> which defines a mapping containing event data. The
       populated value of this map is the <dfn export for=command>event
       parameters</code>.
- - A set of [=remote end steps=] which define how to construct the [=event
-   type=] data.
- - An <dfn export for=event>event trigger</dfn> which is a condition on which
-   the [=remote end steps=] for the event are invoked.
+ - A [=remote end event trigger=] which define when the event is triggered
+   and steps to construct the [=event type=] data.
 
 A [=/session=] has a <dfn export for=event>global event set</dfn> which is a set
-containing containing the [=event types=] that are enabled for all browsing
-contexts. This initially contains events that are <dfn export for=event>in the
-default event set</dfn>.
+containing containing the event names for events that are enabled for all
+browsing contexts. This initially contains the [=event name=] for events that
+are <dfn export for=event>in the default event set</dfn>.
 
 A [=/session=] has a <dfn export for=event>browsing context event map</dfn>,
-which is a map with [=/browsing context=] keys and values that are maps
-from an [=event type=] to a boolean indicating whether the specified event type
-is enabled or disabled for a given browsing context.
+which is a map with [=/browsing context=] keys and values that are maps from an
+[=event name=] to a boolean indicating whether the specified event is enabled or
+disabled for a given browsing context.
 
 <div algorithm>
 
-To determine if an <dfn export>event is enabled</dfn> given an |event type| and
+To determine if an <dfn export>event is enabled</dfn> given an |event name| and
 |browsing context|:
 
   1. While |browsing context| is not null:
 
     1. If [=browsing context event map=] contains an entry for |browsing
-       context|, let |browsing context events| be the result of getting the value of an
-       entry with key |browsing context| from [=browsing context event map=],
-       otherwise let |browsing context events| be <code>undefined</code>.
+       context|, let |browsing context events| be [=browsing context event
+       map=][|browsing context|].  Otherwise let |browsing context events| be
+       null.
 
-    1. If |browsing context events| is not <code>undefined</code>, return the
-       result of getting the value of the entry with key |event type| from
-       |browsing context events|.
+    1. If |browsing context events| is not null, and |browsing context events|
+       contains an entry for |event name| return |browsing context
+       events|[|event name|].
 
     1. Let |browsing context| be the [=parent browsing context=] of |browsing
        context|, if it has one, or null otherwise.
 
-  1. If the [=global event set=] contains |event type| return true.
+  1. If the [=global event set=] contains |event name| return true.
 
   1. Return false.
 
@@ -315,22 +313,23 @@ To determine if an <dfn export>event is enabled</dfn> given an |event type| and
 
 <div algorithm>
 
-To <dfn>obtain a set of events from a name</dfn> given an |event name|:
+To <dfn>obtain a set of event names</dfn> given an |name|:
 
   1. Let |events| be an empty set.
 
-  1. If |event name| contains a U+002E (period):
+  1. If |name| contains a U+002E (period):
 
-    1. If |event name| is the [=event name=] for an event, append that event to
-       |events| and return [=success=] with data |events|.
+    1. If |name| is the [=event name=] for an event, append |name| to |events|
+       and return [=success=] with data |events|.
 
     1. Return an [=error=] with [=error code=] [=Invalid Argument=]
 
-  1. Otherwise |event name| is interpreted as representing all the events in a
-     module. If |event name| is not a [=module name=] return an [=error=] with
+  1. Otherwise |name| is interpreted as representing all the events in a
+     module. If |name| is not a [=module name=] return an [=error=] with
      [=error code=] [=Invalid Argument=].
 
-  1. Add each [=event=] in the module with name |event name| to |events|.
+  1. Append the [=event name=] for each [=event=] in the module with name |name| to
+     |events|.
 
   1. Return [=success=] with data |events|.
 </div>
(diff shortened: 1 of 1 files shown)


jgraham:
This is super sketchy, and probably has some obvious errors, but I wanted to get the PR up early for feedback.

index.bs
```diff
@@ -580,8 +648,35 @@ with parameters |session| and |capabilities| is:
 
 </div>
 
+Common Data Types {#data_types}
```
foolip, line 651:
```suggestion
Common Data Types {#data-types}
```

index.bs
```diff
@@ -580,8 +648,35 @@ with parameters |session| and |capabilities| is:
...
+===============================
+
+This section defines data types which are common to many modules.
+
+## Browsing Context ## {#data_types-browsing_context}
```
foolip, line 656:
```suggestion
## Browsing Context ## {#data-types-browsing-context}
```

I guess this case is why you went with underscores. But generated IDs for similarly hierarchical concepts don't have underscores.

index.bs
```diff
@@ -580,8 +648,35 @@ with parameters |session| and |capabilities| is:
...
+```
+BrowsingContext = text;
+```
+
+Each [=/browsing context=] has an associated <dfn export>browsing context id</dfn>,
+which is a string uniquely identifying that browsing context. For browsing
+contexts with an associated WebDriver [=window handle=] the [=/browsing context
+id=] must be the same as the [=window handle=].
+
+## Realm ## {#data_types-realm}
```
foolip, line 668:
```suggestion
## Realm ## {#data-types-realm}
```

index.bs
```diff
@@ -580,8 +648,35 @@ with parameters |session| and |capabilities| is:
 
 </div>
 
+Common Data Types {#data_types}
+===============================
+
+This section defines data types which are common to many modules.
```
foolip, line 654: These things don't seem like data types to me, they're types of targets, an abstract concept which is yet to be fleshed out. Since these things exist outside of this specification, maybe they can go in https://w3c.github.io/webdriver-bidi/#protocol, similar to what we've done with session? They do have a representation that we'll need to reuse, of course, maybe put that in the CDDL schema?

index.bs
```diff
@@ -593,8 +688,8 @@ events for monitoring the status of the remote end.
 [=remote end schema=]
 
 ```
-
-SessionCommand = (StatusCommand)
+SessionCommand = (StatusCommand //
+                  EnableEventCommand)
```
foolip, line 692:
```suggestion
                  SubscribeCommand)
```

Right?

index.bs
```diff
@@ -606,6 +701,39 @@ SessionResult = (StatusResult)
 
 ```
 
+<div algorithm>
+
+To <dfn lt="updating the event map">update the event map</dfn>, given a |list of
```
foolip, line 706: Having "list" in the variable name feels new and scary to me :) https://fetch.spec.whatwg.org/#convert-header-names-to-a-sorted-lowercase-set is more in the style I'm accustomed to.

index.bs
```diff
@@ -606,6 +701,39 @@ SessionResult = (StatusResult)
...
+
+      1. If there is no browsing context with [=browsing context id=] |context id| return
+         [=error=] with [=error code=] [=invalid argument=]
+
+      1. Let |context| be the browsing context with id |context id|. If the
+         [=browsing context event map=] does not contain an entry for |context|,
+         set the value of the entry for |context| to a new empty map.
+
+      1. Get the entry from the [=browsing context event map=] for |context|
+         and append it to |targets|.
```
foolip, line 727: At this point *targets* hasn't been initialized. Crash!

index.bs
```diff
@@ -606,6 +701,39 @@ SessionResult = (StatusResult)
...
 ```
 
+<div algorithm>
+
+To <dfn lt="updating the event map">update the event map</dfn>, given a |list of
+event names|, a |list of contexts|, and a |target status|:
+
+  1. Let |events| be an empty set.
+
+    1. For each entry |event name| in the |list of event names|,
```
foolip, line 711: The indentation of this and the following is one level too much, resulting in the algorithm having just one main step :)

index.bs
```diff
@@ -606,6 +701,39 @@ SessionResult = (StatusResult)
...
+       let |events| be the union of |events| and the result of [=trying=] to
+       [=obtain a set of events from a name=] with |event name|.
+
+    1. If the |list of contexts| is null, let |targets| be a set containing the [=global
+       event set=]. Otherwise for each entry |context id| in the |list of
+       contexts|, run the following substeps:
```
foolip, line 717:
```suggestion
       contexts|:
```

index.bs
```diff
@@ -651,3 +779,81 @@ The [=remote end steps=] are:
...
+The [=remote end steps=] with |command parameters| are:
+
+    1. Let the |list of event names| be the value of the <code>events</code> field of
+       |command parameters|.
+
+    1. Let the |list of contexts| be the value of the <code>contexts</code>
+       field of |command parameters| if it is present or null if it isn't.
+
+    1. Return the result of [=updating the event map=] with |list of event
+       names|, |list of contexts| and the target status set to true.
```
foolip, line 822:
```suggestion
       names|, |list of contexts| and |target status| true.
```

An argument isn't quite "set to" something, it's just given.

(Modulo suggested renaming elsewhere.)

index.bs
```diff
@@ -651,3 +779,81 @@ The [=remote end steps=] are:
...
+The [=remote end steps=] with |command parameters| are:
+
+    1. Let the |list of event names| be the value of the <code>events</code> field of
+       |command parameters|.
+
+    1. Let the |list of contexts| be the value of the <code>contexts</code>
+       field of |command parameters| if it is present or null if it isn't.
+
+    1. Return the result of [=updating the event map=] with |list of event
+       names|, |list of contexts| and the target status set to false.
```
foolip, line 858:
```suggestion
       names|, |list of contexts| and |target status| false.
```

(See above.)

index.bs
```diff
@@ -606,6 +701,39 @@ SessionResult = (StatusResult)
 
 ```
 
+<div algorithm>
+
+To <dfn lt="updating the event map">update the event map</dfn>, given a |list of
+event names|, a |list of contexts|, and a |target status|:
```
foolip, line 707: I doesn't affect the normative requirements, but I'm not sure if the *target* in *target status* here refers to a browsing context or realm, or the intended status? I think the latter... Maybe *enable* or *enabled* would work?

index.bs
```diff
@@ -606,6 +701,39 @@ SessionResult = (StatusResult)
...
+
+  1. Let |events| be an empty set.
+
+    1. For each entry |event name| in the |list of event names|,
+       let |events| be the union of |events| and the result of [=trying=] to
+       [=obtain a set of events from a name=] with |event name|.
+
+    1. If the |list of contexts| is null, let |targets| be a set containing the [=global
```
foolip, line 715: This case is if the `contexts` parameter is missing. Why shouldn't it be required, though?

index.bs
```diff
@@ -651,3 +779,81 @@ The [=remote end steps=] are:
...
+   <dd>
+      ```
+      SubscribeCommand = {
+        method: "session.subscribe",
+        params: SubscribeParameters
+      }
+
+      SubscribeParameters = {
+        events: [text],
+        contexts?: [BrowsingContext],
```
foolip, line 800:
```suggestion
        targets?: [BrowsingContext],
```

I presume this is where we want to end up? To always give different target types separately seems like it would be tedious, but it does depend on what the client code looks like.

index.bs
```diff
@@ -651,3 +779,81 @@ The [=remote end steps=] are:
...
+        * text => any,
```
foolip, line 801: I'm not familiar with CDDL, what does `* text => any` mean?

index.bs
```diff
@@ -606,6 +701,39 @@ SessionResult = (StatusResult)
...
+    1. If the |list of contexts| is null, let |targets| be a set containing the [=global
+       event set=]. Otherwise for each entry |context id| in the |list of
+       contexts|, run the following substeps:
+
+      1. If there is no browsing context with [=browsing context id=] |context id| return
+         [=error=] with [=error code=] [=invalid argument=]
+
+      1. Let |context| be the browsing context with id |context id|. If the
+         [=browsing context event map=] does not contain an entry for |context|,
+         set the value of the entry for |context| to a new empty map.
```
foolip, line 724: I think the "create if not already there" style can be avoided here. Can't we just say that each target (or browsing context) has an event map which is initially an empty map, and never create it?

index.bs
```diff
@@ -606,6 +701,39 @@ SessionResult = (StatusResult)
...
+
+      1. Get the entry from the [=browsing context event map=] for |context|
+         and append it to |targets|.
+
+    1. For each |target| in |targets|, run the following steps:
```
foolip, line 729:
```suggestion
    1. For each |target| in |targets|:
```

index.bs
```diff
@@ -606,6 +701,39 @@ SessionResult = (StatusResult)
...
+
+      1. For each |event| in events, run the following steps:
```
foolip, line 731:
```suggestion
      1. For each |event| in |events|:
```

index.bs
```diff
@@ -606,6 +701,39 @@ SessionResult = (StatusResult)
...
+
+        1. Set the value of |target| for |event| to |target status|.
```
foolip, line 733: Does this work? It depends on what the type of *target* is (unclear) but *event* is a string so I think it doesn't work? Using https://infra.spec.whatwg.org/#map-set might make this clearer.

index.bs
```diff
@@ -651,3 +779,81 @@ The [=remote end steps=] are:
 
 2. Return [=success=] with data |body|
 
+### Subscribe ### {#command-session-subscribe}
+
+The <dfn export for=commands>subscribe</dfn> command enables certain events either
+globally or for a set of browsing contexts
```
foolip, line 785: Having read through **update the event map** I'm not sure that enabling globally really works. Is the intended model to maintain a global event map, or to enable/disable on all targets which currently exist? (A similar question asked elsewhere, sorry...)

index.bs
```diff
@@ -651,3 +779,81 @@ The [=remote end steps=] are:
...
+
+Issue: This needs to be generalised to work with realms too
```
foolip, line 787:
```suggestion
Issue: This needs to be generalized to work with realms too
```

jk

index.bs
```diff
@@ -651,3 +779,81 @@ The [=remote end steps=] are:
...
+   <dd>
+      ```
+        EmptyResult
+      ```
+   </dd>
+</dl>
+
+The [=remote end steps=] with |command parameters| are:
+
+    1. Let the |list of event names| be the value of the <code>events</code> field of
```
foolip, line 815: A bit of `<div algorithm>` somewhere around here would make this nicer.

index.bs
```diff
@@ -651,3 +779,81 @@ The [=remote end steps=] are:
...
```
foolip, line 851: A bit of `<div algorithm>` somewhere around here would make this nicer, too.

foolip (commented):
Lots of comments, but this is heading in a good direction for sure!

index.bs
```diff
@@ -258,6 +262,67 @@ occurred on the [=remote end=].
     - <code>params</code> which defines a mapping containing event data. The
       populated value of this map is the <dfn export for=command>event
       parameters</code>.
+ - A set of [=remote end steps=] which define how to construct the [=event
+   type=] data.
+ - An <dfn export for=event>event trigger</dfn> which is a condition on which
```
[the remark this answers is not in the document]
jgraham: Well I don't see how we can make it procedural unless we actually patch every other spec to have hooks that call into WebDriver (aiui that's basically what happens with CDP). For example we want to fire an event when a page is loaded. So we need somewhere to write "when an event named 'load' is fired at a global object".

index.bs
bwalderman: `*` means zero or more. `text => any` means a key value pair where the key is a text string and the value is any. So `* text => any` at the end of a map means the map can contain other arbitrary keys in addition to the ones that are required.

bwalderman: I think we can just get rid of this instance of it.

index.bs
```diff
@@ -137,55 +245,117 @@ available [=command|commands=] by module and name.
...
+which is a map with [=/browsing context=] keys and values that are maps
+from an [=event type=] to a boolean indicating whether the specified event type
+is enabled or disabled for a given browsing context.
 
-Event [=event/parameter type|parameter types=] are defined in this
-specification using [[!RFC8610|CDDL]].
+<div algorithm>
 
-## Processing Model ## {#processing-model}
+To determine if an <dfn export>event is enabled</dfn> given an |event type| and
```
bwalderman, line 282: Should this be |event name| instead of |event type|?

index.bs
```diff
@@ -547,12 +773,87 @@ The [=remote end steps=] are:
...
+   <dt>Command Type</dt>
+   <dd>
+      ```
+      SubscribeCommand = {
+        method: "session.subscribe",
+        params: SubscribeParameters
+      }
+
+      SubscribeParameters = {
+        events: [text],
```
bwalderman, line 799: Still learning CDDL, but I believe these need a `*` to indicate multiple occurrences, otherwise it is assumed to be exactly one: https://tools.ietf.org/html/rfc8610#section-3.2

bwalderman (commented):
Looking good so far. Excited to see this taking shape. I just left a few minor comments.

index.bs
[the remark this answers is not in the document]
foolip: We don't need other specs to invoke WebDriver BiDi algorithms, just have the appropriate definitions that we can hook into. For the load event example, it might be https://html.spec.whatwg.org/#completely-finish-loading

Whether we have an explicit "event trigger" concept or not, we'll need the same from other specs. I just don't quite see why we need the concept, when we can say things like "When a document completely finishes loading: do our stuff" instead of defining the trigger and some steps separately.

I might be wrong, but if we could leave out "event trigger" until we define an event, it'll be much easier to judge.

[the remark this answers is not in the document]
jgraham: OK, I've merged these concepts, but I still think this should be called a "remote end event trigger" or something, since it's not a set of steps that get directly invoked by this spec but a COMEFROM for hooks in other specs.

index.bs
```diff
@@ -258,6 +262,67 @@ occurred on the [=remote end=].
...
+   type=] data.
+ - An <dfn export for=event>event trigger</dfn> which is a condition on which
+   the [=remote end steps=] for the event are invoked.
+
+A [=/session=] has a <dfn export for=event>global event set</dfn> which is a set
+containing containing the [=event types=] that are enabled for all browsing
+contexts. This initially contains events that are <dfn export for=event>in the
+default event set</dfn>.
+
+A [=/session=] has a <dfn export for=event>browsing context event map</dfn>,
```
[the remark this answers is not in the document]
jgraham: We haven't defined `target` yet so I think I"ll defer this change.

index.bs
jgraham: The intent is that the events are automatically enabled for all browsing contexts, and will automatically be enabled for contexts that are newly created.

index.bs
```diff
@@ -258,6 +262,67 @@ occurred on the [=remote end=].
     - <code>params</code> which defines a mapping containing event data. The
       populated value of this map is the <dfn export for=command>event
       parameters</code>.
+ - A set of [=remote end steps=] which define how to construct the [=event
+   type=] data.
+ - An <dfn export for=event>event trigger</dfn> which is a condition on which
+   the [=remote end steps=] for the event are invoked.
+
+A [=/session=] has a <dfn export for=event>global event set</dfn> which is a set
```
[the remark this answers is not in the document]
jgraham: That's basically what it's for. I don't know we won't need the pause thing but that seems like a more invasive change in the browser engine and I worry about difficulty of implementation.

index.bs
jgraham: I've updated this and used strings everywhere as you and bwalderman seemed confused by the attempt to use spec-internal objects.

index.bs
```diff
@@ -258,6 +262,67 @@ occurred on the [=remote end=].
...
+       otherwise let |browsing context events| be <code>undefined</code>.
+
+    1. If |browsing context events| is not <code>undefined</code>, return the
+       result of getting the value of the entry with key |event type| from
+       |browsing context events|.
+
+    1. Let |browsing context| be the [=parent browsing context=] of |browsing
+       context|, if it has one, or null otherwise.
+
+  1. If the [=global event set=] contains |event type| return true.
```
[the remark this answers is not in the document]
jgraham: I think I prefer the slightly longer style, but I may change my mind when I'm more used to it.

index.bs
```diff
@@ -258,6 +262,67 @@ occurred on the [=remote end=].
...
+ - An <dfn export for=event>event trigger</dfn> which is a condition on which
+   the [=remote end steps=] for the event are invoked.
+
+A [=/session=] has a <dfn export for=event>global event set</dfn> which is a set
+containing containing the [=event types=] that are enabled for all browsing
+contexts. This initially contains events that are <dfn export for=event>in the
+default event set</dfn>.
+
+A [=/session=] has a <dfn export for=event>browsing context event map</dfn>,
+which is a map with [=/browsing context=] keys and values that are maps
```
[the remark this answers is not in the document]
jgraham: Do you have specific wordning in mind?

index.bs
jgraham: I think that is in the schema. 

I don't think I want this stuff in #protocol because that's basically just the formalism whereas, as you say, these are about the representation in the schema of data types that can be transmitted over the wire. One option is just to put these things into a module like CDP does for everything; modules are going to need a way to define data types in general.

index.bs
jgraham: I think it makes sense to allow events to be enabled globally, and not passing in a list of contexts makes sense as a way to represent that.

index.bs
jgraham: Yeah, probably but I'd rather refactor later once we can actually address realms.

index.bs
jgraham: That means we won't ignore unknown parameters here. I thought that having this everywhere was the plan.

index.bs
jgraham: Hanging custom state off objects we don't control in other specs is *probably*  OK, but I'm not sure it's worth the magic to save a few words (ideally the infra spec would give us better primitives here).

index.bs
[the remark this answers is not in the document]
jgraham: I've asked about how we do this in Gecko, and it turns out the model we have is closer to what's in the spec; we can get a browsingContextCreated event very early in the lifecycle before anything that would cause other events to happen is run. At that point we run the code to enable things that are in a global event list (and do other things that are necessary to ensure that we don't miss anything on startup e.g. in devtools setup the breakpoints that will be applied). AFAICT we don't have any specific concept of pausing here, so I'd be very reluctant to specify something that requires it. AFAICT the main advantage of the pause model is that you could ask the client which events it wants to enable for that specific browsing context, rather than needing a global list. But with the global list model you can enable events you might want for any browsing context and then explicitly disable the ones that you aren't interested in, without requiring a roundtrip to the client in the context opening code. So overall I'd strongly prefer to keep the model in the spec, but I'd expect blink could implement it on top of the pause machinery if necessary.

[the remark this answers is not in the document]
jgraham: s/in the spec/in the PR/g

index.bs
```diff
@@ -269,6 +274,68 @@ occurred on the [=remote end=].
     - <code>params</code> which defines a mapping containing event data. The
       populated value of this map is the <dfn export for=command>event
       parameters</code>.
+ - A <dfn export>remote end event trigger</dfn> which define when the event is
```
mathiasbynens, line 277:
```suggestion
 - A <dfn export>remote end event trigger</dfn> which defines when the event is
```

mathiasbynens (commented):
> The current semantics are that if you subscribe to events for a
> specified browsing context you also get all events from descendent
> browsing contexts unless you specifically unsubscribe in those scopes.

Are `iframe`s considered to be descendent browsing contexts? If so, how do OOPIFs tie in to this? (It seems reasonable to include OOPIFs, but I’d like that decision to be a conscious one, given the implementation challenges it poses.) cc @sadym-chromium

jgraham:
> > The current semantics are that if you subscribe to events for a
> > specified browsing context you also get all events from descendent
> > browsing contexts unless you specifically unsubscribe in those scopes.
> 
> Are `iframe`s considered to be descendent browsing contexts? If so, how do OOPIFs tie in to this? (It seems reasonable to include OOPIFs, but I’d like that decision to be a conscious one, given the implementation challenges it poses.) cc @sadym-chromium

Yes, they are descendent browsing contexts. As I understand it, the way this works in gecko with out of process iframes is we get an event very early in the lifecycle of the browsing context and then pass in a list of things that need to be enabled for that frame so we can load the relevant code and set up event handlers and so on. So I think it's a solvable problem. cc @ochameau

index.bs
```diff
@@ -617,6 +710,52 @@ SessionResult = (StatusResult)
 
 </pre>
 
+<div algorithm>
+
+To <dfn lt="updating the event map">update the event map</dfn>, given
```
bwalderman, line 715: I'm thinking about implementations. For CDP, when a subscribe command is processed, we'll need to actually enable the appropriate CDP domain(s) before updating the event map and completing the command. I'm sure Gecko and WebKit would need to do something similar.

My concern is that the spec as written doesn't guarantee that the underlying platform (CDP, Gecko, WebKit, etc) agrees with the event map before returning from the subscribe command. It might be beneficial to add a step to this algorithm; Something like:

- Take implementation-defined steps to ensure the events listed in `event names` are enabled for the context with id `context id`.

bwalderman (changes requested):

index.bs
jgraham: I don't think this really makes sense spec-wise. I agree that implementations are going to have to do that, but the obvious way to specify this is that you write steps which are triggered by hooks in other specs. Those other steps always run regardless of WebDriver BiDi spec. Then in our even dispatch algorithm we check if the event is enabled given an event name and a browsing context name. 

In a real implementation of course you don't want to process every event and then check at the last moment whether it's enabled for the browsing context, you want to process the minimum possible number of events only for browsing contexts where they will be used, and that probably means updating internal state at this point. But I think trying to work out how to write that down in a spec is complicated and probably not the most useful use of our time. I guess I can add a note.

index.bs
jgraham: Yes. Spec-wise you're going to invoke all events and then filter. Implementations probably won't do that for efficiency reasons, but the models are isomorphic.

index.bs
jgraham: Good call.

bwalderman (approved):

index.bs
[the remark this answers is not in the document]
foolip: Alright, we'll just see how it turns out, given two or three real cases it should be clear which style works best.

index.bs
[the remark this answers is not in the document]
foolip: Thanks for checking what Gecko does. The mechanism in CDP is the `waitForDebuggerOnStart` flag in [`Target.setAutoAttach`](https://chromedevtools.github.io/devtools-protocol/tot/Target/#method-setAutoAttach). It's a very debugger-oriented model, as the name implies, but I suspect there are things one can do other than enable events. It's notable that Puppeteer *doesn't* use this setting, however, which does cause it to miss events. That's how I found out about it, via @sadym-chromium's investigation of [Event Network.loadingFinished is not sent when a frame is loaded from another domain](https://bugs.chromium.org/p/chromium/issues/detail?id=1060080).

All that being said, I'm not familiar enough with the model to have any strong opinions about its advantages over the "global event set". For common use cases I'd imagine the "global event set" to be easy to reason about, so let's go with that, especially given that it's close to what Gecko does.

If we later find a need to do other stuff early in the lifecycle, then maybe we will define a lower-level thing which "global event set" can the redefined on top of. We shall see.

index.bs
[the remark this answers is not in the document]
foolip: Nah, this is mostly a problem when reading the spec in one pass. That's not how most will read it, and it's hard to fend off every misunderstanding regardless of where people start reading. Let's wait for "I'm confused" bug reports.

index.bs
foolip: @jgraham I still stand by these underscore suggestions :)

index.bs
```diff
@@ -591,8 +662,35 @@ with parameters |session| and |capabilities| is:
...
+Each [=/browsing context=] has an associated <dfn export>browsing context id</dfn>,
+which is a string uniquely identifying that browsing context. For browsing
+contexts with an associated WebDriver [=window handle=] the [=/browsing context
+id=] must be the same as the [=window handle=].
+
+## Realm ## {#data_types-realm}
+
+[= remote end schema=] and [= local end schema=]
+```
+Realm = text;
```
foolip, line 686: `Realm` appears to be unused elsewhere. I believe that's known and not an accident, just pointing it out to be sure.

index.bs
foolip: Alright, ultimately this section defined only "browsing context id", "realm id" and the two snippets of CDDL. The two former seem more "mode" than "type", but I agree it's not a clean fit in #protocol either. Let's leave it here until it looks out of place.

index.bs
foolip: Alright, I'm fine leaving this as is.

foolip (approved):
LGTM % remaining suggestions.

PR #54, opened 2020-09-15
chore(.github/deploy): use spec-prod action

sidvishnoi:
I've created a GitHub action ([spec-prod](https://github.com/sidvishnoi/spec-prod/)) that can build Bikeshed (as well as ReSpec) specs, validate them, and deploy the built version to GitHub pages or /TR/. 

I'm hoping we could use a single GitHub action across all W3C specs, rather than every repo creating their own workflows. Presently, I'm sending PRs to a few repositories so we can test it out and fix issues, if any. Please let me know if it is a welcome change.

You can view [sample commits](https://github.com/sidvishnoi/w3c-deploy-test/commits/gh-pages) by this action, and also a [sample deploy run](https://github.com/sidvishnoi/w3c-deploy-test/runs/1117589659).

This PR also enables the action to run as a pull request validator (it presently checks for broken hyperlinks and markup issues). The deployment is done only on push to the default GitHub branch.

~Caveat: The action presently deploys all files to gh-pages (See https://github.com/sidvishnoi/spec-prod/issues/2).~
Edit: The action now deploys only the required files to gh-pages.


jgraham:
I think @foolip might be the most qualified to review this.

.github/workflows/deploy.yml
```diff
@@ -1,21 +1,13 @@
...
-        python-version: 3.8
-    - run: pip install bikeshed
-    - run: bikeshed update
-    - run: bikeshed spec
-    - run: mkdir out && mv index.html out/
-    - uses: peaceiris/actions-gh-pages@v3
-      with:
-        github_token: ${{ secrets.GITHUB_TOKEN }}
-        publish_dir: out
+        GH_PAGES_BRANCH: gh-pages
```
foolip, line 13: Is this really needed, doesn't this variable have a default value?

foolip:
Centralizing spec build stuff into a shared action makes a lot of sense to me.

How many specs are already using this, and what is the reliability of it? Since it installs some dependencies over the network, I assume it sometimes fails with network errors. Are there retries when this happens?

.github/workflows/deploy.yml
sidvishnoi: This variable does two things: enables gh pages deployment as well as specify the branch. Not having a default value as not everyone wants to deploy built version to gh-pages

sidvishnoi:
> How many specs are already using this, and what is the reliability of it?

Presently, [4](https://github.com/search?l=YAML&q=org%3Aw3c+spec-prod&type=Code), all ReSpec based. None are using GitHub pages deployment yet. So, I wanted to start with this spec as first Bikeshed spec if you're comfortable. 

For added safety, it might be GOOD to set `GH_PAGES_BRANCH` to some other branch (and not deploy it to w3c.github.io) and check if things are working fine ([they are in my tests](https://github.com/sidvishnoi/w3c-deploy-test/commits/gh-pages)) and then switch to a real gh-pages branch.

> Since it installs some dependencies over the network, I assume it sometimes fails with network errors. Are there retries when this happens?

It can occasionally fail and there are no retries. But as the whole action is _atomic_, you can re-trigger a run from GitHub Actions UI manually ([docs](https://docs.github.com/en/free-pro-team@latest/actions/managing-workflow-runs/re-running-a-workflow)).

.github/workflows/deploy.yml
foolip: I see, that makes sense!

foolip:
@sidvishnoi I'm happy to try this experimentally and report back problems if you think you'll be around to investigate. I think it's inevitable that some retry mechanism will be needed, nothing that depends on network services ends up being reliable at scale, but yet reliable enough that people assume it will work and have no idea what to do when it doesn't. When we do see these errors, would you be able to add mitigations as needed?

sidvishnoi:
> I'm happy to try this experimentally

I think it would be good to create another workflow and and set the `GH_PAGES_BRANCH` to something else. During the testing phase, we can keep both workflows and let them run independently. I will update the PR if you prefer this approach instead.

> it's inevitable that some retry mechanism will be needed

Agree. Though, as long as we can re-run a failed job with a few clicks, I think it's acceptable.

>  When we do see these errors, would you be able to add mitigations as needed?

I certainly hope so :slightly_smiling_face: 
I'll add contributing guides (and maybe some tests also) soon, in case I'm not available.

foolip:
Sorry for the delay, @sidvishnoi! I'd like to just go ahead and try this, and just fix things on master until it works.

Could you rebase this change so that we can see it working once more before I go ahead and merge it?

sidvishnoi:
@foolip Can we put this on hold for a little longer? I'm not fully available to fix potential issues in spec-prod right now. I'll update here when I'm back.

foolip:
@sidvishnoi sure no problem, just let me know.

sidvishnoi:
Forgot about this PR 😄.
`spec-prod` is stable enough now and under general use.

Renamed workflow file to `auto-publish.yml`, as we plan for all specs under W3C org. Would advice keeping this PR open for 2-3 more days, as we plan to rename `w3c/spec-prod` to `w3c/auto-publish`.

foolip (approved):
I'm going to merge this and see what happens. https://github.com/w3c/webdriver-bidi/pull/105 makes the same changes, but if it doesn't work I'll revert or fix.

foolip:
Deploy worked in https://github.com/w3c/webdriver-bidi/commit/25c12f07b2030a50d50616476bc0b85544c124f8 with no unexpected changes. Success! Thanks @sidvishnoi!

PR #56, opened 2020-09-16
Define how to respond with an error

jgraham:
The main difficulty here is that if we get a command with an invalid id, it's
unclear what to return in the id field. This chooses to return null in that case.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/56.html" title="Last updated on Oct 20, 2020, 1:54 PM UTC (03e5129)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/56/6c72ee4...03e5129.html" title="Last updated on Oct 20, 2020, 1:54 PM UTC (03e5129)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -128,6 +128,7 @@ EmptyParams = { *text }
 <pre class="cddl local-cddl">
 Message = (
   CommandResponse //
+  ErrorResponse //
   Event
 )
 
@@ -137,23 +138,15 @@ CommandResponse = {
   *text => any
 }
 
-ResponseData = (
-  Error //
-  CommandResult
-)
-
-Error = {
+ErrorResponse = {
+  id: uint / null,
   error: (
     "unknown error" /
     "unknown method" /
     "invalid argument"
   ),
   message: text,
-  stacktrace: text,
-}
-
-CommandResult = {
-  value: ResultData,
+  stacktrace?: text,
 }
 
 ResultData = (
@@ -367,9 +360,9 @@ To <dfn>process a command</dfn> given |data|:
       the <code>method</code> property of |data|.
 
    1. If |command| is not in the [=set of all command names=], return an
-      [=Error=] with [=error code=] [=unknown command=].
+      [=error=] with [=error code=] [=unknown command=].
 
-   1. Return an [=Error=] with [=error code=] [=invalid argument=].
+   1. Return an [=error=] with [=error code=] [=invalid argument=].
 </div>
 
 Transport {#transport}
@@ -543,7 +536,7 @@ To <dfn>handle an incoming message</dfn> given a [=WebSocket connection=]
     Issue: Should we instead close |connection| with [=status
     codes|status code=] 1003, or [=respond with an error=]?
 
- 2. [=Assert=]: |data| is a [=scalar value string=], because the
+ 1. [=Assert=]: |data| is a [=scalar value string=], because the
      WebSocket [=handling errors in UTF-8-encoded data=] would already
      have [=fail the WebSocket connection|failed the WebSocket
      connection=] otherwise.
@@ -551,21 +544,13 @@ To <dfn>handle an incoming message</dfn> given a [=WebSocket connection=]
     Issue: Nothing seems to define what [=status codes|status code=]
     is used for UTF-8 errors.
 
- 3. Let |parsed| be the result of [=parse JSON into Infra
-    values|parsing JSON into Infra values=] given |data|. If this
-    throws an exception, then [=respond with an error=] given
-    |connection| and [=error code=] [=invalid argument=], and finally
-    return.
-    <!-- corresponds to Parse error (-32700) in JSON-RPC -->
-
- 5. Let |result| be the result of [=process a command|processing a
+ 1. Let |result| be the result of [=process a command|processing a
     command=] given |data|.
 
- 6. If |result| is an [=Error=], then [=respond with an error=] given
-    |connection|, |result|, and |parsed|["<code>id</code>"], and finally
-    return.
+ 1. If |result| is an [=error=], then [=respond with an error=] given
+    |connection|, |data|, and |result|'s [=error code=], and finally return.
 
- 7. Let |response| be a new [=map=] with the following properties:
+ 1. Let |response| be a new [=map=] with the following properties:
 
    <dl>
       <dt>"id"</dt>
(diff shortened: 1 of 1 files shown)


jgraham:
This depends on #50 so we can't land it before that, but since the error parts weren't defined I made a quick PR.

index.bs
```diff
@@ -514,12 +514,30 @@ To <dfn>handle an incoming message</dfn> given a [=WebSocket connection=]
...
 To <dfn>respond with an error</dfn> given a [=WebSocket connection=]
-|connection| and an [=error code=] |code|:
+|connection|, |command id| and |error code|:
 
- 1. Issue: Form a valid JSON |errorObject| given |code|.
+ 1. If |command id| is not a number, let |command id| be null.
 
- 2. [=Send a WebSocket message=] comprised of |errorObject| over
-    |connection|.
+ 1. Let |error data| be a new map matching the <code>Error</code> production in
```
whimboo, line 521: Is it expected that every step has `1.` assigned to it? Maybe it will be correctly displayed in the rendered page?

index.bs
```diff
@@ -514,12 +514,30 @@ To <dfn>handle an incoming message</dfn> given a [=WebSocket connection=]
...
+ 1. Set the value of <code>id</code> on |error data| to |command id|.
+
+ 1. Assert: |error data| matches the <code>CommandResponse</code> definition in
+    the [=local end definition=].
+
+ 1. Let |response| be the result of [=serialize JSON to bytes|serializing JSON
+    to bytes=] given |error data|.
+
+ 1. Let |serialized| be the result of [=serialize JSON to
+    bytes|serializing JSON to bytes=] given |response|.
```
whimboo, line 538: Does that mean we serialize the `error data` twice?

index.bs
```diff
@@ -126,7 +126,7 @@ Message = (
 )
 
 CommandResponse = (
-  id: uint,
+  id: uint?,
```
bwalderman, line 129: A command response message, by definition, must always have an ID. Can the definitions be split up / reworked so that command responses always have a valid ID and errors have an optional ID? The meaning may end up being the same, but I think this will ultimately be more clear to readers.

bwalderman: Also, I think this should be `uint / null`.

bwalderman (changes requested):

index.bs
jgraham: Yep, Markdown will render it with the correct numbers.

index.bs
jgraham: Good catch.

index.bs
jgraham: Makes sense, fixed.

index.bs
```diff
@@ -122,31 +122,24 @@ EmptyParams = { *text }
...
+CommandResponse = {
   id: uint,
-  ResponseData,
-)
-
-ResponseData = (
-  Error //
-  CommandResult
-)
+  value: ResultData,
```
foolip, line 131:
```suggestion
  result: ResultData,
```

I see it was already `value` in the schema and this is just moving things around, but this seems like an unnecessary divergence from existing protocols and https://www.jsonrpc.org/specification#response_object. Happy to file an issue if you don't want to change it in this PR.

index.bs
whimboo: Could we then stay in sync with the rest of the spec. Under `incoming message` the numbering is incremented for each step.

jgraham: Yeah, I'm slowly converting everything to the `1.` style. We will end up with something consistent, but I don't want to add that to this PR (could be a good candidate for a small editorial PR though)

index.bs
```diff
@@ -514,12 +507,23 @@ To <dfn>handle an incoming message</dfn> given a [=WebSocket connection=]
 
 <div algorithm>
 To <dfn>respond with an error</dfn> given a [=WebSocket connection=]
-|connection| and an [=error code=] |code|:
+|connection|, |command id| and |error code|:
 
- 1. Issue: Form a valid JSON |errorObject| given |code|.
+ 1. If |command id| is not a number, let |command id| be null.
```
foolip, line 512: Our handling of invalid id seems a bit broken/incomplete still. If the number wasn't in the CDDL `uint` range, should we send back that number, or null? I think null would make sense, but I don't think spec makes it so.

This algorithm ("respond with an error") is called by "handle an incoming message", but that just parses JSON and doesn't try to interpret the `id` member. Then we have "Match data against the remote end definition" in "process a command" which operates on the unparsed scalar value string, and that would fail if the id isn't a CDDL `uint`. However, it doesn't extract a known-to-be-uint id if something else in the schema didn't match.

Overall, that we both parse the message payload as JSON and interpret it per CDDL seems suspect, and probably not how one would implement this.

If fixing this in this PR is too much, then a follow-up issue would be OK.

index.bs
```diff
@@ -514,12 +507,23 @@ To <dfn>handle an incoming message</dfn> given a [=WebSocket connection=]
...
 
- 2. [=Send a WebSocket message=] comprised of |errorObject| over
-    |connection|.
+ 1. Let |error data| be a new map matching the <code>ErrorResponse</code>
```
foolip, line 514: Is this a https://infra.spec.whatwg.org/#maps?

Unfortunately, Infra's [serialize JSON to bytes](https://infra.spec.whatwg.org/#serialize-json-to-bytes) doesn't actually work, as it takes a JS value. https://github.com/whatwg/infra/issues/159 tracks (I think) adding more stuff to allow serializing Infra values. I've asked there. A TODO somewhere around here would be good.

foolip (approved):
LGTM % questions and nits

index.bs
jgraham: This was designed to match WebDriver/HTTP. Let's discuss in an issue.

foolip: I've opened https://github.com/w3c/webdriver-bidi/issues/58.

index.bs
foolip: I was wrong and have filed https://github.com/whatwg/infra/issues/336.

jgraham: I tried to fix this a bit. It's definitely not perfect but it's at least not as wrong.

jgraham (commented):
I ended up making a bit of a bigger change in response to the review comments so you may want to look again.

bwalderman (approved):

index.bs
```diff
@@ -543,54 +544,85 @@ To <dfn>handle an incoming message</dfn> given a [=WebSocket connection=]
...
+    production in the [=local end definition=], with the <code>id</code> field
+    set to |command id|, the <code>error</code> field set to |error code|, the
+    <code>message</code> field set to an implementation-defined string
+    containing a human-readable definition of the error that occured and the
+    <code>stacktrace</code> field optionally set to an implementation-defined
+    string containing a stack trace report of the active stack frames at the
+    time when the error occurred.
+
+ 1. Let |response| be the result of [=serialize a map as JSON=] given |error
```
foolip, line 606: We can now use https://infra.spec.whatwg.org/#serialize-an-infra-value-to-a-json-string here.

index.bs
```diff
@@ -543,54 +544,85 @@ To <dfn>handle an incoming message</dfn> given a [=WebSocket connection=]
...
- 6. If |result| is an [=Error=], then [=respond with an error=] given
-    |connection|, |result|, and |parsed|["<code>id</code>"], and finally
-    return.
+ 1. If |result| is an [=error=], then [=respond with an error=] given
+    |connection|, |data|, and |result|'s [=error code=], and finally return.
+
+ 1. Assert: |result| is a [=success=]
 
- 7. Let |response| be a new [=map=] with the following properties:
+ 1. Let |command id| and |value| be the two fields |result|'s data.
```
foolip, line 563: Alright, so this is unpacking the return value from 'Return success with (data parsed["id"], value)' elsewhere.

foolip (approved):
Looks good. Using Infra to serialize a map to JSON is an optional improvement, could just as well do it separately. You decide :)

index.bs
jgraham: Presumably we want to use "as bytes" instead.

foolip: Yes, my bad.

Issue #18, opened 2020-05-29
[BiDi] Specify how to determine and address script execution contexts


jgraham:
WebDriver 1.0 doesn't have any explicit notion of [realms](https://tc39.es/ecma262/#sec-code-realms) (sometimes referred to as "targets" or "execution contexts" although the latter means something different in the context of ES). Instead it always executes script in the realm associated with the active browsing context. There is agreement that the BiDi version should treat realms as first class objects and allow commands to address specific realms.

The platform has several different mechanisms that create realms. Each browsing context has an associated realm in which scripts run (e.g. those provided by `<script>` elements). In addition various kinds of workers also create realms. Worker realms may be associated with zero or many browsing contexts. For example a document may create a dedicated worker which has that document in its [owner set](https://html.spec.whatwg.org/#concept-WorkerGlobalScope-owner-set). That worker may create further workers which have the parent in their owner set. Or a document may access a `SharedWorker` which can have multiple documents (or other workers) in its owner set. Service workers are different again in that they are created by the user agent in response to [service worker events](https://w3c.github.io/ServiceWorker/#service-worker-events), and may be shut down at any time.

[Worklets](https://drafts.css-houdini.org/worklets/) also exist. It's unclear to me what the requirements around those are.

The requirement to be able to run script in a specified realm means that commands which execute script must provide a realm id to use as the target. This is similar to the [Execution Context Id](https://chromedevtools.github.io/devtools-protocol/tot/Runtime/#type-ExecutionContextId) from CDP (note that CDP also has commands which take an object and infer the realm from the passed object).

In addition there must be some way to get a list of accessible realms. For a specific browsing context it may make sense to have a method to get the realm associated with that browsing context, and possibly also workers owned by the browsing context. It probably doesn't make sense to expose service workers in this way since they are more ephemeral.

It also makes sense to provide lifecycle events associated with realms being created and destroyed. This will allow getting handles to realms once they are created. It seems useful for such events to contain information about the kind of execution environment associated with the realm.

bwalderman:
I'm working on updating the relevant sections of the proposal this week:
- [Message Routing](https://github.com/w3c/webdriver-bidi/blob/master/proposals/core.md#message-routing)
- [Target Discovery](https://github.com/w3c/webdriver-bidi/blob/master/proposals/core.md#target-discovery)

bwalderman:
[...]

Issue #16, opened 2020-04-16
Decide on a value serialization format


jgraham:
Remote automation frequently requires serializing js objects and handles to DOM objects over the network. Different serialization formats are possible here with different tradeoffs. We should ensure we make an intentional decision for the format we use for the BiDi protocol; in particular it need not be the same as that for the HTTP protocol, although we *do* have a requirement to be able to share object handles i.e. it must be possible for a client (local end) to transform a WebElement from the HTTP protocol into a form where it can be transmitted over BiDi and will end up refering to the same element.

### WebDriver / HTTP

JS values are serialized as JSON. Handles to elements and windows are serialized using a special object format that's included inline in the JSON. Other values that aren't in JSON (e.g. BigInt, Infinity, etc.) aren't supported at all.

### CDP

Any remote object can be represented via a [RemoteObject](https://chromedevtools.github.io/devtools-protocol/tot/Runtime/#type-RemoteObject) object. This includes a mechaism for returning json serialized values where that is possible, and for representing js primitives that aren't serializable as json. However there isn't a mechanism for representing e.g. an array of elements, which means that something like findElements may require repeated trips over the API.

### Firefox RDP

https://docs.firefox-dev.tools/backend/protocol.html#grips

jgraham:
#57 will address this

foolip:
In order to understand more of how this works in CDP, I tried some things using the approach in https://github.com/w3c/webdriver-bidi/issues/14#issuecomment-633568306. Note that `DEBUG=puppeteer:protocol:*` is needed. CDP's `Runtime.evaluate` has a `returnByValue` parameter which controls the behavior. (I used Puppeteer's `page.evaluate` and `page.evaluateHandle` to test the difference.)

<table>
<tr><th>JS expr</th><th><code>returnByValue</code> true</th><th><code>returnByValue</code> false</th></tr>
<tr><td><code>[1, 2, 3]</code></td><td><code>{"type":"object","value":[1,2,3]}</code></td><td><code>{"type":"object","subtype":"array","className":"Array","description":"Array(3)","objectId":"{\"injectedScriptId\":3,\"id\":1}"}</code></td</tr>
<tr><td><code>{a:1, b:2}</code></td><td><code>{"type":"object","value":{"a":1,"b":2}}</code></td><td><code>{"type":"object","className":"Object","description":"Object","objectId":"{\"injectedScriptId\":3,\"id\":1}"}</code></td</tr>
[...]

PR #57, opened 2020-09-22
Start to specify how to serialize a ES value.

jgraham:
This starts building out the serialization infrastructure for sending
runtime-accessible objects over the wire. The format is based on
what's possible in CDP, with the notable change that the `value` field
is directly able to handle array and object types recursively rather
than using a seperate `preview` field. Also `type` and `subtype` are
merged into a single field and `value` and `unserializableValue` are
merged.

At this point there's no support for directly producing JSON values in
the serialization, we could add that later if required (but it's
arguably less important if the client is able to reconstruct the JSON
output when required).

There are also several other fields from CDP missing; we can add those
later if required.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/57.html" title="Last updated on Oct 26, 2020, 2:17 PM UTC (ab5bb13)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/57/819fd0c...ab5bb13.html" title="Last updated on Oct 26, 2020, 2:17 PM UTC (ab5bb13)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -732,7 +732,7 @@ handle to the object, valid for its lifetime inside the engine. In this case the
 struct) itself containing further remote object values. This allows objects like
 arrays to have a single representation.
 
-DOM Nodes are also represented by <code>RemoteObject</code> instances. These
+[=Nodes=] are also represented by <code>RemoteObject</code> instances. These
 have a partial serialization of the node in the value property.
 
 Note: mirror objects do not keep the original object alive in the runtime. If an
@@ -745,7 +745,7 @@ their corresponding id.
 Issue: Should this be explicitly per realm?
 
 <div algorithm>
-To get the <dfn export>object id for an object</dfn> given an |object|:
+To get the <dfn>object id for an object</dfn> given an |object|:
 
   1. If [=object id map=] for the [=current session=] does not contain |object|
      run the following steps:
@@ -761,7 +761,7 @@ To get the <dfn export>object id for an object</dfn> given an |object|:
 </div>
 
 <div algorithm>
-To get the <dfn export>object for an object id</dfn> given an |object id|:
+To get the <dfn>object for an object id</dfn> given an |object id|:
 
   1. For each |object| → |id value| of [=object id map=]:
 
@@ -808,12 +808,10 @@ ObjectId = text;
 
 UndefinedObject = {
   type: "undefined",
-  value: null,
 }
 
 NullObject = {
   type: "null",
-  value: null,
 }
 
 StringObject = {
@@ -846,13 +844,13 @@ SymbolObject = {
 ArrayObject = {
   type: "array",
   objectId: ObjectId,
-  value?: [RemoteObject]
+  value?: [*RemoteObject]
 }
 
 ObjectObject = {
   type: "object",
   objectId: ObjectId,
-  value?: {*text: RemoteObject}
+  value?: {* text => RemoteObject}
 }
 
 FunctionObject = {
@@ -860,8 +858,8 @@ FunctionObject = {
   objectId: ObjectId,
 }
 
-RegexObject = {
-  type: "regex",
+RegExpObject = {
+  type: "regexp",
   objectId: ObjectId,
   value: text
 }
@@ -947,8 +945,6 @@ WindowProxyObject = {
   type: "window",
   objectId: ObjectId,
 }
-
-
 ```
 
 Issue: Add WASM types?
--- a/index.bs
+++ b/index.bs
@@ -718,7 +718,7 @@ Realm = text;
 Each [=realm=] has an associated <dfn export>realm id</dfn>, which is a string
 uniquely identifying that realm.
 
-## Remote Value ## {#data-types-remote-value}
+## Remote Value ## {#data-types-remote-object}
 
 Values accessible from the ECMAScript runtime are represented by a mirror
 object, specified as <code>RemoteValue</code>. The value's type is specified in
@@ -955,69 +955,69 @@ Issue: Should WindowProxy get attributes in a similar style to Node?
 
 <div algorithm>
 
-To <dfn>serialize as a remote object</dfn> given an |object|, a |max depth|,
+To <dfn>serialize as a remote value</dfn> given an |value|, a |max depth|,
 |node details|, and a |set of known objects|:
 
-  1. Switch on the type of |object|:
+  1. Switch on the type of |value|:
 
     <dl>
       <dt>Undefined
-      <dd>Let |remote object| be an [=object=] matching the `UndefinedValue`
+      <dd>Let |remote value| be an [=object=] matching the `UndefinedValue`
       production in the [=local end definition=].
 
       <dt>Null
-      <dd>Let |remote object| be an [=object=] matching the `NullValue`
+      <dd>Let |remote value| be an [=object=] matching the `NullValue`
       production in the [=local end definition=].
 
       <dt>String
-      <dd>Let |remote object| be an [=object=] matching the `StringValue`
+      <dd>Let |remote value| be an [=object=] matching the `StringValue`
       production in the [=local end definition=], with the <code>value</code>
-      property set to |object|.
+      property set to |value|.
 
       <dt>Number
       <dd>
-      1. Switch on the value of |object|:
+      1. Switch on the value of |value|:
         <dl>
           <dt>NaN
-          <dt>Let |value| be <code>"NaN"</code>
+          <dt>Let |serialized| be <code>"NaN"</code>
           <dt>-0
-          <dt>Let |value| be <code>"-0"</code>
+          <dt>Let |serialized| be <code>"-0"</code>
           <dt>+Infinity
-          <dt>Let |value| be <code>"+Infinity"</code>
+          <dt>Let |serialized| be <code>"+Infinity"</code>
           <dt>-Infinity
-          <dt>Let |value| be <code>"-Infinity"</code>
+          <dt>Let |serialized| be <code>"-Infinity"</code>
           <dt>Otherwise:
-          <dt>Let |value| be |object|
+          <dt>Let |serialized| be |value|
         </dl>
-        Let |remote object| be an [=object=] matching the `NumberValue`
+        Let |remote value| be an [=object=] matching the `NumberValue`
         production in the [=local end definition=], with the <code>value</code>
-        property set to |value|.
+        property set to |serialized|.
 
       <dt>Boolean
-      <dd>Let |remote object| be an [=object=] matching the `BooleanValue`
+      <dd>Let |remote value| be an [=object=] matching the `BooleanValue`
           production in the [=local end definition=], with the <code>value</code>
-          property set to |object|.
+          property set to |value|.
 
       <dt>BigInt
-      <dd>Let |remote object| be an [=object=] matching the `BigIntValue`
+      <dd>Let |remote value| be an [=object=] matching the `BigIntValue`
           production in the [=local end definition=], with the <code>value</code>
           property set to the result of running the [=ToString=] operation on
-      |object|.
+          |value|.
 
       <dt>Symbol
-      <dd>Let |remote object| be an [=object=] matching the `SymbolValue`
+      <dd>Let |remote value| be an [=object=] matching the `SymbolValue`
           production in the [=local end definition=], with the <code>objectId</code>
-          property set to the [=object id for an object=] |object|.
+          property set to the [=object id for an object=] |value|.
 
       <dt>Array
       <dd>
-      1. Let |value| be null.
+      1. Let |serialized| be null.
 
-      1. If |object| is not in the |set of known objects|, and |max depth|
+      1. If |value| is not in the |set of known objects|, and |max depth|
          is not null and greater than 0, run the following steps:
-             1. Append |object| to the |set of known objects|
+             1. Append |value| to the |set of known objects|
 
-             1. Let |value| be a new List.
+             1. Let |serialized| be a new List.
 
              1. For each integer |index| in the range 0 to
                 [=LengthOfArrayLike=](|object|), run the following steps:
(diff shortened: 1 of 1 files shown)


jgraham:
This is a suggestion about the design, and lots of the spec details (e.g. how the type checking is handled) are wrong.

index.bs
```diff
@@ -677,14 +692,515 @@ id=] must be the same as the [=window handle=].
...
+  ArrayBufferObject //
+  NodeObject //
+  WindowProxyObject //
+}
+
+ObjectId = text;
+
+UndefinedObject = {
+  type: "undefined",
+  value: null,
```
foolip, line 793: Including the value for undefined and null seems unnecessary, and in the case of undefined makes it easier to conflate null and undefined. I see that https://tools.ietf.org/html/rfc8610#appendix-D does have `undefined` although I don't understand what it means.

How about just no value?

index.bs
```diff
@@ -677,14 +692,515 @@ id=] must be the same as the [=window handle=].
...
+       |object|
+
+  1. Return [=error=] with [=error code=] [=no such element=]
+
+  Issue: This error code isn't right.
+</div>
+
+[=remote end definition=] and [=local end definition=]
+```
+RemoteObject = {
```
foolip, line 760:
```suggestion
RemoteValue = {
```

Most of these are objects, but the first bunch aren't. Putting the suggestion just here, but it would require changes all over the place of course.

index.bs
```diff
@@ -677,14 +692,515 @@ id=] must be the same as the [=window handle=].
...
+  objectId: ObjectId,
+}
+
+ArrayObject = {
+  type: "array",
+  objectId: ObjectId,
+  value?: [RemoteObject]
+}
+
+ObjectObject = {
```
foolip, line 834: Alright, so this is the final fallback. If we add any new primitive types that are added in https://tc39.es/ecma262/#sec-ecmascript-language-types then we can ensure it will be an object, and never a value.

index.bs
```diff
@@ -677,14 +692,515 @@ id=] must be the same as the [=window handle=].
...
+  type: "object",
+  objectId: ObjectId,
+  value?: {*text: RemoteObject}
+}
+
+FunctionObject = {
```
foolip, line 840: So this how we'd represent a value like `window.Attr`, any Web IDL interface object, which all look like constructors. (No change requested.)

index.bs
```diff
@@ -677,14 +692,515 @@ id=] must be the same as the [=window handle=].
...
+  type: "map",
+  objectId: ObjectId,
+}
+
+WeakSetObject = {
+  type: "set",
+  objectId: ObjectId,
+}
+
+IteratorObject = {
```
foolip, line 877: What JS value would end up being serialized as `IteratorObject`? `Symbol.iterator` it a Symbol and the actual iterators functions, so I'm not sure.

If there is something, how about async iterators?

index.bs
```diff
@@ -677,14 +692,515 @@ id=] must be the same as the [=window handle=].
...
+  type: "iterator",
+  objectId: ObjectId,
+}
+
+GeneratorObject = {
```
foolip, line 882: Can you give an example of a JS value for this?

index.bs
```diff
@@ -677,14 +692,515 @@ id=] must be the same as the [=window handle=].
...
+  type: "iterator",
+  objectId: ObjectId,
+}
+
+ErrorObject = {
```
foolip, line 887: Would the type test be `v instanceof Error` on the remote end? That would include things like `TypeError` and `DOMException`, which I trust is the intention?

But is there any use to splitting this out, if one can't get any more information than from `ObjectObject`?

index.bs
```diff
@@ -677,14 +692,515 @@ id=] must be the same as the [=window handle=].
...
+  type: "error",
+  objectId: ObjectId,
+}
+
+ProxyObject = {
+  type: "proxy",
+  objectId: ObjectId,
+}
+
+PromiseObject = {
```
foolip, line 897: What's the type test for this, and should it be treated differently from non-Promise thenables?

index.bs
```diff
@@ -677,14 +692,515 @@ id=] must be the same as the [=window handle=].
...
+  type: "typedarray",
+  objectId: ObjectId,
+}
+
+ArrayBufferObject = {
+  type: "arraybuffer",
+  objectId: ObjectId,
+}
+
+NodeObject = {
```
foolip, line 912: Wondering about `EventTarget`. The `Node` hierarchy is special of course, but do we need anything special for `EventTarget`? I guess we'll see if we want commands to subscribe to events on specific targets?

index.bs
```diff
@@ -677,14 +692,515 @@ id=] must be the same as the [=window handle=].
...
+  value?: NodeValue,
+}
+
+NodeValue = {
+  nodeType: uint,
+  nodeName: text,
+  localName: text,
+  nodeValue: text,
+  childNodeCount: uint,
+  children?: [*NodeObject],
```
foolip, line 924: Not all nodes have children.

Is this inspired by existing protocols? Taking a close look at them would be good. We could try to represent all node types individually, or flatten everything into a single "node". I expect that elements at least are special in existing protocols?

index.bs
```diff
@@ -677,14 +692,515 @@ id=] must be the same as the [=window handle=].
...
+  attributes?: {*text: text},
+}
+
+WindowProxyObject = {
+  type: "window",
+  objectId: ObjectId,
+}
+
```
foolip, line 932: Some blank lines here.

index.bs
```diff
@@ -677,14 +692,515 @@ id=] must be the same as the [=window handle=].
...
+      <dd>Let |remote object| be an [=object=] matching the `UndefinedObject`
+      production in the [=local end definition=].
+
+      <dt>Null
+      <dd>Let |remote object| be an [=object=] matching the `NullObject`
+      production in the [=local end definition=].
+
+      <dt>String
+      <dd>Let |remote object| be an [=object=] matching the `StringObject`
+      production in the [=local end definition=], with the <code>value</code>
```
foolip, line 960: Here we'll need to handle what happens if the string has lone surrogates :(

index.bs
```diff
@@ -677,14 +692,515 @@ id=] must be the same as the [=window handle=].
...
+      property set to |object|.
+
+      <dt>Number
+      <dd>
+      1. Switch on the value of |object|:
+        <dl>
+          <dt>NaN
+          <dt>Let |value| be <code>"NaN"</code>
+          <dt>-0
+          <dt>Let |value| be <code>"-0"</code>
```
foolip, line 970: Hmm, tricky. This seems like it'll screw things up for people who write their code not expecting any strings, while not being able to represent it is also problematic...

Lazy out: what do other protocols do?

foolip (commented):
I've reviewed carefully up to `Number` in "serialize as a remote object" with sprinkles of attention after that point.

This is really close to the basic structure I had in mind, and there will be many details to sort out :)

index.bs
mathiasbynens: I assumed this was for things like `someIterable[Symbol.iterator]()`, i.e. objects that implement [the `Iterator` interface](https://tc39.es/ecma262/#sec-iterator-interface). Is that what you had in mind, @jgraham?

In that case, @foolip’s suggestion to include something like [`AsyncIteratorObject`](https://tc39.es/ecma262/#sec-asynciterator-interface) makes sense.

index.bs
mathiasbynens: Relevant ES spec section: https://tc39.es/ecma262/#sec-generator-objects

Same question as above: what about `AsyncGenerator`?

index.bs
mathiasbynens: We could spec it so that the serialization re-uses ECMAScript’s `JSON.stringify` — it was recently updated to ensure well-formed output even when lone surrogates are present. https://github.com/tc39/proposal-well-formed-stringify

index.bs
foolip: If that is the intention, then something could be both an iterable and something else at the same time. It's more like a behavior of an object than a type of object.

index.bs
foolip: Brilliant, we should ensure that any serialization we use ends up going through this.

index.bs
```diff
@@ -677,14 +692,515 @@ id=] must be the same as the [=window handle=].
...
+
+SymbolObject = {
+  type: "symbol",
+  objectId: ObjectId,
+}
+
+ArrayObject = {
+  type: "array",
+  objectId: ObjectId,
+  value?: [RemoteObject]
```
bwalderman, line 831:
```suggestion
  value?: [*RemoteObject]
```

bwalderman (changes requested):

index.bs
```diff
@@ -677,14 +692,515 @@ id=] must be the same as the [=window handle=].
...
+  type: "proxy",
+  objectId: ObjectId,
+}
+
+PromiseObject = {
+  type: "promise",
+  objectId: ObjectId,
+}
+
+TypedArrayObject = {
```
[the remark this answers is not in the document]
mathiasbynens: Alternatively, we could introduce dedicated subclasses, i.e. instead of `TypedArrayObject` we'd have `Int8ArrayObject`, `Uint8ArrayObject`, etc.

foolip:
I did a bit of research on what CDP does in https://github.com/w3c/webdriver-bidi/issues/16#issuecomment-703666094. We're not going to try matching CDP exactly, but it's worth considering the differences.

@jgraham when we get around to executing scripts (after https://github.com/w3c/webdriver-bidi/issues/18) do you think we'd have different modes, or is the intention to try avoiding that by always using a simple representation where possible? Plenty of tricky cases come to mind here, like `x = [1,2,3]; x.expando = 4` which would likely serialize to an array of numbers and drop the expando property. Large sparse arrays are another. My suspicion is that we'll need a way to opt out doing the usually-but-not-always helpful thing for cases like these, but the ability to execute arbitrary script to work around limitations in the protocol make me pretty unsure.

index.bs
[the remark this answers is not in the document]
jgraham: I guess if we had a subtype, it would be part of the `value` property. I don't have a strong sense of what's more useful here.

index.bs
jgraham: You're thinking we might have somethink like `Runtime.addEventListener(object_id, script)`? That seems reasonable but idk if that would require anything specific for things that implement `EventTarget`.

index.bs
jgraham: CDP just has https://chromedevtools.github.io/devtools-protocol/tot/DOM/#type-Node WebDriver 1.0 only supports elements, not other node types as return values. Firefox RDP is a bit of a different design, but afiact it only has a node grip type.

index.bs
jgraham: CDP has "value" and "unserializableValue" fields which are used in cases like this.

index.bs
jgraham: It seems a little tricky to reuse directly since afaict the entrypoints all also end up quoting the string, which isn't what we want.

index.bs
jgraham: I think I"m going to pull out these interface types for now. I assume the intent here is that you can write a client which grabs an iterable remote object and provides an API to actually iterate over it, going back to the remote end to actually perform that operation. But I don't think we need to get this right on the first pass.

index.bs
```diff
@@ -681,14 +706,546 @@ id=] must be the same as the [=window handle=].
...
+  StringValue //
+  NumberValue //
+  BooleanValue //
+  BigIntValue //
+  Symbol //
+  ArrayValue //
+  FunctionValue //
+  ObjectValue //
+  FunctionValue //
+  RegexValue //
```
mathiasbynens, line 786:
```suggestion
  RegExpValue //
```

index.bs
```diff
@@ -681,14 +706,546 @@ id=] must be the same as the [=window handle=].
...
+[=remote end definition=] and [=local end definition=]
+```
+RemoteValue = {
+  UndefinedValue //
+  NullValue //
+  StringValue //
+  NumberValue //
+  BooleanValue //
+  BigIntValue //
+  Symbol //
```
mathiasbynens, line 781: `SymbolValue`?

index.bs
```diff
@@ -681,14 +706,551 @@ id=] must be the same as the [=window handle=].
...
+                 with |child|, |max depth| - 1, |node details| and
+                 |set of known objects|.
+
+              1. Append |serialized| to |children|.
+
+            1. [=Set=](|serialized|, "children", |serialized|, false)
+
+            1. Let |attributes| be a new [=Object=].
+
+            1. For each |attribute| in |value|'s <a spec=dom>attributes</a>:
```
foolip, line 1144:
```suggestion
            1. For each |attribute| in |value|'s <a>attribute list</a>:
```

This was linked to https://dom.spec.whatwg.org/#concept-attribute, which isn't the right thing, should be https://dom.spec.whatwg.org/#concept-element-attribute.

Not sure about the syntax, maybe `for=Element` is the right fix.

index.bs
```diff
@@ -681,14 +706,551 @@ id=] must be the same as the [=window handle=].
...
+
+            1. [=Set=](|serialized|, "nodeValue", [=Get=](|value|, "nodeValue"), false)
+
+            1. Let |child node count| be the size of |serialized|'s [=children=].
+
+            1. [=Set=](|serialized|, "childNodeCount", |child node count|, false)
+
+            1. If |max depth| is null and or equal to 0 let |children| be null.
+               Otherwise, let |children| be an empty list and, for each node
+               |child| in the [=children=] of |value|:
```
foolip, line 1132:
```suggestion
               |child| in |value|'s [=children=]:
```

For consistency with the below attribute iteration.

index.bs
```diff
@@ -681,14 +706,551 @@ id=] must be the same as the [=window handle=].
...
+      <dt>Otherwise:
+      <dd>
+      1. let |serialized| be null.
+
+      1. If |value| is not in the |set of known objects|, and |max depth|
+         is not null and greater than 0, run the following steps:
+         1. Append |value| to the |set of known objects|
+
+         1. Let |serialized| be the result of [=serialize as a mapping=] given
+            [=EnumerableOwnPropertyNames=](|value|, key+value), |max depth|, |node
```
foolip, line 1183: key+value looked suspect, but looks like the right way to do invoke [EnumerableOwnPropertyNames](https://tc39.es/ecma262/#sec-enumerableownpropertynames) 👍

But should there be some assert right before this, since we're assuming that *value* is an Object here?

index.bs
```diff
@@ -681,14 +706,551 @@ id=] must be the same as the [=window handle=].
...
+       remote value=] with arguments |child value|, |max
+       depth| - 1, |node details| and |set of known objects|.
+
+    1. Append |serialized child| to |serialized|.
+
+  1. Return [=CreateArrayFromList=](|serialized|)
+</div>
+
+<div algorithm>
+To <dfn>serialize as a mapping</dfn> given |iterable|, |max depth|,
```
foolip, line 1221: This is used to create `ObjectValue`s, and it looks like it creates something on the same form as `Object.entries`, an array of 2-tuples of key+value pairs. What I don't quite get is why, since they keys could only be strings. Is it to preserve iteration order for languages which have unordered dicts, like Python in the past?

Could we just use CBOR objects instead?

index.bs
```diff
@@ -681,14 +706,551 @@ id=] must be the same as the [=window handle=].
...
+  1. For |item| in |iterable|:
+    1. Assert: [=IsArray=](|item|)
+
+    1. Let |property| be [=CreateListFromArrayLike=](|item|)
+
+    1. Assert |property| is a list of size 2
+
+    1. Let |key| be |property|[0] and let |value| be |property|[1]
+
+    1. If [=Type=](|key|) is String, let |serialized key| be |child key|,
```
foolip, line 1235: Could *key* be anything other than a string here?

index.bs
```diff
@@ -681,14 +706,551 @@ id=] must be the same as the [=window handle=].
...
+}
+
+SetValue = {
+  type: "set",
+  objectId: ObjectId,
+  value?: ListValue
+}
+
+WeakMapValue = {
+  type: "map",
```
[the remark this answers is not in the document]
jgraham: That's a typo :)

index.bs
```diff
@@ -681,14 +706,551 @@ id=] must be the same as the [=window handle=].
...
+  objectId: ObjectId,
+  value?: NodeProperties,
+}
+
+NodeProperties = {
+  nodeType: uint,
+  nodeName: text,
+  localName: text,
+  nodeValue: text,
+  childNodeCount: uint,
```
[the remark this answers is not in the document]
jgraham: It might not be if the node has children but we don't include them due to depth limits.

index.bs
jgraham: It can be Symbol.

index.bs
jgraham: It also allows it to work with non-string keys, which are important for maps and synbol-keyed objects.

foolip: Hmm, some bits appear to be missing here then. This "serialize as a mapping" algorithm is only called when making an `ObjectValue` production, not any maps AFAICT. For objects, [EnumerableOwnPropertyNames](https://tc39.es/ecma262/#sec-enumerableownpropertynames) has a "If Type(key) is String" condition which means we could only get string keys here.

For plain objects, I really have to question if it makes sense to complicate the representation in order to preserve more information than you would get from `JSON.stringify` or structured clone as in `postMessage`. Neither preserves symbols.

index.bs
foolip: I think it currently can't, but I'll resolve this in favor of the discussion in https://github.com/w3c/webdriver-bidi/pull/57#discussion_r508596239.

index.bs
```diff
@@ -681,14 +706,574 @@ id=] must be the same as the [=window handle=].
...
+
+To <dfn>serialize as a remote value</dfn> given an |value|, a |max depth|,
+|node details|, and a |set of known objects|:
+
+  1. In the following list of conditions and associated steps, run the first set
+     of steps for which the associated condition is true:
+
+    <dl>
+      <dt>[=Type=](|value|) is Undefined
+      <dd>Let |remote value| be an [=object=] matching the <code>UndefinedValue</code>
```
bwalderman, line 957: `[=object=]` is linking to https://w3c.github.io/FileAPI/#blob-url-entry-object

Should these be https://infra.spec.whatwg.org/#ordered-map instead?

jgraham: I changed things to use infra values now that we can serialize infra values to JSON.

mathiasbynens (approved):

bwalderman (approved):

index.bs
foolip: The condition for this is now that there's a [[ErrorData]] internal slot, which answers my original question.

foolip (approved):
There are some things I still have thoughts/doubts about, but I'd like to see this merged and to file follow-up issues for some things. It's getting very hard for me to follow my own discussion threads, I can't imagine it's easy for anyone with a long-lived PR. Thanks @jgraham, and let's merge!

index.bs
foolip: Let's file a follow-up issue about this.

index.bs
```diff
@@ -681,14 +706,551 @@ id=] must be the same as the [=window handle=].
...
+  ArrayBufferValue //
+  NodeValue //
+  WindowProxyValue //
+}
+
+ObjectId = text;
+
+ListValue = [*RemoteValue];
+
+MappingValue = [*[(RemoteValue / text), RemoteValue]];
```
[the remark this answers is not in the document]
foolip: Do you think this warrants a follow-up issue to discuss? I'm on the fence.

Issue #42, opened 2020-07-01
Specify browsing context discovery


jgraham:
We need a mechanism to discover which top level / other browsing contexts exist and a set of events for the browsing context lifecycle (created/destroyed/etc.).

shs96c:
The mechanism to discover top-level browsing contexts maps cleanly to [Get Window Handles](https://www.w3.org/TR/webdriver/#get-window-handles).

k7z45:
From ChromeDriver side, suppose DevTools is listening on a specific port say 64266.
We would first query http://localhost:64266/json/list via http for the top level browsing contexts and then we have a 
[navigation tracker](https://source.chromium.org/chromium/chromium/src/+/master:chrome/test/chromedriver/chrome/navigation_tracker.cc;l=219) that would process the events on the same port via WebSocket for the frame related events.
For example, for child browsing context, we can listen for [Page.frameAttached ](https://chromedevtools.github.io/devtools-protocol/tot/Page/#event-frameAttached).
I am not sure if there is an equivalent CDP command for querying http://localhost:64266/json/list via WebSocket(maybe [Target.getTargets](https://chromedevtools.github.io/devtools-protocol/tot/Target/#method-getTargets)?).

foolip:
This is being done in https://github.com/w3c/webdriver-bidi/pull/62.

foolip:
Let's called this fixed by https://github.com/w3c/webdriver-bidi/pull/62.

PR #62, opened 2020-10-22
Add a contexts module

jgraham:
This contains a command to get all the child contexts of a given
context, or to get all top-level contexts. It also provides events to
monitor when contexts are created or destroyed.

This is the first module to contain any events, so some infrastructure
for emitting events is also added.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/62.html" title="Last updated on Dec 11, 2020, 4:58 PM UTC (2e9c116)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/62/cc638d0...2e9c116.html" title="Last updated on Dec 11, 2020, 4:58 PM UTC (2e9c116)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -46,6 +46,7 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
     text: invalid argument; url: dfn-invalid-argument
     text: unknown command; url: dfn-unknown-command
     text: no such element; url: dfn-no-such-element
+    text: no such frame; url: dfn-no-such-frame
     text: active sessions; url: dfn-active-session
     text: local end; url: dfn-local-ends
     text: matched capability serialization algorithm; url: dfn-matched-capability-serialization-algorithm
@@ -83,6 +84,12 @@ spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
     text: ToString; url: sec-tostring
     text: Type; url: sec-ecmascript-data-types-and-values
     text: realm; url: sec-code-realms
+spec: HTML; urlPrefix: https://html.spec.whatwg.org/
+  type: dfn
+    text: a browsing context is discarded; url: a-browsing-context-is-discarded
+    text: create a new browsing context; url: creating-a-new-browsing-context
+    text: remove a browsing context; url: bcg-remove
+    text: session history; url: session-history
 </pre>
 
 <pre class="link-defaults">
@@ -140,7 +147,8 @@ Command = {
 }
 
 CommandData = (
-  SessionCommand
+  SessionCommand //
+  BrowsingContextCommand
 )
 
 EmptyParams = { *text }
@@ -157,7 +165,7 @@ Message = (
 
 CommandResponse = {
   id: uint,
-  value: ResultData,
+  result: ResultData,
   *text => any
 }
 
@@ -170,13 +178,19 @@ ErrorResponse = {
 
 ResultData = (
   EmptyResult //
-  SessionResult
+  SessionResult //
+  BrowsingContextResult
 )
 
 EmptyResult = {}
 
-Event = (
+Event = {
+  EventData,
   *text => any
+}
+
+EventData = (
+  BrowsingContextEvent
 )
 </pre>
 
@@ -282,24 +296,29 @@ disabled for a given browsing context.
 <div algorithm>
 
 To determine if an <dfn export>event is enabled</dfn> given |session|,
-|event name| and |browsing context|:
+|event name| and |browsing contexts|:
 
-  1. While |browsing context| is not null:
+Note: |browsing contexts| is a set because a [=shared worker=] can be associated
+      with multiple contexts.
 
-    1. Let |event map| be the [=browsing context event map=] for |session|.
+  1. For each |browsing context| in |browsing contexts|:
 
-    1. If |event map| contains an entry for |browsing context|, let |browsing
-       context events| be |event map|[|browsing context|].  Otherwise let
-       |browsing context events| be null.
+    1. While |browsing context| is not null:
 
-    1. If |browsing context events| is not null, and |browsing context events|
-       contains an entry for |event name| return |browsing context
-       events|[|event name|].
+      1. Let |event map| be the [=browsing context event map=] for |session|.
 
-    1. Let |browsing context| be the [=parent browsing context=] of |browsing
-       context|, if it has one, or null otherwise.
+      1. If |event map| [=contains=] |browsing context|, let |browsing context
+         events| be |event map|[|browsing context|].  Otherwise let |browsing
+         context events| be null.
 
-  1. If the [=global event set=] for |session| contains |event name| return
+      1. If |browsing context events| is not null, and |browsing context events|
+         [=contains=] for |event name| return |browsing context
+         events|[|event name|].
+
+      1. Let |browsing context| be the [=parent browsing context=] of |browsing
+         context|, if it has one, or null otherwise.
+
+  1. If the [=global event set=] for |session| [=contains=] |event name| return
      true.
 
   1. Return false.
@@ -362,7 +381,7 @@ listeners</dfn>, which is initially empty.
 
 A WebDriver [=/session=] has a <dfn>WebSocket connection</dfn> which is
 a network connection that follows the requirements of the
-[[!RFC6455|WebSocket protocol]].
+[[!RFC6455|WebSocket protocol]]. This is initially null.
 
 <div>
 
@@ -517,7 +536,7 @@ To <dfn>handle an incoming message</dfn> given a [=WebSocket connection=]
 
     1. Let |matched| be the map representing the matched data.
 
-    1. Assert: |matched| contains "<code>id</code>", "<code>method</code>", and
+    1. Assert: |matched| [=contains=] "<code>id</code>", "<code>method</code>", and
        "<code>params</code>".
 
     1. Let |command id| be |matched|["<code>id</code>"].
@@ -569,6 +588,27 @@ To <dfn>handle an incoming message</dfn> given a [=WebSocket connection=]
 </div>
 
 <div algorithm>
+To <dfn export>emit an event</dfn> given |body| and |related contexts|:
+
+ 1. [=Assert=]: |body| has [=map/size=] 2 and [=contains=] "<code>method</code>"
+    and "<code>params</code>".
+
+ 1. If the [=current session=] is null, or the [=current session=]'s [=WebSocket
+    Connection=] is null then return.
+
+ 1. If [=event is enabled=] given [=current session=],
+    |body|["<code>method</code>"] and |related contexts|:
+
+   1. Let |connection| be the [=current session=]'s [=WebSocket connection=].
+
+   1. Let |serialized| be the result of [=serialize an infra value to JSON
+        bytes=] given |body|.
+
+   1. [=Send a WebSocket message=] comprised of |serialized| over |connection|.
+
+</div>
+
+<div algorithm>
 To <dfn>respond with an error</dfn> given a [=WebSocket connection=]
 |connection|, |command id|, and |error code|:
 
@@ -590,6 +630,7 @@ To <dfn>respond with an error</dfn> given a [=WebSocket connection=]
  1. [=Send a WebSocket message=] comprised of |response| over |connection|.
 
 </div>
+
 
 <div algorithm>
 To <dfn>handle a connection closing</dfn> given a
@@ -663,18 +704,6 @@ Common Data Types {#data-types}
 ===============================
 
 This section defines data types which are common to many modules.
-
-## Browsing Context ## {#data_types-browsing_context}
-
-[=remote end definition=] and [=local end definition=]
-```
-BrowsingContext = text;
-```
-
-Each [=/browsing context=] has an associated <dfn export>browsing context id</dfn>,
-which is a string uniquely identifying that browsing context. For browsing
-contexts with an associated WebDriver [=window handle=] the [=/browsing context
-id=] must be the same as the [=window handle=].
 
 ## Realm ## {#data-types-realm}
 
@@ -1103,7 +1132,7 @@ To <dfn>serialize as a remote value</dfn> given an |value|, a |max depth|,
               1. Set |serialized|["<code>namespaceURI</code>"] to [=Get=](|value|, "namespaceURI")
 
 
-            1. Let |child node count| be the size of |serialized|'s [=children=].
+            1. Let |child node count| be the [=list/size=] of |serialized|'s [=children=].
 
             1. Set |serialized|["<code>childNodeCount</code>" to |child node count|
 
@@ -1223,7 +1252,7 @@ To <dfn>serialize as a mapping</dfn> given |iterable|, |max depth|,
 
     1. Let |property| be [=CreateListFromArrayLike=](|item|)
 
-    1. Assert: |property| is a list of size 2
+    1. Assert: |property| is a list of [=list/size=] 2
 
     1. Let |key| be |property|[0] and let |value| be |property|[1]
 
@@ -1248,7 +1277,7 @@ To <dfn>serialize as a mapping</dfn> given |iterable|, |max depth|,
 Modules {#modules}
 ==================
 
-## Session ## {#module-session}
+## The session Module ## {#module-session}
 
 The <dfn export for=modules>session</dfn> module contains commands and
 events for monitoring the status of the remote end.
@@ -1258,8 +1287,8 @@ events for monitoring the status of the remote end.
 [=remote end definition=]
 
 <pre class="cddl remote-cddl">
-SessionCommand = (StatusCommand //
-                  SubscribeCommand)
+SessionCommand = (SessionStatusCommand //
+                  SessionSubscribeCommand)
 </pre>
 
 [=local end definition=]
@@ -1289,7 +1318,7 @@ To <dfn lt="updating the event map">update the event map</dfn>, given
 
       1. If |enabled| is true, for each |event name| in |event names|,
          append |event name| to |global event set|. Otherwise for for each
-         |event name| in |event names|, if the |global event set| contains
+         |event name| in |event names|, if the |global event set| [=contains=]
          |event name|, remove |event name| from the |global event set|.
 
       1. Return
@@ -1298,12 +1327,9 @@ To <dfn lt="updating the event map">update the event map</dfn>, given
 
     1. For each entry |context id| in the |list of contexts|:
 
-      1. If there is no browsing context with [=browsing context id=] |context id| return
-         [=error=] with [=error code=] [=invalid argument=]
-
-      1. Let |context| be the browsing context with id |context id|. If the
-         |event map| does not contain an entry for |context|,
-         set the value of the entry for |context| to a new empty map.
+      1. Let |context| be the result of [=trying=] to [=get a browsing context=]
+         with |context id|.  If the |event map| does not contain an entry for
+         |context|, set the value of the entry for |context| to a new empty map.
 
       1. Get the entry from the |event map| for |context| and append it to
          |targets|.
@@ -1323,9 +1349,11 @@ To <dfn lt="updating the event map">update the event map</dfn>, given
     filter only those that ought to be returned to the local end.
 </div>
 
-### Status ### {#command-session-status}
+### Commands ### {#module-session-commands}
 
-The <dfn for=commands>status</dfn> command returns information about
+#### The session.status Command #### {#command-session-status}
+
+The <dfn export for=commands>session.status</dfn> command returns information about
 whether a remote end is in a state in which it can create new sessions,
 but may additionally include arbitrary meta information that is specific
 to the implementation.
@@ -1334,7 +1362,7 @@ to the implementation.
    <dt>Command Type</dt>
    <dd>
       <pre class="cddl remote-cddl">
-      StatusCommand = {
+      SessionStatusCommand = {
         method: "session.status",
         params: EmptyParams,
       }
@@ -1343,7 +1371,7 @@ to the implementation.
    <dt>Return Type</dt>
    <dd>
       <pre class="cddl local-cddl">
-      StatusResult = {
+      SessionStatusResult = {
         ready: bool,
         message: text,
       }
@@ -1366,10 +1394,10 @@ The [=remote end steps=] are:
 
 2. Return [=success=] with data |body|
 
-### Subscribe ### {#command-session-subscribe}
+#### The session.subscribe Command #### {#command-session-subscribe}
 
-The <dfn export for=commands>subscribe</dfn> command enables certain events either
-globally or for a set of browsing contexts
+The <dfn export for=commands>session.subscribe</dfn> command enables certain events
+either globally or for a set of browsing contexts
 
 Issue: This needs to be generalized to work with realms too
 
@@ -1377,12 +1405,12 @@ Issue: This needs to be generalized to work with realms too
    <dt>Command Type</dt>
    <dd>
       <pre class="cddl remote-cddl">
-      SubscribeCommand = {
+      SessionSubscribeCommand = {
         method: "session.subscribe",
         params: SubscribeParameters
       }
 
-      SubscribeParameters = {
+      SessionSubscribeParameters = {
         events: [*text],
         ?contexts: [*BrowsingContext],
       }
@@ -1408,10 +1436,10 @@ The [=remote end steps=] with |command parameters| are:
        |list of event names|, |list of contexts| and enabled true.
 </div>
 
-### Unsubscribe ### {#command-session-unsubscribe}
+#### The session.unsubscribe Command #### {#command-session-unsubscribe}
 
-The <dfn export for=commands>unsubscribe</dfn> command disables events either
-globally or for a set of browsing contexts
+The <dfn export for=commands>session.unsubscribe</dfn> command disables events
+either globally or for a set of browsing contexts
 
 Issue: This needs to be generalised to work with realms too
 
@@ -1419,7 +1447,7 @@ Issue: This needs to be generalised to work with realms too
    <dt>Command Type</dt>
    <dd>
      <pre class="cddl remote-cddl">
-     UnsubscribeCommand = {
+     SessionUnsubscribeCommand = {
        method: "session.unsubscribe",
        params: SubscribeParameters
      }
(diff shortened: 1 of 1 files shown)


bwalderman (changes requested):

index.bs
```diff
@@ -904,3 +922,207 @@ The [=remote end steps=] with |command parameters| are:
     1. Return the result of [=updating the event map=] with [=current session=],
        |list of event names|, |list of contexts| and enabled false.
 </div>
+
+
+## Context ## {#module-contexts}
+
+The <dfn export for=modules>context</dfn> module contains commands and
```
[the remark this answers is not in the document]
jgraham: The intent is that real would be different. Practically I think the implication would be that you could easily enable all context events with a single enable events call, and not also get realm events. That seems like a pretty common requirement for me (e.g. for any automation that doesn't care to interact with workers).

index.bs
```diff
@@ -904,3 +922,207 @@ The [=remote end steps=] with |command parameters| are:
...
+associated WebDriver [=window handle=] the [=/browsing context id=] must be the
+same as the [=window handle=].
+
+<div algorithm>
+To <dfn>get a browsing context</dfn> given |context id|:
+
+ 1. If |context id| is null, return [=success=] with data null.
+
+ 1. If there is no browsing context with [=browsing context id=] |context id| return
+    [=error=] with [=error code=] [=invalid argument=]
```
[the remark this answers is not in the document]
jgraham: Done. I think we probably want a new errors section with some values that map onto the old errors and some that don't.

bwalderman (changes requested):
CDP has Page.frameAttached and Page.frameDetached to indicate when a nested browsing context (frame) is attached or detach from a parent.

Since we're including the `parent` browsing context ID in the `contextCreated` and `contentDestroyed` events, its important to also let the client know if this `parent` ever changes for a given context. I'd suggest adding `contextAttached` and `contextDetached` events that correspond to the CDP events above. The remote end event trigger could be something like:

"When a [browsing context container's](https://html.spec.whatwg.org/#browsing-context-container) [nested browsing context](https://html.spec.whatwg.org/#nested-browsing-context) changes"

jgraham:
> CDP has Page.frameAttached and Page.frameDetached to indicate when a nested browsing context (frame) is attached or detach from a parent.

If I understand the spec correctly we will always generate `contextCreated` and `contextDestroyed` events for these cases. From HTML:
> When an iframe element element is inserted into a document whose browsing context is non-null, the user agent must run these steps:
>
>    Create a new nested browsing context for element.
>
>    Process the iframe attributes for element, with initialInsertion set to true.
>
> When an iframe element is removed from a document, the user agent must discard the element's nested browsing context, if it is not null, and then set the element's nested browsing context to null.

So removing an iframe will cause it to be discarded generating a `contextDestoyed` event, and inserting it into a new parent will cause a new context to be created causing a `contextCreated` event.

index.bs
```diff
@@ -1462,3 +1484,229 @@ The [=remote end steps=] with |command parameters| are:
...
+    |parent|. Otherwise let |parent| and |parent id| be null.
+
+ 1. Let |params| be a map matching the <code>ContextDestroyedParameters</code>
+    production, with the <code>context</code> field set to |context id|,
+    and the <code>parent</code> field set to |parent id|.
+
+ 1. Let |body| be a map matching the <code>ContextDestroyedEvent</code>
+    production, with the <code>params</code> field set to |params|.
+
+ 1. Let |related contexts| be a set.
```
bwalderman, line 1706: Nit: These can be combined into one step like above.

index.bs
```diff
@@ -1462,3 +1484,229 @@ The [=remote end steps=] with |command parameters| are:
...
+
+ 1. If |parent| is not null. append |parent| to |related contexts|.
+
+ 1. [=Emit an event=] with |body| and |parent id|.
```
bwalderman, line 1710:
```suggestion
 1. [=Emit an event=] with |body| and |related contexts|.
```

bwalderman (changes requested):

bwalderman:
> If I understand the spec correctly we will always generate `contextCreated` and `contextDestroyed` events for these cases. From HTML:
> 
> > When an iframe element element is inserted into a document whose browsing context is non-null, the user agent must run these steps:
> > Create a new nested browsing context for element.
> > Process the iframe attributes for element, with initialInsertion set to true.
> > When an iframe element is removed from a document, the user agent must discard the element's nested browsing context, if it is not null, and then set the element's nested browsing context to null.
> 
> So removing an iframe will cause it to be discarded generating a `contextDestoyed` event, and inserting it into a new parent will cause a new context to be created causing a `contextCreated` event.

Thanks. After looking at the HTML spec I agree `contextCreated` and `contextDestroyed` match the behavior defined there.

bwalderman:
I'm a little nervous about use the name `context` in the module and its commands/events instead of `browsingContext`. The word `context` on its own is very generic and we may want other kinds of "contexts" to the protocol. BrowsingContext seems much more clear, and leaves open the possibility of adding any FooContext or BarContext concepts we may need in the future.

jgraham:
> I'm a little nervous about use the name `context` in the module and its commands/events instead of `browsingContext`. The word `context` on its own is very generic and we may want other kinds of "contexts" to the protocol. BrowsingContext seems much more clear, and leaves open the possibility of adding any FooContext or BarContext concepts we may need in the future.

Yeah, that's reasonable. Do we want commands like `BrowsingContext.getBrowsingContexts` or more like `BrowsingContext.getAll`?

jgraham:
> > I'm a little nervous about use the name `context` in the module and its commands/events instead of `browsingContext`. The word `context` on its own is very generic and we may want other kinds of "contexts" to the protocol. BrowsingContext seems much more clear, and leaves open the possibility of adding any FooContext or BarContext concepts we may need in the future.
> 
> Yeah, that's reasonable. Do we want commands like `BrowsingContext.getBrowsingContexts` or more like `BrowsingContext.getAll`?

I decided to go with something more like the latter.

index.bs
```diff
@@ -306,22 +318,24 @@ disabled for a given browsing context.
 <div algorithm>
 
 To determine if an <dfn export>event is enabled</dfn> given |session|,
-|event name| and |browsing context|:
+|event name| and |browsing contexts|:
```
foolip, line 321: Making *browsing contexts* plural here is probably in the service of an anticipated future need, but I wonder what it is? Right now, it will either be an empty set or a set of one thing, the possible  parent browsing context for the context being created or destroyed.

The behavior here is to treat the event as enabled if it's enabled in any of the given browsing contexts, but when will be be passing in multiple browsing contexts here?

This also had the side effect making a browsing context optional, but if that's the main purpose maybe we can make it explicitly optional, but still singular?

index.bs
```diff
@@ -581,6 +595,26 @@ To <dfn>handle an incoming message</dfn> given a [=WebSocket connection=]
 
 </div>
 
+<div algorithm>
+To <dfn export>emit an event</dfn> given |body| and |related contexts|:
+
+ 1. [=Assert=]: |body| has size 2 and contains "<code>method</code>" and "<code>params</code>".
```
foolip, line 601: Linking https://infra.spec.whatwg.org/#map-size and https://infra.spec.whatwg.org/#map-exists here would be good, also making it clear that *body* is a map.

index.bs
```diff
@@ -581,6 +595,26 @@ To <dfn>handle an incoming message</dfn> given a [=WebSocket connection=]
...
+
+ 1. If the [=current session=] is null, or the [=current session=]'s [=WebSocket
```
foolip, line 603: "current session" is used elsewhere without a null check, is it needed here?

index.bs
```diff
@@ -1462,3 +1484,231 @@ The [=remote end steps=] with |command parameters| are:
...
+
+ 1. If |context| has a [=parent browsing context=] let |parent id| be the
+    [=browsing context id=] of that parent. Otherwise let |parent id| be null.
+
+ 1. Let |document| be |context|'s [=active document=].
+
+ 1. Let |url| be the result of running the [=URL serializer=], given
+    |document|'s [=spec=html/URL=].
+
+ 1. Let |context info| be a map matching the <code>BrowsingContextInfo</code>
```
foolip, line 1564:
```suggestion
 1. Let |context info| be a [=map=] matching the <code>BrowsingContextInfo</code>
```

index.bs
```diff
@@ -1462,3 +1484,231 @@ The [=remote end steps=] with |command parameters| are:
...
+To <dfn>get the browsing context info</dfn> given |context|:
+ 1. Let |context id| be the [=browsing context id=] for |context|.
+
+ 1. If |context| has a [=parent browsing context=] let |parent id| be the
+    [=browsing context id=] of that parent. Otherwise let |parent id| be null.
+
+ 1. Let |document| be |context|'s [=active document=].
+
+ 1. Let |url| be the result of running the [=URL serializer=], given
+    |document|'s [=spec=html/URL=].
```
foolip, line 1562:
```suggestion
    |document|'s [=document/URL=].
```

This didn't work, `[=spec=html/URL=]` is in the output. I think something like the suggestion should work, or `<a for=Document>URL</a>`. I think that's nicer than picking the spec directly.

index.bs
```diff
@@ -581,6 +595,26 @@ To <dfn>handle an incoming message</dfn> given a [=WebSocket connection=]
...
+<div algorithm>
+To <dfn export>emit an event</dfn> given |body| and |related contexts|:
+
+ 1. [=Assert=]: |body| has size 2 and contains "<code>method</code>" and "<code>params</code>".
+
+ 1. If the [=current session=] is null, or the [=current session=]'s [=WebSocket
+    Connection=] is null then return.
+
+ 1. If [=event is enabled=] given [=current session=],
+    |body|["<code>method</code>"] and |related contexts|:
```
foolip, line 607: It's a bit unfortunate that "emit an event" needs to take a map with exactly two things, instead of event type and params separately. At the call sites, "emit an event named 'foo.bar' with *params*" would be nicer, and one wouldn't need to peek inside the map here.

The reason for this is basically how the CDDL is written, pairing together event name and the parameters in a way where they can't be mismatched, which is also nice.

I don't have a suggested solution here, but raising it in case someone can see a way to make it nice also at the call sites.

index.bs
```diff
@@ -1266,7 +1289,7 @@ To <dfn>serialize as a mapping</dfn> given |iterable|, |max depth|,
 Modules {#modules}
 ==================
 
-## Session ## {#module-session}
+## session ## {#module-session}
```
foolip, line 1292:
```suggestion
## The session Module ## {#module-session}
```

I'm suggesting this in the common style of "The FooBar Interface", mostly to avoid having a lowercase section title.

index.bs
```diff
@@ -1341,9 +1361,11 @@ To <dfn lt="updating the event map">update the event map</dfn>, given
     filter only those that ought to be returned to the local end.
 </div>
 
-### Status ### {#command-session-status}
+### Commands ### {#module-session-commands}
 
-The <dfn for=commands>status</dfn> command returns information about
+#### Status #### {#command-session-status}
```
foolip, line 1366:
```suggestion
#### The session.status Command #### {#command-session-status}
```

As per above. Or we just uppercase "session" above again :)

Ditto in the follow, I won't make suggestions throughout.

index.bs
```diff
@@ -1462,3 +1484,231 @@ The [=remote end steps=] with |command parameters| are:
     1. Return the result of [=updating the event map=] with [=current session=],
        |list of event names|, |list of contexts| and enabled false.
 </div>
+
+
+## browsingContext ## {#module-browsingContext}
+
+The <dfn export for=modules>Browsing Context</dfn> module contains commands and
```
foolip, line 1491:
```suggestion
The <dfn export for=modules>browsingContext</dfn> module contains commands and
```

For consistency with the session module. One of them should change.

index.bs
```diff
@@ -1462,3 +1484,231 @@ The [=remote end steps=] with |command parameters| are:
...
+events relating to browsing contexts.
+
+### Definition ### {#module-browsingContext-definition}
+
+[=remote end definition=]
+
+<pre class="cddl remote-cddl">
+
+BrowsingContextCommand = (BrowsingContextGetAllCommand)
```
foolip, line 1500:
```suggestion
BrowsingContextCommand = (BrowsingContextListCommand)
```

Right? And doesn't `BrowsingContextCommand` need to be used somewhere for the CDDL to make sense?

index.bs
```diff
@@ -1462,3 +1484,231 @@ The [=remote end steps=] with |command parameters| are:
...
+</pre>
+
+[=local end definition=]
+
+<pre class="cddl local-cddl">
+
+BrowsingContextResult = (BrowsingContextGetAllResult)
```
foolip, line 1507:
```suggestion
BrowsingContextResult = (BrowsingContextListResult)
```

index.bs
```diff
@@ -1462,3 +1484,231 @@ The [=remote end steps=] with |command parameters| are:
...
+
+### Events ### {#module-contexts-events}
+
+#### contextCreated #### {#event-browsingContext-contextCreated}
+
+<dl>
+   <dt>Event Type</dt>
+   <dd>
+      <pre class="cddl local-cddl">
+        BrowsingContextContextCreatedEvent = {
```
foolip, line 1640:
```suggestion
        BrowsingContextCreatedEvent = {
```

To match the earlier CDDL. One of them needs to change.

index.bs
```diff
@@ -1462,3 +1484,231 @@ The [=remote end steps=] with |command parameters| are:
...
+ 1. Let |context| be the newly created browsing context.
+
+ 1. Let |related contexts| be a set containing the [=parent browsing context=]
+    of |context|, if that is not null, or an empty set otherwise.
+
+ 1. Let |params| be the result of [=get the browsing context info=] given
+    |context|.
+
+ 1. Let |body| be a map matching the
+    <code>BrowsingContextContextCreatedEvent</code> production, with the
```
foolip, line 1662:
```suggestion
    <code>BrowsingContextCreatedEvent</code> production, with the
```

index.bs
```diff
@@ -1462,3 +1484,231 @@ The [=remote end steps=] with |command parameters| are:
...
+
+</div>
+
+#### contextDestroyed #### {#event-browsingContext-contextDestroyed}
+
+<dl>
+   <dt>Event Type</dt>
+   <dd>
+      <pre class="cddl local-cddl">
+        BrowsingContextContextDestroyedEvent = {
```
foolip, line 1675:
```suggestion
        BrowsingContextDestroyedEvent = {
```

index.bs
```diff
@@ -1462,3 +1484,231 @@ The [=remote end steps=] with |command parameters| are:
...
+ 1. If |context| has a [=parent browsing context=] let |parent| be that parent
+    browsing context and |parent id| be the [=browsing context id=] of
+    |parent|. Otherwise let |parent| and |parent id| be null.
+
+ 1. Let |params| be a map matching the <code>ContextDestroyedParameters</code>
+    production, with the <code>context</code> field set to |context id|,
+    and the <code>parent</code> field set to |parent id|.
+
+ 1. Let |body| be a map matching the
+    <code>BrowsingContextContextDestroyedEvent</code> production, with the
```
foolip, line 1706:
```suggestion
    <code>BrowsingContextDestroyedEvent</code> production, with the
```

index.bs
```diff
@@ -1462,3 +1484,231 @@ The [=remote end steps=] with |command parameters| are:
...
+</div>
+
+<div algorithm>
+To <dfn>get the browsing context info</dfn> given |context|:
+ 1. Let |context id| be the [=browsing context id=] for |context|.
+
+ 1. If |context| has a [=parent browsing context=] let |parent id| be the
+    [=browsing context id=] of that parent. Otherwise let |parent id| be null.
+
+ 1. Let |document| be |context|'s [=active document=].
```
foolip, line 1559: AFAICT, active document can never be null, so that's good!

index.bs
```diff
@@ -1462,3 +1484,231 @@ The [=remote end steps=] with |command parameters| are:
...
+
+ 1. Let |url| be the result of running the [=URL serializer=], given
```
foolip, line 1561: Can we be explicit about whether or not this should include any fragment? https://chromedevtools.github.io/devtools-protocol/tot/Page/#type-Frame has variants for both, so it would be easy to accidentally use the wrong thing when implementing over CDP.

index.bs
```diff
@@ -1462,3 +1484,231 @@ The [=remote end steps=] with |command parameters| are:
...
+
+ 1. Return |context info|.
+
+</div>
+
+### Commands ### {#module-browsingContext-commands}
+
+#### list ####  {#command-browsingContext-list}
+
+The <dfn export for=commands>browsingContext.list</dfn> command returns a list
```
foolip, line 1577: I see that CDP's equivalent for this lists the whole browsing context tree as one command:
https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-getFrameTree

I believe that's actually better, because there's no risk of getting an inconsistent view of the tree by trying to build it up over multiple roundtrips.

How does this work in Gecko's protocol(s)?

index.bs
```diff
@@ -1462,3 +1484,231 @@ The [=remote end steps=] with |command parameters| are:
...
+        }
+    </pre>
+   </dd>
+</dl>
+
+<div algorithm="remote end steps for browsingContext.list">
+The [=remote end steps=] with |command parameters| are:
+
+  1. Let the |parent id| be the value of the <code>parent</code> field of
+       |command parameters| if present, or null otherwise.
```
foolip, line 1609:
```suggestion
     |command parameters| if present, or null otherwise.
```

(A bit too much indetation.)

foolip: Also let's use/link https://infra.spec.whatwg.org/#map-exists terminology here.

index.bs
```diff
@@ -1462,3 +1484,231 @@ The [=remote end steps=] with |command parameters| are:
...
+
+  1. Let |parent| be the result of [=trying=] to [=get a browsing context=]
+     given |parent id|.
+
+  1. If |parent| is null, let |contexts| be a list containing all [=top-level
```
foolip, line 1614: This will need to change significantly to return the whole tree, but looks right for what it's intended to do now.

index.bs
```diff
@@ -1462,3 +1484,231 @@ The [=remote end steps=] with |command parameters| are:
...
+
+  1. For each |context| of |context|, append the result of [=get the browsing
+     context info=] given |context| to |context info|.
+
+  1. Let |body| be a map matching the <code>BrowsingContextListResult</code>
+     production, with the <code>contexts</code> field set to |context info|.
+
+  1. Return [=success=] with data |body|.
+
+Issue: Should this contain other data like the current URL for the context.
```
foolip, line 1628: It does contain the current URL, doesn't it?

index.bs
```diff
@@ -1462,3 +1484,231 @@ The [=remote end steps=] with |command parameters| are:
...
+         parent: BrowsingContext / null,
+       }
+      </pre>
+   </dd>
+</dl>
+The [=remote end event trigger=] is:
+
+<div algorithm="remote end event trigger for browsingContext.contexDestroyed">
+
+When [=a browsing context is discarded=], run the following steps:
```
foolip, line 1691: https://html.spec.whatwg.org/#a-browsing-context-is-discarded has a funny ID, but like "create a new browsing context" it's an algorithm that's invoked, not an event that happens. So I think this needs to be "When the discard a browsing context algorithm is invoked". Some creative linking may be needed.

index.bs
```diff
@@ -1462,3 +1484,231 @@ The [=remote end steps=] with |command parameters| are:
...
+The [=remote end event trigger=] is:
+
+<div algorithm="remote end event trigger for browsingContext.contextCreated">
+When the [=create a new browsing context=] algorithm is invoked, after the
+[=active document=] of the browsing context is set, run the following steps:
+
+ 1. Let |context| be the newly created browsing context.
+
+ 1. Let |related contexts| be a set containing the [=parent browsing context=]
+    of |context|, if that is not null, or an empty set otherwise.
```
foolip, line 1656: Is it possible for the parent to be null? My understanding was that iframes, for example, get a browsing context created when inserted into the DOM. How does one create an orphaned browsing context, and is that surfaced in other protocols, if at all?

I came across https://chromedevtools.github.io/devtools-protocol/tot/Page/#event-frameAttached, but I don't know if that's the only related event in CDP. @sadym-chromium do you know?

index.bs
```diff
@@ -1462,3 +1484,231 @@ The [=remote end steps=] with |command parameters| are:
...
+
+<div algorithm="remote end event trigger for browsingContext.contexDestroyed">
+
+When [=a browsing context is discarded=], run the following steps:
+
+ 1. Let |context| be the browsing context being discarded.
+
+ 1. Let |context id| be the [=browsing context id=] for |context|.
+
+ 1. If |context| has a [=parent browsing context=] let |parent| be that parent
```
foolip, line 1697: Have you checked how this will end up working if a whole tree of iframes are removed at once? Will events for each browsing context be fired, and in what order?

index.bs
```diff
@@ -1462,3 +1484,231 @@ The [=remote end steps=] with |command parameters| are:
...
+    browsing context and |parent id| be the [=browsing context id=] of
+    |parent|. Otherwise let |parent| and |parent id| be null.
+
+ 1. Let |params| be a map matching the <code>ContextDestroyedParameters</code>
```
foolip, line 1701:
```suggestion
 1. Let |params| be a [=map=] matching the <code>BrowsingContextDestroyedParameters</code>
```

index.bs
```diff
@@ -1462,3 +1484,231 @@ The [=remote end steps=] with |command parameters| are:
...
+  1. If |parent| is null, let |contexts| be a list containing all [=top-level
+     browsing contexts=]. Otherwise let |contexts| be a list containing all
+     [=/browsing contexts=] which are [=child browsing contexts=] of |parent|.
+
+  1. Let |context info| be a list.
+
+  1. For each |context| of |context|, append the result of [=get the browsing
+     context info=] given |context| to |context info|.
+
+  1. Let |body| be a map matching the <code>BrowsingContextListResult</code>
```
foolip, line 1623:
```suggestion
  1. Let |body| be a [=map=] matching the <code>BrowsingContextListResult</code>
```

index.bs
```diff
@@ -1462,3 +1484,231 @@ The [=remote end steps=] with |command parameters| are:
...
+
+ 1. Let |context| be the newly created browsing context.
+
+ 1. Let |related contexts| be a set containing the [=parent browsing context=]
+    of |context|, if that is not null, or an empty set otherwise.
+
+ 1. Let |params| be the result of [=get the browsing context info=] given
+    |context|.
+
+ 1. Let |body| be a map matching the
```
foolip, line 1661:
```suggestion
 1. Let |body| be a [=map=] matching the
```

index.bs
```diff
@@ -1462,3 +1484,231 @@ The [=remote end steps=] with |command parameters| are:
...
+
+ 1. If |context| has a [=parent browsing context=] let |parent| be that parent
+    browsing context and |parent id| be the [=browsing context id=] of
+    |parent|. Otherwise let |parent| and |parent id| be null.
+
+ 1. Let |params| be a map matching the <code>ContextDestroyedParameters</code>
+    production, with the <code>context</code> field set to |context id|,
+    and the <code>parent</code> field set to |parent id|.
+
+ 1. Let |body| be a map matching the
```
foolip, line 1705:
```suggestion
 1. Let |body| be a [=map=] matching the
```

foolip (commented):
Not the fastest review turnaround, but I was at least thorough in the end :)

My main concerns are:
- should we list the whole tree of browsing contexts instead of just one level at a time?
- how do the events for created/discarded browsing contexts match existing protocols?

foolip:
@k7z45 your review of this would be great, given your comments in https://github.com/w3c/webdriver-bidi/issues/42#issuecomment-721969911.

index.bs
jgraham: It's for e.g. shared workers. I actually backported this change from the realms work just to avoid pointless conflicts.

index.bs
jgraham: I made this recurse and added a `maxDepth` parameter in case we want to allow getting just top-level frames for example. I can remove that if we think it's overkill, but it's not much complexity in the spec.

index.bs
jgraham: Apparently I added that and forgot to remove this ;)

index.bs
jgraham: Hmm, so per spec I think the answer is "yes" and "currently top-down". But it depends on the exact details of when the algorithm is invoked. Step 1 discards all documents in the session history which also discards child browsing contexts of those documents. So I guess we actually want to only emit events for contexts that have an active document. And probably want the events to happen bottom up, so that child contexts emit the event before their parent? That seems like it would give stronger invariants.

index.bs
```diff
@@ -306,24 +318,26 @@ disabled for a given browsing context.
 <div algorithm>
 
 To determine if an <dfn export>event is enabled</dfn> given |session|,
-|event name| and |browsing context|:
+|event name| and |browsing contexts|:
```
k7z45, line 321: Given event name and browsing contexts, does the event counts as enabled only if it's enabled for all given browsing contexts?

index.bs
```diff
@@ -1462,3 +1485,278 @@ The [=remote end steps=] with |command parameters| are:
...
+    production with the <code>context</code> field set to |context id|, the
+    <code>parent</code> field set to |parent id|, the <code>url</code> field
+    set to |url|, and the <code>children</code> field set to |child info|.
+
+ 1. Return |context info|.
+
+</div>
+
+<div algorithm>
+  To <dfn>get the ancestor browsing contexts</dfn> given |parent id|, |depth|
```
k7z45, line 1593: Why is it called get the ancestor browsing contexts instead of get the children/descendant browsing contexts?
From step 5, it seems to be collecting the child contexts.

index.bs
```diff
@@ -1462,3 +1485,278 @@ The [=remote end steps=] with |command parameters| are:
...
+
+  1. Let |max depth| be the value of the <code>maxDepth</code> field of |command
+     parameters| if present, or 0 otherwise.
+
+  1. Let |depth| be 0.
+
+  1. Let |contexts| be the result of [=get the ancestor browsing contexts=],
+     given |parent id|, |depth|, and |max depth|.
+
+  1. Let |body| be a [=map=] matching the <code>BrowsingContextListResult</code>
```
k7z45, line 1666: Should it be BrowsingContextGetTreeResult instead of BrowsingContextListResult?

index.bs
```diff
@@ -170,13 +176,19 @@ ErrorResponse = {
...
 EmptyResult = {}
 
-Event = (
+Event = {
+  EventData,
   *text => any
+}
+
+EventData = (
+  ContextEvent
```
k7z45, line 191: This is not defined yet.

index.bs
```diff
@@ -1462,3 +1485,278 @@ The [=remote end steps=] with |command parameters| are:
...
+<div algorithm="remote end event trigger for browsingContext.contextCreated">
+When the [=create a new browsing context=] algorithm is invoked, after the
+[=active document=] of the browsing context is set, run the following steps:
+
+ 1. Let |context| be the newly created browsing context.
+
+ 1. Let |related contexts| be a set containing the [=parent browsing context=]
+    of |context|, if that is not null, or an empty set otherwise.
+
+ 1. Let |params| be the result of [=get the browsing context info=] given
```
k7z45, line 1699: What should be used for depth and max depth here?

index.bs
```diff
@@ -1462,3 +1485,278 @@ The [=remote end steps=] with |command parameters| are:
...
+
+ 1. Let |params| be a [=map=] matching the <code>BrowsingContextDestroyedParameters</code>
+    production, with the <code>context</code> field set to |context id|,
+    and the <code>parent</code> field set to |parent id|.
+
+ 1. Let |body| be a [=map=] matching the
+    <code>BrowsingContextDestroyedEvent</code> production, with the
+    <code>params</code> field set to |params|.
+
+ 1. Let |related contexts| be a set, containing |parent| is that is not null, or
```
k7z45, line 1751: Possible typo: "is" -> "if"

index.bs
```diff
@@ -157,7 +163,7 @@ Message = (
 
 CommandResponse = {
   id: uint,
-  value: ResultData,
+  ResultData,
```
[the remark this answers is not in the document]
bwalderman: `params` is used for command parameters and event parameters, but this is a command's return value. I thought we went with `value` here to match what WebDriver classic does. For reference, JSON-RPC and CDP use `result` here.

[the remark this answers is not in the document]
jgraham: Ugh, sorry I guess I meant result and got confused.

index.bs
jgraham: No; it's for any.

index.bs
jgraham: Good catch.

index.bs
jgraham: I've changed this a bit, but it remains very fragile in the way it integrates with HTML. I feel like we're going to end up revisting this later.

index.bs
```diff
@@ -83,6 +84,10 @@ spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
     text: ToString; url: sec-tostring
     text: Type; url: sec-ecmascript-data-types-and-values
     text: realm; url: sec-code-realms
+spec: HTML; urlPrefix: https://html.spec.whatwg.org/
+  type: dfn
+    text: a browsing context is discarded; url: a-browsing-context-is-discarded
+    text: create a new browsing context; url: creating-a-new-browsing-context
```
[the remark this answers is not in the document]
foolip: OK, so actually that term isn't exported. In HTML things aren't exported by default, the `data-export` attribute is needed, and it's not used. I've filed https://github.com/whatwg/html/issues/6194 to see if we could do this in a cleaner way.

index.bs
```diff
@@ -83,6 +84,10 @@ spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
     text: ToString; url: sec-tostring
     text: Type; url: sec-ecmascript-data-types-and-values
     text: realm; url: sec-code-realms
+spec: HTML; urlPrefix: https://html.spec.whatwg.org/
+  type: dfn
+    text: a browsing context is discarded; url: a-browsing-context-is-discarded
```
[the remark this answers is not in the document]
foolip: Let's leave this for https://github.com/whatwg/html/issues/6194 too.

index.bs
foolip: Ah, shared workers. Is that the only known case, so that for an implementation that doesn't support shared workers, this would always be either an empty set or a set of one? If so, adding a comment to that effect would be nice for those implementers.

index.bs
foolip: Yep, you're right. A notable case is https://w3c.github.io/webdriver-bidi/#object-id-for-an-object, but that's currently unreachable. We'll have to deal with this when we have an equivalent of https://w3c.github.io/webdriver/#dfn-execute-async-script. Nothing is broken right now though, AFAICT.

index.bs
foolip: Note added was "This includes the fragment component of the URL", thanks!

index.bs
foolip: Oh right, obviously, I was too zoomed in on observing a specific page.

annevk:
What's the plan for browsing context groups and top-level browsing contexts getting replaced (due to COOP)?

index.bs
foolip: @k7z45 note that this was resolved by https://github.com/w3c/webdriver-bidi/pull/62/commits/53d42f73673ed55cbe29dc16687b217d25eff3f2 with *depth* 0 and *max depth* 1, causing "get the descendent browsing contexts" to return null in its first step.

index.bs
foolip: I filed it before I got to this comment, but https://github.com/whatwg/html/issues/6194 might be where we figure this out.

Here's my current thinking. Rather than talking about context creation and destruction of individual browsing contexts, express this in terms of modifications to the browsing context tree. Creation/insertion is still pretty simple, I think a newly inserted browsing context can't yet have any descendant browsing siblings. But in the case of removal, send along the whole frame tree that was just removed in a single event.

index.bs
```diff
@@ -1462,3 +1501,283 @@ The [=remote end steps=] with |command parameters| are:
...
+ 1. Let |body| be a [=map=] matching the
+    <code>BrowsingContextDestroyedEvent</code> production, with the
+    <code>params</code> field set to |params|.
+
+ 1. Let |related contexts| be a set, containing |parent| if that is not null, or
+    an empty set otherwise.
+
+ 1. [=Emit an event=] with |body| and |related contexts|.
+
+Issue: the way this hooks into HTML feels very fragile.
```
foolip, line 1777:
```suggestion
Issue: the way this hooks into HTML feels very fragile. See https://github.com/whatwg/html/issues/6194
```

index.bs
```diff
@@ -1462,3 +1501,283 @@ The [=remote end steps=] with |command parameters| are:
...
+
+ 1. Let |related contexts| be a set, containing |parent| if that is not null, or
+    an empty set otherwise.
+
+ 1. [=Emit an event=] with |body| and |related contexts|.
+
+Issue: the way this hooks into HTML feels very fragile. See https://github.com/whatwg/html/issues/6194
+
+Issue: It's unclear if we ought to only fire this event for browsing
+contexts that have active documents; navigation can also cause contexts to
```
foolip, line 1780: CDP has a separate [`Page.frameNavigated`](https://chromedevtools.github.io/devtools-protocol/tot/Page/#event-frameNavigated) event. Navigating a browsing context will indeed result in the tree of browsing contexts changing, but I agree it's not clear how best to represent that.

index.bs
```diff
@@ -1749,38 +1750,35 @@ The [=remote end event trigger=] is:
...
-the [=browsing context group is removed=] (if relevant):
+following steps before running any steps in HTML:
 
-Note: The intent is that when a subtree of browsing contexts is removed, the
-<code>contextDestroyed</code> event is always emitted for a child before the
-parent.
+ 1. If the [=current session=] is null, return.
 
- 1. Let |context| be the browsing context being discarded.
+ 1. If the [=current session=]'s [=context discard=] flag is set, return.
```
foolip, line 1757: Rather than maintaining this flag and "a browsing context is discarded" at the right point in the algorithm to achieve the effect of getting called only once for the topmost browsing context being discarded, can we invent a "tree of browsing contexts discarded" hook that we wish that HTML had, and later go about adding it to HTML or letting HTML call WebDriver BiDi directly?

jgraham: We can of course; do you have a concrete example of what that should look like? Are you basically just imagining that the current discard a browsing context algorithm is split into two parts, one that's called externally and one that's called recursively for nested contexts? (maybe we could just handwave about the event happening when the algorithm is called but it isn't a recursive invocation, until the hook is ready)

index.bs
```diff
@@ -1440,3 +1467,281 @@ The [=remote end steps=] with |command parameters| are:
...
+
+<pre class="cddl local-cddl">
+
+BrowsingContextInfoList = [* BrowsingContextInfo]
+
+BrowsingContextInfo = {
+  context: BrowsingContext,
+  ?parent: BrowsingContext / null,
+  url: text,
+  children: BrowsingContextInfoList / null
```
foolip, line 1544:
```suggestion
  children: BrowsingContextInfoList
```

Being able to represent "no children" as both null and an empty list seems unnecessary.

jgraham: I think they mean slightly different things. `null` is used for "there may be additional contexts but we didn't recurse into them" and the empty list is used for "there are no child contexts". That's a subtle difference but it does seem useful to communicate it somehow.

foolip: I see https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/62.html#get-the-descendent-browsing-contexts returns null if *max depth* is reached. Similar to remote value serialization. OK then :)

index.bs
foolip: https://wicg.github.io/origin-policy/#monkeypatches is the style of handwavery I have in mind, patches that are explicitly intended to graduate into other specs.

We would probably patch https://html.spec.whatwg.org/multipage/window-object.html#a-browsing-context-is-discarded, there are many ways to do it, but the net effect should be a single call to WebDriver BiDi with the topmost browsing context about to be discarded, and no calls for any of its descendant browsing contexts.

Additional examples of monkey patching separated out quite nicely:
https://w3c.github.io/webappsec-csp/#integrations
https://w3c.github.io/webappsec-referrer-policy/#integration-with-html
https://w3c.github.io/webappsec-trusted-types/dist/spec/#integration-with-html

foolip (approved):
The HTML patch looks good to me. There are lots of other open comment threads, but I'm happy to approve this and unblock. If any open comment threads still require action, filing issues for them seems fine to me. (Didn't look closely.)

PR #68, opened 2020-11-04
Make commands run in parallel.

jgraham:
Implement the decision to make all commands async so that there is no
guaranteed ordering between the responses of different commands, and
we don't have to block processing one command on the previous one
having sent the response.

To make this work I inlined the process a command steps so that only
actually running the command happens in parallel and not the initial
error checking. This also means that we don't need infrastructure to
communicate the parallel result back to the main event loop, but can
just write the response directly assuming response writing is
implicitly serialized.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/68.html" title="Last updated on Nov 4, 2020, 3:21 PM UTC (2f79fa0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/68/e9a0822...2f79fa0.html" title="Last updated on Nov 4, 2020, 3:21 PM UTC (2f79fa0)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -160,7 +160,7 @@ Message = (
 
 CommandResponse = {
   id: uint,
-  ResponseData,
+  value: ResultData,
   *text => any
 }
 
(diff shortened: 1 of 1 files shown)


bwalderman (approved):

index.bs
```diff
@@ -575,31 +539,52 @@ To <dfn>handle an incoming message</dfn> given a [=WebSocket connection=]
...
- 1. Let |command id| and |value| be the two fields |result|'s data.
+    1. Let |command id| be |parsed|["<code>id</code>"].
 
- 1. Let |response| be a new [=map=] with the following properties:
+    1. Let |method| be |parsed|["<code>method</code>"]
 
-   <dl>
-      <dt>"id"</dt>
-      <dd>|command id|</dd>
+    1. Run the following steps in parallel:
```
k7z45, line 554: How to run the steps in parallel? They seem to have dependency on the previous step.

bwalderman: Looks like the definition of "parallel" here is borrowed from the HTML spec: https://html.spec.whatwg.org/multipage/infrastructure.html#parallelism

The substeps under "Run the following steps in parallel" start after the previous steps have completed. The substeps are run in order one at a time, but they don't block the main "handle an incoming message" algorithm, so it is free to handle more messages.

PR #69, opened 2020-11-06
Add GitHub workflow to test CDDL syntax

christian-bromann:
Over the last weeks/month I was working on a [CDDL parser](https://www.npmjs.com/package/cddl) as some sort of hobby project. It is now at a stage where it can be used to validate the cddl syntax of all cddl pre formatted texts we have in the spec. My next task is to actually validate the content of the cddl file (e.g. to make sure that all groups that are referenced are also defined etc.).

With the help of this I was able to expose and fix 3 cddl syntax errors we had in the spec already.

***
<a href="https://pr-preview.s3.amazonaws.com/christian-bromann/webdriver-bidi/pull/69.html" title="Last updated on Nov 13, 2020, 2:17 PM UTC (fa45d61)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/69/502ddf9...christian-bromann:fa45d61.html" title="Last updated on Nov 13, 2020, 2:17 PM UTC (fa45d61)">Diff</a>

### DIFF
--- a/scripts/cddl/generate.js
+++ b/scripts/cddl/generate.js
@@ -2,48 +2,65 @@
 
 const fs = require('fs')
 const path = require('path')
+const parse5 = require('parse5')
+
+const { getCDDLNodes } = require('./utils')
 
 const spec = path.resolve(__dirname, '..', '..', 'index.bs')
 const specContent = fs.readFileSync(spec, 'utf-8')
+const specParsed = parse5.parse(specContent)
 
-const preFormattedTexts = [...specContent.match(/<pre .*>([^<]*)<\/pre>/g)]
-    /**
-     * filter for CDDL code
-     */
-    .filter((pre) => pre.split('\n')[0].includes('cddl'))
-    /**
-     * drop HTML tags
-     */
-    .map((pre) => pre.split('\n').slice(1, -1))
-    /**
-     * remove obsolete whitespace
-     */
-    .map((pre) => {
-        const preceedingSpace = pre.reduce((prev, line) => {
-            if (line.length === 0) {
-                return prev
-            }
-    
-            const spaces = line.match(/^\s+/)
-            if (spaces) {
-                return Math.min(prev, spaces[0].length)
-            }
-    
-            return 0
-        }, Infinity)
-        return pre.map((line) => line.slice(preceedingSpace))
-    })
-    /**
-     * drop all cddl snippets without declarations, e.g.
-     * ```
-     * EmptyResult
-     * ```
-     */
-    .filter((pre) => pre.length > 1)
-    /**
-     * join code lines
-     */
-    .map((pre) => pre.join('\n'))
-    .join('\n')
+/**
+ * format the result and sort pre content to their corresponding classes
+ */
+const cddlContent = getCDDLNodes([specParsed]).reduce((val, node) => {
+    if (!val[node.nodeClass]) {
+        val[node.nodeClass] = []
+    }
+    val[node.nodeClass].push(node.content)
+    return val
+}, {})
 
-fs.writeFileSync(path.join(process.cwd(), 'index.cddl'), preFormattedTexts)
+/**
+ * iterate over all pre cddl classes found in the document, sanitize the
+ * content and write them to file.
+ */
+Object.entries(cddlContent).forEach(([cddlClass, content]) => {
+    const fileContent = content
+        /**
+         * remove obsolete whitespace
+         */
+        .map((entry) => {
+            const preceedingSpace = entry.split('\n').reduce((prev, line) => {
+                if (line.length === 0) {
+                    return prev
+                }
+
+                const spaces = line.match(/^\s+/)
+                if (spaces) {
+                    return Math.min(prev, spaces[0].length)
+                }
+
+                return 0
+            }, Infinity)
+            return entry
+                .split('\n')
+                .map((line) => line.slice(preceedingSpace))
+        })
+        /**
+         * drop all cddl snippets without declarations, e.g.
+         * ```
+         * EmptyResult
+         * ```
+         */
+        .filter((pre) => {
+            return pre.filter(Boolean).length > 1
+        })
+        /**
+         * join code lines
+         */
+        .map((pre) => pre.join('\n'))
+        .join('\n')
+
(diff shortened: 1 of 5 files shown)


scripts/cddl/generate.js
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env node
+
+const fs = require('fs')
+const path = require('path')
+
+const spec = path.resolve(__dirname, '..', '..', 'index.bs')
+const specContent = fs.readFileSync(spec, 'utf-8')
+
+const preFormattedTexts = [...specContent.match(/<pre .*>([^<]*)<\/pre>/g)]
```
jgraham, line 9: Parsing HTML with a regex makes me very nervous :) Is there some chance to use e.g. [1] or some other spec-compliant parser for parsing instead? 

[1] https://www.npmjs.com/package/parse5

scripts/cddl/generate.js
```diff
@@ -0,0 +1,49 @@
...
+        }, Infinity)
+        return pre.map((line) => line.slice(preceedingSpace))
+    })
+    /**
+     * drop all cddl snippets without declarations, e.g.
+     * ```
+     * EmptyResult
+     * ```
+     */
+    .filter((pre) => pre.length > 1)
```
jgraham, line 42: Why is this needed?

scripts/cddl/generate.js
```diff
@@ -0,0 +1,49 @@
...
+    /**
+     * join code lines
+     */
+    .map((pre) => pre.join('\n'))
+    .join('\n')
+
+fs.writeFileSync(path.join(process.cwd(), 'index.cddl'), preFormattedTexts)
```
jgraham, line 49: So I'm surprised we end up with a single schema, because afaict there ought to be two roots; one that matches commands (the "remote end schema") and one that matches responses, events and errors (the "local end schema").

jgraham (commented):
I"m really excited abou this, thanks!

A couple of questions:
* Do you support the `foo //= bar` syntax to extend exising groups? We can avoid a lot of boilerplate with that.
* How hard would it be to set up testing with this? e.g. we could check in some JSON that ought to parse and some that out to fail and verify we at least get that.
* Can we adjust the spec production pipeline to publish the schema alongside the spec? That seems useful.

scripts/cddl/generate.js
christian-bromann: We have a lot of
```html
    <pre class="cddl local-cddl">
        EmptyResult
    </pre>
```
which don't really add anything to the CDDL spec, this parses this out

scripts/cddl/generate.js
christian-bromann: We can split it into two CDDL files but all groups or variables used in either of them also need to be defined in both. I am sure we use certain groups in both schemas, don't we?

christian-bromann:
> * Do you support the `foo //= bar` syntax to extend exising groups? We can avoid a lot of boilerplate with that.

Yes, I've [added](https://github.com/christian-bromann/cddl/commit/5db7b32f0eb16950bf535013fd5cb71608664e56) this.

> * How hard would it be to set up testing with this? e.g. we could check in some JSON that ought to validate and some that out to fail and verify we at least get that.

Not sure what you are exactly want to validate but my next task is to validate not only the syntax but also the cddl content, e.g. check if all groups referenced are also defined in the file. Do you have more validation requirements in mind?

> * Can we adjust the spec production pipeline to publish the schema alongside the spec? That seems useful.

Yes. We could have an action that is being triggered on changes to the `index.bs` file and have the action automatically generate and push a new version of the cddl file. I would add this within a separate PR.

scripts/cddl/generate.js
jgraham: Yeah, we do but the spec tries to do this explicitly by having some fragments defined for both the remote and local end schemas.

jgraham:
> > * Do you support the `foo //= bar` syntax to extend exising groups? We can avoid a lot of boilerplate with that.
> 
> Yes, I've [added](https://github.com/christian-bromann/cddl/commit/5db7b32f0eb16950bf535013fd5cb71608664e56) this.

Cool, then I'll rewrite some parts to use it. I didn't use it at first because the previous CDDL lib I tried didn't support it.

> > * How hard would it be to set up testing with this? e.g. we could check in some JSON that ought to validate and some that out to fail and verify we at least get that.
> 
> Not sure what you are exactly want to validate but my next task is to validate not only the syntax but also the cddl content, e.g. check if all groups referenced are also defined in the file. Do you have more validation requirements in mind?

I mean having documents like `{"id": 1, "method": "session.subscribe", "params": {"contexts": ["a"], "events": ["Log"]}` and verifying whether they match the CDDL definitions. I'm not saying that should be part of this PR, but being able to add in some test fragments to verify that the schema works as expected and to provide examples for implementors seems valuable.

> > * Can we adjust the spec production pipeline to publish the schema alongside the spec? That seems useful.
> 
> Yes. We could have an action that is being triggered on changes to the `index.bs` file and have the action automatically generate and push a new version of the cddl file. I would add this within a separate PR.

Yep, in a different PR for sure.

christian-bromann:
> I mean having documents like `{"id": 1, "method": "session.subscribe", "params": {"contexts": ["a"], "events": ["Log"]}` and verifying whether they match the CDDL definitions.

Sounds good to me. Added it to christian-bromann/cddl#2 as next task.

scripts/cddl/generate.js
christian-bromann: currently I only see `pre` tags with `local-cddl` and `remote-cddl`. We could add the content of `pre` tags that only have the class `cddl`, e.g. `<pre class="cddl remote-cddl">` to both files. WDYT?

christian-bromann:
@jgraham updated the parsing script to use `parse5`. The script now generates two files `remote.cddl` and `local.cddl` and is designed to create more types of cddl files based on the `pre` class. Is there anything you would like to add?

scripts/cddl/generate.js
jgraham: I've used `class="local-cddl remote-cddl cddl"` in e.g. the contexts PR. Happy to do whatever is easiest for you here.

jgraham (approved):
I definitely think there's more to do here e.g. I think we ought to run on the post-bikeshed index.html document, but I'm so excited about this I"d rather get it landed and start using it than blocking on more work now.

PR #70, opened 2020-11-09
Initial cut at the script module

jgraham:
This adds a method to get a list of existing realms and an event when
a new realm is created. It is anticipated that the module will also be
used for further script-related commands such as script execution in
the future.

Realms are given an id and a type parameter according to the kind of
global they contain. The set of paramters is extensible to allow for
the possibility of further kinds in the future, but the anticipation
is they will be added to the specification, so it's a requirement that
unknown kinds are prefixed.

The getRealms method allows filtering by context to only get realms
related to the active document in that context (i.e. workers with that
document in the owner set or the window script for that
document). There isn't currently a way to get only nested workers or
similar; it's not really clear how useful this is.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/70.html" title="Last updated on Jan 14, 2021, 10:03 AM UTC (8685079)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/70/90c819d...8685079.html" title="Last updated on Jan 14, 2021, 10:03 AM UTC (8685079)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1858,7 +1858,7 @@ To <dfn>get the realm info</dfn> given |environment settings|:
  1. Return |realm info|
 
 Issue: We currently don't provide information about realms of unknown
-       types. That might be a problem foe e.g. extension-related realms.
+       types. That might be a problem for e.g. extension-related realms.
 
 
 Note: Future variations of this specification will retain the invariant that
@@ -1953,7 +1953,7 @@ The [=remote end steps=] with |command parameters| are:
 
 Issue: Extend this to also allow realm parents e.g. for nested workers? Or get all ancestor workers.
 
-Issue: We might want to have a more sophicticated filter system than just a
+Issue: We might want to have a more sophisticated filter system than just a
        literal match.
 
 </div>
@@ -2115,4 +2115,3 @@ To discard a browsing context |browsingContext|, run these steps:
 Issue: The actual patch might be better to split the algorithm into an outer algorithm that is
 called by external callers and an inner algorithm that's used for recursive calls. That's quite hard
 to express as a patch to the specification since it requires changing multiple parts.
-
--- a/index.bs
+++ b/index.bs
@@ -1621,7 +1621,7 @@ top-level contexts when no parent is provided.
    <dd>
     <pre class="cddl local-cddl">
         BrowsingContextGetTreeResult = {
-          contexts: [*BrowsingContextInfo]
+          contexts: BrowsingContextInfoList
         }
     </pre>
    </dd>
@@ -1898,9 +1898,11 @@ realm associated with the [=document=] currently loaded in a specified
    <dt>Return Type</dt>
    <dd>
     <pre class="cddl local-cddl">
-        ScriptGetRealmsResult = {
-          realms: [*RealmInfo]
-        }
+      RealmInfoList = [* RealmInfo]
+
+      ScriptGetRealmsResult = {
+        realms: RealmInfoList
+      }
     </pre>
    </dd>
 </dl>
--- a/index.bs
+++ b/index.bs
@@ -1816,9 +1816,6 @@ To <dfn>get the realm info</dfn> given |environment settings|:
    <dl>
      <dt>The [=Realm/global object=] specified by |environment settings| is a [=Window=] object
      <dd>
-      1. If [=Realm/global object=] specified by |environment settings| has a [=Document=]
-         that is not yet fully active, return null.
-
       1. Let |type| be "<code>window</code>".
 
      <dt>The [=Realm/global object=] specified by |environment settings| is a {{DedicatedWorkerGlobalScope}} object
--- a/index.bs
+++ b/index.bs
@@ -1747,9 +1747,9 @@ ScriptCommand = (ScriptGetRealmsCommand)
 
 ScriptResult = (ScriptGetRealmsResult)
 
-BrowsingContextEvent = (
-    BrowsingContextCreatedEvent //
-    BrowsingContextDestroyedEvent
+ScriptEvent = (
+    ScriptRealmCreatedEvent //
+    ScriptRealmDestroyedEvent
 )
 
 </pre>
@@ -1794,12 +1794,10 @@ Issue: This has the wrong error code
 RealmInfo = {
   realm: Realm,
   type: RealmType,
-  origin: text,
+  origin: text
 }
 
-RealmType = (
-  "window" / "dedicated-worker" / "shared-worker" / "service-worker" / "worker" / "paint-worklet" / "audio-worklet" / "worklet" / text
-)
+RealmType = "window" / "dedicated-worker" / "shared-worker" / "service-worker" / "worker" / "paint-worklet" / "audio-worklet" / "worklet" / text
 </pre>
 
 The <code>RealmInfo</code> type represents the properties of a realm.
--- a/index.bs
+++ b/index.bs
@@ -88,11 +88,13 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
   type: dfn
     text: a browsing context is discarded; url: a-browsing-context-is-discarded
     text: create a new browsing context; url: creating-a-new-browsing-context
+    text: environment settings object's Realm; url: environment-settings-object's-realm
     text: remove a browsing context; url: bcg-remove
     text: session history; url: session-history
     text: set up a window environment settings object; url: set-up-a-window-environment-settings-object
     text: set up a worker environment settings object; url: set-up-a-worker-environment-settings-object
-
+    text: worker event loop; url: worker-event-loop-2
+    text: worklet global scopes; url: concept-document-worklet-global-scopes
 </pre>
 
 <pre class="link-defaults">
@@ -116,7 +118,7 @@ This specification depends on the Infra Standard. [[!INFRA]]
 
 Network protocol messages are defined using CDDL. [[!RFC8610]]
 
-# Protocol {#protocol}
+# Protocol # {#protocol}
 
 This section defines the basic concepts of the WebDriver BiDi
 protocol. These terms are distinct from their representation at the
@@ -351,7 +353,7 @@ To <dfn>obtain a set of event names</dfn> given an |name|:
   1. Return [=success=] with data |events|.
 </div>
 
-# Transport {#transport}
+# Transport # {#transport}
 
 Message transport is provided using the WebSocket protocol.
 [[!RFC6455]]
@@ -702,9 +704,9 @@ with parameters |session| and |capabilities| is:
 
 </div>
 
-# Common Data Types {#data-types}
+# Common Data Types # {#data-types}
 
-## Remote Value ## {#data-types-remote-value}
+## Remote Value ## {#type-common-RemoteValue}
 
 Values accessible from the ECMAScript runtime are represented by a mirror
 object, specified as <code>RemoteValue</code>. The value's type is specified in
@@ -1261,7 +1263,7 @@ To <dfn>serialize as a mapping</dfn> given |iterable|, |max depth|,
   1. Return |serialized|
 </div>
 
-# Modules {#modules}
+# Modules # {#modules}
 
 ## The session Module ## {#module-session}
 
@@ -1489,7 +1491,7 @@ BrowsingContextEvent = (
 
 ### Types ### {#module-browsingcontext-types}
 
-#### The browsingContext.BrowsingContext Type #### {#types-browsingcontext}
+#### The browsingContext.BrowsingContext Type #### {#type-browsingContext-Browsingcontext}
 
 [=remote end definition=] and [=local end definition=]
 
@@ -1517,7 +1519,7 @@ To <dfn>get a browsing context</dfn> given |context id|:
 
 </div>
 
-#### The browsingContext.BrowsingContextInfo Type #### {#types-browsingcontextinfo}
+#### The browsingContext.BrowsingContextInfo Type #### {#type-browsingContext-BrowsingContextInfo}
 
 [=local end definition=]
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1469,3 +1479,446 @@ The [=remote end steps=] with |command parameters| are:
...
+
+ 1. Let |realm id| be the [=realm id=] for |realm|.
+
+ 1. Run the steps under the first matching condition:
+
+   <dl>
+     <dt>The [=Realm/global object=] specified by |environment settings| is a [=Window=] object
+     <dd>
+      1. If [=Realm/global object=] specified by |environment settings| has a [=Document=]
+         is not yet fully active, return null.
```
bwalderman, line 1766:
```suggestion
         that is not yet fully active, return null.
```

index.bs
```diff
@@ -1469,3 +1479,446 @@ The [=remote end steps=] with |command parameters| are:
...
+     <dd>
+       1. Let |type| be "<code>worker</code>".
+     <dt>The [=Realm/global object=] specified by |environment settings| is a {{SharedWorkerGlobalScope}} object
+
+    <dd>
+       1. Let |type| be "<code>sharedworker</code>".
+     <dt>The [=Realm/global object=] specified by |environment settings| is a {{ServiceWorkerGlobalScope}} object
+
+    <dd>
+       1. Let |type| be "<code>sharedworker</code>".
```
bwalderman, line 1780:
```suggestion
       1. Let |type| be "<code>serviceworker</code>".
```

index.bs
```diff
@@ -1469,3 +1479,446 @@ The [=remote end steps=] with |command parameters| are:
...
+  1. Return [=success=] with data |body|.
+
+
+Issue: Extend this to also allow realm parents e.g. for nested workers? Or get all ancestor workers.
+
+</div>
+
+### Events ### {#module-script-events}
+
+#### realmCreated #### {#event-realm-created}
```
bwalderman, line 1883: What about a `realmDestroyed` event? We could potentially fire this during the steps to [discard a document](https://html.spec.whatwg.org/#garbage-collection-and-browsing-contexts) for "window" and "worklet" realms, and during [terminate a worker](https://html.spec.whatwg.org/#terminate-a-worker) for the worker types.

bwalderman (dismissed):

index.bs
```diff
@@ -177,12 +178,14 @@ ErrorResponse = {
   ),
   message: text,
   stacktrace?: text,
+  *text => any
```
foolip, line 181: This is an addition to `ErrorResponse`. Does "respond with an error" also need to be updated to populate this, or how does it fit together?

index.bs
```diff
@@ -1715,3 +1705,220 @@ When [=a browsing context is discarded=], run the following steps:
...
+)
+
+RealmInfo = {
+  realm: Realm,
+  type: RealmType,
+  origin: text,
+}
+
+RealmType = (
+  "window" / "worker" / "sharedworker" / "serviceworker" / "worklet" / text
```
foolip, line 1729: Since specific types of workers are enumerated, should different worklets also not be their own types?

I'm assume that "worker" here means only dedicated workers, it's not a generic type. Should we promise in a note that all workers will always end with the string "worker" and nothing else ever will, so that clients can test if something is any kind of worker without getting nervous?

The same setup should then be OK for worklets too.

index.bs
```diff
@@ -1715,3 +1705,220 @@ When [=a browsing context is discarded=], run the following steps:
...
+ 1. If |realm id| is null, return [=success=] with data null.
+
+ 1. If there is no realm with [=realm id=] |realm id| return
+    [=error=] with [=error code=] [=no such frame=]
+
+ 1. Let |realm| be the browsing context with id |realm id|.
+
+ 1. Return [=success=] with data |realm|
+
+Issue: This has the wrong error code
```
foolip, line 1749: How do we add a "no such realm" error code? Can we define our own error codes in the spec, even if explicitly as monkey-patching that needs upstreaming?

index.bs
```diff
@@ -1715,3 +1705,220 @@ When [=a browsing context is discarded=], run the following steps:
...
+
+<div algorithm>
+To <dfn>get a realm</dfn> given |realm id|:
+
+ 1. If |realm id| is null, return [=success=] with data null.
+
+ 1. If there is no realm with [=realm id=] |realm id| return
+    [=error=] with [=error code=] [=no such frame=]
+
+ 1. Let |realm| be the browsing context with id |realm id|.
```
foolip, line 1745:
```suggestion
 1. Let |realm| be the [=realm=] with [=realm id|id=] |realm id|.
```

index.bs
```diff
@@ -1715,3 +1705,220 @@ When [=a browsing context is discarded=], run the following steps:
...
+Each [=realm=] has an associated <dfn export>realm id</dfn>, which is a string
+uniquely identifying that realm. This is implicitly set when the realm is
+created.
+
+<div algorithm>
+To <dfn>get a realm</dfn> given |realm id|:
+
+ 1. If |realm id| is null, return [=success=] with data null.
+
+ 1. If there is no realm with [=realm id=] |realm id| return
```
foolip, line 1742:
```suggestion
 1. If there is no [=realm=] with [=realm id|id=] |realm id| return
```

index.bs
```diff
@@ -1715,3 +1705,220 @@ When [=a browsing context is discarded=], run the following steps:
...
+       1. Let |type| be "<code>serviceworker</code>".
+
+    <dt>The [=Realm/global object=] specified by |environment settings| is a {{WorkletGlobalScope}} object
+    <dd>
+       1. Let |type| be "<code>worklet</code>".
+
+    <dt>Otherwise:
+    <dd>
+      1. Let |type| be an implementation-defined string, consisting of a prefix,
+         followed by a colon ("<code>:</code>") followed by a type name.
```
foolip, line 1789: I think including this here is risky. Some implementer way well think we're suggesting that they use "goog:newkindofworklet" here down the line, assuming we do in fact separate out types of worklets.

I'd rather we assert that this is unreachable, so that it's explicitly unexpected and something that needs updating here if it ever happens.

index.bs
```diff
@@ -1715,3 +1705,220 @@ When [=a browsing context is discarded=], run the following steps:
...
+ 1. Return [=success=] with data |realm|
+
+Issue: This has the wrong error code
+
+</div>
+
+<div algorithm>
+To <dfn>get the realm info</dfn> given |environment settings|:
+
+ 1. Let |realm| be |environment settings|' [=realm execution context=]'s Realm component.
```
foolip, line 1756: Funny, this unliked "Realm component" is used in HTML as well. I guess it's a realm :)

index.bs
```diff
@@ -1715,3 +1705,220 @@ When [=a browsing context is discarded=], run the following steps:
...
+
+ 1. Let |realm id| be the [=realm id=] for |realm|.
+
+ 1. Run the steps under the first matching condition:
+
+   <dl>
+     <dt>The [=Realm/global object=] specified by |environment settings| is a [=Window=] object
+     <dd>
+      1. If [=Realm/global object=] specified by |environment settings| has a [=Document=]
```
foolip, line 1765: Why this early return? When will this scenario occur? If it happens very rarely, then it's likely clients will come to assume the return value is a string, and just break in rare conditions.

If there are ways of using a realm that don't make sense when the document isn't yet fully active, maybe those commands should fail in some way instead? But if it's rare, the problem will remain.

index.bs
```diff
@@ -1715,3 +1705,220 @@ When [=a browsing context is discarded=], run the following steps:
...
+ 1. Return |realm info|
+
+</div>
+
+### Commands ### {#module-script-commands}
+
+#### getRealms ####  {#command-script-getrealms}
+
+The <dfn export for=commands>getRealms</dfn> command returns a list of all
+realms, optionally filtered to realms of a specific type, or to the realm
```
foolip, line 1807:
```suggestion
realms, optionally filtered to [=realms=] of a specific type, or to the realm
```

Perhaps we can also define type separately, so that it's linkable here?

index.bs
```diff
@@ -1715,3 +1705,220 @@ When [=a browsing context is discarded=], run the following steps:
...
+associated with the document currently loaded in a specified browsing context.
```
foolip, line 1808:
```suggestion
associated with the [=document=] currently loaded in a specified [=browsing context=].
```

index.bs
```diff
@@ -1715,3 +1705,220 @@ When [=a browsing context is discarded=], run the following steps:
...
+   <dd>
+    <pre class="cddl local-cddl">
+        GetRealmsResult = {
+          realms: [*RealmInfo]
+        }
+    </pre>
+   </dd>
+</dl>
+
+<div algorithm="remote end steps for script.getRealms">
```
foolip, line 1835: I haven't reviewed this in detail, but wonder if it will change in light of the discussion of browsing context trees in https://github.com/w3c/webdriver-bidi/pull/62?

@sadym-chromium do you know what represents a JavaScript realm in CDP, and how they're discovered? Is it a `Runtime.something`, a `Target.something`, or something else? I'm not sure what to look for to see if this seems similar to CDP or not :)

index.bs
```diff
@@ -1715,3 +1705,220 @@ When [=a browsing context is discarded=], run the following steps:
...
+    [=responsible document=]'s [=Document/browsing context=] to |related contexts|.
+
+    Otherwise if the [=Realm/global object=] specified by |settings| is a
+    {{WorkerGlobalScope}}, for each |owner| in the [=Realm/global object=]'s [=owner
+    set=], if |owner| is a [=Document=], append |owner| to |related contexts|.
+
+ 1. Let |body| be a map matching the <code>RealmCreatedEvent</code>
+    production, with the <code>params</code> field set to |realm info|.
+
+ 1. [=Emit an event=] with |body| and |related contexts|.
```
foolip, line 1922: Oops, *related contexts* is a set of documents, not a set of browsing contexts, will need adjusting.

index.bs
```diff
@@ -1715,3 +1705,220 @@ When [=a browsing context is discarded=], run the following steps:
...
+ 1. Let |realm info| be be the result of [=get the realm info=] given
+    |environment settings|.
+
+ 1. Let |related contexts| be an empty set.
+
+ 1. If the [=responsible document=] of |settings| is a [=Document=], append the
+    [=responsible document=]'s [=Document/browsing context=] to |related contexts|.
+
+    Otherwise if the [=Realm/global object=] specified by |settings| is a
+    {{WorkerGlobalScope}}, for each |owner| in the [=Realm/global object=]'s [=owner
```
foolip, line 1916: OK, so the "for each" here is how we can end up with multiple members in *related contexts*, which can happen for shared workers.

However, a shared worker is only created once right, and some browsing context has to be the one that brought it about. Can there in fact be multiple owners at the time a realm is created?

In the case that a browsing context later joins the party with a shared worker, is an event needed to signal that, or do we pretent it was created again?

index.bs
bwalderman: These are "ExecutionContexts" in CDP: https://chromedevtools.github.io/devtools-protocol/tot/Runtime/#event-executionContextCreated

You get a batch of `Runtime.executionContextCreated` events when the `Runtime` domain is first enabled. I don't recall the exact details but I believe the [ExecutionContextDescription](https://chromedevtools.github.io/devtools-protocol/tot/Runtime/#type-ExecutionContextDescription) has the necessary info to associate it with a Frame.

index.bs
jgraham: I think we should just define error codes explicitly in this spec, in a way that's broadly compatible with WebDriver/HTTP. But I don't think we should do that in this PR.

index.bs
jgraham: I think we could at least have multiple contexts associated with a worker if we are emitting events because we subscribed to the `script.realmCreated` event for some existing contexts and we're backfilling. The spec doesn't cover that yet but I think it will.

I don't know what CDP does with shared workers; could be interesting to investigate.

index.bs
jgraham: Let's handle the extensible discussion in a seperate PR.

index.bs
jgraham: I don't really know how to get a list of all defined worklet types, but I agree having some convention around the naming of subtypes will help.

index.bs
jgraham: So for now I've mad it just skip unknown realm types that aren't a subtype of worker or worklet. That makes some sense because we don't have spec text for how to handle them in other cases.

index.bs
jgraham: I'm not sure quite what kind of change you're thinking of, so I haven't changed anything yet.

index.bs
jgraham: This is the extensibility stuff we were discussing elsewhere. I suggest we don't block this PR on it, but clean up all the productions that ought to be extensible later.

index.bs
jgraham: I added this. Working out where it's should be invoked is hard because that's intentionally unexposed to content...

index.bs
jgraham: I removed this condition for now.

index.bs
foolip: When posting this comment I didn't understand what `*text => any` was and thought it was a new `text` field. But yeah, let's discuss in https://github.com/w3c/webdriver-bidi/issues/79.

index.bs
foolip: Sounds good!

index.bs
foolip: Thanks @bwalderman! I don't know anything that needs to be changed at this point.

index.bs
```diff
@@ -1816,21 +1857,32 @@ To <dfn>get the realm info</dfn> given |environment settings|:
 
  1. Return |realm info|
 
+Issue: We currently don't provide information about realms of unknown
+       types. That might be a problem foe e.g. extension-related realms.
```
foolip, line 1861:
```suggestion
       types. That might be a problem for e.g. extension-related realms.
```

index.bs
```diff
@@ -1816,21 +1857,32 @@ To <dfn>get the realm info</dfn> given |environment settings|:
...
+
+
+Note: Future variations of this specification will retain the invariant that
```
foolip, line 1864: Excellent!

index.bs
```diff
@@ -1895,17 +1953,20 @@ The [=remote end steps=] with |command parameters| are:
 
 Issue: Extend this to also allow realm parents e.g. for nested workers? Or get all ancestor workers.
 
+Issue: We might want to have a more sophicticated filter system than just a
```
foolip, line 1956:
```suggestion
Issue: We might want to have a more sophisticated filter system than just a
```

foolip (approved):

index.bs
foolip: OK so let's not block this PR on investigating that, seems like a fine thing to look at while prototyping an implementation.

foolip (approved):
Still approved.

PR #73, opened 2020-11-17
Start adding log module

jgraham:
This adds a domain with a single log.entryAdded event. It currently handles
both console logs and javascript execptions, but further additions are
going to be required for things like network errors.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/73.html" title="Last updated on Feb 12, 2021, 3:50 PM UTC (aab50a3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/73/3c7a3b3...aab50a3.html" title="Last updated on Feb 12, 2021, 3:50 PM UTC (aab50a3)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2141,8 +2141,6 @@ related to logging.
 
 <pre class="cddl remote-cddl">
 
-Command = (GetCommand)
-
 LogEvent = (
   LogEntryAddedEvent
 )
@@ -2155,6 +2153,12 @@ LogEvent = (
 <pre class="cddl local-cddl">
 
 LogLevel = "debug" / "info" / "warning" / "error"
+
+LogEntry = (
+  GenericLogEntry //
+  ConsoleLogEntry //
+  JavascriptLogEntry
+)
 
 BaseLogEntry = {
   level: LogLevel,
@@ -2187,8 +2191,9 @@ Each log event is represented by a <code>LogEntry</code> object. This has a
 <code>type</code> property which represents the type of log entry added, a
 <code>level</code> property representing severity, a <code>text</code> property
 with the log message string itself, and a <code>timestamp</code> property
-corresponding to the time the log entry was generated. The <code>extra</code>
-property contains additional data specific to the given type.
+corresponding to the time the log entry was generated. Specific variants of the
+<code>LogEntry</code> are used to represent logs from different sources, and
+provide additional fields specific to the entry type.
 
 #### log.StackFrame #### {#types-log-stackframe}
 
@@ -2206,7 +2211,7 @@ StackFrame = {
 A frame in a stacktrace is represented by a <code>StackFrame</code> object. This
 has a <code>url</code> property, which represents the URL of the script, a
 <code>functionName</code> property which represents the name of the executing
-function, an <code>lineNumber</code> and <code>columnNumber</code> properties,
+function, and <code>lineNumber</code> and <code>columnNumber</code> properties,
 which represent the line and column number of the executed code.
 
 The <dfn>current stack trace</dfn> is a representation of the stack of the
@@ -2223,7 +2228,7 @@ behaviour here is implementation defined, but the general process is as follows:
 
     1. Let |functionName| be the name of |frame|'s associated function.
 
-    1. Let |lineNumber| and |columnNumber| be the on-based line and zero-based
+    1. Let |lineNumber| and |columnNumber| be the one-based line and zero-based
        column numbers, respectively, of the location in |frame|'s associated
        script resource corresponding to |frame|.
 
@@ -2247,11 +2252,7 @@ behaviour here is implementation defined, but the general process is as follows:
       <pre class="cddl local-cddl">
         LogEntryAddedEvent = {
          method: "log.entryAdded",
-         params: (
-             GenericLogEntry //
-             ConsoleLogEntry //
-             JavascriptLogEntry
-         )
+         params: LogEntry,
        }
       </pre>
    </dd>
@@ -2293,7 +2294,7 @@ ignore>options</var>:
 
 1. Let |entry| be a map matching the <code>ConsoleLogEntry</code> production,
    with the the <code>level</code> field set to |level|, the <code>text</code>
-   field set to |text|, the <code>timestap</code> field set to |timestamp|, the
+   field set to |text|, the <code>timestamp</code> field set to |timestamp|, the
    <code>stackTrace</code> field set to |stack| if |stack| is not null, or
    omitted otherwise, the |method| field set to |method|, the <code>realm</code>
    field set to |realm| and the <code>args</code> field set to |serialized
@@ -2301,8 +2302,6 @@ ignore>options</var>:
 
 1. Let |body| be a map matching the <code>LogEntryAddedEvent</code> production, with
    the <code>params</code> field set to |entry|.
-
-1. Let |related contexts| be an empty set.
 
 1. Let |settings| be the [=current settings object=]
 
@@ -2322,7 +2321,7 @@ ignore>line number</var>, <var ignore>column number</var>, |message| and
 
 1. Let |entry| be a map matching the <code>JavascriptLogEntry</code> production,
    with <code>level</code> set to "<code>error</code>", <code>text</code> set to
-   |message|, and the <code>timestap</code> field set to |timestamp|.
+   |message|, and the <code>timestamp</code> field set to |timestamp|.
 
 1. Let |related contexts| be the result of [=get related contexts=] given |settings|.
 
--- a/index.bs
+++ b/index.bs
@@ -74,7 +74,6 @@ spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
     text: EnumerableOwnPropertyNames; url: sec-enumerableownpropertynames
     text: Get; url: sec-get
     text: HasProperty; url: sec-hasproperty
-    text: internal slot; url: sec-object-internal-methods-and-internal-slots
     text: IsArray; url: sec-isarray
     text: IsCallable; url: sec-iscallable
     text: IsPromise; url: sec-ispromise
@@ -87,6 +86,8 @@ spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
     text: ToString; url: sec-tostring
     text: Type; url: sec-ecmascript-data-types-and-values
     text: current realm record; url: current-realm
+    text: internal slot; url: sec-object-internal-methods-and-internal-slots
+    text: primitive value; url: sec-primitive-value
     text: realm; url: sec-code-realms
     text: time value; url: sec-time-values-and-time-range
 spec: HTML; urlPrefix: https://html.spec.whatwg.org/
@@ -1159,8 +1160,10 @@ To <dfn>serialize as a remote value</dfn> given an |value|, a |max depth|,
                Otherwise, let |children| be an empty list and, for each node
                |child| in the [=children=] of |value|:
 
+              1. Let |child depth| be |max depth| - 1 if |max depth| is not null, or null otherwise.
+
               1. Let |serialized| be the result of [=serialize as a remote value=]
-                 with |child|, |max depth| - 1, |node details| and
+                 with |child|, |child depth|, |node details| and
                  |set of known objects|.
 
               1. Append |serialized| to |children|.
@@ -1184,9 +1187,14 @@ To <dfn>serialize as a remote value</dfn> given an |value|, a |max depth|,
                1. Let |shadow root| be |value|'s [=Element/shadow root=].
 
                1. If |shadow root| is null, let |serialized shadow| be null.
-                  Otherwise let |serialized shadow| be the result of
-                  [=serialize as a remote value=] with |shadow root|, |max depth| - 1,
-                  false and |set of known objects|.
+                  Otherwise run the following substeps:
+
+                 1.  Let |child depth| be |max depth| - 1 if |max depth| is not
+                    null, or null otherwise.
+
+                 1. Let |serialized shadow| be the result of
+                    [=serialize as a remote value=] with |shadow root|, |child depth|,
+                    false and |set of known objects|.
 
                   Note: this means the <code>objectId</code> for the shadow root
                   will be serialized irrespective of whether the shadow is open or closed,
@@ -1249,9 +1257,12 @@ To <dfn>serialize as a list</dfn> given |iterable|, |max depth|,
 
   1. For each |child value| in |iterable|:
 
-    1. Let |serialized child| be the result of [=serialize as a
-       remote value=] with arguments |child value|, |max
-       depth| - 1, |node details| and |set of known objects|.
+    1. Let |child depth| be |max depth| - 1 if |max depth| is not null, or null
+       otherwise.
+
+    1. Let |serialized child| be the result of [=serialize as a remote value=]
+       with arguments |child value|, |child depth|, |node details| and |set of
+       known objects|.
 
     1. Append |serialized child| to |serialized|.
 
@@ -1275,14 +1286,17 @@ To <dfn>serialize as a mapping</dfn> given |iterable|, |max depth|,
 
     1. Let |key| be |property|[0] and let |value| be |property|[1]
 
+    1. Let |child depth| be |max depth| - 1 if |max depth| is not null, or null
+       otherwise.
+
     1. If [=Type=](|key|) is String, let |serialized key| be |child key|,
        otherwise let |serialized key| be the result of [=serialize as a remote
-       value=] with arguments |child key|, |max depth| - 1, |node details| and
+       value=] with arguments |child key|, |child depth|, |node details| and
        |set of known objects|.
 
-    1. Let |serialized value| be the result of [=serialize as a
-       remote value=] with arguments |value|, |max
-       depth| - 1, |node details| and |set of known objects|.
+    1. Let |serialized value| be the result of [=serialize as a remote value=]
+       with arguments |value|, |child depth|, |node details| and |set of known
+       objects|.
 
     1. Let |serialized child| be («|serialized key|, |serialized value|»).
 
@@ -2189,13 +2203,17 @@ optionally, <var ignore>options</var>:
 
 1. Let |timestamp| be a [=time value=] representing the current date and time in UTC.
 
-1. Let |text| be an implementation-defined string representing the arguments
-   passed as |args|, or null if there is no appropriate string representation.
+1. Let |text| be an empty string.
+
+1. For each |arg| in |args|:
+
+  1. If |arg| is a [=primitive value=], append [=ToString=](|arg|) to
+     |text|. Otherwise append an implementation-defined string to |text|.
 
 1. Let |serialized args| be a new list.
 
 1. For each |arg| of |args|, append the result of [=serialize as a remote
-   value=] given |arg| to |serialized args|.
+   value=] given |arg|, null, true, and an empty [=set=] to |serialized args|.
 
 1. Let |realm| be the [=realm id=] of the [=current Realm Record=].
 
@@ -2221,7 +2239,7 @@ optionally, <var ignore>options</var>:
 
 1. [=Emit an event=] with |body| and |related contexts|.
 
-After an implementation is required to [=report an exception=] from |script|,
+Immediately after an implementation is required to [=report an exception=] from |script|,
 and the error is [=not handled=]:
 
 1. Let |settings| be |script|'s [=script/settings object=].
@@ -2253,7 +2271,10 @@ and the error is [=not handled=]:
 
 Issue: Add support for stacktraces
 
-Issue: Lots more things require logging
+Issue: Lots more things require logging. CDP has LogEntryAdded types xml,
+javascript, network, storage, appcache, rendering, security, deprecation,
+worker, violation, intervention, recommendation, other. These are in addition to
+the js exception and console API types that are represented by different methods.
 
 Issue: Allow implementation-defined log types
 
--- a/index.bs
+++ b/index.bs
@@ -2278,7 +2278,7 @@ ignore>options</var>:
    field set to |stack| if |stack| is not null, or omitted otherwise, and the
    <code>extra</code> field set to |extra|.
 
-1. Let |body| be a map matching the <code>LogEntryAdded</code> production, with
+1. Let |body| be a map matching the <code>LogEntryAddedEvent</code> production, with
    the <code>params</code> field set to |entry|.
 
 1. Let |related contexts| be an empty set.
--- a/index.bs
+++ b/index.bs
@@ -2243,7 +2243,7 @@ ignore>options</var>:
 1. If |logLevel| is "<code>error</code>" or "<code>assert</code>", let |level|
    be "<code>error</code>". If |logLevel| is "<code>debug</code>" or
    "<code>trace</code>" let |level| be "<code>debug</code>". If |logLevel| is
-   "<code>warn</code>", let |level| be "<code>warning</code>". Otherwise let
+   "<code>warn</code>" or "<code>warning</code>", let |level| be "<code>warning</code>". Otherwise let
    |level| be "<code>info</code>".
 
 1. Let |timestamp| be a [=time value=] representing the current date and time in UTC.
--- a/index.bs
+++ b/index.bs
@@ -2252,6 +2252,8 @@ ignore>options</var>:
 
 1. For each |arg| in |args|:
 
+  1. If |arg| is not the first entry in |args|, append a U+0020 SPACE to |text|.
+
   1. If |arg| is a [=primitive value=], append [=ToString=](|arg|) to
      |text|. Otherwise append an implementation-defined string to |text|.
 
--- a/index.bs
+++ b/index.bs
@@ -62,6 +62,8 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
     text: window handle; url: dfn-window-handle
 spec: CONSOLE; urlPrefix: https://console.spec.whatwg.org
   type: dfn
+    text: formatter; url: formatter
+    text: formatting specifier; url: formatting-specifiers
     text: printer; url: printer
 spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
   type: dfn
@@ -2279,7 +2281,14 @@ ignore>options</var>:
 
 1. Let |text| be an empty string.
 
-1. For each |arg| in |args|:
+1. If [=Type=](||args|[0]) is String, and |args|[0] contains a [=formatting
+   specifier=], let |formatted args| be [=Formatter=](|args|). Otherwise let
+   |formatted args| be |args|.
+
+   Issue: This is underdefined in the console spec, so it's unclar if we can get
+   interoperable behaviour here.
+
+1. For each |arg| in |formatted args|:
 
   1. If |arg| is not the first entry in |args|, append a U+0020 SPACE to |text|.
 
--- a/index.bs
+++ b/index.bs
@@ -60,6 +60,9 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
     text: WebDriver new session algorithm; url: dfn-webdriver-new-session-algorithm
     text: web element reference; url: dfn-web-element-reference
     text: window handle; url: dfn-window-handle
+spec: CONSOLE; urlPrefix: https://console.spec.whatwg.org
+  type: dfn
+    text: printer; url: printer
 spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
   type: dfn
     text: Array; url: sec-array-objects
@@ -83,12 +86,17 @@ spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
     text: ToDateString; url: sec-todatestring
     text: ToString; url: sec-tostring
     text: Type; url: sec-ecmascript-data-types-and-values
+    text: current realm record; url: current-realm
     text: realm; url: sec-code-realms
+    text: time value; url: sec-time-values-and-time-range
 spec: HTML; urlPrefix: https://html.spec.whatwg.org/
   type: dfn
+    text: WindowProxy; url: windowproxy
     text: a browsing context is discarded; url: a-browsing-context-is-discarded
     text: create a new browsing context; url: creating-a-new-browsing-context
     text: environment settings object's Realm; url: environment-settings-object's-realm
+    text: not handled; url: concept-error-nothandled
+    text: report an error; url: report-the-error
     text: remove a browsing context; url: bcg-remove
     text: session history; url: session-history
     text: set up a window environment settings object; url: set-up-a-window-environment-settings-object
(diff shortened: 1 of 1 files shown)


jgraham:
More to do here but this is for early feedback on the general form.

index.bs
```diff
@@ -1710,3 +1738,161 @@ When [=a browsing context is discarded=], run the following steps:
...
+}
+
+</pre>
+
+Each log event is represented by a <code>LogEntry</code> object. This has a
+<code>kind</code> property which represents the kind of log entry added, a
+<code>level</code> property representing serverity, a <code>text</code> property
+with the log message string itself, and a <code>timesteamp</code> property
+corresponding to the time the log entry was generated. The <code>extra</code>
+property contains additional data specific to the given type.
```
christian-bromann, line 1800: I wonder if we could describe the properties as part of the CDDL definition instead of spec text, e.g.

```
LogEntry = {
  type: text,             # represents the kind of log entry added
  level: LogLevel,        # log severity
  text: text / null,      # log message
  timestamp: int,         # time the log entry was generated
  extra: { *text => any } # additional data specific to the given type
}
```

This could be useful later on for code generation.

index.bs
```diff
@@ -1710,3 +1738,161 @@ When [=a browsing context is discarded=], run the following steps:
...
+JavascriptLogExtra = {
+  lineNumber: int,
+  columnNumber: int,
+  url: text
+}
+
+</pre>
+
+Each log event is represented by a <code>LogEntry</code> object. This has a
+<code>kind</code> property which represents the kind of log entry added, a
```
christian-bromann, line 1796:
```suggestion
<code>type</code> property which represents the type of log entry added, a
```

index.bs
```diff
@@ -1710,3 +1738,161 @@ When [=a browsing context is discarded=], run the following steps:
...
+<code>level</code> property representing serverity, a <code>text</code> property
```
christian-bromann, line 1797:
```suggestion
<code>level</code> property representing severity, a <code>text</code> property
```

christian-bromann (commented):
some cddl related comments

index.bs
```diff
@@ -1710,3 +1738,161 @@ When [=a browsing context is discarded=], run the following steps:
...
+<pre class="cddl local-cddl">
+
+LogLevel = "debug" / "info" / "warning" / "error"
+
+LogEntry = {
+  type: text,
+  level: LogLevel,
+  text: text / null,
+  timestamp: int,
+  extra: { *text => any }
```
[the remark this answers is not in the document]
christian-bromann: this has been fixed.

index.bs
jgraham: Yeah, we definitely could do that, but it would be quite different from the way that IDL fragments work where all the non-normative descriptive stuff is in the text. I'd be interested in other opinions here.

whimboo:
In the last days I implemented the `Runtime.consoleAPICalled` from CDP for Firefox. And as noticed this kind of event is used for calls into the Console API (eg. console.log(), console.error(), ...), and for console messages. Any Javascript error is instead routed through `Log.entryAdded`.

As seen here we want all to be reported through the same kind event, and allow pre-filtering for types? How would we differentiate between a Javascript error and a call to `console.error()`? For now those would be identical, right? And I wonder if we have to differentiate between those.

index.bs
```diff
@@ -1710,3 +1738,161 @@ When [=a browsing context is discarded=], run the following steps:
...
+         params: LogEntry
+       }
+      </pre>
+   </dd>
+</dl>
+The [=remote end event trigger=] is:
+
+<div algorithm="remote end event trigger for log.entryAdded">
+
+When the [=Printer=] algorithm is invoked with arguments |logLevel|, |args| and,
```
foolip, line 1819:
```suggestion
When the [=Printer=] operation is invoked with arguments |logLevel|, |args| and,
```

https://console.spec.whatwg.org/#printer calls itself an operation.

index.bs
```diff
@@ -60,6 +60,9 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
     text: WebDriver new session algorithm; url: dfn-webdriver-new-session-algorithm
     text: web element reference; url: dfn-web-element-reference
     text: window handle; url: dfn-window-handle
+spec: CONSOLE; urlPrefix: https://console.spec.whatwg.org
```
foolip, line 63: Want to file an issue about integration hooks for this?

index.bs
```diff
@@ -1710,3 +1738,161 @@ When [=a browsing context is discarded=], run the following steps:
...
+   </dd>
+</dl>
+The [=remote end event trigger=] is:
+
+<div algorithm="remote end event trigger for log.entryAdded">
+
+When the [=Printer=] algorithm is invoked with arguments |logLevel|, |args| and,
+optionally, <var ignore>options</var>:
+
+1. If |logLevel| is "<code>error</code>" or "<code>assert</code>", let |level|
```
foolip, line 1822: Good stuff, this takes of the sort of problem I had in https://github.com/puppeteer/puppeteer/issues/6255.

index.bs
```diff
@@ -1710,3 +1738,161 @@ When [=a browsing context is discarded=], run the following steps:
...
+   be "<code>error</code>". If |logLevel| is "<code>debug</code>" or
+   "<code>trace</code>" let |level| be "<code>debug</code>". If |logLevel| is
+   "<code>warn</code>", let |level| be "<code>warning</code>". Otherwise let
+   |level| be "<code>info</code>".
+
+1. Let |timestamp| be a [=time value=] representing the current date and time in UTC.
+
+1. Let |text| be an implementation-defined string representing the arguments
```
foolip, line 1830: Oh, well, that's unfortunate. How large are the differences between implementations here? Could tests written for this event assert anything at all about what the text is? It seems like `console.log("foo")` should have the text "foo", at the very least.

index.bs
```diff
@@ -1710,3 +1738,161 @@ When [=a browsing context is discarded=], run the following steps:
...
+   passed as |args|, or null if there is no appropriate string representation.
+
+1. Let |serialized args| be a new list.
+
+1. For each |arg| of |args|, append the result of [=serialize as a remote
+   value=] given |arg| to |serialized args|.
```
foolip, line 1836: This needs additional arguments for "serialize as a remote value".

index.bs
```diff
@@ -1710,3 +1738,161 @@ When [=a browsing context is discarded=], run the following steps:
...
+
+1. Let |realm| be the [=realm id=] of the [=current Realm Record=].
+
+1. Let |extra| be a map matching the <code>ConsoleLogExtra</code> production,
+   with the <code>type</code> field set to |logLevel|, the <code>real</code>
```
foolip, line 1841:
```suggestion
   with the <code>type</code> field set to |logLevel|, the <code>realm</code>
```

index.bs
```diff
@@ -1710,3 +1738,161 @@ When [=a browsing context is discarded=], run the following steps:
...
+   field set to |realm| and the <code>args</code> field set to |serialized
+   args|.
+
+1. Let |entry| be a map matching the <code>LogEntry</code> production, with the
```
foolip, line 1845: This is a lot of glue code. As we keep going, is there a path to abstracting away most of this, so that events could just return maps and something else turns those into the right productions? Something a bit handwavy like https://dom.spec.whatwg.org/#concept-event-create which nonetheless everyone knows how to interpret.

index.bs
```diff
@@ -1710,3 +1738,161 @@ When [=a browsing context is discarded=], run the following steps:
...
+
+1. Let |related contexts| be an empty set.
+
+1. Let |settings| be the [=current settings object=]
+
+1. Let |related contexts| be the result of [=get related contexts=] given |settings|.
+
+1. [=Emit an event=] with |body| and |related contexts|.
+
+After an implementation is required to [=report an exception=] from |script|,
```
foolip, line 1862: How long after? :) Hooks TBD?

index.bs
```diff
@@ -1710,3 +1738,161 @@ When [=a browsing context is discarded=], run the following steps:
...
+   used as the message in [=report an error=].
+
+1. Let |line number| and |column number| be the one-based line number and
+   zero-based column number, respectively, in the resource containing |script|,
+   at the line where the error originated.
+
+1. Let |url| be the result of running the [=URL serializer=], given
+   |settings|'s [=creation URL=].
+
+1. Let |extra| a map matching the <code>JavascriptLogExtra</code> production,
```
foolip, line 1877: How about other kinds of errors that also show up in the console, or say deprecation messages? Will we need to write additional prose to handle those cases?

index.bs
```diff
@@ -1710,3 +1738,161 @@ When [=a browsing context is discarded=], run the following steps:
...
+   <code>timestap</code> field set to |timestamp| and the <code>extra</code>
+   field set to |extra|.
+
+1. Let |related contexts| be the result of [=get related contexts=] given |settings|.
+
+1. [=Emit an event=] with |body| and |related contexts|.
+
+Issue: Add support for stacktraces
+
+Issue: Lots more things require logging
```
foolip, line 1894: Yep. Can you list the ones you know of already here?

index.bs
jgraham: Yeah, I think asserting that string arguments are represented as themselves makes sense.

index.bs
jgraham: Yes, I think basically every type needs to be handled specially. It's a bit annoying but we can probably have a single set of steps for all the ones which are basically implementation-defined just explaining which types to use for the various common kinds of message.

index.bs
jgraham: So there's no difference between a map and a map that matches a production other that the latter is guaranteed to match a production. I was thinking we could have some pseudo-varargs function like:

When required to <dfn>make a map matching CDDL production</dfn> |production| with a variable number of arguments |args|:

1. Let |result| be an [=map=] matching the production |production|, specifying the fields according to the values of |args| as follows:
  * If |arg| is not a [=tuple=], set the field with the name of |arg| to the value of arg.
  * If |arg| is a tuple, set the field named by the first item of |arg| to the second item of |arg|.

1. Assert: |result| matches |production|

1. Return |result|

And then you would use it like:

1. [=Make a map matching the CDDL production=] <code>LogEntry</code> with ("type", "console"), |level|, |text|, |timestamp| and |extra|/

jgraham:
OK, I fied a few issues, but need to come back and improve the way we hook into other specs.

index.bs
```diff
@@ -1739,6 +1781,171 @@ contexts that have active documents; navigation can also cause contexts to
...
+)
+</pre>
+
+### Types ### {#module-log-types}
+
+#### Log Entry #### {#types-log-log-entry}
+
+<pre class="cddl local-cddl">
+
+LogLevel = "debug" / "info" / "warning" / "error"
```
whimboo, line 1813: All these entries are related to console API calls, but how would we handle Javascript errors and warnings as raised by the browser? Do we also want to have that as a separate event like `Runtime.exceptionThrown` vs `Runtime.consoleAPICalled` in CDP?

Also there are way more types available like `dir`, `timeEnd`, and many others. Do we want to fold all of these into the `info` level? It will make it impossible to watch for these specific events.

index.bs
```diff
@@ -1739,6 +1781,171 @@ contexts that have active documents; navigation can also cause contexts to
...
+
+LogEntry = {
+  type: text,
+  level: LogLevel,
+  text: text / null,
+  timestamp: int,
+  extra: { *text => any }
```
whimboo, line 1820: What about the stacktrace? I miss that part here.

index.bs
jgraham: The additional detail is provided on the extra field; the `ConsoleLogExtra` has a `type` field that allows all the detail for the different console types. I think having a very flat design in which any log type could add to the top-level log level would make it very hard to extend because clients would get top-level types they didn't understand and that don't really fit into the "give me all events above level X" paradigm.

Although we haven't designed filtering yet, I agree that it's going to be harder to filter on extra data. In practice for console events in tests the author is probably pretty in control of what's emitted anyway, so hopefully just being able to filter by type and level is enough.

index.bs
jgraham: Yeah, that might make sense to lift to the top level.

index.bs
whimboo: So what about Javascript exceptions then? Where would those fold into? Are those equal to `console.error()` calls?

jgraham: No, this is the difference between `type: console` and `type: javascript` at the top level.

index.bs
```diff
@@ -1739,6 +1782,219 @@ contexts that have active documents; navigation can also cause contexts to
...
+### Types ### {#module-log-types}
+
+#### log.LogEntry #### {#types-log-logentry}
+
+<pre class="cddl local-cddl">
+
+LogLevel = "debug" / "info" / "warning" / "error"
+
+LogEntry = {
+  type: text,
```
sadym-chromium, line 1817:
```suggestion
  context: BrowsingContext,
  type: text,
```

index.bs
```diff
@@ -1739,6 +1782,219 @@ contexts that have active documents; navigation can also cause contexts to
...
+  level: LogLevel,
+  text: text / null,
```
sadym-chromium, line 1819: Log message can have an array of parameters. Should the BiDi hide it and scalar all of them into a single string?

index.bs
```diff
@@ -1739,6 +1782,219 @@ contexts that have active documents; navigation can also cause contexts to
...
+  timestamp: int,
+  ?stackTrace: [*StackFrame],
+  extra: { *text => any }
+}
+
+ConsoleLogExtra = {
+  type: text,
```
sadym-chromium, line 1826: `type` is already declared in the `LogEntry`

index.bs
```diff
@@ -1739,6 +1782,219 @@ contexts that have active documents; navigation can also cause contexts to
...
+  realm: Realm,
+  args: [*RemoteValue],
+}
+
+JavascriptLogExtra = {
+  lineNumber: int,
+  columnNumber: int,
+  url: text
+}
```
sadym-chromium, line 1835: Is it somehow different from the `?stackTrace: [*StackFrame]` in the `LogEntry`?

index.bs
```diff
@@ -1739,6 +1782,219 @@ contexts that have active documents; navigation can also cause contexts to
...
+
+</pre>
+
+Each log event is represented by a <code>LogEntry</code> object. This has a
+<code>type</code> property which represents the type of log entry added, a
+<code>level</code> property representing severity, a <code>text</code> property
+with the log message string itself, and a <code>timestamp</code> property
+corresponding to the time the log entry was generated. The <code>extra</code>
+property contains additional data specific to the given type.
```
sadym-chromium, line 1844: What is the `type` and is it different from the `level`? In [CDP there is only `type` in the meaning of `level`](https://chromedevtools.github.io/devtools-protocol/tot/Runtime/#event-consoleAPICalled).

index.bs
```diff
@@ -1739,6 +1782,219 @@ contexts that have active documents; navigation can also cause contexts to
...
+
+A frame in a stacktrace is represented by a <code>StackFrame</code> object. This
+has a <code>url</code> property, which represents the URL of the script, a
+<code>functionName</code> property which represents the name of the executing
+function, an <code>lineNumber</code> and <code>columnNumber</code> properties,
+which represent the line and column number of the executed code.
+
+The <dfn>current stack trace</dfn> is a representation of the stack of the
+[=running execution context=]. The details of this are unspecified, and so the
+bheaviour here is implementation defined, but the general process is as follows:
```
sadym-chromium, line 1867:
```suggestion
behaviour here is implementation defined, but the general process is as follows:
```

index.bs
jgraham: That's handled in `ConsoleLogExtra` for the console case. I assume things like js exceptions really are just strings (this was based on https://chromedevtools.github.io/devtools-protocol/tot/Log/#event-entryAdded )

index.bs
jgraham: Is the concern that having `logEntry.type` and `logEntry.extra.type` is confusing? It doesn't actually conflict. Do you have a proposed alternative (for concreteness, we'd have something like `{type: "console", [...], extra: {type: "log", [...]}}` for a `console.log` call).

index.bs
jgraham: I think this makes sense, although of course the proposed change isn't enough to actually populate the context id. I was looking at what CDP has and it doesn't seem to have this for `Log.LogEntry`. How do existing clients/implementations work here?

index.bs
jgraham: It definitely feels like this might duplicate the stack information. We can remove it.

index.bs
jgraham: The API here covers both `Runtime.ConsoleAPICalled` and `Log.EntryAdded` events from CDP. So the `type` tells you about the kind of log entry it is (e.g. from a `console` method call in js, an unhandled js exception, or something else), `level` is a coarse levelling to allow filtering across all types, and the type-specific information is in `extra`. In particular to get the particular console method call you need to look in `extra`, not `level`. Does that make sense?

index.bs
sadym-chromium: Should we consider merging `ConsoleLogExtra` and `JavascriptLogExtra` and write the corresponding type into the top-level field `type`? The differences between them two are:
* `realm`, which can be made optional,
* `args`, which can contain text as an only argument in case of `JavascriptLogExtra`.

It would allow to remove the `extra` field, and make log entry more straight-forward. I don't see much profit in the differentiation.

index.bs
sadym-chromium: I got it, thanks.

index.bs
```diff
@@ -1739,6 +1782,219 @@ contexts that have active documents; navigation can also cause contexts to
...
+   |level| be "<code>info</code>".
+
+1. Let |timestamp| be a [=time value=] representing the current date and time in UTC.
+
+1. Let |text| be an empty string.
+
+1. For each |arg| in |args|:
+
+  1. If |arg| is a [=primitive value=], append [=ToString=](|arg|) to
+     |text|. Otherwise append an implementation-defined string to |text|.
```
sadym-chromium, line 1928: Current `console.log` implementation append arguments with a space:
```
> console.log(1, 2)
1 2
```

index.bs
sadym-chromium: Got it, thanks.

index.bs
```diff
@@ -1739,6 +1782,219 @@ contexts that have active documents; navigation can also cause contexts to
...
+
+<div algorithm="remote end event trigger for log.entryAdded">
+
+Define the following [=console Printer steps=] with |logLevel|, |args|, and <var
+ignore>options</var>:
+
+1. If |logLevel| is "<code>error</code>" or "<code>assert</code>", let |level|
+   be "<code>error</code>". If |logLevel| is "<code>debug</code>" or
+   "<code>trace</code>" let |level| be "<code>debug</code>". If |logLevel| is
+   "<code>warn</code>", let |level| be "<code>warning</code>". Otherwise let
```
sadym-chromium, line 1918:
```suggestion
   "<code>warn</code>" or "<code>warning</code>", let |level| be "<code>warning</code>". Otherwise let
```

index.bs
sadym-chromium: It looks like the `Log.LogEntry` event is global per connection, while [`Runtime.consoleAPICalled`](https://chromedevtools.github.io/devtools-protocol/tot/Runtime/#event-consoleAPICalled) has an `executionContextId`.

jgraham: I remembered the reason we don't provide a context id but just a realm id is that something like a service worker has a realm but not a context. We could of course provide both, but I expect that we'll end up with a generic way of converting between them, and lots of places will take either a realm id or context id and work with both.

index.bs
jgraham: Thanks, fixed.

index.bs
```diff
@@ -1739,6 +1782,212 @@ contexts that have active documents; navigation can also cause contexts to
...
+   args|.
+
+1. Let |entry| be a map matching the <code>LogEntry</code> production, with the
+   <code>type</code> field set to "<code>console</code>", the <code>level</code>
+   field set to |level|, the <code>text</code> field set to |text|, the
+   <code>timestap</code> field set to |timestamp|, the <code>stackTrace</code>
+   field set to |stack| if |stack| is not null, or omitted otherwise, and the
+   <code>extra</code> field set to |extra|.
+
+1. Let |body| be a map matching the <code>LogEntryAdded</code> production, with
```
foolip, line 1947:
```suggestion
1. Let |body| be a map matching the <code>LogEntryAddedEvent</code> production, with
```

index.bs
foolip: Regarding the original question of `LogEntry.type` vs. `ConsoleLogExtra.type`, @jgraham already gave the `{type: "console", [...], extra: {type: "log", [...]}}` example, and in https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/73.html#event-log-entryAdded it looks like "console" is currently the only possible value for `LogEntry.type` and `ConsoleLogExtra.type` is the *logLevel*, so I guess actually "info" for `console.log`. There are things in https://console.spec.whatwg.org/ I'm not sure what it should map to, like `console.groupEnd()`.

On the `extra` parameter, if we anticipate more than one kind of log entry, I wonder if we couldn't instead go with plain inheritance, where a `ConsoleLogEntry` production inherits from `LogEntry` and adds fields? Clients would type test using `LogEntry.type`, but we'd avoid the extra nesting via `extra`.

index.bs
```diff
@@ -1739,6 +1782,212 @@ contexts that have active documents; navigation can also cause contexts to
...
+#### log.LogEntry #### {#types-log-logentry}
+
+<pre class="cddl local-cddl">
+
+LogLevel = "debug" / "info" / "warning" / "error"
+
+LogEntry = {
+  type: text,
+  level: LogLevel,
+  text: text / null,
```
foolip, line 1819: Is it possible for `text` to be null here? In the "console Printer steps" it won't ever be null but rather an empty string. Maybe make this non-nullable until we find a case where null makes sense?

`console.log()` with no arguments is an interesting case here, but letting text be the empty string in this case seems OK to me. One could still distinguish it from `console.log('')` based on the arguments, if that were ever important.

jgraham: I was thinking of cases like `groupEnd()` where there isn't a meaningful log message.

foolip: Ah, that makes sense. Will we need to patch https://console.spec.whatwg.org/#groupend to support that case?

index.bs
jgraham: I'm going to resolve this on the basis the comment isn't going to block this PR, but please feel free to open an Issue with this suggestion.

index.bs
jgraham: https://github.com/whatwg/console/issues/195

index.bs
jgraham: Yeah, the patch I have isn't quite enough.

index.bs
jgraham: I changed this to be more like inheritance. I'm still not sure what the ideal tradeoff specwise between it looking like inheritance and looking like mixins is, but we can change the formalism later if we're happy with the message structure.

index.bs
jgraham: I changed the proposed patch to intercept anything that calls a method on the console interface.

index.bs
```diff
@@ -2086,6 +2130,211 @@ worker=] algorithm:
...
+The <dfn export for=modules>log</dfn> module contains functionality and events
+related to logging.
+
+### Definition ### {#module-log-definition}
+
+[=remote end definition=]
+
+<pre class="cddl remote-cddl">
+
+Command = (GetCommand)
```
bwalderman, line 2144: Can this be removed?

index.bs
```diff
@@ -2086,6 +2130,211 @@ worker=] algorithm:
...
+  type: "javascript",
+}
+
+</pre>
+
+Each log event is represented by a <code>LogEntry</code> object. This has a
+<code>type</code> property which represents the type of log entry added, a
+<code>level</code> property representing severity, a <code>text</code> property
+with the log message string itself, and a <code>timestamp</code> property
+corresponding to the time the log entry was generated. The <code>extra</code>
```
bwalderman, line 2190: Looks like we don't have an `extra` property anymore.

index.bs
```diff
@@ -2086,6 +2130,211 @@ worker=] algorithm:
...
+  lineNumber: int,
+  columnNumber: int,
+}
+
+</pre>
+
+A frame in a stacktrace is represented by a <code>StackFrame</code> object. This
+has a <code>url</code> property, which represents the URL of the script, a
+<code>functionName</code> property which represents the name of the executing
+function, an <code>lineNumber</code> and <code>columnNumber</code> properties,
```
bwalderman, line 2209:
```suggestion
function, and <code>lineNumber</code> and <code>columnNumber</code> properties,
```

index.bs
```diff
@@ -2086,6 +2130,211 @@ worker=] algorithm:
...
+
+ 1. For each stack frame |frame| in the stack of the running execution context,
+    starting from the most recently executed frame, run the following steps:
+
+    1. Let |url| be the result of running the [=URL serializer=], given
+       the <a spec=url for=/>URL</a> of |frame|'s associated script resource.
+
+    1. Let |functionName| be the name of |frame|'s associated function.
+
+    1. Let |lineNumber| and |columnNumber| be the on-based line and zero-based
```
bwalderman, line 2226:
```suggestion
    1. Let |lineNumber| and |columnNumber| be the one-based line and zero-based
```

index.bs
```diff
@@ -2086,6 +2130,211 @@ worker=] algorithm:
...
+#### entryAdded #### {#event-log-entryAdded}
+
+<dl>
+   <dt>Event Type</dt>
+   <dd>
+      <pre class="cddl local-cddl">
+        LogEntryAddedEvent = {
+         method: "log.entryAdded",
+         params: (
+             GenericLogEntry //
```
bwalderman, line 2251: Nit: Could we define a LogEntry production for this grouping up above and reference LogEntry here? That way the list of supported LogEntry types appears closer to their actual definitions.

index.bs
```diff
@@ -2086,6 +2130,211 @@ worker=] algorithm:
...
+   field set to |text|, the <code>timestap</code> field set to |timestamp|, the
+   <code>stackTrace</code> field set to |stack| if |stack| is not null, or
+   omitted otherwise, the |method| field set to |method|, the <code>realm</code>
+   field set to |realm| and the <code>args</code> field set to |serialized
+   args|.
+
+1. Let |body| be a map matching the <code>LogEntryAddedEvent</code> production, with
+   the <code>params</code> field set to |entry|.
+
+1. Let |related contexts| be an empty set.
```
bwalderman, line 2305: Can this step be removed?

index.bs
```diff
@@ -2086,6 +2130,211 @@ worker=] algorithm:
...
+1. For each |arg| of |args|, append the result of [=serialize as a remote
+   value=] given |arg|, null, true, and an empty [=set=] to |serialized args|.
+
+1. Let |realm| be the [=realm id=] of the [=current Realm Record=].
+
+1. Let |stack| be the [=current stack trace=].
+
+1. Let |entry| be a map matching the <code>ConsoleLogEntry</code> production,
+   with the the <code>level</code> field set to |level|, the <code>text</code>
+   field set to |text|, the <code>timestap</code> field set to |timestamp|, the
```
bwalderman, line 2296:
```suggestion
   field set to |text|, the <code>timestamp</code> field set to |timestamp|, the
```

index.bs
```diff
@@ -2086,6 +2130,211 @@ worker=] algorithm:
...
+
+1. If |handled| is true return.
+
+1. Let |settings| be |script|'s [=script/settings object=].
+
+1. Let |stack| be the [=current stack trace=] for the exception.
+
+1. Let |entry| be a map matching the <code>JavascriptLogEntry</code> production,
+   with <code>level</code> set to "<code>error</code>", <code>text</code> set to
+   |message|, and the <code>timestap</code> field set to |timestamp|.
```
bwalderman, line 2325:
```suggestion
   |message|, and the <code>timestamp</code> field set to |timestamp|.
```

bwalderman (changes requested):

index.bs
jgraham: I think that was a rebase error :)

index.bs
jgraham: Yes :)

index.bs
```diff
@@ -2086,6 +2130,210 @@ worker=] algorithm:
...
+   args|.
+
+1. Let |body| be a map matching the <code>LogEntryAddedEvent</code> production, with
+   the <code>params</code> field set to |entry|.
+
+1. Let |settings| be the [=current settings object=]
+
+1. Let |related contexts| be the result of [=get related contexts=] given |settings|.
+
+1. [=Emit an event=] with |body| and |related contexts|.
```
sadym-chromium, line 2310: NIT
```suggestion
1. [=Emit an event=] with |body| and |related browsing contexts|.
```

index.bs
```diff
@@ -2115,3 +2363,29 @@ To discard a browsing context |browsingContext|, run these steps:
 Issue: The actual patch might be better to split the algorithm into an outer algorithm that is
 called by external callers and an inner algorithm that's used for recursive calls. That's quite hard
 to express as a patch to the specification since it requires changing multiple parts.
+
+
+The [=report an error=] algorithm is modified with an additional step at the
+end:
+
+<div algorithm>
+11. Call any <dfn>error reporting steps</dfn> decined in external specifications
```
sadym-chromium, line 2372: Maybe `defined`? 
```suggestion
11. Call any <dfn>error reporting steps</dfn> defined in external specifications
```

sadym-chromium (approved):
LGTM with few comments

index.bs
```diff
@@ -2086,6 +2130,210 @@ worker=] algorithm:
...
+
+  1. If |arg| is not the first entry in |args|, append a U+0020 SPACE to |text|.
+
+  1. If |arg| is a [=primitive value=], append [=ToString=](|arg|) to
+     |text|. Otherwise append an implementation-defined string to |text|.
+
+1. Let |serialized args| be a new list.
+
+1. For each |arg| of |args|, append the result of [=serialize as a remote
+   value=] given |arg|, null, true, and an empty [=set=] to |serialized args|.
```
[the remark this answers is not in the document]
jgraham: So, I think we're going to need to improve the control of serialization for the whole spec at some point. I think I'd prefer to work on that when we specify script execution. So maybe file an issue for now?

index.bs
```diff
@@ -2086,6 +2130,210 @@ worker=] algorithm:
...
+1. Let |timestamp| be a [=time value=] representing the current date and time in UTC.
+
+1. Let |text| be an empty string.
+
+1. For each |arg| in |args|:
+
+  1. If |arg| is not the first entry in |args|, append a U+0020 SPACE to |text|.
+
+  1. If |arg| is a [=primitive value=], append [=ToString=](|arg|) to
+     |text|. Otherwise append an implementation-defined string to |text|.
```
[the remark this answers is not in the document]
jgraham: I had no idea that worked. I'll add an issue to the spec. I think we should support it, but I also think we shouldn't block this PR on supporting it.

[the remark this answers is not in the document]
jgraham: OK, I figured out how to at least make it clear we intend to support this.

index.bs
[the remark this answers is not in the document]
sadym-chromium: #86

bwalderman (approved):

PR #76, opened 2020-12-04
Fix nesting of final "handle an incoming message" steps

foolip:
These were indented one space too much and ended up under the "If this
results in a match" branch, which made no sense.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/76.html" title="Last updated on Dec 4, 2020, 9:14 AM UTC (75220ba)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/76/e533bf1...75220ba.html" title="Last updated on Dec 4, 2020, 9:14 AM UTC (75220ba)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -545,15 +545,15 @@ To <dfn>handle an incoming message</dfn> given a [=WebSocket connection=]
       1. [=Send a WebSocket message=] comprised of |serialized| over
          |connection| and return.
 
-  1. Otherwise there is no match.
+ 1. Otherwise there is no match.
 
-     Let |error code| be [=invalid argument=].
+    Let |error code| be [=invalid argument=].
 
-  1. If |data| is a map and |data|["<code>method</code>"] exists and is a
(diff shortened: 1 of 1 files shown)


christian-bromann (approved):

foolip:
@sadym-chromium here's something I came across while trying to write some tests for this algorithm.

@jgraham for context, we're playing with a prototype in https://github.com/sadym-chromium/WebDriverBiDiServerExperiments and I'm trying to write tests using py3 asyncio, something that could plausibly morph into wpt tests.

@AutomatedTester would you be able to add @sadym-chromium as a collaborator so I can assign reviews to him? We haven't written it down, but it of course would remain the case that big/consequential changes aren't made with input only from one vendor, but small fixes like this could be reviewed between the two of us.

foolip:
Thanks for the review, @christian-bromann!

PR #77, opened 2020-12-07
Respond with an error for binary messages

foolip:
Silently doing nothing is not very easy to debug, and more importantly
if in the future do support some binary messages, we should respond with
an error for any we do not recognize, just as we do for text messages.

Doing this required refactoring since "respond with an error" previous
assumed a text |data| argument, but we should not attempt to parse a
binary message as JSON.

Lift out the JSON parsing to before try to match against the CDDL, to
avoid parsing twice. Previously parsing was implicit in the "match"
verb, but we can just as well "match" against a parsed representation,
it is no more or less defined what that means.

One typo was fixed: occured

The only intended observable change is for non-text messages.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/77.html" title="Last updated on Dec 10, 2020, 11:57 PM UTC (1f316b2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/77/fe54519...1f316b2.html" title="Last updated on Dec 10, 2020, 11:57 PM UTC (1f316b2)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -495,10 +495,9 @@ typically be "<code>localhost</code>".
 To <dfn>handle an incoming message</dfn> given a [=WebSocket connection=]
 |connection|, type |type| and data |data|:
 
- 1. If |type| is not [=%x1 denotes a text frame|text=], return.
-
-    Issue: Should we instead close |connection| with [=status
-    codes|status code=] 1003, or [=respond with an error=]?
+ 1. If |type| is not [=%x1 denotes a text frame|text=], [=respond with an
+    error=] given |connection|, null, and [=invalid argument=], and finally
+    return.
 
  1. [=Assert=]: |data| is a [=scalar value string=], because the
      WebSocket [=handling errors in UTF-8-encoded data=] would already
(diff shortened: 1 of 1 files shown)


foolip:
This was suggested by @sadym-chromium in https://github.com/sadym-chromium/WebDriverBiDiServerExperiments/pull/12 and I'm going to update the tests for it there. My hope is that those tests will eventually turn into something suitable for WPT.

sadym-chromium (approved):

index.bs
```diff
@@ -508,26 +507,32 @@ To <dfn>handle an incoming message</dfn> given a [=WebSocket connection=]
...
+
+ 1. Match |parsed| against the [=remote end definition=]. If this results in a
     match:
 
-    1. Let |parsed| be the map representing the matched data.
+    1. Let |matched| be the map representing the matched data.
 
-    1. Assert: |parsed| contains "<code>id</code>", "method", and
-       "<code>params</code>"
+    1. Assert: |matched| contains "<code>id</code>", "<code>method</code>", and
```
jgraham, line 520: Historically we have ignored unknown keys for forward-compat.

foolip: That's precisely what I'm trying to express here, but have apparently failed. My intention was that |matched| effectively be the subset of |parsed| that matches the CDDL.

jgraham: But if the remote end definition has `{*text => text}` then it will match the other fields too. If it doesn't have that then I think it will reject extra fields.

foolip: So that's what `{*text => any}` means... and you're saying we need that for the definition to match at all? This sounds like a problem, I'll file an issue.

What should I do here for now? A note saying how |matched| is different from |parsed| would be good. But it sounds like actually they would be identical if there was a match, right?

jgraham: Right, I think it would be identical (or maybe not quite if we start using the default value features). You could handwave in some way about matched containing only the explictly named fields. Or just remove "and no others".

foolip: I've filed https://github.com/w3c/webdriver-bidi/issues/79.

jgraham: FWIW I'm happy to land this if you just delete "and no others" for now.

foolip: OK, I'll do just that.

jgraham (approved):

PR #80, opened 2020-12-16
Remote end behavior for duplicate command id

k7z45:
This adds an issue to handling of commands with duplicate id
for same or different commands.

***
<a href="https://pr-preview.s3.amazonaws.com/k7z45/webdriver-bidi/pull/80.html" title="Last updated on Dec 16, 2020, 6:55 PM UTC (d175a3f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/80/16f1bec...k7z45:d175a3f.html" title="Last updated on Dec 16, 2020, 6:55 PM UTC (d175a3f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -254,6 +254,12 @@ Note: This is because the command id is entirely controlled by the [=local end=]
 and isn't necessarily unique over the course of a session. For example a [=local
 end=] which ignores all responses could use the same command id for each command.
 
+Issue: Do we want to say the behavior is undefined if remote end receives different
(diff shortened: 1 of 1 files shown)


jgraham:
I think we converged on the idea that the command id is opaque to the remote end; it just copies it from the command to the result without caring about semantics. It's up to the local end to ensure that it produces sufficiently unique command ids to distinguish the responses. So as far as I'm concerned these issues are resolved already.

k7z45:
Got it, thanks for the response.

PR #85, opened 2021-02-10
Allow emitting events on subscription

jgraham:
This adds support for backfilling events when subscribing to a new
event type either for the first time, or for a new set of
contexts. The use case is allowing local ends to update themselves with
the current state of the remote end, just by listening for events,
without needing to be very careful about the exact order for which
subscriptions happen.

If an event is backfilled, it must define remote end subscribe steps,
which take a set of browsing contexts where the event is newly active,
and an |include global| flag which indicates whether the subscription
is global.

These steps have a priority order so that we can create invariants
like browsing context events being emitted before the realms or log
messages relating to those contexts.

In order to make it easier to backfill events, this also changes the
subscription model to only allow subscibing per top-level context, not
per frame.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/85.html" title="Last updated on Mar 17, 2021, 4:01 PM UTC (1e54361)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/85/9f81644...1e54361.html" title="Last updated on Mar 17, 2021, 4:01 PM UTC (1e54361)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1366,8 +1366,12 @@ disabled, the return value is always empty.
   1. Let |global event set| be a [=set/clone=] of the [=global event set=] for
      |session|.
 
-  1. Let |event map| be a [=map/clone=] the [=browsing context event map=] for
-     |session|.
+  1. Let |event map| be a new map.
+
+  1. For each |key| → |value| of the [=browsing context event map=] for
+     |session|:
+
+    1. Set |event map|[|key|] to a [=set/clone=] of |value|.
 
   1. Let |enabled events| be a new map.
 
@@ -1388,7 +1392,7 @@ disabled, the return value is always empty.
              1. Let |event enabled contexts| be the [=event enabled browsing
                 contexts=] given |session| and |event name|
 
-             1. |event name| to |global event set|.
+             1. Add |event name| to |global event set|.
 
              1. For each |context| of |event enabled contexts|, remove |event
                 name| from |event map|[|context|].
@@ -1399,8 +1403,8 @@ disabled, the return value is always empty.
 
         1. For each |event name| in |event names|:
 
-          1. If the |global event set| [=contains=] |event name|, remove |event
-             name| from the |global event set|. Otherwise return [=error=] with
+          1. If |global event set| [=contains=] |event name|, remove |event
+             name| from |global event set|. Otherwise return [=error=] with
              [=error code=] [=invalid argument=].
 
   1. Otherwise, if |browsing contexts| is not null:
@@ -1427,7 +1431,7 @@ disabled, the return value is always empty.
 
         1. If |enabled| is true and |target| does not contain |event name|:
 
-          1. |event name| to |target|.
+          1. Add |event name| to |target|.
 
           1. If |enabled events| does not contain |event name|, set |enabled
              events|[|event name|] to a new set.
--- a/index.bs
+++ b/index.bs
@@ -299,6 +299,11 @@ occurred on the [=remote end=].
       parameters</code>.
  - A <dfn export>remote end event trigger</dfn> which defines when the event is
    triggered and steps to construct the [=event type=] data.
+ - Optionally, a set of <dfn>remote end subscribe steps</dfn>, which define
+   steps to take when a local end subscribes to an event. Where defined these
+   steps have an associated <dfn>subscribe priority</dfn> which is an integer
+   controlling the order in which the steps are run when multiple events are
+   enabled at once, with lower integers indicating steps that run earlier.
 
 A [=/session=] has a <dfn export for=event>global event set</dfn> which is a set
 containing the event names for events that are enabled for all
@@ -306,9 +311,24 @@ browsing contexts. This initially contains the [=event name=] for events that
 are <dfn export for=event>in the default event set</dfn>.
 
 A [=/session=] has a <dfn export for=event>browsing context event map</dfn>,
-which is a map with [=/browsing context=] keys and values that are maps from an
-[=event name=] to a boolean indicating whether the specified event is enabled or
-disabled for a given browsing context.
+which is a map with [=/top-level browsing context=] keys and values that are a
+set of [=event name=]s for events that are enabled in the given browsing
+context.
+
+<div algorithm>
+
+To obtain a list of <dfn>event enabled browsing contexts</dfn> given
+|session| and |event name|:
+
+1. Let |contexts| be an empty set.
+
+1. For each |context| → |events| of |session|'s [=browsing context event map=]:
+
+  1. If |events| contains |event name|, append |context| to |contexts|
+
+1. Return |contexts|.
+
+</div>
 
 <div algorithm>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -306,27 +326,27 @@ To determine if an <dfn export>event is enabled</dfn> given |session|,
 Note: |browsing contexts| is a set because a [=shared worker=] can be associated
       with multiple contexts.
 
-  1. For each |browsing context| in |browsing contexts|:
+1. Let |top-level browsing contexts| be an empty set.
 
-    1. While |browsing context| is not null:
+1. For each |browsing context| of |browsing contexts|, append the [=top-level
+   browsing context=] for |browsing context| to in |top-level browsing
```
k7z45, line 332: nit. Add "the" before |browsing context| and remove in before |top-level ...

index.bs
```diff
@@ -1289,52 +1309,101 @@ SessionResult = (StatusResult)
...
-    1. If the |list of contexts| is null:
+  1. If |browsing contexts| is null:
+
+      1. If |enabled| is true:
+
+        1. For each |event name| of |event names|:
+
+           1. If |global event set| doesn't contain |event name|:
+
+             1. Let |context enabled events| be the [=event enabled browsing
```
k7z45, line 1342: Should |context enabled events| be called |event enabled contexts| instead?

index.bs
```diff
@@ -1289,52 +1309,101 @@ SessionResult = (StatusResult)
...
-      1. Get the entry from the |event map| for |context| and append it to
-         |targets|.
+         1. Append |event name| to |target|.
 
-    1. For each |target| in |targets|:
+         1. If |enabled events| does not contain |event name|, set |enabled
+            events|[|event name|] to a new set.
 
-      1. For each |event name| in |event names|:
+         1. Append |context| to |context enabled events|[|event name|].
```
k7z45, line 1391: Shoud |context enabled events| here just be |enabled events|?

index.bs
```diff
@@ -1289,52 +1309,101 @@ SessionResult = (StatusResult)
...
-    1. Return [=success=] with data null.
+  1. Return [=success=] with data |enabled events|.
+
+Note: Implementations that do additional work when an event is enabled,
+e.g. subscribing to the relevant engine-internal events, will likely perform
+those additional steps when updating the event map. This specification uses
+a model where hooks are always called and then the event map is used to
+filter only those that ought to be returned to the local end.
+
+Issue: Currently if an event is enabled globally and then disabled for a
```
k7z45, line 1404: Should be an error unless we have a disabled event context map as well.

index.bs
```diff
@@ -306,27 +326,27 @@ To determine if an <dfn export>event is enabled</dfn> given |session|,
...
-      1. If |browsing context events| is not null, and |browsing context events|
-         [=contains=] for |event name| return |browsing context
-         events|[|event name|].
+    1. If |event map| [=contains=] |browsing context|, let |browsing context
+       events| be |event map|[|browsing context|].  Otherwise let |browsing
+       context events| be null.
 
-      1. Let |browsing context| be the [=parent browsing context=] of |browsing
-         context|, if it has one, or null otherwise.
+   1. If |browsing context events| is not null, and |browsing context events|
```
k7z45, line 343: This is not aligned with the previous "1.".

index.bs
jgraham: OK, so I updated this to emit an error in the case that you try to disable an event that is not in fact enabled. idk if this is the right call here, or if we should just do this for the global case, but not the context-specific case.

index.bs
```diff
@@ -1289,52 +1308,105 @@ SessionResult = (StatusResult)
...
+value are those for which the event was already enabled. When the events are
+enabled for specific contexts, the contexts in the return value are those for
+which the event are now enabled but were not previously. When events are
+disabled, the return value is always empty.
 
-  1. Let |event map| be the [=browsing context event map=] for |session|.
+  1. Let |global event set| be a [=set/clone=] of the [=global event set=] for
+     |session|.
+
+  1. Let |event map| be a [=map/clone=] the [=browsing context event map=] for
```
bwalderman, line 1324:
```suggestion
  1. Let |event map| be a [=map/clone=] of the [=browsing context event map=] for
```

index.bs
```diff
@@ -1289,52 +1308,105 @@ SessionResult = (StatusResult)
...
 
-      1. Return
+           1. If |global event set| doesn't contain |event name|:
 
-    1. Let |targets| be an empty list.
+             1. Let |event enabled contexts| be the [=event enabled browsing
+                contexts=] given |session| and |event name|
 
-    1. For each entry |context id| in the |list of contexts|:
+             1. |event name| to |global event set|.
```
bwalderman, line 1346:
```suggestion
             1. Add |event name| to |global event set|.
```

index.bs
```diff
@@ -1289,52 +1308,105 @@ SessionResult = (StatusResult)
...
+             1. For each |context| of |event enabled contexts|, remove |event
+                name| from |event map|[|context|].
+
+             1. Set |enabled events|[|event name|] to |event enabled contexts|.
+
+      1. If |enabled| is false:
+
+        1. For each |event name| in |event names|:
+
+          1. If the |global event set| [=contains=] |event name|, remove |event
```
bwalderman, line 1357: Nit: This step should refer to the local variable named "global event set", and not **the** global event set.

index.bs
```diff
@@ -1289,52 +1308,105 @@ SessionResult = (StatusResult)
...
+      1. If |enabled| is true and |global event set| contains |event name|, continue.
 
-      1. For each |event name| in |event names|:
+      1. For each |context| → |target| in |targets|:
 
-        1. Set |target|[|event name|] to |enabled|.
+        1. If |enabled| is true and |target| does not contain |event name|:
 
-    1. Return [=success=] with data null.
+          1. |event name| to |target|.
```
bwalderman, line 1385:
```suggestion
          1. Add |event name| to |target|.
```

bwalderman (changes requested):

index.bs
```diff
@@ -2007,6 +2150,40 @@ object:
...
+
+The [=remote end subscribe steps=] with [=subscribe priority=] 2, given
+|contexts| and |include global| are:
+
+1. Let |environment settings| be a list of all the [=environment settings objects=]
+   that have their [=execution ready flag=] set.
+
+1. For each |settings| of |environment settings|:
+
+  1. Let |related contexts| be an empty set.
```
bwalderman, line 2163: This looks like a good place to re-use the "get related contexts" algorithm defined in #73.

jgraham: Agreed, but we need to merge that one first (or factor out that part, but it seems like it will be more owrk overall). In any case the logging module will need an update when this merges to buffer log events.

bwalderman: Sounds good. Thanks.

bwalderman (approved):

PR #92, opened 2021-03-04
Use “WebDriver Bidirectional protocol”, not “BiDi”

sideshowbarker:
In W3C fora, the term *bidi* is used often in discussions about Internationalization, where it is widely understood to refer to bidirectional text — with associated specifications such as the Unicode Bidirectional Algorithm spec https://unicode.org/reports/tr9/

So a number of people who reviewed the Browser Testing and Tools WG charter were thrown off by the use of the term *BiDi* in the title of the WebDriver Bidirectional protocol spec. Based on the title, they expected the spec to have something to do with bidirectional text — which it doesn’t of course.

Thus to preempt ongoing confusion from future readers (especially those involved in the W3C and in Unicode work), it seems prudent *throughout the specification itself* to just consistently write out the full name of the protocol, WebDriver Bidirectional protocol, and avoid the “BiDi” shorthand.

But making that change doesn’t preclude continuing to use the “WebDriver BiDi” shorthand term in documents other than the spec itself, and in mailing-list discussions, etc.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/92.html" title="Last updated on Mar 4, 2021, 7:21 AM UTC (c8f7b63)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/92/9f81644...c8f7b63.html" title="Last updated on Mar 4, 2021, 7:21 AM UTC (c8f7b63)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1,5 +1,5 @@
 <pre class=metadata>
-Title: WebDriver BiDi
+Title: WebDriver Bidirectional Protocol
 Shortname: webdriver-bidi
 Level: 1
 Status: ED
@@ -7,7 +7,7 @@ Group: browser-testing-tools
 URL: https://w3c.github.io/webdriver-bidi/
 Repository: w3c/webdriver-bidi
 No Editor: true
-Abstract: This document defines the BiDirectional WebDriver Protocol, a mechanism for remote control of user agents.
+Abstract: This document defines the WebDriver Bidirectional Protocol, a mechanism for remote control of user agents.
 Boilerplate: conformance no
 Complain About: accidental-2119 yes, missing-example-ids yes
 Default Ref Status: current
@@ -132,7 +132,7 @@ Network protocol messages are defined using CDDL. [[!RFC8610]]
 
 # Protocol # {#protocol}
 
-This section defines the basic concepts of the WebDriver BiDi
+This section defines the basic concepts of the WebDriver Bidirectional
 protocol. These terms are distinct from their representation at the
 <a href=#transport>transport</a> layer.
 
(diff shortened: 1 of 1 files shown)


foolip:
How about the spec short name, will that need to change too?

sideshowbarker:
> How about the spec short name, will that need to change too?

No, let’s just keep that as it is. The rationale for that is easily defendable: Shortnames should be short.

jgraham:
I don't want to be difficult, but changing the name the spec uses to refer to itself away from what we actually in practice use to refer to the spec in the hope of avoiding confusion doesn't seem like an obvious win.

I agree that this in an unfortunate case where the same term is used in multiple contexts, and previously W3C has mostly only had specifications dealing with one of those contexts and so from that perspective it seems like a novel form of ambiguity. But even in the wider web space, the use of Bidirectional/BiDi is already common in both text processing and network protocol domains and in general it seems like it doesn't cause too many problems. So whilst I am sympathetic to the fact that we've unintentionally wasted some people's time, it seems like it's something that will solve itself without any action from us. If that isn't the case, expanding out "BiDi" to "Bidirectional" doesn't seem like it will be sufficient to resove the ambiguity, given that the Unicode consortium also uses "Bidirectional" in formal contexts.

sideshowbarker:
> I don't want to be difficult, but changing the name the spec uses to refer to itself away from what we actually in practice use to refer to the spec in the hope of avoiding confusion doesn't seem like an obvious win.

So to be clear: An acceptable resolution is for those of you actively working on the spec to close this PR without merging it.

In the issue description, I tried in good faith to articulate as persuasively as I could the argument for making the change, based on the comments I saw from reviewers. But it’s up to the discretion of the spec editors to make your own evaluation about how compelling that argument is. And if you find that the argument isn’t compelling enough to merit making the change — then as I said, an acceptable resolution is to close this PR without merging it.

jgraham:
I think the conclusion from the telecon is that we're reluctant to stop using "BiDi" but are happy for the spec to be called "[The] WebDriver BiDi Protocol" and to use the word protocol consistently when the spec refers to itself, if that helps avoid confusion.

sideshowbarker:
OK, thanks — closing per https://github.com/w3c/webdriver-bidi/pull/92#issuecomment-796045483

PR #98, opened 2021-04-11
Try rewriting the explainer

gsnedders:
This attempts to follow the TAG-given advice at https://w3ctag.github.io/explainers

It also tries to bring the explainer in-line with today's spec.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -1,14 +1,26 @@
-# Bidirectional WebDriver Protocol
+# Bidirectional WebDriver Protocol Explainer
 
-## Overview
+WebDriver BiDi (Bi-Directional) is being developed to allow web developers to migrate from
+Chromium-only [CDP](https://chromedevtools.github.io/devtools-protocol/)-based (Chrome DevTools
+Protocol) developer tooling to cross-browser tooling.
 
-This document presents a possible design for a bidirectional WebDriver protocol, incorporating scenarios and resolutions discussed at the TPAC 2019 working group meeting. The protocol uses JSON-RPC messaging over WebSockets as the transport mechanism. WebDriver's current model of the browser is extended to include service workers and other non-page targets and make it possible for clients to target these additional contexts. We also discuss how the new protocol can interoperate with the existing protocol. Sample protocol messages illustrating how the protocol would work are included, and an JSON API specification is included alongside the document.
+## Motivation
+
+Many developer tools are exclusively targetting Chrome, relying on CDP, resulting in websites being
+better tested against Chrome and leading to site compatibility bugs for other browsers.
+
+WebDriver BiDi introduces a new protocol, designed to be used in conjunction with the existing
+WebDriver protocol, allowing new functionality to be introduced to reduce the gap in functionality
+to CDP and allowing developer tooling to target a wider variety of browsers.
+
+WebDriver BiDi should help ensure a better end-user experience across all browsers by allowing
+developers to use the same tooling across all browsers.
 
 ## Goals
 
-The protocol is designed with the following goals in mind:
-
-- **Support for the top customer scenarios identified at TPAC 2019:**
+- **Support for the top customer scenarios
+  [identified](https://www.w3.org/2019/09/19-webdriver-minutes.html#item03) at
+  [TPAC 2019](https://www.w3.org/2019/09/TPAC/):**
     - Listen for DOM events
     - Log what's going on in the browser including console and JS errors
     - Fail fast on any JS error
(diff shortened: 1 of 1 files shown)


explainer.md
```diff
@@ -31,19 +43,50 @@ The protocol is designed with the following goals in mind:
...
+### Choice of Protocol Layer
+
+\[FIXME. Why JSON?\]
+
+CDDL is used to describe the protocol layer because it provides formal semantics, accomplishing the
+"machine-readable API specification" goal while being similar to JSON-RPC used in the prior art.
+
+### Considered alternatives
+
+\[FIXME: Adopting CDP wholesale\]
```
jgraham, line 77: Maybe mention: reluctant to standardise, exposes implementation details, chatty protocol only designed for local usage.

explainer.md
```diff
@@ -31,19 +43,50 @@ The protocol is designed with the following goals in mind:
...
+
+## Design
+
+WebDriver BiDi defines a transport layer (built on top of WebSockets) and a protocol on top of that
+(using JSON, where the messages are described in the standard using
+[CDDL](https://tools.ietf.org/html/rfc8610)).
+
+### Choice of Transport Layer
+
+\[FIXME\]
```
jgraham, line 66: Dunno what you want here, but JSON-over-WebSockets is close the the prior art and the most standard thing for a bidi communication protocol on the web stack.

explainer.md
gsnedders: Yeah, some of the FIXME was definitely "it's Sunday and I'm meant to be tidying my flat, not trying to improve the explainer due to WebDriver BiDi getting mentioned on orange site" (there are currently no interesting comments).

explainer.md
gsnedders: Ignoring the prior art for a second, it's not immediately obvious why JSON-over-WebSockets is a better solution than CBOR-over-TCP, especially when it comes to non-local usage.

jgraham: I think "this is not where we want to spend our complexity budget; tooling in this space uniformly has access to JSON parsing, and WebSockets is the common denominator in existing implementations", but also "this is something we could change later if we end up in a place where the transport is having a noticable effect on performance" (but I'd be surprised if we end up in that place; if the deserialization cost is that high the protocol got too chatty anyway).

explainer.md
```diff
@@ -31,19 +43,50 @@ The protocol is designed with the following goals in mind:
...
+
+### Choice of Protocol Layer
+
+\[FIXME. Why JSON?\]
+
+CDDL is used to describe the protocol layer because it provides formal semantics, accomplishing the
+"machine-readable API specification" goal while being similar to JSON-RPC used in the prior art.
```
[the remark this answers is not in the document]
foolip: This could do with some tweaking I think. CDDL describes the shapes of messages that can be encoded as JSON, and we're using CDDL to describe something very similar to JSON-RPC, but CDDL *itself* is not similar to JSON-RPC, they're on different layers.

Perhaps "while allowing us to define a protocol similar to JSON-RPC used in the prior art"?

explainer.md
foolip: Maybe this is where we mention WebTransport and CBOR, two newer things we're not going to use right now.

explainer.md
```diff
@@ -31,19 +43,50 @@ The protocol is designed with the following goals in mind:
...
+\[FIXME: Adopting CDP wholesale\]
 
-## Proposals
+## Privacy and Security Concerns
 
-- [Core Functionality](./proposals/core.md)
-- [Bootstrap Scripts](./proposals/bootstrap-scripts.md)
+Any protocol that can be used for web testing automation can also open browsers up to malicious
+actors. It is vital that any functionality cannot be accessed from web platform content; browsers
+with multi-process architectures may want to minimise the amount of functionality within the web
```
foolip, line 83: A very good point!

Aside: This is actually relevant to the question of whether the `navigate` command should support relative URLs, because to do so would require going to the content process to resolve the URL, or using a cached-and-maybe-stale URL in another process.

foolip (commented):
This is a great improvement, there are some FIXMEs, but "try" is too humble a description. With just a few small fixes I'd be happy for us to land this.

explainer.md
```diff
@@ -1,14 +1,26 @@
...
+Many developer tools are exclusively targetting Chrome, relying on CDP, resulting in websites being
+better tested against Chrome and leading to site compatibility bugs for other browsers.
+
+WebDriver BiDi introduces a new protocol, designed to be used in conjunction with the existing
+WebDriver protocol, allowing new functionality to be introduced to reduce the gap in functionality
+to CDP and allowing developer tooling to target a wider variety of browsers.
 
-The protocol is designed with the following goals in mind:
+WebDriver BiDi should help ensure a better end-user experience across all browsers by allowing
+developers to use the same tooling across all browsers.
```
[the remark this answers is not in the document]
whimboo: As such we should really do better in creating detailed WebDriver-bidi tests that will cover nearly all (important) cases for individual commands and events. For WebDriver HTTP I still see a lot of untested areas, especially around navigation. :(

explainer.md
```diff
@@ -1,14 +1,26 @@
-# Bidirectional WebDriver Protocol
+# Bidirectional WebDriver Protocol Explainer
 
-## Overview
+WebDriver BiDi (Bi-Directional) is being developed to allow web developers to migrate from
+Chromium-only [CDP](https://chromedevtools.github.io/devtools-protocol/)-based (Chrome DevTools
```
whimboo, line 4: Shall we remove the `Chromium-only` bit here given that we also have (partial) support for it in Firefox?

explainer.md
```diff
@@ -1,14 +1,26 @@
...
-## Overview
+WebDriver BiDi (Bi-Directional) is being developed to allow web developers to migrate from
+Chromium-only [CDP](https://chromedevtools.github.io/devtools-protocol/)-based (Chrome DevTools
+Protocol) developer tooling to cross-browser tooling.
 
-This document presents a possible design for a bidirectional WebDriver protocol, incorporating scenarios and resolutions discussed at the TPAC 2019 working group meeting. The protocol uses JSON-RPC messaging over WebSockets as the transport mechanism. WebDriver's current model of the browser is extended to include service workers and other non-page targets and make it possible for clients to target these additional contexts. We also discuss how the new protocol can interoperate with the existing protocol. Sample protocol messages illustrating how the protocol would work are included, and an JSON API specification is included alongside the document.
+## Motivation
 
-## Goals
+Many developer tools are exclusively targetting Chrome, relying on CDP, resulting in websites being
```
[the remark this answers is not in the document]
foolip: @whimboo you might be interested in this wording as well, given your comment at https://github.com/w3c/webdriver-bidi/pull/98#discussion_r628046799

explainer.md
```diff
@@ -31,19 +43,50 @@ The protocol is designed with the following goals in mind:
...
+## Non-goals
+
+Feature parity with CDP is a non-goal at this time; many features of CDP are rarely used by
+existing developer tooling, and being able to entirely supplant CDP is not a goal at this time.
+
+## Prior Art
+
+\[FIXME: [CDP](https://chromedevtools.github.io/devtools-protocol/),
+         [Firefox Remote Debug Protocol](https://firefox-source-docs.mozilla.org/devtools/backend/protocol.html),
+         [Firefox Remote Protoco](https://wiki.mozilla.org/WebDriver/RemoteProtocol)l,
```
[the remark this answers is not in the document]
gsnedders: This isn't really prior art in the same sense, given it's never been publicly exposed.

Issue #97, opened 2021-03-30
Define session creation without HTTP driver


jgraham:
Some implementations want to support connecting without requiring an initial HTTP request. This is required for feature parity with CDP-based clients e.g. Puppeteer that are able to establish a connection to the browser directly without going through a seperate driver binary.

The following discussion will assume a WebSockets based transport, but the same issues would apply to an implementation that wanted to allow connections over e.g. a unix pipe without the initial HTTP handshake.

Currently once a session is created, the HTTP layer returns a websocket url of the form `ws://localhost:<port>/session/<sessionid>` for the client to connect to. Since this requires the session id to be known it's clear that this doesn't work well for establishing a session directly over WebSockets. An obvious implementation would be to allow connecting to `ws://localhost:<port>/session/` and defining a command like

```
SessionNewCommand = {
  method: "session.new",
  params: SessionNewParameters
}

SessionSubscribeParameters = {
  ? alwaysMatch: Capabilities,
  ? firstMatch: [*Capabilities],
}

Capabilities = {
  *text: any
}
```

Then, if you send this command when there's no existing session it would create a session and return a response with the matched capabilities and otherwise it would error.

One question is whether the session itself should reuse the connection; it would be the "wrong" resource since the session id wouldn't be in the path. But I don't immediately see a practical problem with reusing it in this case, and the alternative would require the client to establish a new connection which adds latency.

One wart might be if you want to allow reconnecting to the session once it dropped; in this case it might be necessary to connect ot the URL including the session id (to account for nodes accepting multiple sessions). That is also probably sufficient reason not to change the spec to make `/session` the only supported ws resource; in a node that supports multiple sessions that would make it hard to work out which session to reconnect to.

Another question is with this setup is how to to communicate the ws port to the local end. This is analogous to the problem of how to communicate the HTTP server address to local ends, and is usually solved either by putting the local end in control and allowing the client to select the address through remote-specific options (with some risk of races) or by communicating the address back through stdout of the client.

whimboo:
> Some implementations want to support connecting without requiring an initial HTTP request. This is required for feature parity with CDP-based clients e.g. Puppeteer that are able to establish a connection to the browser directly without going through a seperate driver binary.
>
> The following discussion will assume a WebSockets based transport, but the same issues would apply to an implementation that wanted to allow connections over e.g. a unix pipe without the initial HTTP handshake.
 
[...]

PR #99, opened 2021-04-13
Add the ability to start a session directly with the bidi connection

jgraham:
The case for this is:

* Some clients might be uninterested in using the classic WebDriver
HTTP connection, and want to avoid depending on a HTTP client just for
session creation.

* Some implementations might ship the HTTP server as a seperate
component, but have the bidi connection built-in to the browser
itself. This will allow clients to use bidi without a seperate driver
binary (similar to devtools) without requiring the browser to ship a
full HTTP implementation just for new session.

* It is needed to "explain" the behaviour of classic WebDriver in
terms of the BiDI protocol; in particular the new session and session
status command.

Implementation-wise the model is that a remote end is allowed to start
a WebSockets server that accepts connections to the `/session`
resource. This connection can then be used for "static" commands
i.e. those which don't require a session. Currently this is just
`session.status` and `session.new`. Once `session.new` is called, the
same WS connection is reused for subsequent commands that are part of
the session.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/99.html" title="Last updated on Sep 22, 2021, 2:27 PM UTC (495cce0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/99/fbafbf3...495cce0.html" title="Last updated on Sep 22, 2021, 2:27 PM UTC (495cce0)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1605,7 +1605,6 @@ Capabilities = {
     ?browserName: text,
     ?browserVersion: text,
     ?platformName: text,
-    ?pageLoadStrategy: "none" / "eager" / "normal",
     ?proxy: {
         ?proxyType: "pac" / "direct" / "autodetect" / "system" / "manual",
         ?proxyAutoconfigUrl: text,
@@ -1615,15 +1614,7 @@ Capabilities = {
         ?sslProxy: text,
         ?socksProxy: text,
         ?socksVersion: int,
-    }
-    ?strictFileInteractability: bool,
-    ?timeouts: {
-        ?script: int / null,
-        ?pageLoad: int,
-        ?implicit: int
     },
-    ?unhandledPromptBehaviour: "dismiss" / "accept" / "dismiss and notify" / "accept and notify" / "ignore",
-    ?webSocketUrl: bool,
     *text => any
 };
 </pre>
@@ -1716,7 +1707,6 @@ This is a [=static command=].
           browserName: text,
           browserVersion: text,
           platformName: text,
-          pageLoadStrategy: "none" / "eager" / "normal",
           proxy: {
             ?proxyType: "pac" / "direct" / "autodetect" / "system" / "manual",
             ?proxyAutoconfigUrl: text,
@@ -1728,14 +1718,6 @@ This is a [=static command=].
             ?socksVersion: int,
           },
           setWindowRect: bool,
-          strictFileInteractability: bool,
-          timeouts: {
-            script: int / null,
-            pageLoad: int,
-            implicit: int
-          },
-          ?unhandledPromptBehaviour: "dismiss" / "accept" / "dismiss and notify" / "accept and notify" / "ignore",
-          ?webSocketUrl: text,
           *text => any
         }
       }
--- a/index.bs
+++ b/index.bs
@@ -2258,7 +2258,7 @@ browsing context to the given URL.
 </dl>
 
 <div algorithm="remote end steps for browsingContext.navigate">
-The [=remote end steps=] with |command parameters| are:
+The [=remote end steps=] with <var ignore>session</var> and |command parameters| are:
 
   1. Let |context id| be the value of the <code>context</code> field of
      |command parameters|.
@@ -2479,8 +2479,6 @@ become inaccessible but not yet get discarded because bfcache.
 The [=remote end event trigger=] is the <dfn export>WebDriver BiDi navigation
 started</dfn> steps given |context| and |navigation status|:
 
- 1. If the [=current session=] is null, return.
-
  1. Let |params| be the result of [=get the navigation info=] given |context|
     and |navigation status|.
 
@@ -2519,8 +2517,6 @@ started</dfn> steps given |context| and |navigation status|:
 <div algorithm>
 The [=remote end event trigger=] is the <dfn export>WebDriver BiDi fragment
 navigated</dfn> steps given |context| and |navigation status|:
-
- 1. If the [=current session=] is null, return.
 
  1. Let |params| be the result of [=get the navigation info=] given |context|
     and |navigation status|.
@@ -2561,8 +2557,6 @@ navigated</dfn> steps given |context| and |navigation status|:
 The [=remote end event trigger=] is the <dfn export>WebDriver BiDi DOM content
 loaded</dfn> steps given |context| and |navigation status|:
 
- 1. If the [=current session=] is null, return.
-
  1. Let |params| be the result of [=get the navigation info=] given |context|
     and |navigation status|.
 
@@ -2602,8 +2596,6 @@ loaded</dfn> steps given |context| and |navigation status|:
 The [=remote end event trigger=] is the <dfn export>WebDriver BiDi load
 complete</dfn> steps given |context| and |navigation status|:
 
- 1. If the [=current session=] is null, return.
-
  1. Let |params| be the result of [=get the navigation info=] given |context|
     and |navigation status|.
 
@@ -2641,8 +2633,6 @@ complete</dfn> steps given |context| and |navigation status|:
 <div algorithm>
 The [=remote end event trigger=] is the <dfn export>WebDriver BiDi download
 started</dfn> steps given |context| and |navigation status|:
-
- 1. If the [=current session=] is null, return.
 
  1. Let |params| be the result of [=get the navigation info=] given |context|
     and |navigation status|.
@@ -2682,8 +2672,6 @@ started</dfn> steps given |context| and |navigation status|:
 The [=remote end event trigger=] is the <dfn export>WebDriver BiDi navigation
 aborted</dfn> steps given |context| and |navigation status|:
 
- 1. If the [=current session=] is null, return.
-
  1. Let |params| be the result of [=get the navigation info=] given |context|
     and |navigation status|.
 
@@ -2722,8 +2710,6 @@ aborted</dfn> steps given |context| and |navigation status|:
 <div algorithm>
 The [=remote end event trigger=] is the <dfn export>WebDriver BiDi navigation
 failed</dfn> steps given |context| and |navigation status|:
-
- 1. If the [=current session=] is null, return.
 
  1. Let |params| be the result of [=get the navigation info=] given |context|
     and |navigation status|.
--- a/index.bs
+++ b/index.bs
@@ -46,13 +46,17 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
     text: getting a property; url: dfn-get-a-property
     text: intermediary node; url: dfn-intermediary-node
     text: invalid argument; url: dfn-invalid-argument
+    text: invalid session id; url: dfn-invalid-session-id
     text: local end; url: dfn-local-ends
     text: matched capability serialization algorithm; url: dfn-matched-capability-serialization-algorithm
+    text: maximum active sessions; url: dfn-maximum-active-sessions
     text: no such element; url: dfn-no-such-element
     text: no such frame; url: dfn-no-such-frame
+    text: process capabilities; url: dfn-processing-capabilities
     text: remote end steps; url: dfn-remote-end-steps
     text: remote end; url: dfn-remote-ends
     text: session ID; url: dfn-session-id
+    text: session not created; url: dfn-session-not-created
     text: session; url: dfn-sessions
     text: set a property; url: dfn-set-a-property
     text: success; url: dfn-success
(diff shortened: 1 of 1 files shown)


jgraham:
There is a big open issue about how to deal with capabilities that are maybe not relevant for non-HTTP sessions, but I think this is ready for some wider review.

jgraham:
I've reworked this to depend on the changes in https://github.com/w3c/webdriver/pull/1608

This allows the possibility of both BiDi sessions and multiple sessions (though doesn't require support for either).

index.bs
```diff
@@ -619,10 +675,16 @@ To <dfn>handle an incoming message</dfn> given a [=WebSocket connection=]
 
     1. Let |method| be |matched|["<code>method</code>"]
 
+    1. Let |command| be the command with [=command name=] |method|.
+
+    1. If |session| is null and |command| is not a [=static command=], then
+       [=respond with an error=] given |connection|, |command id|, and [=session
+       not created=], and return.
```
whimboo, line 682: Please not that we have to return an `invalid session id` error here.

index.bs
```diff
@@ -455,9 +489,12 @@ established from |listener| must be TLS encrypted.
...
 listeners</dfn>, which is initially empty.
 
-A WebDriver [=/session=] has a <dfn>WebSocket connection</dfn> which is
-a network connection that follows the requirements of the
-[[!RFC6455|WebSocket protocol]]. This is initially null.
+A [=remote end=] has a [=set=] of <dfn>WebSocket connections not associated with a
+session</dfn>, which is initially empty.
+
+A [=WebDriver BiDi session=] has a <dfn>WebSocket connection</dfn> which is a
+network connection that follows the requirements of the [[!RFC6455|WebSocket
```
whimboo, line 496: Didn't we aim to allow multiple connections per WebDriver session to allow re-connections after possible network drops? Or will that be handled in a different (follow-up) PR?

index.bs
```diff
@@ -781,23 +843,41 @@ with parameter |value| is:
...
+
+1. Let |webSocketUrl| be the result of [=getting a property=] named
+   "<code>webSocketUrl</code>" from |capabilities|.
 
- 1. Let |webSocketUrl| be the result of [=getting a property=] named
-    "<code>webSocketUrl</code>" from |capabilities|.
+1. If |webSocketUrl| is undefined, or false, return.
 
- 2. If |webSocketUrl| is undefined, return.
+1. [=Assert=]: |webSocketUrl| is true.
```
whimboo, line 855: Shouldn't we specify the type of error to be returned if it fails?

index.bs
```diff
@@ -1548,6 +1663,86 @@ The [=remote end steps=] are:
...
+Note: A session created this way will not be accessible via HTTP.
+
+This is a [=static command=].
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+      <pre class="cddl remote-cddl">
+      SessionNewCommand = {
+        method: "session.status",
```
whimboo, line 1680: s/session.status/session.new/

index.bs
```diff
@@ -1503,6 +1583,36 @@ filter only those that ought to be returned to the local end.
...
+        ?proxyType: "pac" / "direct" / "autodetect" / "system" / "manual",
+        ?proxyAutoconfigUrl: text,
+        ?ftpProxy: text,
+        ?httpProxy: text,
+        ?noProxy: [*text],
+        ?sslProxy: text,
+        ?socksProxy: text,
+        ?socksVersion: int,
+    },
+    *text => any
```
whimboo, line 1608: What about `setWindowRect`?

index.bs
```diff
@@ -1548,6 +1663,86 @@ The [=remote end steps=] are:
...
+
+1. Let |session| be the result of [=trying=] to [=create a session=] with
+   |capabilities| and |flags|.
+
+1. Set |session|'s [=BiDi flag=] to true.
+
+   Note: the connection for this session will be set to the current connection
+   in the caller.
+
+1. Let |body| be a new map matching the <code>SessionNewResult<code> production,
```
juliandescottes, line 1740:
```suggestion
1. Let |body| be a new map matching the <code>SessionNewResult</code> production,
```

whimboo:
Fixes issue #97

index.bs
```diff
@@ -473,21 +510,33 @@ accept the incoming connection:
    running these steps and act as if the requested service is not
    available.
 
-2. [=Get a session ID for a WebSocket resource=] with |resource name|
+1. If |resource name| is the byte string "<code>/session</code>",
+   and the implementation [=supports BiDi-only sessions=]:
```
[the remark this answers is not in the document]
jgraham: I was letting it be optional at the moment to allow the possibility that e.g. Safari won't support this. I don't think in practice it's going to make much difference if we frame it as optional in the spec or not, if everyone does support it we can simplify a bit by removing the possibility of not supporting it, but making it mandatory won't be enough to make it supported by anyone who wants to require creating a HTTP session first.

index.bs
jgraham: There is no error. If the flag isn't set we didn't request a BiDi session, so we don't do anything for BiDi but could still create a HTTP-only session.

index.bs
jgraham: I believe that only appears in the response; it doesn't seem to be something that's handled as part of the request. Maybe it should be?

index.bs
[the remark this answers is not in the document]
whimboo: Sounds fine.

index.bs
whimboo: Right, but that is covered by the former check on line 874. Here it would still be possible to pass in any type, which might not be a boolean. As such we can fail.

index.bs
whimboo: I'm not exactly sure how the capability matching works from the client side. There might be cases when there is a requirement for tests to actually enforce moving and resizing of windows. I doubt this could be covered by just the platform capability? There might still be window managers (?) that on desktop that don't allow it?

But this would also require a change in the WebDriver HTTP spec, so maybe we leave it as is (at least for now).

index.bs
```diff
@@ -262,7 +267,16 @@ EventData = (
...
-WebDriver BiDi uses the same [=/session=] concept as WebDriver.
+WebDriver BiDi extends the [=/session=] concept from WebDriver.
+
+A [=/session=] has a <dfn>BiDi flag</dfn>, which is false unless otherwise
+stated.
+
+A <dfn>WebDriver BiDi Session</dfn> is a [=/session=] which has the [=BiDi flag=]
+set to true.
+
+An implementation has a list of <dfn>active BiDI sessions</dfn>, which is
```
foolip, line 278: It looks like nothing is ever added to this list. Rather than adding that bookkeeping, however, can we just handwave "all WebDriver BiDi sessions" or something?

Typo fix if we keep it:

```suggestion
An implementation has a list of <dfn>active BiDi sessions</dfn>, which is
```

index.bs
```diff
@@ -262,7 +267,16 @@ EventData = (
 
 ## Session ## {#session}
 
-WebDriver BiDi uses the same [=/session=] concept as WebDriver.
+WebDriver BiDi extends the [=/session=] concept from WebDriver.
+
+A [=/session=] has a <dfn>BiDi flag</dfn>, which is false unless otherwise
+stated.
+
+A <dfn>WebDriver BiDi Session</dfn> is a [=/session=] which has the [=BiDi flag=]
```
foolip, line 275: Since it's just "session" in https://w3c.github.io/webdriver/#dfn-sessions I figure the more consistent name would be just "BiDi session" without "WebDriver":

```suggestion
A <dfn>BiDi session</dfn> is a [=/session=] which has the [=BiDi flag=]
```

index.bs
```diff
@@ -481,23 +527,35 @@ accept the incoming connection:
...
+   and the implementation [=supports BiDi-only sessions=]:
+
+    1. Run any other implementation-defined steps to decide if the
+       connection should be accepted, and if it is not stop running these
+       steps and act as if the requested service is not available.
+
+    1. Add the connection to the set of [=WebSocket connections not associated
+       with a session=].
+
+    1. Return
```
foolip, line 540:
```suggestion
    1. Return.
```

index.bs
```diff
@@ -745,10 +813,17 @@ To <dfn>respond with an error</dfn> given a [=WebSocket connection=]
...
- 1. If there is a WebDriver [=/session=] with |connection| as its [=connection=],
-    set the [=connection=] on that [=/session=] to null.
+ 1. If there is a [=WebDriver BiDi session=] [=associated with connection=]
+    |connection|:
+
+   1. Let |session| be the [=WebDriver BiDi session=] [=associated with
+      connection=] |connection|.
 
-Issue: This should also reset any internal state
+   1. Remoce |connection| from |session|'s [=session WebSocket
```
foolip, line 822:
```suggestion
   1. Remove |connection| from |session|'s [=session WebSocket
```

index.bs
```diff
@@ -1557,6 +1685,86 @@ The [=remote end steps=] are:
...
+        ?firstMatch: [*CapabilitiesRequest]
+      }
+      </pre>
+   </dd>
+   <dt>Return Type</dt>
+   <dd>
+      <pre class="cddl local-cddl">
+      SessionNewResult = {
+        sessionId: text,
+        capabilities: {
```
foolip, line 1717: Some of these make sense in the context of WebDriver over HTTP, but not really for BiDi if connecting directly to a browser. `browserName` in particular :)

I'm also not sure about whether it might be too late for `acceptInsecureCertificates` and `proxy`, if those might be things that are implemented as command line arguments.

Are there any issues for Firefox to implement this as written?

index.bs
```diff
@@ -1557,6 +1685,86 @@ The [=remote end steps=] are:
...
+1. Let |capabilities| be the result of [=trying=] to [=process capabilities=]
+   with |command parameters| and |flags|.
+
+1. Let |session| be the result of [=trying=] to [=create a session=] with
+   |capabilities| and |flags|.
+
+1. Set |session|'s [=BiDi flag=] to true.
+
+   Note: the connection for this session will be set to the current connection
+   in the caller.
```
foolip, line 1760:
```suggestion
   by the caller.
```

index.bs
```diff
@@ -789,23 +864,41 @@ with parameter |value| is:
...
+
+1. If |flags| contains "<code>bidi</code>", return.
 
- 1. Let |webSocketUrl| be the result of [=getting a property=] named
-    "<code>webSocketUrl</code>" from |capabilities|.
+1. Let |webSocketUrl| be the result of [=getting a property=] named
+   "<code>webSocketUrl</code>" from |capabilities|.
 
- 2. If |webSocketUrl| is undefined, return.
+1. If |webSocketUrl| is undefined, or false, return.
```
foolip, line 874: I don't have a strong position on the Oxford comma, but this comma made it harder to read IMHO:

```suggestion
1. If |webSocketUrl| is undefined or false, return.
```

index.bs
foolip: I'm not sure if `webSocketUrl` is guaranteed to be true, but if it's not we shouldn't use an assert here at all.

index.bs
jgraham: Well for a direct connection to a browser `browserName` doesn't make any sense over HTTP either. The use case is something like Selenium Grid, and it at least seems plausible that the same thing can work without requiring a HTTP session: the grid server provides a WebSocket endpoint to connect to, and handles the browser selection and setup.

For for `acceptInsecureCertificates` and `proxy`, if you can't set those at runtime, and the browser is already started, you'd reject the session. But in a case like Puppeteer where the same software starts the browser and then generates a new session request that won't happen. It does means duplicating some information between the command line arguments (or profile settings, or whatever), but it also allows the same setup to work when the software that starts the browser isn't the one generating the new session message (again consider the case of a setup with middleware like a grid server; in this case the client doesn't start the browser but connects to the server and sends it the new session message, which it uses to configure the browser and then forwards on).

For Firefox in particular I think we can set those things at runtime and not just at startup time. But of course there are other things (like command line arguments) which we can't set at runtime.

index.bs
jgraham: It is, hence the assert :) If `webSocketUrl` was set to something other than a boolean we'd fail during capabilites validation, see the `webSocketUrl capability deserialization algorithm` slightly further up.

index.bs
jgraham: Yes, I think that's out of scope for this issue. If there's a problem here let's file an issue on the WebDriver classic spec.

index.bs
foolip: I see, the behavior when connecting directly to a browser is defined then and not a mystery. Some of this might be tricky to implement when the protocol is built into the browser binary and we might not notice until it's built in, but anyway the model seems sensible.

foolip (approved):

PR #105, opened 2021-05-07
CI: Autopublish w/ GitHub Actions + w3c/spec-prod

sideshowbarker:


foolip:
See also https://github.com/w3c/webdriver-bidi/pull/54, existing workflow needs to be removed.

foolip:
I've merged https://github.com/w3c/webdriver-bidi/pull/54 instead.

PR #107, opened 2021-05-25
Specify a browsingContext.reload command

jgraham:
This works just like navigation, except it has no URL and instead
takes an additional parameter to allow specifying a
force-reload. However adding this into HTML looks like a pretty
significant task, so for now the spec just handwaves how it will
behave.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/107.html" title="Last updated on Sep 14, 2021, 7:34 PM UTC (ea398f9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/107/916796d...ea398f9.html" title="Last updated on Sep 14, 2021, 7:34 PM UTC (ea398f9)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -101,12 +101,14 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
     text: create a new browsing context; url: creating-a-new-browsing-context
     text: environment settings object's Realm; url: environment-settings-object's-realm
     text: handled; url: concept-error-handled
+    text: history handling behavior; url: history-handling-behavior
     text: navigation id; url: concept-navigation-id
     text: report an error; url: report-the-error
     text: remove a browsing context; url: bcg-remove
     text: session history; url: session-history
     text: set up a window environment settings object; url: set-up-a-window-environment-settings-object
     text: set up a worker environment settings object; url: set-up-a-worker-environment-settings-object
+    text: set up a worklet environment settings object; url: set-up-a-worklet-environment-settings-object
     text: worker event loop; url: worker-event-loop-2
     text: worklet global scopes; url: concept-document-worklet-global-scopes
 </pre>
@@ -1702,7 +1704,8 @@ navigation status</dfn> struct, which has the following items:
 
 BrowsingContextCommand = (
     BrowsingContextGetTreeCommand //
-    BrowsingContextNavigateCommand
+    BrowsingContextNavigateCommand //
+    BrowsingContextReloadCommand
 )
 </pre>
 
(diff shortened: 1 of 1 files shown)


jgraham:
Note that this depends on #93

whimboo:
> <AutomatedTester> whimboo (IRC):are we only allowing this for top level browser context to match browsers
> <AutomatedTester> jgraham (IRC): I assumed it was any context and I don't think I have that limitation in the PR

What's missed in the notes is that I also mentioned that in WebDriver HTTP we only allow reloads for the top-level browsing context: https://w3c.github.io/webdriver/#refresh.

So we can call this an enhancement? Should the HTTP spec also be updated?

bwalderman (approved):
LGTM. Looks like specifying the "ignore cache" behavior will happen mostly in the HTML spec and there would only be minimal changes here once the exact mechanism for passing the flag is known.

index.bs
```diff
@@ -1830,6 +1831,106 @@ To <dfn>get the browsing context info</dfn> given |context|,
...
+
+<div algorithm>
+To <dfn>await a navigation</dfn> given |context|, |request| and optionally
+|history handling| (default: "<code>default</code>") and |ignore cache| (default: false):
+
+1. Let |navigation id| be the string representation of a
+   [[!RFC4122|UUID]] based on truly random, or pseudo-random numbers.
+
+1. [=Navigate=] |context| with resource |request|, and using |context| as the
+   [=source browsing context=], with navigation id |navigation id|, and history
```
foolip, line 1844: The name of the parameter in https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigate is `navigationId`, I think we should either use that or link the concept:

```suggestion
   [=source browsing context=], with <var ignore>navigationId</var> |navigation id|, and history
```

or

```suggestion
   [=source browsing context=], with [=navigation id=] |navigation id|, and history
```

Same for `historyHandling`.

index.bs
```diff
@@ -1830,6 +1831,106 @@ To <dfn>get the browsing context info</dfn> given |context|,
...
+
+   Issue: property specify how the |ignore cache| flag works. This needs to
+   consider whether only the first load of a resource bypasses the cache
+   (i.e. whether this is like initially clearing the cache and proceeding like
+   normal), or whether resources not directly loaded by the HTML parser
+   (e.g. loads initiated by scripts or stylesheets) also bypass the cache.
+
+1.  Let (|event received|, |navigate status|) be [=await=] given
+    «"<code>navigation started</code>", "<code>navigation failed</code>",
+    and "<code>fragment navigated</code>"» and |navigation id|.
```
foolip, line 1856:
```suggestion
   "<code>fragment navigated</code>"» and |navigation id|.
```

The "and" inside the « » notation doesn't follow https://infra.spec.whatwg.org/#lists.

(And then I noticed this was just moved, not new, so I won't comment on more things in this block.)

index.bs
```diff
@@ -1994,93 +2095,62 @@ The [=remote end steps=] with |command parameters| are:
...
-     where local ends parse the implementation-defined strings to figure out
-     what actually went wrong?
+1. Let |URL| be |document|'s <a spec=DOM>URL</a>.
 
-  1. Let |body| be a [=map=] matching the
-     <code>BrowsingContextNavigateResult</code> production, with the
-     <code>navigation</code> field set to |status|'s id, and the
-     <code>url</code> field set to the result of the [=URL serializer=] given
-     |status|'s url.
+1. Let |request| be a new [=/request=] whose URL is |URl|.
```
foolip, line 2146:
```suggestion
1. Let |request| be a new [=/request=] whose URL is |URL|.
```

Although, just lowercasing the variable name would be more idiomatic I think.

foolip:
@jgraham I had only minor comments, the most important thing here I'd say is rebasing to make sure this still builds correctly.

index.bs
```diff
@@ -1997,93 +2082,64 @@ The [=remote end steps=] with |command parameters| are:
...
-     where local ends parse the implementation-defined strings to figure out
-     what actually went wrong?
+1. Let |url| be |document|'s <a spec=DOM>URL</a>.
 
-  1. Let |body| be a [=map=] matching the
-     <code>BrowsingContextNavigateResult</code> production, with the
-     <code>navigation</code> field set to |status|'s id, and the
-     <code>url</code> field set to the result of the [=URL serializer=] given
-     |status|'s url.
+1. Let |request| be a new [=/request=] whose URL is |url|.
```
foolip, line 2134: I looked around to see if browsing contexts have URLs, but it seems like not. That's good, this seems right.

foolip (approved):

PR #109, opened 2021-06-04
Add a browsingContext.traverseHistory command

jgraham:
This traverses the history by a specified `delta` (e.g. +1 for forward,
or -1 for back). Like other navigation commands the `wait` parameter
controls whether to return as soon as the history traversal starts, or
wait for a specified level of completeness.

Mush like other navigation commands, history traversal is tracked with
a unique navigation id.

Typically history traversal results in a navigation, but in the case
of the history entry being in the bfcache, no navigation is
necessary. In this case any non-"none" value for wait will return once
the `pageshow` event is fired, and the `persisted` attribute of the
return value is set to `true`.

This model differs somewhat from the CDP model where one navigates to
an explicit history entry and failure is only possible if the entry id
is invalid. But back/forwward only seems closer to the use cases we
have, and allowing events to be traced to a specific traversal seems
consistent with the way we handle other navigation-related events.


jgraham:
This depends on #107 plus some additional changes to the HTML spec that I haven't created a PR for yet.

index.bs
```diff
@@ -1993,6 +1995,116 @@ The [=remote end steps=] with |command parameters| are:
...
+   <dd>
+      <pre class="cddl remote-cddl">
+      BrowsingContextTraverseHistoryCommand = {
+        method: "browsingContext.traverseHistory",
+        params: BrowsingContextTraverseHistoryParameters
+      }
+
+      BrowsingContextReloadParameters = {
+        context: BrowsingContext,
+        delta: int,
```
whimboo, line 2014: Using a delta to define the number of history entries to move back or forward means that we also need an API to actually retrieve the maximum amount of entries.

Maybe pages might also be interested in to go back/forward to a specific URL.

index.bs
```diff
@@ -1993,6 +1995,116 @@ The [=remote end steps=] with |command parameters| are:
...
+  1. Assert: |context| is not null.
+
+  1. Let |delta| be the value of the <code>delta</code> field of |command
+     parameters|.
+
+  1. Let |wait condition| be the value of the <code>wait</code> field of |command
+     parameters| if present, or "<code>none</code>" otherwise.
+
+  1. Let |navigate status| be the result of running the [=traverse the history
+     by a delta steps=] given |delta|, and |context|.
```
whimboo, line 2049: What should happen if `delta` is outside of the maximum available pages in the history? Should it be silently ignored as in this case, or a specific error raised?

index.bs
```diff
@@ -1993,6 +1995,116 @@ The [=remote end steps=] with |command parameters| are:
...
+     "<code>load</code>".
+
+ 1. Let (|event received|, |status|) be [=await=] given «|event name|,
+    "<code>download started</code>", "<code>navigation failed</code>",
+    "<code>page show</code>"» and |navigation id|.
+
+ 1. If |event received| is "<code>navigation failed</code>"
+    return [=error=] with [=error code=] [=unknown error=].
+
+ 1. If |event received| is "<code>page show</code>", let |persisted| be true,
```
whimboo, line 2076: Shouldn't this be set to the value of the `persisted` property of `pageshow`?

index.bs
```diff
@@ -1993,6 +1995,116 @@ The [=remote end steps=] with |command parameters| are:
...
+  1. Return [=success=] with data |body|.
+
+</div>
+
+<div algorithm>
+
+The <dfn export>WebDriver-BiDi page show</dfn> steps given <var
+ignore>context</var> and |navigation status| are:
+
+Issue: Do we want to expose a `browsingContext.pageShow event? In that case we'd
```
whimboo, line 2094: Isn't that what we will do for navigation anyway when no wait parameter has been given and the command returns immediately?

index.bs
jgraham: It doesn't mean we need that if there aren't use cases for it. HTTP-based WebDriver just has `/back` and `/forward`. Puppeteer has [goBack](https://pptr.dev/#?product=Puppeteer&version=v9.1.1&show=api-pagegobackoptions) and `goForward`, which [PlayWright](https://playwright.dev/docs/api/class-page/#pagegobackoptions) seems to have kept unchanged. Cypress has [go](https://playwright.dev/docs/api/class-page/#pagegobackoptions) which takes a delta. I can't see any of them having an API to go back/forward to a specific URL or to get a the size of the session history.

I think if those use cases come up later we'd be better off adding an API that returns a list of session history entries, like CDP. Such an PI seems useful for testing the platform itself, but it's hard to argue it's part of the MVP for replacing existing automation tools given none of them provide it.

index.bs
jgraham: There will be an error although at this point there's a lot of `unknown error`. I (or someone else!) need to put some time into rewriting the error codes for BiDi.

index.bs
jgraham: Not as currently written because the hook in the HTML spec (which needs a seperate PR) is only invoked if the page came out of the bfcache, not if the page was loaded normally. Therefore `persisted` is always true in this case. If we change things so there's a BiDi `context.pageShow` event then we'll need to set the persisted property in the hook and reflect that here.

jgraham:
https://github.com/whatwg/html/pull/6921 for the HTML side of this change.

index.bs
jgraham: The question is whether we want a specific event for it so that clients can wait for pageShow even if it's triggered by something other than a navigation command. I'm inclined to think it would probably be a good idea, given pageshow/pagehide are typically the most reliable things to use in the face of bfcache. I don't quite know if it's covered by `Page.lifecycleEvent` in CDP since the docs don't mention when the events are actually emitted.

index.bs
whimboo: @jgraham would you mind filing a new issue for that with details? Would be good to keep that on our radar.

jgraham: #112 covers this.

index.bs
```diff
@@ -1830,6 +1834,106 @@ To <dfn>get the browsing context info</dfn> given |context|,
...
+
+1. If |navigate status|'s status is "<code>canceled</code>" return [=error=]
+   with [=error code=] [=unknown error=].
+
+   TODO: is this the right way to handle errors here?
+
+1. Assert: |navigate status|'s status is "<code>pending</code>" and
+   |navigation id| is not null.
+
+1. If |wait condition| is "<code>none</code>":
```
bwalderman, line 1889: `wait condition` needs to be passed as a parameter to this algorithm.

index.bs
```diff
@@ -1994,93 +2098,203 @@ The [=remote end steps=] with |command parameters| are:
...
+1. Let |context| be the result of [=trying=] to [=get a browsing context=]
+   with |context id|.
 
-  1. If |navigate status|'s status is "<code>canceled</code>" return [=error=]
-     with [=error code=] [=unknown error=].
+1. Assert: |context| is not null.
 
-     TODO: is this the right way to handle errors here?
+1. Let |ignore cache| be the the value of the <code>ignoreCache</code> field of |command
+   parameters| if present, or false otherwise.
```
bwalderman, line 2141:
```suggestion
1. Let |wait condition| be the value of the <code>wait</code> field of |command
   parameters| if present, or "<code>none</code>" otherwise.
```

bwalderman: Also need to pass this to `await a navigation` below.

bwalderman (approved):
LGTM. Just left some minor comments.

index.bs
```diff
@@ -1699,7 +1700,9 @@ navigation status</dfn> struct, which has the following items:
 
 BrowsingContextCommand = (
     BrowsingContextGetTreeCommand //
-    BrowsingContextNavigateCommand
+    BrowsingContextNavigateCommand //
+    BrowsingContextReloadCommand //
```
foolip, line 1704: I guess this will go away when rebasing.

index.bs
```diff
@@ -1830,6 +1834,107 @@ To <dfn>get the browsing context info</dfn> given |context|,
...
+
+1. If |navigate status|'s status is "<code>complete</code>":
+
+  1. Let |body| be a [=map=] matching the
+     <code>BrowsingContextNavigateResult</code> production, with the
+     <code>navigation</code> field set to |navigation id|, and the
+     <code>url</code> field set to the result of the [=URL serializer=] given
+     |navigate status|'s url.
+
+  1. Return [=success=] with data |body|, and then run the following steps [=in
```
foolip, line 1872: Will the "in parallel" here be replaced with "queue a task" after https://github.com/w3c/webdriver-bidi/pull/115?

index.bs
```diff
@@ -1994,93 +2099,208 @@ The [=remote end steps=] with |command parameters| are:
...
+ 1. Let |navigation id| be |navigation status|'s id.
+
+ 1. [=Resume=] with "<code>page show</code>", |navigation id|, and
+    |navigation status|.
+
+</div>
+
+<div algorithm>
+
+The <dfn export>WebDriver-BiDi pop state</dfn> steps given <var
```
foolip, line 2291:
```suggestion
The <dfn export>WebDriver BiDi pop state</dfn> steps given <var
```

index.bs
```diff
@@ -1830,6 +1834,107 @@ To <dfn>get the browsing context info</dfn> given |context|,
...
+  1. Let |body| be a [=map=] matching the
+     <code>BrowsingContextNavigateResult</code> production, with the
+     <code>navigation</code> field set to |navigation id|, and the
+     <code>url</code> field set to the result of the [=URL serializer=] given
+     |navigate status|'s url.
+
+  1. Return [=success=] with data |body|, and then run the following steps [=in
+     parallel=]:
+
+     1. Run the [=WebDriver-BiDi fragment navigated=] steps given |context|
```
foolip, line 1875: Matching https://github.com/whatwg/html/pull/6921:

```suggestion
     1. Run the [=WebDriver BiDi fragment navigated=] steps given |context|
```

index.bs
```diff
@@ -1830,6 +1834,107 @@ To <dfn>get the browsing context info</dfn> given |context|,
...
+  1. Let |body| be a [=map=] matching the
+     <code>BrowsingContextNavigateResult</code> production, with the
+     <code>navigation</code> field set to |navigation id|, and the
+     <code>url</code> field set to the result of the [=URL serializer=] given
+     |navigate status|'s url.
+
+  1. Return [=success=] with data |body|, and then run the following steps [=in
+     parallel=]:
+
+     1. Run the [=WebDriver-BiDi navigation started=] steps given |context|
```
foolip, line 1901: This is all a bit hard to grasp initially, so to spell out what's going on with all the returning and in parallel: it's to ensure that the command response comes before any events. The "WebDriver BiDi navigation started" algorithm just emits the event, it doesn't call into HTML or anything else.

Also good ol' hyphen suggestion:

```suggestion
     1. Run the [=WebDriver BiDi navigation started=] steps given |context|
```

index.bs
```diff
@@ -1830,6 +1834,107 @@ To <dfn>get the browsing context info</dfn> given |context|,
...
+        and |navigate status|
+
+1. Run the [=WebDriver-BiDi navigation started=] steps given |context|
```
foolip, line 1904:
```suggestion
1. Run the [=WebDriver BiDi navigation started=] steps given |context|
```

index.bs
```diff
@@ -1994,93 +2099,208 @@ The [=remote end steps=] with |command parameters| are:
...
+   <code>persisted</code> field set to |persisted|, and the <code>url</code>
+   field set to the result of the [=URL serializer=] given |status|'s url.
+
+1. Return [=success=] with data |body|.
+
 </div>
 
+<div algorithm>
+
+The <dfn export>WebDriver-BiDi page show</dfn> steps given <var
```
foolip, line 2273:
```suggestion
The <dfn export>WebDriver BiDi page show</dfn> steps given <var
```

index.bs
```diff
@@ -1830,6 +1834,107 @@ To <dfn>get the browsing context info</dfn> given |context|,
...
+     <code>url</code> field set to the result of the [=URL serializer=] given
+     |navigate status|'s url.
+
+  1. Return [=success=] with data |body|, and then run the following steps [=in
+     parallel=]:
+
+     1. Run the [=WebDriver-BiDi fragment navigated=] steps given |context|
+        and |navigate status|
+
+   Note: this is the case if the navigation only caused the fragment to
```
foolip, line 1878: I'm not sure I follow what's going on here. This happens unconditionally, but history traversal isn't always a fragment navigation. Does this perfectly and intricately balance the now conditional invocation of "WebDriver BiDi fragment navigated" on the HTML side that I wondered about in https://github.com/whatwg/html/pull/6921#discussion_r714528869?

Also, capitalization nit since the note is forever:
```suggestion
   Note: This is the case if the navigation only caused the fragment to
```

index.bs
```diff
@@ -1830,6 +1834,107 @@ To <dfn>get the browsing context info</dfn> given |context|,
...
+  1. Return [=success=] with data |body|, and then run the following steps [=in
+     parallel=]:
+
+     1. Run the [=WebDriver-BiDi navigation started=] steps given |context|
+        and |navigate status|
+
+1. Run the [=WebDriver-BiDi navigation started=] steps given |context|
+   and |navigate status|
+
+   Note: this event was previously suppressed to ensure that it would come
```
foolip, line 1907: Closing the door to my initial reading that this is about a previous state of the spec:

```suggestion
   Note: This event was previously suppressed to ensure that it comes
```

index.bs
```diff
@@ -1830,6 +1834,107 @@ To <dfn>get the browsing context info</dfn> given |context|,
...
+   after the command response in the case that |wait condition| is
+   "<code>none</code>".
+
+   Issue: Replace this suppression mechanism with an event queue.
```
foolip, line 1911: Yes please 👍

index.bs
```diff
@@ -1994,93 +2099,208 @@ The [=remote end steps=] with |command parameters| are:
...
   1. Let |request| be a new [=/request=] whose URL is |url record|.
 
-  1. Let |navigation id| be the string representation of a
-     [[!RFC4122|UUID]] based on truly random, or pseudo-random numbers.
+  1. Return the result of [=await a navigation=] with |context|, |request|, and
+     |wait condition|.
+
+</div>
+
+#### The browsingContext.reload Command ####  {#command-browsingContext-reload}
```
foolip, line 2107: What's the browsingContext.reload addition doing here? I assume it'll go away with a rebase and will ignore here.

whimboo:
Hi @jgraham. I just stumbled over this PR and wonder what's actually left to do here. There are some review comments you might wanna have a look at.

foolip:
@jgraham I've merged to resolve conflicts, but further changes are still needed. The "child browsing contexts" concept is no longer in HTML, and perhaps other things.

index.bs
```diff
@@ -3062,6 +3065,146 @@ The [=remote end steps=] with |command parameters| are:
...
+1. Let |navigable| be the [=/navigable=] whose [=navigable/active
+   document=] is |browsing context|'s [=browsing context/active document=].
+
+1. Let |delta| be |command parameters|["<code>delta</code>"].
+
+1. If |command parameters| [=map/contains=] "<code>wait</code>, let |wait
+   condition| be |command parameters|["<code>wait</code>"]. Otherwise let |wait
+   condition| be "<code>none</code>".
+
+1. Let |navigation id| be the string representation of a
```
whimboo, line 3118: Could we maybe integrate all the following in `await a navigation`? We may be able to pass in another argument like `delta` to differentiate between a normal navigation and a history traversal. That way we wouldn't have to duplicate that much code.

index.bs
```diff
@@ -3062,6 +3065,146 @@ The [=remote end steps=] with |command parameters| are:
...
+</div>
+
+<div algorithm>
+
+The <dfn export>WebDriver BiDi page show</dfn> steps given <var
+ignore>context</var> and |navigation status| are:
+
+Issue: Do we want to expose a `browsingContext.pageShow event? In that case we'd
+need to call this whenever `pageshow` is going to be emitted, not just on
+bfcache restore, and also add the persisted status to the data.
```
whimboo, line 3187: @OrKoN would that be beneficial / helpful for Puppeteer?

whimboo (changes requested):

index.bs
OrKoN: I think it might be we currently do not support bfcache in Puppeteer. I think it's related to the prerendering topic too https://github.com/w3c/webdriver-bidi/issues/321 So far there is nothing in WebDriver BiDi that would cover navigations that are bfcache restore/prerendering activation. I think we will need to come up with something before restoring would result in a different browsing context becoming the primary context in a tab.

jgraham:
Not quite, let's move this one back to draft and it can cover the addition of `wait` and a navigation id.

Lightning00Blade:
I guess this was superseded by https://github.com/w3c/webdriver-bidi/pull/584.

Issue #116, opened 2021-07-14
Specify creating/opening a browsing context


foolip:
Before one can [navigate](https://github.com/w3c/webdriver-bidi/issues/43) a browsing context, one has to be able to open one.

whimboo:
When the browser gets started there is always a default tab open, which implies that also a top-level browsing context (navigable) exists. When the last tab gets closed, the session automatically ends and the browser quits. So under which condition do we have to block the navigation? Isn't it just that we have to retrieve the current or a list of known to be open browsing contexts?

jgraham:
I think the concern is that if one doesn't specify how to create a browsing context, one can't be _sure_ it already exists. Nothing really requires browsers to open a context at startup.

In practice a) browsers always do create a context at startup, so it's hard to get into a situation where there isn't one and b) we do need this feature for the same reasons we needed it in WebDriver classic. So I don't think we need to block on it but I also don't think we want to delay it too mcuh given it's not complex.

whimboo:
> I think the concern is that if one doesn't specify how to create a browsing context, one can't be _sure_ it already exists. Nothing really requires browsers to open a context at startup.

Sure but there is [browsingContext.getTree()](https://w3c.github.io/webdriver-bidi/#command-browsingContext-getTree) to get the list of open browsing contexts. Maybe we also just need a command to return the currently active navigable / browsing context.

foolip:
My concern stems from the fact that most Puppeteer examples one can find all open a new page and navigate it, rather than finding the already open page and using that. It's true that this doesn't block https://github.com/w3c/webdriver-bidi/issues/43, but it seems like something we want sooner rather than later.

Even with such a command, it would still be good to specify that there should be exactly one browsing context for a new session, if we want clients to be able to depend on that and not waste that browsing context.

jgraham:
We could speccify that there must be at least one browsing context at session creation, but we can't specify that there must be exactly one (consider connecting to an already running browser, or a tool that does passive monitoring of network traffic and rendering timing for performance analysis. rather than driving tests itself).

foolip:
At least one would do then :)

PR #133, opened 2021-09-08
Define the browsingContext.create command

foolip:
Fixes https://github.com/w3c/webdriver-bidi/issues/116.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/133.html" title="Last updated on Nov 30, 2021, 2:56 PM UTC (1ae5d85)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/133/9bfad74...1ae5d85.html" title="Last updated on Nov 30, 2021, 2:56 PM UTC (1ae5d85)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -120,6 +120,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
     text: set up a window environment settings object; url: set-up-a-window-environment-settings-object
     text: set up a worker environment settings object; url: set-up-a-worker-environment-settings-object
     text: set up a worklet environment settings object; url: set-up-a-worklet-environment-settings-object
+    text: window open steps; url: multipage/window-object.html#window-open-steps
     text: worker event loop; url: worker-event-loop-2
     text: worklet global scopes; url: concept-document-worklet-global-scopes
 </pre>
@@ -2232,11 +2233,31 @@ The <dfn export for=commands>browsingContext.create</dfn> command creates a new
 The [=remote end steps=] with |command parameters| are:
 
   1. Let |type| be the value of the <code>type</code> field of
-     |command parameters| if present, or "<code>window</code>" otherwise.
+     |command parameters|.
 
-  1. TODO: insert step 5 of https://w3c.github.io/webdriver/#new-window verbatim here?
+  <!-- This is based on step 5 of https://w3c.github.io/webdriver/#new-window,
+       but without using the "current browsing context" concept. -->
+  1. Create a new [=top-level browsing context=] by running the [=window open
+     steps=] with <var ignore>url</var> set to "<code>about:blank</code>",
+     <var ignore>target</var> set to the empty string, and
+     <var ignore>features</var> set to "<code>noopener</code>". This must be
+     done without invoking the [=/focusing steps=] for the created browsing
+     context. If |type| is "<code>tab</code>", and the implementation supports
+     multiple browsing contexts in the same OS window, the new browsing context
+     should share an OS window with any other [=top-level browsing context=]. If
+     |type| is "<code>window</code>", and the implementation supports multiple
+     browsing contexts in separate OS windows, the created browsing context
+     should be in a new OS window. In all other cases the details of how the
+     browsing context is presented to the user are implementation defined.
 
-  1. If |context id| be the [=browsing context id=] of the newly created [=/browsing context=].
+     Issue: HTML's [=window open steps=] depend on the "entry global object"
+     and WebDriver's <a href="https://w3c.github.io/webdriver/#new-window">New
+     Window</a> command depends on the "current browsing context". Given that we
+     don't have any such implicit context here, what OS window should new tabs
+     be opened in?
+
+  1. Let |context id| be the [=browsing context id=] of the newly created
+     [=/browsing context=].
 
   1. Let |body| be a [=map=] matching the <code>BrowsingContextCreateResult</code>
      production, with the <code>context</code> field set to |context id|.
--- a/index.bs
+++ b/index.bs
@@ -1930,6 +1930,7 @@ navigation status</dfn> struct, which has the following items:
 <pre class="cddl remote-cddl">
 
 BrowsingContextCommand = (
+    BrowsingContextCreate //
     BrowsingContextGetTreeCommand //
     BrowsingContextNavigateCommand //
     BrowsingContextReloadCommand
@@ -1941,6 +1942,7 @@ BrowsingContextCommand = (
 <pre class="cddl local-cddl">
 
 BrowsingContextResult = (
+    BrowsingContextCreateResult //
     BrowsingContextGetTreeResult //
     BrowsingContextNavigateResult
 )
@@ -2194,6 +2196,55 @@ To <dfn>get the navigation info</dfn>, given |context| and |navigation status|:
 
 ### Commands ### {#module-browsingContext-commands}
 
+#### The browsingContext.create Command ####  {#command-browsingContext-create}
+
+The <dfn export for=commands>browsingContext.create</dfn> command creates a new
+[=/browsing context=], either in a new window or in a new tab, and returns its
+[=browsing context id=].
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+      <pre class="cddl remote-cddl">
+      BrowsingContextCreateCommand = {
+        method: "browsingContext.create",
+        params: BrowsingContextCreateParameters
+      }
+
+      BrowsingContextCreateType = "tab" / "window"
+
+      BrowsingContextCreateParameters = {
+        type: BrowsingContextCreateType
+      }
+      </pre>
+   </dd>
+   <dt>Return Type</dt>
+   <dd>
+    <pre class="cddl local-cddl">
+        BrowsingContextCreateResult = {
+          context: BrowsingContext
+        }
+    </pre>
+   </dd>
+</dl>
+
+<div algorithm="remote end steps for browsingContext.create">
+The [=remote end steps=] with |command parameters| are:
+
+  1. Let |type| be the value of the <code>type</code> field of
+     |command parameters| if present, or "<code>window</code>" otherwise.
+
+  1. TODO: insert step 5 of https://w3c.github.io/webdriver/#new-window verbatim here?
+
+  1. If |context id| be the [=browsing context id=] of the newly created [=/browsing context=].
+
+  1. Let |body| be a [=map=] matching the <code>BrowsingContextCreateResult</code>
+     production, with the <code>context</code> field set to |context id|.
+
+  1. Return [=success=] with data |body|.
+
+</div>
+
 #### The browsingContext.getTree Command ####  {#command-browsingContext-getTree}
 
 The <dfn export for=commands>browsingContext.getTree</dfn> command returns a
@@ -2228,7 +2279,7 @@ top-level contexts when no parent is provided.
 <div algorithm="remote end steps for browsingContext.getTree">
 The [=remote end steps=] with <var ignore>session</var> and |command parameters| are:
 
-1. Let the |parent id| be the value of the <code>parent</code> field of
+1. Let |parent id| be the value of the <code>parent</code> field of
    |command parameters| if present, or null otherwise.
 
 1. Let |max depth| be the value of the <code>maxDepth</code> field of |command
--- a/index.bs
+++ b/index.bs
@@ -2200,7 +2200,7 @@ To <dfn>get the navigation info</dfn>, given |context| and |navigation status|:
 #### The browsingContext.create Command ####  {#command-browsingContext-create}
 
 The <dfn export for=commands>browsingContext.create</dfn> command creates a new
-[=/browsing context=], either in a new window or in a new tab, and returns its
+[=/browsing context=], either in a new tab or in a new window, and returns its
 [=browsing context id=].
 
 <dl>
--- a/index.bs
+++ b/index.bs
@@ -1931,7 +1931,7 @@ navigation status</dfn> struct, which has the following items:
 <pre class="cddl remote-cddl">
 
 BrowsingContextCommand = (
-    BrowsingContextCreate //
+    BrowsingContextCreateCommand //
     BrowsingContextGetTreeCommand //
     BrowsingContextNavigateCommand //
     BrowsingContextReloadCommand


index.bs
```diff
@@ -1882,6 +1884,55 @@ To <dfn>get the navigation info</dfn>, given |context| and |navigation status|:
...
+   </dd>
+</dl>
+
+<div algorithm="remote end steps for browsingContext.create">
+The [=remote end steps=] with |command parameters| are:
+
+  1. Let |type| be the value of the <code>type</code> field of
+     |command parameters| if present, or "<code>window</code>" otherwise.
+
+  1. TODO: insert step 5 of https://w3c.github.io/webdriver/#new-window verbatim here?
```
foolip, line 1925: This is really the main question of this PR, everything I've done is mostly the boilerplate.

https://w3c.github.io/webdriver/#new-window involves the "current browsing context" which we probably want to avoid in BiDi. But without that, presumably we should pass in another browsing context which the new one should be opened after in the tab case?

sideshowbarker:
I guess https://labs.w3.org/repo-manager/pr/id/w3c/webdriver-bidi/133 indicates that y’all should have @cwilso add you to the Browser Testing and Tools WG

index.bs
jgraham: Yeah I guess it makes sense to have an optional context, although absent that there's probably a way to talk about using the browsing context which has, or most recently had, OS-level focus as the relevant context in the tab case.

It occurs to me this should be accompanied by a PR to provide a switch-to-context method c.f. https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-bringToFront

foolip:
For my own review later, https://chromedevtools.github.io/devtools-protocol/tot/Target/#method-createTarget seems to be the CDP equivalent command. It's tabs by default, boolean `newWindow` param to open in a new window.

foolip:
Looks like CDP has a `url` parameter, which makes sense, but WebDriver classic doesn't have that and always opens about:blank. We should probably have a URL parameter here.

jgraham:
https://github.com/puppeteer/puppeteer/blob/e94a6b73d3014487b48a018fdcfb5f9874c86bdf/src/common/Browser.ts#L431-L443 seems to be the only use of `createTarget` in Puppeteer and https://github.com/microsoft/playwright/blob/768a97cfdc4b7425d164a13139ed2d4df358f740/src/server/chromium/crBrowser.ts#L347 in PlayWright. It at least appears that constraining the URL to `about:blank` and not having width/height parameters for windows is compatible with those uses (I think the problem with non-about:blank URLs is that once that's possible you also want to have the same ability to wait for the page to load as in other navigation commands).

foolip:
OK, so maybe we should start with about:blank for simplicity. Even so we'll need to decide whether `browsingContext.contextCreated`, `browsingContext.navigationStarted` and other navigation/loading-related events are fired. As long as the command response is guaranteed to come before any of the events then it's probably best to simply fire the events.

foolip:
> I guess https://labs.w3.org/repo-manager/pr/id/w3c/webdriver-bidi/133 indicates that y’all should have @cwilso add you to the Browser Testing and Tools WG

Yes, I've seen some email about that :)

index.bs
```diff
@@ -1882,6 +1884,55 @@ To <dfn>get the navigation info</dfn>, given |context| and |navigation status|:
...
+
+  1. If |context id| be the [=browsing context id=] of the newly created [=/browsing context=].
```
whimboo, line 1927:
```suggestion
  1. Let |context id| be the [=browsing context id=] of the newly created [=/browsing context=].
```

whimboo (changes requested):

sadym-chromium (approved):

sadym-chromium:
As asked offline, here is the 
<details>
  <summary>raw CDP log when new page with a given URL is open</summary>

```
puppeteer:protocol:SEND ► {"method":"Target.createTarget","params":{"url":"data:text/html,<h2>test</h2>","newWindow":true},"id":2} +8ms
puppeteer:protocol:RECV ◀ {"method":"Target.targetCreated","params":{"targetInfo":{"targetId":"05A991FBEA1BF19BCBF846A07FF3A1F1","type":"page","title":"","url":"","attached":false,"canAccessOpener":false,"browserContextId":"8177D676175D58796AAED92BA41EA52C"}}} +6ms
puppeteer:protocol:RECV ◀ {"id":2,"result":{"targetId":"05A991FBEA1BF19BCBF846A07FF3A1F1"}} +2ms
puppeteer:protocol:RECV ◀ {"method":"Target.targetInfoChanged","params":{"targetInfo":{"targetId":"739DC90364B192A143B762598BCEFF61","type":"page","title":"about:blank","url":"about:blank","attached":false,"canAccessOpener":false,"browserContextId":"8177D676175D58796AAED92BA41EA52C"}}} +51ms
puppeteer:protocol:RECV ◀ {"method":"Target.targetInfoChanged","params":{"targetInfo":{"targetId":"05A991FBEA1BF19BCBF846A07FF3A1F1","type":"page","title":"data:text/html,<h2>test</h2>","url":"data:text/html,<h2>test</h2>","attached":false,"canAccessOpener":false,"browserContextId":"8177D676175D58796AAED92BA41EA52C"}}} +1ms
puppeteer:protocol:SEND ► {"method":"Target.attachToTarget","params":{"targetId":"05A991FBEA1BF19BCBF846A07FF3A1F1","flatten":true},"id":3} +56ms
puppeteer:protocol:RECV ◀ {"method":"Target.targetInfoChanged","params":{"targetInfo":{"targetId":"05A991FBEA1BF19BCBF846A07FF3A1F1","type":"page","title":"data:text/html,<h2>test</h2>","url":"data:text/html,<h2>test</h2>","attached":true,"canAccessOpener":false,"browserContextId":"8177D676175D58796AAED92BA41EA52C"}}} +2ms
puppeteer:protocol:RECV ◀ {"method":"Target.attachedToTarget","params":{"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5","targetInfo":{"targetId":"05A991FBEA1BF19BCBF846A07FF3A1F1","type":"page","title":"data:text/html,<h2>test</h2>","url":"data:text/html,<h2>test</h2>","attached":true,"canAccessOpener":false,"browserContextId":"8177D676175D58796AAED92BA41EA52C"},"waitingForDebugger":false}} +0ms
puppeteer:protocol:RECV ◀ {"id":3,"result":{"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5"}} +1ms
puppeteer:protocol:SEND ► {"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5","method":"Page.enable","params":{},"id":4} +3ms
puppeteer:protocol:SEND ► {"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5","method":"Page.getFrameTree","params":{},"id":5} +0ms
puppeteer:protocol:SEND ► {"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5","method":"Target.setAutoAttach","params":{"autoAttach":true,"waitForDebuggerOnStart":false,"flatten":true},"id":6} +1ms
puppeteer:protocol:SEND ► {"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5","method":"Performance.enable","params":{},"id":7} +0ms
puppeteer:protocol:SEND ► {"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5","method":"Log.enable","params":{},"id":8} +0ms
puppeteer:protocol:RECV ◀ {"id":4,"result":{},"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5"} +36ms
puppeteer:protocol:RECV ◀ {"id":5,"result":{"frameTree":{"frame":{"id":"05A991FBEA1BF19BCBF846A07FF3A1F1","loaderId":"31A2E14BA8313C3EA897A2A94F9FE01D","url":"data:text/html,<h2>test</h2>","domainAndRegistry":"","securityOrigin":"://","mimeType":"text/html","adFrameType":"none","secureContextType":"InsecureScheme","crossOriginIsolatedContextType":"NotIsolated"}}},"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5"} +0ms
puppeteer:protocol:RECV ◀ {"id":6,"result":{},"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5"} +0ms
puppeteer:protocol:SEND ► {"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5","method":"Page.setLifecycleEventsEnabled","params":{"enabled":true},"id":9} +35ms
puppeteer:protocol:SEND ► {"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5","method":"Runtime.enable","params":{},"id":10} +0ms
puppeteer:protocol:SEND ► {"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5","method":"Network.enable","params":{},"id":11} +0ms
puppeteer:protocol:RECV ◀ {"id":7,"result":{},"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5"} +2ms
puppeteer:protocol:RECV ◀ {"id":8,"result":{},"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5"} +0ms
puppeteer:protocol:RECV ◀ {"method":"Page.lifecycleEvent","params":{"frameId":"05A991FBEA1BF19BCBF846A07FF3A1F1","loaderId":"31A2E14BA8313C3EA897A2A94F9FE01D","name":"commit","timestamp":1236561.748665},"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5"} +0ms
puppeteer:protocol:RECV ◀ {"method":"Page.lifecycleEvent","params":{"frameId":"05A991FBEA1BF19BCBF846A07FF3A1F1","loaderId":"31A2E14BA8313C3EA897A2A94F9FE01D","name":"DOMContentLoaded","timestamp":1236561.800613},"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5"} +0ms
puppeteer:protocol:RECV ◀ {"method":"Page.lifecycleEvent","params":{"frameId":"05A991FBEA1BF19BCBF846A07FF3A1F1","loaderId":"31A2E14BA8313C3EA897A2A94F9FE01D","name":"load","timestamp":1236561.800961},"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5"} +0ms
puppeteer:protocol:RECV ◀ {"method":"Page.lifecycleEvent","params":{"frameId":"05A991FBEA1BF19BCBF846A07FF3A1F1","loaderId":"31A2E14BA8313C3EA897A2A94F9FE01D","name":"networkAlmostIdle","timestamp":1236561.832845},"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5"} +0ms
puppeteer:protocol:RECV ◀ {"method":"Page.lifecycleEvent","params":{"frameId":"05A991FBEA1BF19BCBF846A07FF3A1F1","loaderId":"31A2E14BA8313C3EA897A2A94F9FE01D","name":"networkIdle","timestamp":1236561.832845},"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5"} +0ms
puppeteer:protocol:RECV ◀ {"id":9,"result":{},"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5"} +0ms
puppeteer:protocol:RECV ◀ {"method":"Runtime.executionContextCreated","params":{"context":{"id":1,"origin":"://","name":"","auxData":{"isDefault":true,"type":"default","frameId":"05A991FBEA1BF19BCBF846A07FF3A1F1"}}},"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5"} +4ms
puppeteer:protocol:RECV ◀ {"id":10,"result":{},"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5"} +0ms
puppeteer:protocol:RECV ◀ {"id":11,"result":{},"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5"} +0ms
puppeteer:protocol:SEND ► {"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5","method":"Page.addScriptToEvaluateOnNewDocument","params":{"source":"//# sourceURL=__puppeteer_evaluation_script__","worldName":"__puppeteer_utility_world__"},"id":12} +6ms
puppeteer:protocol:RECV ◀ {"id":12,"result":{"identifier":"1"},"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5"} +1ms
puppeteer:protocol:SEND ► {"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5","method":"Page.createIsolatedWorld","params":{"frameId":"05A991FBEA1BF19BCBF846A07FF3A1F1","grantUniveralAccess":true,"worldName":"__puppeteer_utility_world__"},"id":13} +1ms
puppeteer:protocol:RECV ◀ {"method":"Runtime.executionContextCreated","params":{"context":{"id":2,"origin":"","name":"__puppeteer_utility_world__","auxData":{"isDefault":false,"type":"isolated","frameId":"05A991FBEA1BF19BCBF846A07FF3A1F1"}}},"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5"} +3ms
puppeteer:protocol:RECV ◀ {"id":13,"result":{"executionContextId":2},"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5"} +1ms
puppeteer:protocol:SEND ► {"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5","method":"Emulation.setDeviceMetricsOverride","params":{"mobile":false,"width":800,"height":600,"deviceScaleFactor":1,"screenOrientation":{"angle":0,"type":"portraitPrimary"}},"id":14} +4ms
puppeteer:protocol:SEND ► {"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5","method":"Emulation.setTouchEmulationEnabled","params":{"enabled":false},"id":15} +0ms
puppeteer:protocol:RECV ◀ {"id":14,"result":{},"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5"} +2ms
puppeteer:protocol:RECV ◀ {"id":15,"result":{},"sessionId":"3E4D0B0F6910E3754F2F49B052BE45B5"} +0ms
```

</details>

So the `Target.createTarget` causes 2 rounds of the events. 1 for an empty page, 2 for the desired page afterwards.

index.bs
bwalderman: Calling CDP's Target.createTarget is a lot like a user opening a new unrelated tab, in which case there's no need for an existing/current browsing context.

It almost feels like we should have two commands; One that invokes the `window.open` steps as WebDriver/HTTP's New Window command does, and one that just creates a new tab/window without an opener relationship as Target.createTarget does.

foolip: Aha, so rather than a `type` argument, two commands for these two operations? https://chromedevtools.github.io/devtools-protocol/tot/Target/#method-createTarget does have a `newWindow` boolean though, so the precedent seems to be a single command, both in CDP and WebDriver. Am I misunderstanding the suggestion?

foolip: @jgraham regarding `Page.bringToFront` in CDP, do you think the equivalent in BiDi should be `browsingContext.bringToFront` then?

Perhaps we need to make the concept of a group of browsing contexts (tabs) explicit? Playwright has a notion of [`BrowserContext`](https://playwright.dev/docs/api/class-browsercontext) which seems to group together their `Page` objects. The main reason for it seems to be "independent browser sessions" rather than representing a visible window, but there might be something to this?

index.bs
```diff
@@ -1882,6 +1884,55 @@ To <dfn>get the navigation info</dfn>, given |context| and |navigation status|:
...
+        }
+    </pre>
+   </dd>
+</dl>
+
+<div algorithm="remote end steps for browsingContext.create">
+The [=remote end steps=] with |command parameters| are:
+
+  1. Let |type| be the value of the <code>type</code> field of
+     |command parameters| if present, or "<code>window</code>" otherwise.
```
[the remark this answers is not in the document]
foolip: If we keep `type` as an argument rather than having two commands, I think we should use "tab" as the default then.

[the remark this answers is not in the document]
foolip:
```suggestion
     |command parameters| if present, or "<code>tab</code>" otherwise.
```

index.bs
jgraham: Note that WebDriver is creating the window with `noopener`, so it's really a new tab that's not connected to any other tabs. So I think one command that opens a new top level browsing context in either a tab or a window should be fine.

I think BrowserContext in Playwrihgt is trying to solve a different kind of problem; I think it's an abstraction over the idea of multiple containers (in gecko) or private windows (in Blink), which don't share state. It seems like a reasonable idea to surface that concept, but I don't think it's apping 1:1 with OS windows, and is probably a more advanced feature. It's also not a 1:1 mapping with OS windows, which I think is what you're looking for here.

For my money, a `browsingContext.bringToFront` that would take the context parameter, get the associated top-level-browsing context, and do [something] makes sense. The question is whether [something] should require that it ends up with OS-level focus or just that it's selected context in whatever window it's in without requiring anything at the OS level.

foolip: OK, let's put Playwright's `BrowserContext` to the side for now.

For the purposes of this PR, my main question is if we want a `browsingContext.create` command to work without any reference to which window it should open the new browsing context in. The slightly unsatisfying bit is going to be if we add a notion of windows later, there would have to be an implicit window here and we've tried to avoid things like WebDriver's "current top-level browsing context" in BiDi so far.

jgraham: I think my proposal from https://github.com/w3c/webdriver-bidi/pull/133#discussion_r704346287 still makes sense: if you provide a context, the new context should be created as if the user was creating a new context starting at that one (that's hard to specify in detail because to some extent it's an implementation detail, but roughly it would mean a new tab would share an os window with the tab containing that context; I think the text will be necessarily handwavy here).

index.bs
```diff
@@ -1882,6 +1884,55 @@ To <dfn>get the navigation info</dfn>, given |context| and |navigation status|:
...
+      <pre class="cddl remote-cddl">
+      BrowsingContextCreateCommand = {
+        method: "browsingContext.create",
+        params: BrowsingContextCreateParameters
+      }
+
+      BrowsingContextCreateType = "tab" / "window"
+
+      BrowsingContextCreateParameters = {
+        type: BrowsingContextCreateType
```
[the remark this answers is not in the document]
foolip: In the "WebDriver Bidi - Editorial Meeting to review PRs" meeting yesterday we decided to leave it as required. Clients can pick a default, and not all implementations will support both, so a default at a protocol level would still require handling the case where the requested type isn't supported.

index.bs
foolip: I've left this as an inline issue in the spec now, reading:

>  HTML’s window open steps depend on the "entry global object" and WebDriver’s New Window command depends on the "current browsing context", but we have neither here. What OS window should new tabs be opened in?

index.bs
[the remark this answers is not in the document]
foolip: I undid this in https://github.com/w3c/webdriver-bidi/pull/133/commits/ec54f423f0444be19b3d0d88533f44b679156497, instead making *type* required.

foolip:
@whimboo do you want to take another look?

jgraham (approved):

index.bs
```diff
@@ -2194,6 +2197,75 @@ To <dfn>get the navigation info</dfn>, given |context| and |navigation status|:
...
+     done without invoking the [=/focusing steps=] for the created browsing
+     context. If |type| is "<code>tab</code>", and the implementation supports
+     multiple browsing contexts in the same OS window, the new browsing context
+     should share an OS window with any other [=top-level browsing context=]. If
+     |type| is "<code>window</code>", and the implementation supports multiple
+     browsing contexts in separate OS windows, the created browsing context
+     should be in a new OS window. In all other cases the details of how the
+     browsing context is presented to the user are implementation defined.
+
+     Issue: HTML's [=window open steps=] depend on the "entry global object"
```
jgraham, line 2253: We could have an optional context parameter that would be used as a reference? Like

```
1. Let |reference context id| be the value of the <code>context</code> field of parameters, if present, or null otherwise.
1. If |reference context id| is not null:
   1. Let |reference context| be the result of [=trying=] to [=get a browsing context=] with |reference context id|.
 1. Otherwise let |reference context| be null.
```

and then the "tab" case we say the new browsing context should share a window with the top-level browsing context of reference context, if it's not null, or with an implementation-defined existing context, if it is.

And then

foolip: Yes, something like that sounds good! I'll do that in a follow-up PR.

index.bs
```diff
@@ -120,6 +120,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
     text: set up a window environment settings object; url: set-up-a-window-environment-settings-object
     text: set up a worker environment settings object; url: set-up-a-worker-environment-settings-object
     text: set up a worklet environment settings object; url: set-up-a-worklet-environment-settings-object
+    text: window open steps; url: multipage/window-object.html#window-open-steps
```
whimboo, line 123: None of the other definitions use `multipage`. Why is that needed here?

index.bs
```diff
@@ -1701,6 +1701,7 @@ navigation status</dfn> struct, which has the following items:
 <pre class="cddl remote-cddl">
 
 BrowsingContextCommand = (
+    BrowsingContextCreate //
```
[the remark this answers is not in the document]
whimboo: I don't see that this requested change has been applied.

index.bs
```diff
@@ -2194,6 +2197,75 @@ To <dfn>get the navigation info</dfn>, given |context| and |navigation status|:
 
 ### Commands ### {#module-browsingContext-commands}
 
+#### The browsingContext.create Command ####  {#command-browsingContext-create}
+
+The <dfn export for=commands>browsingContext.create</dfn> command creates a new
+[=/browsing context=], either in a new window or in a new tab, and returns its
```
whimboo, line 2203: Given that the tab is the default now, maybe we should switch the order?

whimboo (changes requested):
When I read it right there is still a missing item to get fixed which was inappropriately marked as resoved.

bwalderman (approved):

index.bs
foolip: All the terms that we don't need to add custom link text for use the multipage URLs, and in context of most links this looked inconsistent. But I'll send a PR to fix this whole block instead.

foolip: https://github.com/w3c/webdriver-bidi/pull/152

index.bs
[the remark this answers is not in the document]
foolip: Hmm, I guess I missed it among other comments. Applying now!

index.bs
foolip: There is no default, but I'll change the order anyway to match the alphabetical order below.

index.bs
whimboo: This is covered by https://github.com/w3c/webdriver-bidi/issues/151 now. Thanks @bwalderman for filing that.

whimboo (approved):
Thanks @foolip! This looks fine now. I assume you will merge yourself.

index.bs
foolip: That's now merged and I've merged to resolve the resulting conflicts.

index.bs
foolip: https://github.com/w3c/webdriver-bidi/pull/153

PR #136, opened 2021-09-14
Add script.evaluate command

jgraham:
This enables evaluation of scripts in an existing js realm, using a similar command/response format to `runtime.evaluate` in CDP. See https://github.com/w3c/webdriver-bidi/commit/a1a8742e0031f127c51de43191b52e0abba46c5f for more details.


juliandescottes (commented):
Read through the patch, I just have some minor nits and questions. Thanks for working on this!

index.bs
```diff
@@ -2602,8 +2655,247 @@ Note: Future variations of this specification will retain the invariant that
...
+
+1. Let |value| be |evaluation status|.\[[Value]]
+
+1. If |evaluation status|.\[[Type]] is <code>throw</code>, let |exception
+   details| be the result of [=get exception details=] given |evaluation
+   status|. Otherwise let |exception details| be null.
+
+1. Let |result| be the result of [=serialize as a remote value=] given |value|.
+
+1. Return a new map matching the <code>ScriptEvaluateResult</code> production,
```
bwalderman, line 2894: Is it possible for a script evaluation to both throw and return a result? Otherwise, I think ScriptEvaluateResult would be better expressed as a sum type which can be either { result } or { exceptionDetails }.

jgraham: I thought so too, but that's at least not quite how CDP works. It has both a `result` field with the object that was thrown and also a `exceptionDetails` field with details of the error. For example `throw 2` ends up as:
```
{
    "result": {
        "type": "number",
        "value": 2,
        "description": "2"
    },
    "exceptionDetails": {
        "exceptionId": 11,
        "text": "Uncaught",
        "lineNumber": 0,
        "columnNumber": 0,
        "scriptId": "78",
        "stackTrace": {
            "callFrames": [
                {
                    "functionName": "",
                    "scriptId": "78",
                    "url": "",
                    "lineNumber": 0,
                    "columnNumber": 0
                }
            ]
        },
        "exception": {
            "type": "number",
            "value": 2,
            "description": "2"
        }
    }
}
```

With a non-primitive value things are a little more interesting, `throw {}` gives:
```
{
    "result": {
        "type": "object",
        "className": "Object",
        "description": "Object",
        "objectId": "5263362622729908126.1.21",
        "preview": {
            "type": "object",
            "description": "Object",
            "overflow": false,
            "properties": []
        }
    },
    "exceptionDetails": {
        "exceptionId": 18,
        "text": "Uncaught",
        "lineNumber": 0,
        "columnNumber": 0,
        "scriptId": "91",
        "stackTrace": {
            "callFrames": [
                {
                    "functionName": "",
                    "scriptId": "91",
                    "url": "",
                    "lineNumber": 0,
                    "columnNumber": 0
                }
            ]
        },
        "exception": {
            "type": "object",
            "className": "Object",
            "description": "Object",
            "objectId": "5263362622729908126.1.22",
            "preview": {
                "type": "object",
                "description": "Object",
                "overflow": false,
                "properties": []
            }
        }
    }
}
```
`result` and `exceptionDetails.exception` refer to the same object but the `objectID` is different. I don't know if CDP is packing some information into the id so it's possible to tell it's the same object, rather than using it as an opaque identifier.

Anyway, I haven't found a case where `result` and `exceptionDetails.exception` can be different, so I think we should do one or the other rather than both. We could do that as "either `result` or `exceptionDetails`" or we could always value `result` and just use `exceptionDetails` for additional information.

index.bs
```diff
@@ -2602,8 +2655,247 @@ Note: Future variations of this specification will retain the invariant that
...
+      <pre class="cddl remote-cddl">
+      ScriptEvaluateCommand = {
+        method: "script.evaluate",
+        params: ScriptEvaluateParameters
+      }
+
+      ScriptEvaluateParameters = {
+        expression: text,
+        ?awaitPromise: bool,
+        target: RealmOrBrowsingContext,
```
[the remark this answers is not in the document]
jgraham: Yes, this was left over from an earlier name. Good catch!

foolip:
This bit of code from Puppeteer that @sadym-chromium found is relevant for how Puppeteer uses `exceptionDetails` and `result`:
https://github.com/puppeteer/puppeteer/blob/7341d9fadd1466a5b2f2bde8631f3b02cf9a7d8a/src/common/ExecutionContext.ts#L220-L227

index.bs
sadym-chromium: FWIW
There are [3 types of IDs in CDP](https://github.com/WICG/devtools-protocol/issues/9#issuecomment-367831808):
1. `RemoteObjectId`
1. `NodeId`
1. `BackendNodeId`

According to the [`bindObject` logic](https://source.chromium.org/chromium/chromium/src/+/main:v8/src/inspector/injected-script.cc;drc=9481da96106e5863fdd22bf54a0a6c264e6c56f4;l=1003), all the objects in the same context will have the same first 2 numbers in the ID, and the incremental last number even for the same object.

index.bs
```diff
@@ -2781,6 +2792,54 @@ ScriptEvent = (
...
+[=Remote end definition=] and [=local end definition=]
+
+<pre class="cddl local-cddl remote-cddl">
+ExceptionDetails = {
+  columnNumber: uint,
+  exception: RemoteValue,
+  lineNumber: uint.
+  stackTrace: StackTrace
+  text: text,
+};
```
sadym-chromium, line 2806: I started implementing the `script.evaluate` proto, and figured out the fields `exception` and `stackTrace` are optional in the CDP. Was making them obligated intentional?
```suggestion
ExceptionDetails = {
  columnNumber: uint,
  exception?: RemoteValue,
  lineNumber: uint,
  stackTrace?: StackTrace,
  text: text,
};
```

jgraham: Do yo uhave an example of a script that would return an error with no `result` or `exceptionDetails.exception` value, or no `exceptionDetails.stackTrace`? Currently the prose part always populates those fields, so making them optional in the CDDL doesn't change anything.

sadym-chromium: About **`exception `**: you are right. I can think only of the scenario when the script times out. But in this case the command failed on the protocol level:
```
{
  "id":9,
  "error":{
    "code":-32000,
    "message":"Execution was terminated"},
  "sessionId":"8DC93214FDAEE6DE8EC5D57AA35A0BE7"}
```

The **`stacktrace`** is added by the [`Debuger domain`](https://chromedevtools.github.io/devtools-protocol/tot/Debugger/), which is off by default, as it's quite expensive.

sadym-chromium:
```suggestion
ExceptionDetails = {
  columnNumber: uint,
  exception: RemoteValue,
  lineNumber: uint,
  stackTrace?: StackTrace,
  text: text,
};
```

sadym-chromium:
It's unclear, if the `expression` in the `ScriptEvaluateParameters` can contain serialised objects or not.
I'd restrict `expression` as a valid JS, which return a function, and add serialised arguments as an additional field in `ScriptEvaluateParameters`.

index.bs
jgraham: Hmm, so if one throws an actual exception object, that gets a `stack` property. So I don't know quite how lazilly that can be constructed, but it seems like in the case an actual exception is thrown (which I guess is the majority case vs just throwing a random object), the engine has to at least store enough information to allow constructing a stack anyway. So my inclination is to say we always have to provide a stack trace for an exception as part of the protocol.

sadym-chromium: I'm not sure if I got the idea. Do you mean in case of throwing a random object BiDi should return an empty stack? Or treat the case somehow differently?

jgraham: No, I just mean that `throw new Error()` is going to return an object with a stack anyway. Given that, it seems totally reasonable to unconditionally put the stack into the response; it doesn't seem likely to actually have a significant extra perf cost in the engine given you're already generating a stack for most exceptions. For comparison, aiui Firefox's RDP always includes a stack when scipt evaluation results in an exception.

sadym-chromium: I'm still not sure what should be the result of evaluating the following script. Could you please clarify?
```
const throwException = ()=>{
      throw {a:'a'};
  }
throwException();
```

index.bs
```diff
@@ -2893,8 +2950,255 @@ Note: Future variations of this specification will retain the invariant that
...
+        target: Target,
+      }
+      </pre>
+   </dd>
+   <dt>Return Type</dt>
+   <dd>
+    <pre class="cddl local-cddl">
+      ScriptEvaluateResult = (
+        ScriptEvaluateSuccessResult //
+        ScriptEvaluateExceptionResult
```
sadym-chromium, line 3135:
```suggestion
        ScriptEvaluateResultSuccess //
        ScriptEvaluateResultException
```

index.bs
juliandescottes: This code snippet generates a stacktrace in both Chrome or Firefox DevTools' console. Why should we have no stacktrace with BiDi? Is it also for performance reasons?

Or is it unclear because we should somehow make a difference between a script which fails to evaluate for something like a syntax error vs something that throws a custom error?

jgraham: Yes, I'd expect a stacktrace in the response. As @juliandescottes says it matches what you'd expect to happen in devtools. For Chrome in particular, evaluating that in the console gives a CDP response:
```
{
    "result": {
        "type": "object",
        "className": "Object",
        "description": "Object",
        "objectId": "5064291994948395578.1.3",
        "preview": {
            "type": "object",
            "description": "Object",
            "overflow": false,
            "properties": [
                {
                    "name": "a",
                    "type": "string",
                    "value": "a"
                }
            ]
        }
    },
    "exceptionDetails": {
        "exceptionId": 2,
        "text": "Uncaught",
        "lineNumber": 1,
        "columnNumber": 6,
        "scriptId": "34",
        "stackTrace": {
            "callFrames": [
                {
                    "functionName": "throwException",
                    "scriptId": "34",
                    "url": "",
                    "lineNumber": 1,
                    "columnNumber": 6
                },
                {
                    "functionName": "",
                    "scriptId": "34",
                    "url": "",
                    "lineNumber": 3,
                    "columnNumber": 0
                }
            ]
        },
        "exception": {
            "type": "object",
            "className": "Object",
            "description": "Object",
            "objectId": "5064291994948395578.1.4",
            "preview": {
                "type": "object",
                "description": "Object",
                "overflow": false,
                "properties": [
                    {
                        "name": "a",
                        "type": "string",
                        "value": "a"
                    }
                ]
            }
        }
    }
}
```
which clearly includes the stack. For BiDi I'd expect something like:
```
{
    "exceptionDetails": {
        "text": "Uncaught",
        "lineNumber": 1,
        "columnNumber": 6,
        "stackTrace": {
            "callFrames": [
                {
                    "functionName": "throwException",
                    "scriptId": "34",
                    "url": "",
                    "lineNumber": 1,
                    "columnNumber": 6
                },
                {
                    "functionName": "",
                    "scriptId": "34",
                    "url": "",
                    "lineNumber": 3,
                    "columnNumber": 0
                }
            ]
        },
        "exception": {
            "type": "object",
            "objectId": "5064291994948395578",
            "value": {
               "a": {
                   "type": "string",
                   "value": "a"
               }
           }
    }
}
```

sadym-chromium: Indeed, if you make a described experiment in Chrome DevTools, you will get a stacktrace. But, getting into implementation details, it is because of the fact `Debugger` domain is enabled when DevTools is open. By default it is disabled, and no stacktrace will be provided.

Anyway, having or not a stacktrace in case of exception seems more a "debuggability" feature, so why not to keep it up to implementation?

Experiment to see difference between having `Debugger` domain enabled and disabled in Chrome:
0. Get a Chrome tab without an open DevTools.
1. Navigate to `data:text/html,<script>const a=()=>{throw 1;}; const b=()=>{a();};const c=()=>{b();};c();</script>`.
2. Open DevTools and go to console.
3. Verify there is no detailed stacktrace. Only `Uncaught 1`.
4. Reload the page with DevTools open.
5. Verify there exception is extendable with detailed stacktrace:
```
Uncaught 1
a @ ...
b @ ...
c @ ...
(anonymous) @ ...
```

jgraham: Leaving things up to the imlementation is usually not ideal because it means that consumers can't rely on a specific behaviour.

In this case, what's the argument from the user perspective for not providing a stacktrace for some exceptions? If my evaluated script fails and I don't expect it to, it seems like I'm always going to want a stacktrace to help debug the failure. Given that exception objects have a stack anyway, it doesn't seem like it can be too expensive to produce the stack in the js engine, so performance doesn't seem like a string argument from a user point of view.

sadym-chromium: UPD: my bad. It's not a `Debugger` but a `Runtime` domain enriches CDP response with a stacktrace. Anyway, it still seems to be reasonable to keep it up for implementation.

sadym-chromium: Ok, I agree. We can keep stackTrace obligatory.

```suggestion
ExceptionDetails = {
  columnNumber: uint,
  exception: RemoteValue,
  lineNumber: uint,
  stackTrace: StackTrace,
  text: text,
};
```

index.bs
```diff
@@ -2781,6 +2792,54 @@ ScriptEvent = (
 
 ### Types ### {#module-script-types}
 
+#### The script.ExceptionDetails type #### {#type-script-Exception}
```
whimboo, line 2795:
```suggestion
#### The script.ExceptionDetails type #### {#type-script-ExceptionDetails}
```

index.bs
```diff
@@ -2781,6 +2792,54 @@ ScriptEvent = (
...
+ExceptionDetails = {
+  columnNumber: uint,
+  exception: RemoteValue,
+  lineNumber: uint.
+  stackTrace: StackTrace
+  text: text,
+};
+</pre>
+
+The <code>ExceptionDetails</code> type represents a javascript exception.
```
whimboo, line 2809:
```suggestion
The <code>ExceptionDetails</code> type represents a JavaScript exception.
```

Maybe this also applies to other parts of the document. I haven't checked that.

index.bs
```diff
@@ -2893,8 +2950,255 @@ Note: Future variations of this specification will retain the invariant that
 
 </div>
 
+#### The script.StackFrame type #### {#types-script-stackframe}
```
whimboo, line 2953:
```suggestion
#### The script.StackFrame type #### {#types-script-StackFrame}
```

index.bs
```diff
@@ -2893,8 +2950,255 @@ Note: Future variations of this specification will retain the invariant that
...
+
+<pre class="cddl local-cddl">
+StackFrame = {
+  url: text,
+  functionName: text,
+  lineNumber: int,
+  columnNumber: int,
```
whimboo, line 2960: Similar to other type definitions we might want to order it alphabetically.

index.bs
```diff
@@ -2893,8 +2950,255 @@ Note: Future variations of this specification will retain the invariant that
...
+}
+</pre>
+
+A frame in a stack trace is represented by a <code>StackFrame</code>
+object. This has a <code>url</code> property, which represents the URL of the
+script, a <code>functionName</code> property which represents the name of the
+executing function, and <code>lineNumber</code> and <code>columnNumber</code>
+properties, which represent the line and column number of the executed code.
+
+#### The script.StackTrace type #### {#types-script-stacktrace}
```
whimboo, line 2970:
```suggestion
#### The script.StackTrace type #### {#types-script-StackTrace}
```

index.bs
```diff
@@ -2893,8 +2950,255 @@ Note: Future variations of this specification will retain the invariant that
...
+      ScriptEvaluateCommand = {
+        method: "script.evaluate",
+        params: ScriptEvaluateParameters
+      }
+
+      ScriptEvaluateParameters = {
+        expression: text,
+        ?awaitPromise: bool,
+        target: Target,
+      }
```
whimboo, line 3127: Does it matter in which order types are defined? In other places sub types are  defined first. Also sort alphabetically?

Issue #140, opened 2021-09-30
Add `script.callFunction` command


sadym-chromium:
As a BiDi user, I want to be able to invoke some custom function with some custom arguments on the page context.
For example, I want to be able to invoke `(text) => {this.innerHtml=text}` with custom `this` and `text`.

It can be done:
1. As a part of the `script.evaluate` command by adding optional parameters `thisObject` and `argument`. It can be quite ugly, as `expression` changes meaning from **expression to be evaluated and returned as a result** to **expression to return a function which will be invoked with the given params and each result will be returned to user**.
2. As another command `script.invoke` with an `expression` string returning function to be invoked, `thisObject` and `argument`.

I'm inclined to go with the second way.

There is another open question:
1. How should the arguments be presented? As a user, I want to be ale to use both some constants, simple objects like `{a: 1}`, and `RemoteObject` return by e.g. `script.evaluate`.

For primitive type it is logical to use the [`remote value`](https://w3c.github.io/webdriver-bidi/#type-common-RemoteValue) serialisation. But for non-primitive it's enough to have only `objectId`, even without type. Should we send `{type: 'object', objectId: 'SOME_OBJECT_ID'}`, or we don'r require `type`?

For non-primitive objects I want to use as an argument, if I want to pass some new object as an argument, which was not presented in BiDi yet (like `{a: 1}`), I want to use be able to use BiDi serialisation protocol. But it requires making an `objectId` optional.

# UPD:
As discussed, command `script.callFunction` would be more readable.

jgraham:
Yes, I was assuming we'd have something like `script.call[FunctionOn]`.

For arguments, we could indeed allow a serialization like `{objectId: <object id string>}` to refer to an object on the remote end. The question is whether to allow `type` and `value` fields. The advantage of that would be allowing the exact same object from a response to be used in an argument. The disadvantage is that it's unclear what to do if there's a mismatch (e.g. if you pass in `type: "object"` but once you lookup the object id get a `map`). 

I think I favour minimisng the amount of validation having a schema like:
```
Argument = (
   RemoteValueArgument //
   ValueArgument
)

RemoteValueArgument = {
    objectId: text
}

ValueArgument = {
    type: text,
    value: Argument
}
```

sadym-chromium:
Merged in #142

PR #142, opened 2021-10-12
`script.callFunction` spec

sadym-chromium:
Specification of the `script.callFunction` command #140.
* #136 used as a blue print for `callFunction` steps.
* `RemoteValue` is split into `PrimitiveProtocolValue` and `RemoteValue`.
* Described `PrimitiveProtocolValue` deserialization.
* Introduced `LocalValue` type.
* Described `LocalValue` deserialization.
* Calling function is done in 3 steps: 
  1. Deserialize arguments.
  1. Evaluate function declaration.
  1. Call function with arguments.

***
<a href="https://pr-preview.s3.amazonaws.com/sadym-chromium/webdriver-bidi/pull/142.html" title="Last updated on Dec 2, 2021, 4:22 PM UTC (7c73549)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/142/f5ce465...sadym-chromium:7c73549.html" title="Last updated on Dec 2, 2021, 4:22 PM UTC (7c73549)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -91,8 +91,8 @@ spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
     text: IsRegExp; url: sec-isregexp
     text: LengthOfArrayLike; url: sec-lengthofarraylike
     text: Object; url: sec-object-objects
-    text: InstantiateOrdinaryFunctionObject; url: #sec-runtime-semantics-instantiateordinaryfunctionobject
-    text: EvaluateFunctionBody; url: #sec-runtime-semantics-evaluatefunctionbody
+    text: IsCallable; url: #sec-iscallable
+    text: OrdinaryCallEvaluateBody; url: #sec-ordinarycallevaluatebody
     text: ScriptEvaluation; url: #sec-runtime-semantics-scriptevaluation
     text: Set; url: sec-set
     text: ThisTimeValue; url: thistimevalue
@@ -941,7 +941,7 @@ Issue: Defined <dfn>deserialize a reference</dfn> steps.
 
 ### Primitive Protocol Value ### {#data-types-protocolValue-primitiveValue}
 
-<dfn>Primitive Protocol Value</dfn> represents primitive protocol values which can be serialized to /
+Represents primitive protocol values which can be serialized to /
 deserialized from ECMAScript without reference to existing objects.
 
 ```
@@ -3428,6 +3428,28 @@ The <dfn>deserialize an argument value</dfn>(|argumentValue|) is the following:
 1. If |argumentValue| is [=RemoteReference=], return [=deserialize a reference=](|argumentValue|).
 1. Return [=deserialize a local value=](|argumentValue|).
 
+The steps to <dfn>get function object</dfn> with a given |function declaration|,
+|environment settings|, |base URL|, and |options| are the following:
+
+1. Let |function script| be the result of [=create a classic script=] with |function declaration|,
+   |environment settings|, |base URL|, and |options|.
+
+1. [=Prepare to run script=] with |environment settings|.
+
+1. Set |function evaluation status| to [=ScriptEvaluation=](|function script|'s record).
+
+1. [=Clean up after running script=] with |environment settings|.
+
+1. Assert: |function evaluation status|.\[[Type]] is not <code>throw</code>.
+
+1. Let |function object| be |evaluation status|.\[[Value]].
+
+1. Assert [=IsCallable=](|function object|) is true.
+
+1. Return |function object|.
+
+Issue: Define behavior in case of exception during script evaluation.
+
 The [=remote end steps=] with |command parameters| are:
 
 1. Let |realm| be the result of [=trying=] to [=get a realm from a target=]
@@ -3435,11 +3457,6 @@ The [=remote end steps=] with |command parameters| are:
 
 1. Let |environment settings| be the [=environment settings object=] whose
    [=realm execution context=]'s Realm component is |realm|.
-
-1. [=Prepare to run script=] with |environment settings|.
-
-1. Let |function declaration| be the value of the
-   <code>functionDeclaration</code> field of |command parameters|.
 
 1. Let |arguments declaration| be the value of the <code>arguments</code> field
    of |command parameters|.
@@ -3453,11 +3470,16 @@ The [=remote end steps=] with |command parameters| are:
 
   1. Append |deserialized argument| to the |deserialized arguments list|.
 
-1. Let |function object| be the result of
-   [=InstantiateOrdinaryFunctionObject=](|function declaration|).
+1. Let |function declaration| be the value of the
+   <code>functionDeclaration</code> field of |command parameters|.
+
+1. Let |function object| be the result of [=get function object=](|function declaration|,
+   |environment settings|, |base URL|, and |options|).
+
+1. [=Prepare to run script=] with |environment settings|.
 
 1. Set |function evaluation status| to
-   [=EvaluateFunctionBody=](|function object|, |deserialized arguments list|).
+   [=OrdinaryCallEvaluateBody=](|function object|, |deserialized arguments list|).
 
 1. If |function evaluation status|.\[[Type]] is <code>normal</code>,
    |await promise| is true, and
--- a/index.bs
+++ b/index.bs
@@ -79,7 +79,7 @@ spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
     text: Await; url: await
     text: BigInt; url: sec-bigint-constructor
     text: boolean; url: sec-terms-and-definitions-boolean-value
-    text: Call; url: #sec-call
+    text: Call; url: sec-call
     text: Completion Record; url: sec-completion-record-specification-type
     text: CreateArrayFromList; url: sec-createarrayfromlist
     text: CreateArrayIterator; url: sec-createarrayiterator
--- a/index.bs
+++ b/index.bs
@@ -1164,17 +1164,17 @@ SetLocalValue = {
 
 <div algorithm>
 
-To <dfn>map serialized key-value list to deserialized key-value list</dfn> given a
+To <dfn>deserialize key-value list</dfn> given a
 |serialized key-value list|:
 
 1. Let |deserialized key-value list| be a new list.
 
 1. For each |serialized key-value| in the |serialized key-value list|:
 
-   1. If |serialized key-value| has not 2 elements,
+   1. If [=list/size=] of |serialized key-value| is not 2,
       return [=error=] with [=error code=] [=invalid argument=].
 
-   1. Let |serialized key| be the first element of |serialized key-value|.
+   1. Let |serialized key| be |serialized key-value|[0].
 
    1. If |serialized key| is a <code>string</code>,
       let |deserialized key| be |serialized key|.
@@ -1182,7 +1182,7 @@ To <dfn>map serialized key-value list to deserialized key-value list</dfn> given
    1. Otherwise let |deserialized key| be result of [=trying=] to [=deserialize local value=]
       with given |serialized key|.
 
-   1. Let |serialized value| be the second element of |serialized key-value|.
+   1. Let |serialized value| be |serialized key-value|[1].
 
    1. Let |deserialized value| be result of [=trying=] to [=deserialize local value=] with given
       |serialized value|.
@@ -1239,15 +1239,15 @@ To <dfn>deserialize local value</dfn> given a |local protocol value|:
 
     <dt>|type| is the string "<code>map</code>"
     <dd>
-      1. Let |deserialized key-value list| be a result of [=trying=]
-         [=map serialized key-value list to deserialized key-value list=] with given |value|.
+      1. Let |deserialized key-value list| be a result of [=trying=] to
+         [=deserialize key-value list=] with given |value|.
 
       1. Return [=success=] with data <code>new [=Map=](|deserialized key-value list|)</code>.
 
     <dt>|type| is the string "<code>object</code>"
     <dd>
-      1. Let |deserialized key-value list| be a result of [=trying=]
-         [=map serialized key-value list to deserialized key-value list=] with given |value|.
+      1. Let |deserialized key-value list| be a result of [=trying=] to
+         [=deserialize key-value list=] with given |value|.
 
       1. Return [=success=] with data
          <code>[=Object.fromEntries=](|deserialized key-value list|)</code>.
--- a/index.bs
+++ b/index.bs
@@ -3494,105 +3494,6 @@ Issue: This has the wrong error code
 
 ### Commands ### {#module-script-commands}
 
-#### The script.evaluate Command ####  {#command-script-evaluate}
-
-The <dfn export for=commands>script.evaluate</dfn> command evaluates a provided
-script in a given realm. For convenience a browsing context can be provided in
-place of a realm, in which case the realm used is the realm of the browsing
-context's active document.
-
-The method returns the value of executing the provided script, unless it returns
-a promise and <code>awaitPromise</code> is true (which is the default), in which
-case the resolved value of the promise is returned.
-
-<dl>
-   <dt>Command Type</dt>
-   <dd>
-      <pre class="cddl remote-cddl">
-      ScriptEvaluateCommand = {
-        method: "script.evaluate",
-        params: ScriptEvaluateParameters
-      }
-
-      ScriptEvaluateParameters = {
-        expression: text,
-        target: Target,
-        ?awaitPromise: bool,
-      }
-      </pre>
-   </dd>
-   <dt>Return Type</dt>
-   <dd>
-    <pre class="cddl local-cddl">
-      ScriptEvaluateResult = (
-        ScriptEvaluateResultSuccess //
-        ScriptEvaluateResultException
-      )
-
-      ScriptEvaluateResultSuccess = {
-         result: RemoteValue
-      }
-
-      ScriptEvaluateResultException = {
-        exceptionDetails: ExceptionDetails
-      }
-    </pre>
-   </dd>
-</dl>
-
-TODO: Add timeout argument. It's not totally clear how this ought to work; in
-Chrome it seems like the timeout doesn't apply to the promise resolve step, but
-that likely isn't what clients want.
-
-The [=remote end steps=] with |command parameters| are:
-
-1. Let |realm| be the result of [=trying=] to [=get a realm from a target=]
-   given the value of the <code>target</code> field of |command parameters|.
-
-1. Let |environment settings| be the [=environment settings object=] whose
-   [=realm execution context=]'s Realm component is |realm|.
-
-1. Let |source| be the value of the <code>expression</code> field of |command
-   parameters|.
-
-1. Let |await promise| be the value of the <code>awaitPromise</code> field of
-   |command parameters|, if present, or <code>true</code> otherwise.
-
-1. Let |options| be the [=default classic script fetch options=].
-
-1. Let |base URL| be the [=API base URL=] of |environment settings|.
-
-1. Let |script| be the result of [=create a classic script=] with |source|,
-   |environment settings|, |base URL|, and |options|.
-
-1. [=Prepare to run script=] with |environment settings|.
-
-1. Set |evaluation status| to [=ScriptEvaluation=](|script|'s record).
-
-1. If |evaluation status|.\[[Type]] is <code>normal</code>, |await promise| is
-   true, and [=IsPromise=](|evaluation status|.\[[Value]]):
-
-   1. Set |evaluation status| to <a spec=ECMASCRIPT>Await</a>(|evaluation status|.\[[Value]]).
-
-1. [=Clean up after running script=] with |environment settings|.
-
-1. If |evaluation status|.\[[Type]] is <code>throw</code>:
-
-  1. Let |exception details| be the result of [=get exception details=] given |evaluation
-     status|
-
-  1. Return a new map matching the <code>ScriptEvaluateResultException</code>
-     production, with the <code>exceptionDetails</code> field set to |exception
-     details|.
-
-1. Assert: |evaluation status|.\[[Type]] is <code>normal</code>.
-
-1. Let |result| be the result of [=serialize as a remote value=] given
-   |evaluation status|.\[[Value]].
-
-1. Return a new map matching the <code>ScriptEvaluateResultSuccess</code>
-   production, with the <code>result</code> field set to |result|.
-
 #### The script.callFunction Command ####  {#command-script-callFunction}
 
 The <dfn export for=commands>script.callFunction</dfn> command calls a provided
@@ -3761,6 +3662,105 @@ The [=remote end steps=] with |command parameters| are:
    |evaluation status|.\[[Value]].
 
 1. Return a new map matching the <code>ScriptCallFunctionResultSuccess</code>
+   production, with the <code>result</code> field set to |result|.
+
+#### The script.evaluate Command ####  {#command-script-evaluate}
+
+The <dfn export for=commands>script.evaluate</dfn> command evaluates a provided
+script in a given realm. For convenience a browsing context can be provided in
+place of a realm, in which case the realm used is the realm of the browsing
+context's active document.
+
+The method returns the value of executing the provided script, unless it returns
+a promise and <code>awaitPromise</code> is true (which is the default), in which
+case the resolved value of the promise is returned.
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+      <pre class="cddl remote-cddl">
+      ScriptEvaluateCommand = {
+        method: "script.evaluate",
+        params: ScriptEvaluateParameters
+      }
+
+      ScriptEvaluateParameters = {
+        expression: text,
+        target: Target,
+        ?awaitPromise: bool,
+      }
+      </pre>
+   </dd>
+   <dt>Return Type</dt>
+   <dd>
+    <pre class="cddl local-cddl">
+      ScriptEvaluateResult = (
+        ScriptEvaluateResultSuccess //
+        ScriptEvaluateResultException
+      )
+
+      ScriptEvaluateResultSuccess = {
+         result: RemoteValue
+      }
+
+      ScriptEvaluateResultException = {
+        exceptionDetails: ExceptionDetails
+      }
+    </pre>
+   </dd>
+</dl>
+
+TODO: Add timeout argument. It's not totally clear how this ought to work; in
+Chrome it seems like the timeout doesn't apply to the promise resolve step, but
+that likely isn't what clients want.
+
+The [=remote end steps=] with |command parameters| are:
+
+1. Let |realm| be the result of [=trying=] to [=get a realm from a target=]
+   given the value of the <code>target</code> field of |command parameters|.
+
+1. Let |environment settings| be the [=environment settings object=] whose
+   [=realm execution context=]'s Realm component is |realm|.
+
+1. Let |source| be the value of the <code>expression</code> field of |command
+   parameters|.
+
+1. Let |await promise| be the value of the <code>awaitPromise</code> field of
+   |command parameters|, if present, or <code>true</code> otherwise.
+
+1. Let |options| be the [=default classic script fetch options=].
+
+1. Let |base URL| be the [=API base URL=] of |environment settings|.
+
+1. Let |script| be the result of [=create a classic script=] with |source|,
+   |environment settings|, |base URL|, and |options|.
+
+1. [=Prepare to run script=] with |environment settings|.
+
+1. Set |evaluation status| to [=ScriptEvaluation=](|script|'s record).
+
+1. If |evaluation status|.\[[Type]] is <code>normal</code>, |await promise| is
+   true, and [=IsPromise=](|evaluation status|.\[[Value]]):
+
+   1. Set |evaluation status| to <a spec=ECMASCRIPT>Await</a>(|evaluation status|.\[[Value]]).
+
+1. [=Clean up after running script=] with |environment settings|.
+
+1. If |evaluation status|.\[[Type]] is <code>throw</code>:
+
+  1. Let |exception details| be the result of [=get exception details=] given |evaluation
+     status|
+
+  1. Return a new map matching the <code>ScriptEvaluateResultException</code>
+     production, with the <code>exceptionDetails</code> field set to |exception
+     details|.
+
+1. Assert: |evaluation status|.\[[Type]] is <code>normal</code>.
+
+1. Let |result| be the result of [=serialize as a remote value=] given
+   |evaluation status|.\[[Value]].
+
+1. Return a new map matching the <code>ScriptEvaluateResultSuccess</code>
    production, with the <code>result</code> field set to |result|.
 
 #### The script.getRealms Command ####  {#command-script-getRealms}
--- a/index.bs
+++ b/index.bs
@@ -1369,13 +1369,13 @@ SymbolRemoteValue = {
    objectId: ObjectId,
 }
 
-ArrayValue = {
+ArrayRemoteValue = {
   type: "array",
   objectId: ObjectId,
   ?value: ListRemoteValue,
 }
 
-ObjectValue = {
+ObjectRemoteValue = {
   type: "object",
   objectId: ObjectId,
   ?value: MappingRemoteValue,
@@ -1418,8 +1418,23 @@ WeakSetRemoteValue = {
   objectId: ObjectId,
 }
 
+IteratorRemoteValue = {
+  type: "iterator",
+  objectId: ObjectId,
+}
+
+GeneratorRemoteValue = {
+  type: "generator",
+  objectId: ObjectId,
+}
+
 ErrorRemoteValue = {
   type: "error",
+  objectId: ObjectId,
+}
+
+ProxyRemoteValue = {
+  type: "proxy",
   objectId: ObjectId,
 }
 
@@ -1464,6 +1479,12 @@ WindowProxyRemoteValue = {
 Issue: Add WASM types?
 
 Issue: Should WindowProxy get attributes in a similar style to Node?
+
+Issue: Describe `IteratorRemoteValue` deserialization.
+
+Issue: Describe `GeneratorRemoteValue` deserialization.
+
+Issue: Describe `ProxyRemoteValue` deserialization.
 
 Issue: handle String / Number / etc. wrapper objects specially?
 
@@ -3549,7 +3570,10 @@ The [=remote end steps=] with |command parameters| are:
 The <dfn export for=commands>script.callFunction</dfn> command calls a provided
 function with given arguments in a given realm.
 
- <code>RealmInfo</code> can be either a realm or a browsing context.
+<code>RealmInfo</code> can be either a realm or a browsing context.
+
+Note: <code>this</code> argument is ignore in case of arrow function in
+<code>functionDeclaration</code>.
 
 <dl>
    <dt>Command Type</dt>
--- a/index.bs
+++ b/index.bs
@@ -1020,7 +1020,7 @@ To <dfn>serialize primitive protocol value</dfn> given a |value|:
 1. Let |remote value| be undefined.
 
 1. In the following list of conditions and associated steps, run the first set
-   of steps for which the associated condition is true:
+   of steps for which the associated condition is true, if any:
 
   <dl>
     <dt>[=Type=](|value|) is undefined
@@ -1235,8 +1235,6 @@ To <dfn>deserialize local value</dfn> given a |local protocol value|:
 1. In the following list of conditions and associated steps, run the first set of steps for which
    the associated condition is true. If this throws exception, return [=error=] with [=error code=]
    [=invalid argument=]:
-
-  Issue: TODO: add regex options.
 
   <dl>
 
--- a/index.bs
+++ b/index.bs
@@ -101,7 +101,7 @@ spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
     text: Type; url: sec-ecmascript-data-types-and-values
     text: current realm record; url: current-realm
     text: internal slot; url: sec-object-internal-methods-and-internal-slots
-    text: primitive value; url: sec-primitive-value
+    text: primitive ECMAScript value; url: sec-primitive-value
     text: realm; url: sec-code-realms
     text: running execution context; url: running-execution-context
     text: time value; url: sec-time-values-and-time-range
@@ -937,11 +937,11 @@ RemoteReference = {
 
 Issue: Defined <dfn>deserialize a reference</dfn> steps.
 
-## Value ## {#data-types-value}
+## Protocol Value ## {#data-types-protocolValue}
 
-### Primitive Value ### {#data-types-value-PrimitiveValue}
+### Primitive Protocol Value ### {#data-types-protocolValue-primitiveValue}
 
-<dfn>Primitive Value</dfn> represents primitive values which can be serialized to /
+<dfn>Primitive Protocol Value</dfn> represents primitive protocol values which can be serialized to /
 deserialized from ECMAScript without reference to existing objects.
 
 ```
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3199,6 +3201,143 @@ The [=remote end steps=] with |command parameters| are:
...
+   <dd>
+    <pre class="cddl remote-cddl">
+      ScriptInvokeCommand = {
+        method: "script.invoke",
+        params: ScriptInvokeParameters
+      }
+
+      ScriptInvokeParameters = {
+        functionDeclaration: text;
+        arguments: [InvokeArgument];
```
jgraham, line 3226: Arguably we should call this `LocalValue` since it's a value from what WebDriver calls the "local end". But also I think we need to think a bit about requirements. This setup assumes an asymmetry between the client and the browser, in that the client can refer to objects owned by the browser, but the browser can't refer to objects owned by the client. That makes some amount of sense, but let's say for sandboxed script execution you end up with some WebIDL containing a `webdriver.postMessage()` function that emits a `script.postMessage` event containing data. Then let's say, depending on the type of message you want to append to a different array on the local end (client). Obviously one way to do this is to just identify each target array with strings. But it could also be possible to write something like (on the local end, pseudocode):

```js
context.addEventListener("script.postMessage", msg => {
   {target, data} = message;
   target.append(data);
});
let fooTarget = [];
let barTarget = [];
let func = (msgName, target) => {
  window.addEventListener(msgName, event => webdriver.postMessage({target, data: event.data}));
};
context.callFunctionInSandbox(func, ["foo", fooTarget]);
context.callFunctionInSandbox(func, ["bar", barTarget]);
```

I don't know if that's a case that's worth optimising for, but it seems at least worth considering the possibility of being able to pass local-end objects through the remote.

index.bs
```diff
@@ -3199,6 +3201,143 @@ The [=remote end steps=] with |command parameters| are:
...
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl remote-cddl">
+      ScriptInvokeCommand = {
+        method: "script.invoke",
+        params: ScriptInvokeParameters
+      }
+
+      ScriptInvokeParameters = {
```
jgraham, line 3224: Should there be a way to specify the `thisObject` for the call?

index.bs
```diff
@@ -3199,6 +3201,143 @@ The [=remote end steps=] with |command parameters| are:
...
+        functionDeclaration: text;
+        arguments: [InvokeArgument];
+        awaitPromise?: bool;
+        target: Target;
+      }
+
+      InvokeArgument = RemoteValueArgument | LocalValueArgument;
```
jgraham, line 3231: I think this would be
InvokeArgument = (
    RemoteValueArgument //
    LocalValueArgument
);
AFAICT `|` isn't an operator in CDDL.

index.bs
```diff
@@ -3199,6 +3201,143 @@ The [=remote end steps=] with |command parameters| are:
...
+1. Let |options| be the [=default classic script fetch options=].
+
+1. Let |base URL| be the [=API base URL=] of |environment settings|.
+
+1. Let |function script| be the result of [=create a classic script=] with |function declaration|,
+   |environment settings|, |base URL|, and |options|.
+
+1. [=Prepare to run function script=] with |environment settings|.
+
+1. Set |function evaluation status| to [=ScriptEvaluation=](|function script|'s record).
```
jgraham, line 3293: Do we actually want the function to end up in the window global? e.g. if I send `functionDeclaration: "function foo() {}"`, should `window.foo` end up existing?

jgraham: Also, what happens if this doesn't result in a callable? For example if I pass in ``functionDeclaration: "2"`? Or something with side effects like `functionDeclaration: "window.foo` = 1"`?

index.bs
sadym-chromium: thanks!

index.bs
sadym-chromium: > Do we actually want the function to end up in the window global

No! I'm still working on step definitions, and hope to fix it soon.

index.bs
```diff
@@ -2895,6 +2901,143 @@ Note: Future variations of this specification will retain the invariant that
 
 ### Commands ### {#module-script-commands}
 
+#### The script.invoke Command ####  {#command-script-invoke}
```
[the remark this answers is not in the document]
sadym-chromium: Changed spec to `script.callFunction`

index.bs
sadym-chromium: After a [discussion](https://github.com/w3c/webdriver-bidi/issues/140#issuecomment-942498390) decided to keep the page a single source of truth.

index.bs
foolip: My take is that `this` will be null by default, so adding an optional `thisObject` later will be easy. It's not strictly necessary since one can always pass it as an argument instead, but might be nice to have for parity with other protocols. Fine to defer for me.

index.bs
foolip: What does CDP do here?

index.bs
```diff
@@ -3199,6 +3203,129 @@ The [=remote end steps=] with |command parameters| are:
...
+        RemoteValueArgument //
+        LocalValueArgument
+      );
+
+      RemoteValueArgument = {
+        objectId: string;
+      };
+
+      LocalValueArgument = {
+        type: string;
```
jgraham, line 3243: Presumably "type" is going to be a closed set of values that we know how to deserialize.

index.bs
```diff
@@ -3199,6 +3203,129 @@ The [=remote end steps=] with |command parameters| are:
...
+   of |command parameters|.
+
+1. Let |JS arguments list| be an empty list.
+
+1. For each |arg declaration| of |arguments declaration|:
+
+  Issue: describe |deserialize CallFunctionArgument|.
+
+  1. Let |JS arg| be the result of
+     |deserialize CallFunctionArgument|(|arg declaration|).
```
jgraham, line 3293: This will end up like `[=deserialize a local value=] given |argument|`.

index.bs
```diff
@@ -3199,6 +3203,129 @@ The [=remote end steps=] with |command parameters| are:
...
+
+1. Let |function declaration| be the value of the
+   <code>functionDeclaration</code> field of |command parameters|.
+
+1. Let |arguments declaration| be the value of the <code>arguments</code> field
+   of |command parameters|.
+
+1. Let |JS arguments list| be an empty list.
+
+1. For each |arg declaration| of |arguments declaration|:
```
jgraham, line 3288: I'd rather use `argument` than `arg declaration`

index.bs
```diff
@@ -3199,6 +3203,129 @@ The [=remote end steps=] with |command parameters| are:
...
+
+  Issue: describe |deserialize CallFunctionArgument|.
+
+  1. Let |JS arg| be the result of
+     |deserialize CallFunctionArgument|(|arg declaration|).
+
+  1. Append |JS arg| to the |JS arguments list|.
+
+1. Let |JS function object| be the result of
+   [=InstantiateOrdinaryFunctionObject=](|function declaration|).
```
jgraham, line 3298: So if I understand ECMAScript correctly, this only works for the `function [name]() {}` form and not, for example `() => {}`. I also wonder if we should support async functions here.

`InstantiateOrdinaryFunctionObject` also requires explicitly setting the scope of the function, which seems to be missing here.

sadym-chromium: Valid point. 
It looks like it would make more sense just to evaluate the `functionDeclaration`, [assert the result is `callable`](https://tc39.es/ecma262/#sec-iscallable), and [ [[call]] ](https://tc39.es/ecma262/#sec-ecmascript-function-objects-call-thisargument-argumentslist) with deserialized arguments.

index.bs
```diff
@@ -1046,84 +1005,431 @@ BigIntValue = {
...
+
+1. Return the result of getting the value for |object| in |session|'s [=object
+   id map=].
+
+</div>
+
+[=remote end definition=] and [=local end definition=]
+```
+RemoteValue = {
+  PrimitiveProtocolValue //
```
whimboo, line 1332: Should we call this `PrimitiveRemoteValue`?

index.bs
```diff
@@ -1046,84 +1005,431 @@ BigIntValue = {
...
+  MapLocalValue //
+  ObjectLocalValue //
+  RegExpLocalValue //
+  SetLocalValue //
+}
+
+ListLocalValue = [*LocalValue];
+
+ArrayLocalValue = {
+  type: "array",
```
whimboo, line 1134: So for serialization we have `typedarray` and `arraybuffer`. How does that map to deserialization? I don't see that we could deserialize such RemoteValue instances.

Also why do we need two different (remote/local) types for the same RemoteValue definition? That's not clear to me and I haven't seen a discussion about that yet. In general `serialize()` produces a `RemoteValue` and deserialize takes exactly such an object as input.

whimboo (commented):
Before having a closer look at the PR I have a basic question regarding the type definitions (see inline comment), and I would like to get that discussed / answered first. Thanks.

index.bs
jgraham: I think the idea is it's not representing a value that's transmitted only by copy, never by reference. Since the same types are by-copy going from remote-to-local and local-to-remote they can have a single representation in the schema. On the other hand by-reference values have an asymmetric representation; going from remote to local you need to transmit the type, and the handle, whereas going local to remote you only need to transmit the handle. For types where we support sending both a reference and a copy of the data (e.g. arrays, objects, regexp) remote to local sends the type, the handle, and a copy of the data, whereas local to remote sends either a handle _or_ the data required to construct a new object on the remote end.

index.bs
```diff
@@ -3199,6 +3464,177 @@ The [=remote end steps=] with |command parameters| are:
...
+</div>
+
+<div algorithm>
+To <dfn>evaluate function body</dfn> with a given |function declaration|,
+|environment settings|, |base URL|, and |options|:
+
+Note: the |function declaration| is parenthesized and evaluated.
+
+1. Let |function script| be the result of [=create a classic script=] with
+   <code>(</code> + |function declaration| + <code>)</code>, |environment settings|, |base URL|, and |options|.
```
jgraham, line 3557: I think you need to use [string concatenate](https://infra.spec.whatwg.org/#string-concatenate) from the infra spec here, something like `Let |body| be [=concatenate=] «"<code>(</code>", |function declaration|, "<code>)</code>"»`.

index.bs
```diff
@@ -3199,6 +3464,177 @@ The [=remote end steps=] with |command parameters| are:
...
+
+1. Let |function body evaluation status| be the result of [=evaluate function body=]
+   (|function declaration|, |environment settings|, |base URL|, and |options|).
+
+1. If |function body evaluation status|.\[[Type]] is <code>throw</code>:
+
+  1. Let |exception details| be the result of [=get exception details=] given
+     |function body evaluation status|.
+
+  1. Return a new map matching the <code>ScriptCallFunctionResultException</code>
```
jgraham, line 3598: I think we should just make passing in something that doesn't successfully evaluate as js a protocol error

jgraham (commented):
This generally looks good.

There's quite a few cases where some spec constructs have been misused (notably `try`, but also some invented syntaxes for interacting with schema values) and I haven't marked all of them; I assume you can find all the other cases that are like the marked case.

sadym-chromium (commented):
Forgot to publish my responses

index.bs
```diff
@@ -1046,84 +1005,431 @@ BigIntValue = {
...
+1. Let |value| be undefined.
+
+1. If |primitive protocol value| has field <code>value</code>:
+   1. Let |value| be the value of the <code>value</code> field of |primitive protocol value|.
+
+1. In the following list of conditions and associated steps, run the first set
+   of steps for which the associated condition is true:
+
+  <dl>
+    <dt>|type| is undefined
```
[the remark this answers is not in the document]
sadym-chromium: If the type is unknown, deserialization return an `error` with `invalid argument`:
> `1. Return [=error=] with [=error code=] [=invalid argument=]`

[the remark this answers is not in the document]
jgraham: Right, so we could in fact just assert that list of conditions in the definition list is exhaustive since this can only be called with data that matched the schema.

index.bs
```diff
@@ -1046,84 +1005,431 @@ BigIntValue = {
...
+
+</div>
+
+<div algorithm>
+
+To <dfn>map serialized value list to deserialized value list</dfn> given a |serialized value list|:
+
+1. Let |deserialized values| be a new list.
+
+1. For each |serialized value| in the |serialized value list|:
```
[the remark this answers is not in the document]
sadym-chromium: Added TODO

index.bs
```diff
@@ -1046,84 +1005,431 @@ BigIntValue = {
...
+
+1. Return [=success=] with data |deserialized values|.
+
+</div>
+
+<div algorithm>
+
+To <dfn>deserialize a local value</dfn> given a |local protocol value|:
+
+1. If is <code>PrimitiveProtocolValue</code>, return [=deserialize primitive protocol value=]
```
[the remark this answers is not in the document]

index.bs
```diff
@@ -1046,84 +1005,431 @@ BigIntValue = {
...
+   of |local protocol value|.
```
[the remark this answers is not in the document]

index.bs
```diff
@@ -1046,84 +1005,431 @@ BigIntValue = {
...
+To <dfn>map serialized value list to deserialized value list</dfn> given a |serialized value list|:
+
+1. Let |deserialized values| be a new list.
+
+1. For each |serialized value| in the |serialized value list|:
+
+   1. Let |deserialized value| be result of [=try=] to [=deserialize a local value=] with given
+      |serialized value|.
+
+   1. If |deserialized value| is [=error=], return |deserialized value|.
```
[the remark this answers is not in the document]
sadym-chromium: removed

index.bs
```diff
@@ -1046,84 +1005,431 @@ BigIntValue = {
...
+      given |serialized key|.
+
+   1. If |deserialized key| is [=error=], return |deserialized key|.
+
+   1. Let |serialized value| be the second element of |serialized key-value|.
+
+   1. Let |deserialized value| be result of [=try=] to [=deserialize a local value=] with given
+      |serialized value|.
+
+   1. If |deserialized value| is [=error=], return |deserialized value|.
```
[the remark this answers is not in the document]
sadym-chromium: removed

index.bs
```diff
@@ -1046,84 +1005,431 @@ BigIntValue = {
...
+      let |deserialized key| be |serialized key|.
+
+   1. Otherwise let |deserialized key| be result of [=try=] to [=deserialize a local value=] with
+      given |serialized key|.
+
+   1. If |deserialized key| is [=error=], return |deserialized key|.
+
+   1. Let |serialized value| be the second element of |serialized key-value|.
+
+   1. Let |deserialized value| be result of [=try=] to [=deserialize a local value=] with given
```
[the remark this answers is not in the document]
sadym-chromium: (hopefully) fixed

index.bs
```diff
@@ -1046,84 +1005,431 @@ BigIntValue = {
...
+
+1. Let |type| be the value of the <code>type</code> field of |local protocol value|.
+
+1. Let |value| be the value of the <code>value</code> field of |local protocol value|.
+
+1. In the following list of conditions and associated steps, run the first set of steps for which
+   the associated condition is true:
+
+  <dl>
+    <dt>|type| is undefined
```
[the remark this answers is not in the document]
sadym-chromium: removed

index.bs
```diff
@@ -1046,84 +1005,431 @@ BigIntValue = {
...
+    <dd>Return [=error=] with [=error code=] [=invalid argument=].
+
+    <dt>|type| equals <code>ArrayLocalValue.type</code>
```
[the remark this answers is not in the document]
sadym-chromium: replaced with string literals

index.bs
```diff
@@ -3199,6 +3464,177 @@ The [=remote end steps=] with |command parameters| are:
...
+    </pre>
+   </dd>
+</dl>
+
+Issue: TODO: Add timeout argument as described in the script.evaluate.
+
+<div algorithm>
+To <dfn>deserialize an argument value</dfn> given an |serialized argument|:
+
+1. If |serialized argument| is [=RemoteReference=], return [=deserialize a remote reference=] of
```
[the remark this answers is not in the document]

index.bs
```diff
@@ -1046,84 +1005,431 @@ BigIntValue = {
...
+   1. Let |value| be the value of the <code>value</code> field of |primitive protocol value|.
+
+1. In the following list of conditions and associated steps, run the first set
+   of steps for which the associated condition is true:
+
+  <dl>
+    <dt>|type| is undefined
+    <dd>Return [=error=] with [=error code=] [=invalid argument=].
+
+    <dt>|type| equals <code>UndefinedValue.type</code>
```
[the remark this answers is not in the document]
sadym-chromium: Replaced, thanks

index.bs
sadym-chromium: Return `invalid argument` in case of function evaluation failed.

index.bs
sadym-chromium: done, thanks!

index.bs
```diff
@@ -1046,84 +1005,431 @@ BigIntValue = {
...
+
+   1. If |serialized key-value| has not 2 elements,
+      return [=error=] with [=error code=] [=invalid argument=].
+
+   1. Let |serialized key| be the first element of |serialized key-value|.
+
+   1. If |serialized key| is a <code>string</code>,
+      let |deserialized key| be |serialized key|.
+
+   1. Otherwise let |deserialized key| be result of [=try=] to [=deserialize a local value=] with
```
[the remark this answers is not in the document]

index.bs
```diff
@@ -3199,6 +3464,177 @@ The [=remote end steps=] with |command parameters| are:
...
+  1. Let |deserialized argument| be the result of [=try=] to [=deserialize an argument value=] with
+     given |serialized argument|.
+
+  1. If |deserialized argument| is [=error=], return |deserialized argument|.
+
+  1. Append |deserialized argument| to the |deserialized arguments list|.
+
+1. Return [=success=] with data |deserialized arguments list|.
+
+Issue: Clarify behaviour in case of deserialization failed.
```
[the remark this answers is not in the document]
sadym-chromium: true. Removed TODO

index.bs
```diff
@@ -3199,6 +3464,177 @@ The [=remote end steps=] with |command parameters| are:
...
+To <dfn>deserialize arguments</dfn> given |serialized arguments list|:
+
+1. Let |deserialized arguments list| be an empty list.
+
+1. For each |serialized argument| of |serialized arguments list|:
+
+  1. Let |deserialized argument| be the result of [=try=] to [=deserialize an argument value=] with
+     given |serialized argument|.
+
+  1. If |deserialized argument| is [=error=], return |deserialized argument|.
```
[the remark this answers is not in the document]
sadym-chromium: Wasn't sure in that. Removed here and in other places. Thanks!

index.bs
```diff
@@ -3199,6 +3464,177 @@ The [=remote end steps=] with |command parameters| are:
...
+
+<div algorithm>
+
+To <dfn>deserialize arguments</dfn> given |serialized arguments list|:
+
+1. Let |deserialized arguments list| be an empty list.
+
+1. For each |serialized argument| of |serialized arguments list|:
+
+  1. Let |deserialized argument| be the result of [=try=] to [=deserialize an argument value=] with
```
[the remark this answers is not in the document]

index.bs
```diff
@@ -1046,84 +1011,437 @@ BigIntValue = {
...
+   no such a field.
+
+1. Let |value| be the value of the <code>value</code> field of |local protocol value| or undefined
+   if no such a field.
+
+1. In the following list of conditions and associated steps, run the first set of steps for which
+   the associated condition is true. If this throws exception, return [=error=] with [=error code=]
+   [=invalid argument=]:
+
+  Issue: TODO: add regex options.
```
jgraham, line 1239: Is this done, or what exactly is missing?

index.bs
```diff
@@ -1046,84 +1011,437 @@ BigIntValue = {
...
+
+1. Let |type| be the value of the <code>type</code> field of |primitive protocol value|.
+
+1. Let |value| be undefined.
+
+1. If |primitive protocol value| has field <code>value</code>:
+   1. Let |value| be the value of the <code>value</code> field of |primitive protocol value|.
+
+1. In the following list of conditions and associated steps, run the first set of steps for which
+   the associated condition is true. If this throws exception, return [=error=] with [=error code=]
```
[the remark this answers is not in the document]
sadym-chromium: Theoretically the value can be not parsable like this:
```
{
  type: "bigint",
  value: "SOME_NONSENSE"
}
```

We can either:
1. explicitly describe the value should match a regex `-?\d+` to match the serialization format
2. delegate it to ECMAScript parser (like now).

Do I miss some options?

index.bs
sadym-chromium: It's done. Removed.

index.bs
```diff
@@ -1046,84 +1011,437 @@ BigIntValue = {
...
-  type: "symbol",
-  objectId: ObjectId,
+<div algorithm>
+
+To <dfn>serialize primitive protocol value</dfn> given a |value|:
+
+1. Let |remote value| be undefined.
+
+1. In the following list of conditions and associated steps, run the first set
+   of steps for which the associated condition is true:
```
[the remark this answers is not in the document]

sadym-chromium:
There is a discrepancy in the way classic and arrow functions interpret `this`:
```
// Returns `globalThis`:
(() => this).apply(1);

// Returns `1`:
(function () { return this; }).apply(1);
```

Transposition the logic to classic function, the result of the following command should be `{ "objectId": "SOME_OBJECT_ID" }`:
```
{
   ...
   "functionDeclaration": "function(){return this}",
   "this": { "objectId": "SOME_OBJECT_ID" }
   ...
}
```

The question is what the result should be in case of having an arrow function. Should it be `globalThis` (which I personally find illogical), or the same as classic function (which goes against the ECMAScript implementation, but is more user-friendly)?
```
{
   ...
   "functionDeclaration": "()=>{return this}",
   "this": { "objectId": "SOME_OBJECT_ID" }
   ...
}
```

mathiasbynens:
IMHO it would be surprising to deviate from the ECMAScript spec.

jgraham:
Yes, I think we should probably just match the bhaviour of [`Function.prototype.call`](https://tc39.es/ecma262/#sec-function.prototype.call) here. But happily if you look at the spec, it also passes in the `thisObj` to `[call](https://tc39.es/ecma262/#sec-call)` and the fact that an arrow function already has a bound `thisObj` is handled later on. So we don't need any changes to achieve that. But we should maybe have a note in the spec that the `thisObj` will be ignored for arrow functions.

foolip:
Letting `thisObject` silently have no effect sounds acceptable. But since this is likely a programmer error, is there a way to detect this situation and treat it as an invalid parameter situation?

jgraham:
> Letting `thisObject` silently have no effect sounds acceptable. But since this is likely a programmer error, is there a way to detect this situation and treat it as an invalid parameter situation?

I don't think so? As far as I can tell doing something in ECMAScript like `(() => {return this}).call({})` will work without error, but won't do what you expect. I question the value of trying to be more clever than the underlying language at trying to detect an error here.

sadym-chromium:
Thanks for all the inputs. Let's keep the spec as is, and keep the behaviour `ECMAScript`-like with ignoring `this` for arrow functions.

whimboo:
@sadym-chromium I would like to go through the full PR again but there a lot of left open conversations that seem to have been done. Would you mind to close all of these? Thanks a lot.

sadym-chromium:
@whimboo sure. I wasn't sure what is the protocol, and expected conversations to be resolved by their openers. Closed conversations I consider resolved.

index.bs
```diff
@@ -1046,101 +1010,467 @@ BigIntValue = {
...
+   [=invalid argument=]:
+
+  <dl>
+
+    <dt>|type| is the string "<code>array</code>"
+    <dd>
+      1. Let |deserialized value list| be a result of [=trying=] to [=deserialize value list=] with
+         given |value|.
+
+      1. Return [=success=] with data set to a new [=Array=] object constructed with
```
jgraham, line 1245: Can't we just say `return [=CreateArrayFromList=](|deserialized value list|)`?

index.bs
```diff
@@ -1046,101 +1010,467 @@ BigIntValue = {
...
+   [=deserialize primitive protocol value=] with |local protocol value|.
+
+1. Let |type| be the value of the <code>type</code> field of |local protocol value| or undefined if
+   no such a field.
+
+1. Let |value| be the value of the <code>value</code> field of |local protocol value| or undefined
+   if no such a field.
+
+1. In the following list of conditions and associated steps, run the first set of steps for which
+   the associated condition is true. If this throws exception, return [=error=] with [=error code=]
```
jgraham, line 1235: Again, I think we need to handle failure in each case.

index.bs
```diff
@@ -1046,101 +1010,467 @@ BigIntValue = {
...
+
+      1. Return [=success=] with data set to a new [=Array=] object constructed with
+         |deserialized value list| as <code>values</code>.
+
+    <dt>|type| is the string "<code>date</code>"
+    <dd>
+      1. If |value| does not match [=Date Time String Format=], return [=error=] with
+         [=error code=] [=invalid argument=].
+
+      1. Return [=success=] with data set to a result of [=Date.parse=] with |value| as
```
jgraham, line 1253: We can write `[=Date.parse=](|value|)`; ECMAScript algorithms can be called with parens.

index.bs
```diff
@@ -1046,101 +1010,467 @@ BigIntValue = {
...
+         <code>string</code>.
+
+    <dt>|type| is the string "<code>map</code>"
+    <dd>
+      1. Let |deserialized key-value list| be a result of [=trying=] to
+         [=deserialize key-value list=] with |value|.
+
+      1. Return [=success=] with data set to a new [=Map=] object constructed with
```
jgraham, line 1261:
```
1. Let |iterable| be [=CreateArrayFromList=](|deserialized key-value list|)
1. `Return [=success=] with data [=Map=](|iterable|)
```

index.bs
```diff
@@ -1046,101 +1010,467 @@ BigIntValue = {
...
+         |deserialized key-value list| as <code>iterable</code>.
+
+    <dt>|type| is the string "<code>object</code>"
+    <dd>
+      1. Let |deserialized key-value list| be a result of [=trying=] to
+         [=deserialize key-value list=] with |value|.
+
+      1. Return [=success=] with data set to a result of [=Object.fromEntries=] with
```
jgraham, line 1269: Similar to above.

index.bs
```diff
@@ -1046,101 +1010,467 @@ BigIntValue = {
...
+         |deserialized key-value list| as <code>iterable</code>.
+
+    <dt>|type| is the string "<code>regexp</code>"
+    <dd>
+      1. Let |pattern| be the value of the <code>pattern</code> field of |local protocol value|.
+
+      1. Let |flags| be the value of the <code>flags</code> field of |local protocol value| or
+         undefined if no such a field.
+
+      1. Return [=success=] with data set to a new [=RegExp=] object constructed with |pattern| and
```
jgraham, line 1279: `[=Regexp=](|pattern|, |flags|)`

index.bs
```diff
@@ -1046,101 +1010,467 @@ BigIntValue = {
...
+         |flags|.
+
+    <dt>|type| is the string "<code>set</code>"
+    <dd>
+      1. Let |deserialized value list| be a result of [=trying=] to
+         [=deserialize value list=] with |value|.
+
+      1. Return [=success=] with data set to a new [=Set object=] constructed with
```
jgraham, line 1287: Similar to above

index.bs
```diff
@@ -1046,101 +1010,467 @@ BigIntValue = {
...
+of the data in the ECMAScript object; for container types this can contain
+further <code>RemoteValue</code> instances. The <code>value</code> property can
+be null if there is a duplicate object i.e. the object has already been
+serialized in the current <code>RemoteValue</code>, perhaps as part of a
+cycle, or otherwise when the maximum serialization depth is reached.
+
+[=Nodes=] are also represented by <code>RemoteValue</code> instances. These have
+a partial serialization of the node in the value property.
+
+Note: mirror objects do not keep the original object alive in the runtime. If an
```
jgraham, line 1316: I know you just moved this but I don't think we've actually agreed that's true.

index.bs
```diff
@@ -3199,6 +3494,176 @@ The [=remote end steps=] with |command parameters| are:
...
    production, with the <code>result</code> field set to |result|.
 
+#### The script.callFunction Command ####  {#command-script-callFunction}
+
+The <dfn export for=commands>script.callFunction</dfn> command calls a provided
+function with given arguments in a given realm.
+
+<code>RealmInfo</code> can be either a realm or a browsing context.
+
+Note: <code>this</code> argument is ignore in case of arrow function in
```
jgraham, line 3504: `The <code>this</code> argument`

Also, it's nto quite ignored; if it doesn't resolve to an object we throw an error, but it doesn't affect the `this` binding inside the function.

index.bs
sadym-chromium: I'm about to fix all the arrays with `CreateArrayFromList`.

Do you mean returning value instead of `[=success=] with data` as well? Primitive deserialization returns `[=success=] with data`.

jgraham: I mean `return [=success=] with data…`, yes. Sorry for the confusion :)

index.bs
sadym-chromium: Fixed note.

index.bs
sadym-chromium: Added an issue to reconsider the lifecycle.

index.bs

index.bs

index.bs

index.bs

index.bs

index.bs
sadym-chromium: done. Exception is expected only in Regex parsing.

index.bs

index.bs
```diff
@@ -1046,101 +1010,467 @@ BigIntValue = {
...
+
+   1. Let |serialized value| be |serialized key-value|[1].
+
+   1. Let |deserialized value| be result of [=trying=] to [=deserialize local value=] with
+      |serialized value|.
+
+   1. Append a <code>new [=Array=](|deserialized key|, |deserialized value|)</code>
+      to |deserialized key-value list|.
+
+1. Return [=success=] with data |deserialized key-value list|.
```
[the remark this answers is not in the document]

index.bs
```diff
@@ -1046,101 +1010,467 @@ BigIntValue = {
...
+
+   1. Otherwise let |deserialized key| be result of [=trying=] to [=deserialize local value=]
+      with |serialized key|.
+
+   1. Let |serialized value| be |serialized key-value|[1].
+
+   1. Let |deserialized value| be result of [=trying=] to [=deserialize local value=] with
+      |serialized value|.
+
+   1. Append a <code>new [=Array=](|deserialized key|, |deserialized value|)</code>
```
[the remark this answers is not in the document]

index.bs
```diff
@@ -919,104 +932,55 @@ requests to start a WebDriver session, it must:
...
+
+1. Let |object id| be the value of the <code>objectId</code> field of |remote reference|.
 
 1. For each |object| → |id value| of [=object id map=]:
 
   1. If |id value| is equal to |object id|, return [=success=] with data
      |object|
 
-1. Return [=error=] with [=error code=] [=no such element=]
+1. Return [=error=] with [=error code=] [=invalid argument=].
```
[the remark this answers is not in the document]
sadym-chromium: added an issue

sadym-chromium:
@jgraham let me know if you prefer me closing conversations, or you want to do it yourself. I find it more logical a conversation to be closed by the opener to make sure it was understood in the right way, but still not sure what is the convention here.

index.bs
```diff
@@ -1046,101 +1012,480 @@ BigIntValue = {
...
+
+<div algorithm>
+To get the <dfn>object id for an object</dfn> given a |session| and |object|:
+
+1. If |session|'s [=object id map=] does not contain |object|, run the
+   following steps:
+
+  1. Let |object id| be a new, unique, string identifier for |object|. If
+     |object| is an [=/element=] this must be the [=web element reference=] for
+     |object|; if it's a {{WindowProxy}} object, this must be the [=window
```
whimboo, line 1343: What about a `shadow root identifier`? It seems to be missing here.

I'm not sure if we still need a `frame identifier`. @jgraham?

index.bs
```diff
@@ -1046,101 +1012,480 @@ BigIntValue = {
...
+     handle=] for |object|.
+
+  1. Set the value of |object| in |session|'s [=object id map=] to |object id|.
+
+1. Return the result of getting the value for |object| in |session|'s [=object
+   id map=].
+
+</div>
+
+[=remote end definition=] and [=local end definition=]
```
whimboo, line 1353: Is that expected to be a heading? It looks a bit lost right now.

index.bs
```diff
@@ -2768,6 +3078,7 @@ relating to script realms and execution.
 
 ScriptCommand = (
   ScriptEvaluateCommand //
+  ScriptCallFunctionCommand //
```
whimboo, line 3092:
```suggestion
  ScriptCallFunctionCommand //
  ScriptEvaluateCommand //
```

index.bs
```diff
@@ -2779,6 +3090,7 @@ ScriptCommand = (
 
 ScriptResult = (
   ScriptEvaluateResult //
+  ScriptCallFunctionResult //
```
whimboo, line 3104:
```suggestion
  ScriptCallFunctionResult //
  ScriptEvaluateResult //
```

index.bs
```diff
@@ -3199,6 +3511,176 @@ The [=remote end steps=] with |command parameters| are:
...
+   <dd>
+    <pre class="cddl remote-cddl">
+      ScriptCallFunctionCommand = {
+        method: "script.callFunction",
+        params: ScriptCallFunctionParameters
+      }
+
+      ScriptCallFunctionParameters = {
+        functionDeclaration: text;
+        arguments: [ArgumentValue];
```
whimboo, line 3546: Shouldn't these be optional?

index.bs
```diff
@@ -3199,6 +3511,176 @@ The [=remote end steps=] with |command parameters| are:
 1. Return a new map matching the <code>ScriptEvaluateResultSuccess</code>
    production, with the <code>result</code> field set to |result|.
 
+#### The script.callFunction Command ####  {#command-script-callFunction}
```
whimboo, line 3525: Please move up before `script.evaluate`.

index.bs
```diff
@@ -3199,6 +3511,176 @@ The [=remote end steps=] with |command parameters| are:
...
+   </dd>
+</dl>
+
+Issue: TODO: Add timeout argument as described in the script.evaluate.
+
+<div algorithm>
+To <dfn>deserialize a parameter value</dfn> given an |serialized value|:
+
+1. If |serialized value| matches the [=RemoteReference=] production, return
+   [=deserialize remote reference=] of |serialized value|.
```
whimboo, line 3584: Ah, so here we actually check for the remote reference. Should we repeat that for each and every command, or better handle it within `[=deserialize local value=]`?

whimboo (commented):
This is great work @sadym-chromium! Sorry that it has been taken a bit longer to go through this PR but finally I found the time. I hope the inline comments are helpful to get the remaining issues clarified and fixed. It would be great to have it merged soon.

I'll leave the final review to @jgraham.

index.bs
sadym-chromium: @whimboo I'm not sure how to handle the shadow roots. Could you please create a dedicated issue to keep the question out of scope of the current PR?

index.bs
sadym-chromium: `[=deserialize local value=]` now cares about `[=RemoteReference=]`

index.bs
sadym-chromium: done

index.bs

index.bs
sadym-chromium: I'm not sure why this line is there at all. Removed.

index.bs
```diff
@@ -919,104 +932,57 @@ requests to start a WebDriver session, it must:
...
 
-</div>
+Issue: handle "stale object reference" case.
 
-<!--
 <div algorithm>
-To get the <dfn>object for an object id</dfn> given an |object id|:
+To <dfn>deserialize remote reference</dfn> given a |remote reference|:
+
+1. Let |object id| be the value of the <code>objectId</code> field of |remote reference|.
```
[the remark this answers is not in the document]
sadym-chromium: It looks like I'm not quite understand the comment. The `remote reference` is a variable, so no links to types are expected.

index.bs
```diff
@@ -919,104 +932,57 @@ requests to start a WebDriver session, it must:
...
-serialized in the current <code>RemoteValue</code>, perhaps as part of a
-cycle, or otherwise when the maximum serialization depth is reached.
-
-[=Nodes=] are also represented by <code>RemoteValue</code> instances. These have
-a partial serialization of the node in the value property.
-
-Note: mirror objects do not keep the original object alive in the runtime. If an
-object is discarded in the runtime, subsequent attempts to access it via the
-protocol will result in an error.
+<dfn>RemoteReference</dfn> represents reference to existing ECMAScript object in
```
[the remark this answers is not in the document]

index.bs
```diff
@@ -1046,101 +1013,483 @@ BigIntValue = {
...
+    <dd>Return [=success=] with data [=undefined=].
+
+    <dt>|type| is the string "<code>null</code>"
+    <dd>Return [=success=] with data [=null=].
+
+    <dt>|type| is the string "<code>string</code>"
+    <dd>Return [=success=] with data |value|.
+
+    <dt>|type| is the string "<code>number</code>"
+    <dd>Return [=success=] with data [=Number=](|value|).
```
jgraham, line 1110: Similarly to the BigInt case I think it's clearer to use https://tc39.es/ecma262/#sec-stringtonumber directly.

jgraham (changes requested):
Just a handful of last comments.

index.bs
```diff
@@ -3171,6 +3498,170 @@ Issue: This has the wrong error code
...
+
+1. Let |function object| be |function body evaluation status|.\[[Value]].
+
+1. If [=IsCallable=](|function object|) is <code>false</code>:
+
+   1. Return an [=error=] with [=error code=] [=Invalid Argument=]
+
+1. [=Prepare to run script=] with |environment settings|.
+
+1. Set |evaluation status| to
```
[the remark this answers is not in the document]
sadym-chromium: done

index.bs
```diff
@@ -1046,101 +1013,483 @@ BigIntValue = {
...
+<div algorithm>
+To get the <dfn>object id for an object</dfn> given a |session| and |object|:
+
+1. If |session|'s [=object id map=] does not contain |object|, run the
+   following steps:
+
+  1. Let |object id| be a new, unique, string identifier for |object|. If
+     |object| is an [=/element=] this must be the [=web element reference=] for
+     |object|; if it's a {{WindowProxy}} object, this must be the [=window
+     handle=] for |object|.
```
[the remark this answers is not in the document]
sadym-chromium: changed

index.bs

sadym-chromium:
@jgraham done, thanks for such a precise view!

jgraham (approved):

index.bs
```diff
@@ -1046,101 +1012,480 @@ BigIntValue = {
...
+
+    <dt>|type| is the string "<code>number</code>"
+    <dd>Return [=success=] with data [=Number=](|value|).
+
+    <dt>|type| is the string "<code>boolean</code>"
+    <dd>Return [=success=] with data |value|.
+
+    <dt>|type| is the string "<code>bigint</code>"
+    <dd>
+      1. Let |bigint_result| be [=BigInt=](|value|). If this throws exception, return [=error=]
```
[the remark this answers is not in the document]
whimboo: @jgraham Given that this PR got merged, is there anything that we should follow-up on this in a new issue / PR?

index.bs
whimboo: I would have done but missed to do that because this issue was marked as resolved. Filed https://github.com/w3c/webdriver-bidi/issues/156 now.

index.bs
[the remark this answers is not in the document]
whimboo: Sorry, cannot remember myself. Marking as resolved.

index.bs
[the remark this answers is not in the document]
whimboo: Actually this has been done: `Let bigint_result be StringToBigInt(value).`. So it just didn't get resolved.

PR #152, opened 2021-11-30
Editorial: use HTML multipage URLs everywhere

foolip:
All automatically linked terms use the multipage URLs, so do that for
our custom terms as well.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/152.html" title="Last updated on Nov 30, 2021, 12:39 PM UTC (326d772)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/152/c0e2c44...326d772.html" title="Last updated on Nov 30, 2021, 12:39 PM UTC (326d772)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -103,25 +103,24 @@ spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
     text: realm; url: sec-code-realms
     text: running execution context; url: running-execution-context
     text: time value; url: sec-time-values-and-time-range
-spec: HTML; urlPrefix: https://html.spec.whatwg.org/
+spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
   type: dfn
-    text: WindowProxy; url: windowproxy
-    text: a browsing context is discarded; url: a-browsing-context-is-discarded
-    text: create a classic script; url: creating-a-classic-script
-    text: create a new browsing context; url: creating-a-new-browsing-context
-    text: default classic script fetch options; url: default-classic-script-fetch-options
-    text: environment settings object's Realm; url: environment-settings-object's-realm
-    text: handled; url: concept-error-handled
-    text: history handling behavior; url: history-handling-behavior
-    text: navigation id; url: concept-navigation-id
-    text: report an error; url: report-the-error
-    text: remove a browsing context; url: bcg-remove
-    text: session history; url: session-history
-    text: set up a window environment settings object; url: set-up-a-window-environment-settings-object
-    text: set up a worker environment settings object; url: set-up-a-worker-environment-settings-object
-    text: set up a worklet environment settings object; url: set-up-a-worklet-environment-settings-object
-    text: worker event loop; url: worker-event-loop-2
-    text: worklet global scopes; url: concept-document-worklet-global-scopes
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -103,25 +103,24 @@ spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
...
   type: dfn
-    text: WindowProxy; url: windowproxy
-    text: a browsing context is discarded; url: a-browsing-context-is-discarded
-    text: create a classic script; url: creating-a-classic-script
-    text: create a new browsing context; url: creating-a-new-browsing-context
-    text: default classic script fetch options; url: default-classic-script-fetch-options
-    text: environment settings object's Realm; url: environment-settings-object's-realm
-    text: handled; url: concept-error-handled
-    text: history handling behavior; url: history-handling-behavior
-    text: navigation id; url: concept-navigation-id
```
foolip, line 116: Note that this URL was broken before.

index.bs
```diff
@@ -103,25 +103,24 @@ spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
     text: realm; url: sec-code-realms
     text: running execution context; url: running-execution-context
     text: time value; url: sec-time-values-and-time-range
-spec: HTML; urlPrefix: https://html.spec.whatwg.org/
+spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
   type: dfn
-    text: WindowProxy; url: windowproxy
```
foolip, line 108: `WindowProxy` linking works without this.

index.bs
```diff
@@ -103,25 +103,24 @@ spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
...
-    text: report an error; url: report-the-error
-    text: remove a browsing context; url: bcg-remove
-    text: session history; url: session-history
-    text: set up a window environment settings object; url: set-up-a-window-environment-settings-object
-    text: set up a worker environment settings object; url: set-up-a-worker-environment-settings-object
-    text: set up a worklet environment settings object; url: set-up-a-worklet-environment-settings-object
-    text: worker event loop; url: worker-event-loop-2
-    text: worklet global scopes; url: concept-document-worklet-global-scopes
+    text: a browsing context is discarded; url: window-object.html#a-browsing-context-is-discarded
+    text: create a classic script; url: multipage/webappapis.html#creating-a-classic-script
```
jgraham, line 109: Is this `multipage/` prefix correct?

jgraham (commented):
I would very strongly prefer that we always linked to the single page spec, if that's possible to configure instead.

index.bs
foolip: No! :)

foolip:
@jgraham I don't think it is possible to configure. I can't find any issues about this in https://github.com/tabatkins/bikeshed and also nothing about "multipage" in the source that would suggest support for switching to single-page.

jgraham (approved):
OK, that's disappointing. Using the multipage spec both makes it harder to add links manually when required and also makes the destination of the link less useful. But agreed that we should be consistent.

foolip:
FWIW, I'm also in team single-page because that makes searching for uses of a term much easier, but I suspect that's not a majority preference.

PR #153, opened 2021-11-30
Add a context parameter for browsingContext.create

foolip:
This is based on @jgraham's suggestion:
https://github.com/w3c/webdriver-bidi/pull/133#discussion_r757432132

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/153.html" title="Last updated on Nov 30, 2021, 3:23 PM UTC (f9f81f6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/153/f5ce465...f9f81f6.html" title="Last updated on Nov 30, 2021, 3:23 PM UTC (f9f81f6)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/153.html" title="Last updated on Oct 5, 2022, 10:28 AM UTC (54abb51)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/153/01c8d75...54abb51.html" title="Last updated on Oct 5, 2022, 10:28 AM UTC (54abb51)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2876,7 +2876,8 @@ The <dfn export for=commands>browsingContext.create</dfn> command creates a new
       BrowsingContextCreateType = "tab" / "window"
 
       BrowsingContextCreateParameters = {
-        type: BrowsingContextCreateType
+        type: BrowsingContextCreateType,
+        ?referenceContext: BrowsingContext
       }
       </pre>
    </dd>
@@ -2896,6 +2897,19 @@ The [=remote end steps=] with |command parameters| are:
   1. Let |type| be the value of the <code>type</code> field of
      |command parameters|.
 
+  1. Let |reference context id| be the value of the <code>referenceContext</code>
+     field of |command parameters|, if present, or null otherwise.
+
+  1. If |reference context id| is not null, let |reference context| be the
+     result of [=trying=] to [=get a browsing context=] with
+     |reference context id|. Otherwise let |reference context| be null.
+
+ 1. If |reference context| is not null and is not a [=top-level browsing context=],
+    return [=error=] with [=error code=] [=invalid argument=].
+
+  1. If the implementation is unable to create a new browsing context for any
+     reason then return [=error=] with [=error code=] [=unsupported operation=].
+
   <!-- This is based on step 5 of https://w3c.github.io/webdriver/#new-window,
        but without using the "current browsing context" concept. -->
   1. Create a new [=top-level browsing context=] by running the [=window open
@@ -2903,19 +2917,26 @@ The [=remote end steps=] with |command parameters| are:
      <var ignore>target</var> set to the empty string, and
      <var ignore>features</var> set to "<code>noopener</code>". This must be
      done without invoking the [=/focusing steps=] for the created browsing
-     context. If |type| is "<code>tab</code>", and the implementation supports
-     multiple browsing contexts in the same OS window, the new browsing context
-     should share an OS window with any other [=top-level browsing context=]. If
-     |type| is "<code>window</code>", and the implementation supports multiple
-     browsing contexts in separate OS windows, the created browsing context
-     should be in a new OS window. In all other cases the details of how the
-     browsing context is presented to the user are implementation defined.
+     context. Which OS window the new [=/browsing context=] is created in
+     depends on |type| and |reference context|:
 
-     Issue: HTML's [=window open steps=] depend on the "entry global object"
-     and WebDriver's <a href="https://w3c.github.io/webdriver/#new-window">New
-     Window</a> command depends on the "current browsing context". Given that we
-     don't have any such implicit context here, what OS window should new tabs
-     be opened in?
+     * If |type| is "<code>tab</code>" and the implementation supports
+       multiple browsing contexts in the same OS window:
+
+       * The new browsing context should reuse an existing OS window, if any.
+
+       * If |reference context| is not null, the new browsing context should
+         reuse the window containing |reference context|, if any. If the
+         top-level browsing contexts inside an OS window have a definite ordering,
+         the new browsing context should be immediately after
+         |reference context|'s [=top-level browsing context=] in that ordering.
+
+     * If |type| is "<code>window</code>", and the implementation supports
+       multiple browsing contexts in separate OS windows, the created browsing
+       context should be in a new OS window.
+
+     * Otherwise, the details of how the browsing context is presented to the
+       user are implementation defined.
 
   1. Let |context id| be the [=browsing context id=] of the newly created
      [=/browsing context=].


index.bs
```diff
@@ -2234,26 +2235,34 @@ The [=remote end steps=] with |command parameters| are:
...
-     and WebDriver's <a href="https://w3c.github.io/webdriver/#new-window">New
-     Window</a> command depends on the "current browsing context". Given that we
-     don't have any such implicit context here, what OS window should new tabs
-     be opened in?
+     context. Which OS window the new [=/browsing context=] is created in
+     depends on |type| and |reference context|:
+
+     * If |type| is "<code>tab</code>" and the implementation supports
+       multiple browsing contexts in the same OS window, the new browsing context
+       should share an OS window with |reference context| if not null, and any
```
foolip, line 2257: Can we require anything beyond sharing a window? For example, can we require that the new browsing context is directly after _reference context_ in a list of tabs? If so, should we call the parameter something less generic than `context`, like `afterContext`?

jgraham: Well we're not currently exposing the tab ordering in any way, but maybe we should say something like, 
```
* If |type| is "<code>tab</code> and the implementation supports multiple browsing contexts in the same OS window:
   * The new browsing context should re-use an existing OS window, if any.
   * If |reference context| is not null, the should be the window containing |reference context|, if any. If the top level browsing contexts inside an OS window have a definite ordering, the new created top level browsing context should be immediately after |reference context|'s [=top level browsing context=] in that ordering.
 ```
 
 I don't know if the ordering thing also makes sense for OS windows? I don't really know if the concept of OS windows having an order exists anywhere, or if you can choose where new windows appear in that ordering. In any case it seems pretty hard to observe.

Comment (whimboo) [reply index.bs:2257]: In Marionette we have the feature to open a new context (tab and window) in the background. The ordering is then checked, but I doubt that this is something we actually would have to offer via WebDriver BiDi. See also the recently closed issue https://github.com/whatwg/html/issues/5872.

Comment (bwalderman) [reply index.bs:2257]: As a start, I think it makes sense to offer the same guarantees that the classic New Window command has, which is just that the new context shares the same window if requested. There's some other relevant comments in #151.

File: index.bs [hunk 2]
```diff
@@ -2234,26 +2235,34 @@ The [=remote end steps=] with |command parameters| are:
...
+       other [=top-level browsing context=] otherwise.
+
+     * If |type| is "<code>window</code>", and the implementation supports
+       multiple browsing contexts in separate OS windows, the created browsing
+       context should be in a new OS window.
+
+     * Otherwise, the details of how the browsing context is presented to the
+       user are implementation defined.
```
whimboo, line 2265: What's not clear to me here is if this could also be a possible error. Or should we assume that the reverse type might always work? Probably not, so I would propose to add a paragraph that describes what needs to happen if no new browsing context can be created.

Given that the above mentioned issue is not tightly related to the changes in this PR maybe we should cover that in a separate one?

foolip: An implementation that can't create new browsing contexts at all would just not implement `browsingContext.create`, right? Or are you thinking of some situation where it can succeed in some cases but fail in others?

jgraham: I think you can at least imagine an implementation that can create N browsing contexts, but not N+1. But I don't think we handle that in the WebDriver classic case and I'm not sure it should be a slocker to landing improvements here.

index.bs
```diff
@@ -2234,26 +2235,34 @@ The [=remote end steps=] with |command parameters| are:
   1. Let |type| be the value of the <code>type</code> field of
      |command parameters|.
 
+  1. Let |reference context id| be the value of the <code>context</code> field
+     of |command parameters|, if present, or null otherwise.
+
+  1. If |reference context id| is not null, let |reference context| be the
```
whimboo, line 2241: Question: Should we enforce the `reference context` to be a top-level browsing context? Right now it would accept any browsing context.

index.bs
```diff
@@ -2234,26 +2235,34 @@ The [=remote end steps=] with |command parameters| are:
   1. Let |type| be the value of the <code>type</code> field of
      |command parameters|.
 
+  1. Let |reference context id| be the value of the <code>context</code> field
+     of |command parameters|, if present, or null otherwise.
```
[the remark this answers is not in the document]
foolip: Let's make this an invalid argument error.

index.bs
foolip: I've pushed a change using James's wording.

foolip:
This is ready for review again now.

index.bs
```diff
@@ -2780,26 +2781,46 @@ The [=remote end steps=] with |command parameters| are:
...
-
-     Issue: HTML's [=window open steps=] depend on the "entry global object"
-     and WebDriver's <a href="https://w3c.github.io/webdriver/#new-window">New
-     Window</a> command depends on the "current browsing context". Given that we
-     don't have any such implicit context here, what OS window should new tabs
-     be opened in?
+     context. Which OS window the new [=/browsing context=] is created in
+     depends on |type| and |reference context|:
+
+     * If |type| is "<code>tab</code>" and the implementation supports
```
whimboo, line 2807: We don't use quotes around code.

```suggestion
     * If |type| is <code>tab</code> and the implementation supports
```

index.bs
```diff
@@ -2780,26 +2781,46 @@ The [=remote end steps=] with |command parameters| are:
...
+     of |command parameters|, if present, or null otherwise.
+
+  1. If |reference context id| is not null, let |reference context| be the
+     result of [=trying=] to [=get a browsing context=] with
+     |reference context id|. Otherwise let |reference context| be null.
+
+ 1. If |reference context| is not null and is not a [=top-level browsing context=],
+    return [=error=] with [=error code=] [=invalid argument=].
+
+  1. If the implementation is unable to creating a new browsing context for any
```
juliandescottes, line 2794:
```suggestion
  1. If the implementation is unable to create a new browsing context for any
```

juliandescottes (approved):
Looks good to me, just one nit on top of the suggestions from @whimboo , thanks!

index.bs
```diff
@@ -2760,7 +2760,8 @@ The <dfn export for=commands>browsingContext.create</dfn> command creates a new
       BrowsingContextCreateType = "tab" / "window"
 
       BrowsingContextCreateParameters = {
-        type: BrowsingContextCreateType
+        type: BrowsingContextCreateType,
+        ?context: BrowsingContext
```
sadym-chromium, line 2764: Maybe
```suggestion
        ?referenceContext: BrowsingContext
```

sadym-chromium, line 2764: Maybe like this to avoid confusion? E.g. users can expect the given context to be replaced with the new one.
```suggestion
        ?referenceContext: BrowsingContext
```

index.bs
```diff
@@ -2780,26 +2781,46 @@ The [=remote end steps=] with |command parameters| are:
...
-     be opened in?
+     context. Which OS window the new [=/browsing context=] is created in
+     depends on |type| and |reference context|:
+
+     * If |type| is "<code>tab</code>" and the implementation supports
+       multiple browsing contexts in the same OS window:
+
+       * The new browsing context should reuse an existing OS window, if any.
+
+       * If |reference context| is not null, the new browsing context should
```
sadym-chromium, line 2812: Should we add `if implementation supports it` disclaimer here as well?

sadym-chromium (approved):
Couple of minor comments

whimboo:
@foolip would you have the time to take a look at this PR again?

foolip:
I'll come back to this soon.

index.bs
foolip: I don't think so, because this is a sub-bullet of 'If _type_ is "<code>tab</code>" and the implementation supports multiple browsing contexts in the same OS window'.

foolip:
I'll go ahead and merge, or I'm sure to forget about this for another month. If I got something wrong, I'll make it a priority to fix it, just poke me.

whimboo:
@foolip actually such a change would have required a wpt test addition as well. Would you have the chance to follow-up with that?

foolip:
@whimboo yes, I've asked @sadym-chromium to add a test when he implements this.

whimboo:
Sounds good. So whoever is first with the implementation might do it. We can let you know in case we start to implement at first. Please do the same. Thanks.

whimboo:
@sadym-chromium actually yesterday I checked if we already have a bug and found https://bugzilla.mozilla.org/show_bug.cgi?id=1765619. There are already wdspec tests attached. So maybe @juliandescottes could finish the set of patches and we can land it first.

juliandescottes:
> would be great, thanks!
> 
> Actually, I'm not sure if there is a way to test if the parameter actually works, only the signature.

Yes we had the same issue. I only added tests for the signature at https://phabricator.services.mozilla.com/D143653 , and I also added some gecko specific tests at https://phabricator.services.mozilla.com/D143654, but you need to assume a specific implementation in order to "test" the feature.

My patches were written before the PR got its last few updates, so I guess they are a bit outdated. But I can update them next week.

PR #154, opened 2021-12-02
Line number of stack frame needs to be zero-based

whimboo:
***
<a href="https://pr-preview.s3.amazonaws.com/whimboo/webdriver-bidi/pull/154.html" title="Last updated on Dec 2, 2021, 9:03 AM UTC (538c8cd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/154/f5ce465...whimboo:538c8cd.html" title="Last updated on Dec 2, 2021, 9:03 AM UTC (538c8cd)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3065,7 +3065,7 @@ frame</dfn> corresponding to execution of a statement or expression in a script
  <dt><dfn for=stackframe>function</dfn>
  <dd>The name of the function being executed
  <dt><dfn for=stackframe>line number</dfn>
- <dd>The one-based line number of the executed code, relative to the top
+ <dd>The zero-based line number of the executed code, relative to the top
  of the resource containing |script|.
  <dt><dfn for=stackframe>column number</dfn>
  <dd>The zero-based column number of the executed code, relative to the start of


foolip:
What's the precedent from existing debugging protocols on this?

jgraham:
CDP is [zero based for everything](https://chromedevtools.github.io/devtools-protocol/tot/Runtime/#type-CallFrame). 

HTML error reporting [seems to be](https://software.hixie.ch/utilities/js/live-dom-viewer/saved/9864) seems to be different per browser: In Firefox I get one-based line numbers, and  zero based for column numbers assuming that it's using the position right before"new" as the column. In Chrome the line numbers seem to be zero based and the column numbers seem to be one-based assuming it's using the position right before "throw" as the column.

Puppeteer seems to use the CDP values directly.

whimboo:
Sorry, for not having added the details why I'm proposing the change. I was side-tracked until now. So thanks a lot @jgraham for doing it.

whimboo:
@foolip did the reply from @jgraham help to understand the current situation? Is there something else that we need to make progress on this particular change? Currently I'm implementing stack handling for `log.entryAdded` for Firefox, so adding the right set of WebDriver tests would be good. 

Please let us know if something is missing or unclear. Thanks a lot.

foolip:
Alright, let's go with zero-based!

foolip (approved):

PR #155, opened 2021-12-03
Add some security checks when handling a websocket connection.

jgraham:
Check the Host and Origin headers for the incoming connection to
verify the connection is allowed.

The language is intended to allow the specific behaviour to be largely
implementation defined, whilst recommending a default
behaviour that prevents CSRF-type attacks (reject host headers that
aren't an IP address or the server hostname, reject any requests with
an origin header).

Hopefully adding this text will ensure that implementations consider
the security issues accepting a connection, even though it's not
possible to give precise requirements that apply to all
implementations.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -552,11 +552,37 @@ implementation must proceed according to the WebSocket [=server-side
 requirements=], with the following steps run when deciding whether to
 accept the incoming connection:
 
-1. Let |resource name| be the resource name from [=reading the
-   client's opening handshake=]. If |resource name| is not in
-   |listener|'s [=list of WebSocket resources=], then stop
-   running these steps and act as if the requested service is not
+1. Using the data from [=reading the client's opening handshake=]:
+
+  1. Let |resource name| be the resource name.
+
+  1. Let |host| be the value of the Host header, if any, or undefined otherwise.
+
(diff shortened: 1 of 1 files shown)


jgraham:
I made an equivalent HTTP PR at https://github.com/w3c/webdriver/pull/1634 I think that one might have better wording, but I'll leave this as is until one or the other is reviewed to see what people prefer.

whimboo:
CC @sadym-chromium, @bwalderman

index.bs
```diff
@@ -552,12 +552,38 @@ implementation must proceed according to the WebSocket [=server-side
...
+   hostname of the WebSocket server or another hostname the implementation has
+   been configured to allow, or in any case if the implementation wants to
+   reject connections with |host| as the Host header, then stop running these
+   steps and act as if the requested service is not available.
+
+   Note: rejecting connections with unexpected values in the Host header
+   prevents DNS rebinding attacks. Implementations can opt to provide more
+   stringent controls where appropriate, for example only accepting connections
+   when the |host| value corresponds to a loopback interface [[!RFC5735]].
+
```
sadym-chromium, line 578: I'd add a note about Docker edge case here as well: https://github.com/w3c/webdriver/pull/1634#discussion_r762931438

PR #158, opened 2021-12-08
Support sandboxed script execution

jgraham:
Allow running scripts in a sandbox environment, such that they have access to the DOM of  page, but don't share js-level state with the page.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4029,6 +4029,63 @@ Issue: We might want to have a more sophisticated filter system than just a
 
 ### Events ### {#module-script-events}
 
+#### The script.postMessage Event #### {#event-script-postMessage}
+
+<dl>
+   <dt>Event Type</dt>
+   <dd>
+      <pre class="cddl local-cddl">
+        ScriptPostMessageEvent = {
+         method: "script.postMessage",
+         params: MessageInfo
+       }
+
+       MessageInfo = {
+         realm: Realm,
+         message: RemoteValue
+      }
+      </pre>
+   </dd>
+</dl>
+
+<div algorithm>
+The [=remote end event trigger=] is the <dfn>script postMessage</dfn> steps given |message| are:
+
+1. Let |serialize result| be the result of [=serialize as a remote value=] given
+   |message|.
+
+1. If |serialize result| is an error, then throw a [=TypeError=] exception and return.
+
+1. Let |serialized message| be |serialize result|'s [=data=].
+
+1. Let |related browsing contexts| be a set containing the
+   {{SandboxGlobalScope}}'s [=asssociated browsing context=].
+
+1. Let |realm| be the {{SandboxGlobalScope}}'s [=relevant realm=].
+
+1. [=Emit an postMessage event=] with |realm|, |serialized message| and |related
+   browsing contexts|.
+
+</div>
+
+<div algorithm>
+To <dfn>emit a postMessage event</dfn> given |realm|, |remote value|, and |related
+browsing contexts|:
+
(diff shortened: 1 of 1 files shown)


sadym-chromium:
Started discussion offline:
Should the `ObjectID`s for JS Node wrapper objects (and other DOM objects) be consistent in different sandboxes? 
* yes, as user can use those nodes both in page realm and in sandbox.
* no, as the JS wrapper objects are different and pretending they are the same (share the same objectId) seems odd.

Potential solution for that would be excluding DOM instances from the serialization and consider serialization only from the ECMAScript realm perspective. Having that, another way to describe the node is needed.

Another solution is to add something like `domID` to DOM objects so that they can be referenced when needed across sandboxes.

Implementation-wise in CDP we face the need of somehow merging domains (Runtime (V8) runs `evaluate` but cannot serialize the result, as it doesn't know of DOM).

Having that, we'd consider split DOM instances from it's JS wrappers, leaving the last in default serialization. And introducing DOM-specific serialization like `describeDomObjects` or `decribeNodes`.

jgraham:
I created https://github.com/w3c/webdriver-bidi/issues/176 for the discussion in that comment.

index.bs
```diff
@@ -932,6 +932,159 @@ requests to start a WebDriver session, it must:
...
+## Sandbox Realms ## {#sandbox-realm}
+
+Each sandbox is a unique EMCAScript [=Realm=]. However the sandbox realm
+provides access to platform objects in an existing {{Window}} realm via
+{{SandboxProxy}} objects.
+
+<div algorithm>
+To <dfn>get or create a sandbox realm</dfn> given |name| and |source browsing context|:
+
+1. If [=context sandbox map=] does not contain |name|, set |source browsing
```
bwalderman, line 967: Looks like this algorithm doesn't quite align with the type of `context sandbox map` described above. Should this step check if `context sandbox map` contains `source browsing context` instead of `name` since it is a weak map whose keys are browsing contexts?

index.bs
```diff
@@ -932,6 +932,159 @@ requests to start a WebDriver session, it must:
...
+
+1. Let |environment settings| be an [=environment settings object=] whose
+   algorithms are defined as follows:
+
+   <dl>
+       <dt>The [=realm execution context=]
+       <dd>
+
+       <dt>The [=module map=]
+       <dd>Issue: Not sure if this ought to be the module map of the assocaited
```
bwalderman, line 997: I suspect this should be an empty module map. If I'm understanding correctly, using the document's module map would mean that any names exported by module scripts on the page are now accessible in the sandbox if it imports the modules. This would break the isolation between the two realms.

bwalderman (changes requested):

foolip:
@jgraham this is in draft still, do you want review, or what's needed to unblock it?

index.bs
```diff
@@ -932,6 +932,159 @@ requests to start a WebDriver session, it must:
...
+1. If |wrapper property| is not <code>undefined</code>, return |wrapper
+   property|.
+
+1. Let |wrapped| be |this|.\[[Wrapped]].
+
+1. If |wrapped| is a [=platform object=]:
+
+  1. If |P| is a [=member=] of an [=interface=] implemented by |wrapped|:
+
+    1. If |P| is an [=atrthtmlibute=], let |getter| be the [=attribute getter=]
```
foolip, line 1073: Is "atrthtmlibute" a typo?

juliandescottes:
> @jgraham this is in draft still, do you want review, or what's needed to unblock it?

It seems that part of this proposal was moved to https://github.com/w3c/webdriver-bidi/pull/169 ?

PR #162, opened 2021-12-15
`SessionSubscribeParameters` -> `SubscribeParameters`

sadym-chromium:
Rename `SessionSubscribeParameters` -> `SubscribeParameters` to keep consistency.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/162.html" title="Last updated on Dec 15, 2021, 3:48 PM UTC (ee78bba)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/162/6b23251...ee78bba.html" title="Last updated on Dec 15, 2021, 3:48 PM UTC (ee78bba)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2136,10 +2136,10 @@ Issue: This needs to be generalized to work with realms too
       <pre class="cddl remote-cddl">
       SessionSubscribeCommand = {
         method: "session.subscribe",
-        params: SubscribeParameters
+        params: SessionSubscribeParameters
       }
 
-      SubscribeParameters = {
+      SessionSubscribeParameters = {
         events: [*text],
         ?contexts: [*BrowsingContext],
       }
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2139,7 +2139,7 @@ Issue: This needs to be generalized to work with realms too
         params: SubscribeParameters
       }
 
-      SessionSubscribeParameters = {
+      SubscribeParameters = {
```
whimboo, line 2142: I wonder how it would look like when several modules would have a `subscribe` command that need parameters. Maybe then it would make sense to prefix with the module name?

`script.callFunction` has btw the same issue.

sadym-chromium: Agree. Module prefix is a good idea. Renamed to 

Issue with `script.callFunction`, do you mean `ArgumentValue`, which is too vogue and can interfere with other modules?

Issue #146, opened 2021-11-19
"log.entryAdded" event misses reference to browsing context or realm in its payload


whimboo:
Events for `log.entryAdded` currently only contain the actual data related to the log entry, but clearly miss their origin. Like say executing a `console.log('foo')` script across different browsing contexts doesn't allow to differentiate the individual log events when having subscribed to the event globally. That makes it impossible for the client to filter out and assign these entries. 

As such I would propose to get the browsing context id, or a realm id added to the payload of the log event.

whimboo:
@CC @jgraham

whimboo:
See also https://github.com/w3c/webdriver-bidi/pull/73#discussion_r564461083 why a browsing context id hasn't been added (yet). I'll let @jgraham decide here.

bwalderman:
Looks like the original reason stated in [the comment from #73](https://github.com/w3c/webdriver-bidi/pull/73#discussion_r564461083) was because it wasn't yet clear what a "context" property should look like. It could be either a browsing context, or a realm in cases where there is no browsing context (e.g. a ServiceWorker).

Since that comment was posted, we've defined the [script.Target](https://w3c.github.io/webdriver-bidi/#type-script-Target) type which could fit this role. The `log.entryAdded` event could provide an origin" property of type Target that has a `context` when it comes from a context, or a `realm` otherwise.

whimboo:
Note that PR #167 will add an `origin` property to the `RealmInfo` type which then could be used here.

And just noticed that for `JavascriptLogEntry` entries we do not have a `realm` set because it's only available for `ConsoleLogEntry`. Should we maybe move the `realm` property to the `BaseLogEntry`?

whimboo:
There is now the question if we could re-use the `Target` type from the script module for the purpose of telling the client which realm or browsing context a given log message is originating from. Then there is also the optional `sandbox` parameter for this type, which actually might be helpful as well if a script running inside a sandbox logs these events.

If we want to allow the additional `sandbox` information within the event's payload, should we re-use this type, and add a new `target` property to the `Log.BaseLogEntry` type?

whimboo:
This was fixed with PR #209.

PR #167, opened 2021-12-27
Additional info for specific Realm types

bwalderman:
This changed restructures the `RealmInfo` type definition to allow us to add extra properties describing certain realm types. The first such addition is a `context` property for Window realms. This is important to allow clients to match realms discovered through `script.realmCreated` events with their corresponding browsing contexts.

**Motivation**

CDP exposes this information in `Runtime.executionContextCreated` events, through the `"frameId"` property:

```
puppeteer:protocol:RECV ◀ {"method":"Runtime.executionContextCreated","params":{"context":{"id":1,"origin":"://","name":"","uniqueId":"8785519071475702595.5679421674668736962","auxData":{"isDefault":true,"type":"default","frameId":"B6B81A7C8BB83D58F3DBAD1834FFD825"}}},"sessionId":"0DCC8B2872D54EB59741FC80EC3FE398"} +2ms
```

Puppeteer uses this information to associate the execution context with the correct frame so that it can present a single unified `Page` object to users and avoid needing to expose separate user-facing concepts for browsing contexts and script execution.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/167.html" title="Last updated on Apr 29, 2022, 1:49 PM UTC (e7b3afc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/167/ed5198c...e7b3afc.html" title="Last updated on Apr 29, 2022, 1:49 PM UTC (e7b3afc)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3419,7 +3419,6 @@ RealmInfo = {
   PaintWorkletRealmInfo //
   AudioWorkletRealmInfo //
   WorkletRealmInfo //
-  GenericRealmInfo
 }
 
 BaseRealmInfo = {
@@ -3466,11 +3465,6 @@ AudioWorkletRealmInfo = {
 WorkletRealmInfo = {
   BaseRealmInfo,
   type: "worklet"
-}
-
-GenericRealmInfo = {
-  BaseRealmInfo,
-  type: text
 }
 </pre>
 
@@ -3544,9 +3538,7 @@ To <dfn>get the realm info</dfn> given |environment settings|:
 
    <dt>Otherwise:
    <dd>
-      1. Let |realm info| be a map matching the <code>GenericRealmInfo</code> production,
-        with the <code>realm</code> field set to |realm id|, the <code>type</code>
-        field set to "<code>unknown</code>" and the <code>origin</code> field set to |origin|.
+      1. Let |realm info| be null.
   </dl>
 
 1. Return |realm info|
--- a/index.bs
+++ b/index.bs
@@ -3411,12 +3411,67 @@ Issue: This has the wrong error code
 
 <pre class="cddl local-cddl">
 RealmInfo = {
+  WindowRealmInfo //
+  DedicatedWorkerRealmInfo //
+  SharedWorkerRealmInfo //
+  ServiceWorkerRealmInfo //
+  WorkerRealmInfo //
+  PaintWorkletRealmInfo //
+  AudioWorkletRealmInfo //
+  WorkletRealmInfo //
+  GenericRealmInfo
+}
+
+BaseRealmInfo = {
   realm: Realm,
-  type: RealmType,
   origin: text
 }
 
-RealmType = "window" / "dedicated-worker" / "shared-worker" / "service-worker" / "worker" / "paint-worklet" / "audio-worklet" / "worklet" / text
+WindowRealmInfo = {
+  BaseRealmInfo,
+  type: "window",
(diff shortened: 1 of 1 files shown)


sadym-chromium:
Sorry for delay, I need a bit more time to review this.

index.bs
```diff
@@ -3303,58 +3358,72 @@ To <dfn>get the realm info</dfn> given |environment settings|:
 
 1. Let |realm id| be the [=realm id=] for |realm|.
 
+1. Let |origin| be the [=serialization of an origin=] given |environment settings|'s |origin|.
+
 1. Run the steps under the first matching condition:
 
   <dl>
     <dt>The [=Realm/global object=] specified by |environment settings| is a [=Window=] object
```
sadym-chromium, line 3366: Looks like the question is out of the scope of the PR, but still I ask: what the check if `smth is a Window object` will look like? Should it be specified based on some internal fields or something like that?

sadym-chromium, line 3366: Maybe extract `The [=Realm/global object=] specified by |environment settings|` into a variable to avoid redundancy?

sadym-chromium (approved):
couple of comments almost out of the scope of the given PR

index.bs
```diff
@@ -3303,58 +3358,74 @@ To <dfn>get the realm info</dfn> given |environment settings|:
...
    <dd>
-      1. Let |type| be "<code>worklet</code>".
+      1. Let |realm info| be a map matching the <code>WorkletRealmInfo</code> production,
+        with the <code>realm</code> field set to |realm id|, the <code>type</code>
+        field set to "<code>worklet</code>" and the <code>origin</code> field set to |origin|.
 
    <dt>Otherwise:
    <dd>
-      1. Return null.
+      1. Let |realm info| be a map matching the <code>GenericRealmInfo</code> production,
```
jgraham, line 3422: I'm a bit worried about this. For example do we end up with an event for extension content script realms? It makes more sense to me to specifcally have a list of supported realm types.

bwalderman: Removing GenericRealmInfo.

index.bs
```diff
@@ -3428,58 +3477,72 @@ To <dfn>get the realm info</dfn> given |environment settings|:
...
     <dd>
-     1. Let |type| be "<code>window</code>".
-
-    <dt>The [=Realm/global object=] specified by |environment settings| is a {{DedicatedWorkerGlobalScope}} object
+      1. Let |document| be |environment settings|' [=responsible document=].
+      1. Let |browsing context| be |document|'s [=browsing context=].
+      1. Let |context id| be the [=browsing context id=] for |browsing context|.
+      1. Let |realm info| be a map matching the <code>WindowRealmInfo</code> production,
+        with the <code>realm</code> field set to |realm id|, the <code>type</code>
+        field set to "<code>window</code>", the <code>origin</code> field set to |origin|,
```
jgraham, line 3494: In other cases we haven't re-specified the value of fully constrained fields (i.e. ones whose value is a single literal) when constructing the object.

whimboo:
@bwalderman as it looks like this PR seems to be in a good shape and close to be done. Would you have the time to finish the remaining request? I would love to re-use parts of it for the `context` reference of `log.entryAdded` events (issue #146). Thanks.

whimboo:
With the latest changes from @jgraham that looks fine to me.

PR #169, opened 2022-01-12
Add support for running script inside a sandbox

jgraham:
This adds the API surface for running script inside a sandbox whilst omitting all detail about how a sandbox might work.

The API model is that when executing script in a context you can provide a sandbox name, which will cause a sandbox with that name to be used, creating it if necessary. If you already created the sandobx, you can also use the realm id directly.

This currently doesn't add any addition functionality inside the sandbox except for the sandboxing.


index.bs
```diff
@@ -933,6 +933,91 @@ requests to start a WebDriver session, it must:
...
+</div>
+
+## Sandbox Proxy Objects ## {#sandbox-proxy}
+
+A <dfn interface>SandboxProxy</dfn> object is an exotic object that mediates sandboxed access to
+objects from another realm. Sandbox proxy objects are designed to enforce the
+following restrictions:
+
+* Platform objects are accessible, but property access returns only
+  WebIDL-defined properties and not ECMAScript defined properties (either
```
foolip, line 996:
```suggestion
  Web IDL-defined properties and not ECMAScript-defined properties (either
```

index.bs
```diff
@@ -933,6 +933,91 @@ requests to start a WebDriver session, it must:
...
+  "expando" properties that are not present in the underlying interface, or
+  ECMAScript defined properties that shadow a property in the underlying
```
foolip, line 998:
```suggestion
  ECMAScript-defined properties that shadow a property in the underlying
```

foolip (approved):

index.bs
```diff
@@ -933,6 +933,91 @@ requests to start a WebDriver session, it must:
...
+  are not accessible, but interface members can be used as normal.
+
+There is no {{SandboxProxy}} interface object.
+
+Issue: Define in detail how {{SandboxProxy}} works
+
+## SandboxWindowProxy ## {#sandbox-sandboxwindowproxy}
+
+<xmp class=idl>
+[Exposed=]
```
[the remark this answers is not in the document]
jgraham: I'm not sure. It isn't actually constructable from JS, or accessible from a Window global. But of course you can access as the global of a sandbox.

Issue #170, opened 2022-01-31
Specify closing a browsing context


whimboo:
Beside issue #116 (PR #133) we would also need a command to close a browsing context.

whimboo:
We should be clear about what should happen when the last browsing context gets closed. I assume it will be similar to the WebDriver HTTP specification, so that the browser quits and the session will be closed?

CC @jgraham

jgraham:
This also needs to fail if the context isn't a top level browsing context.

juliandescottes:
PR #172 was open to fix this, but there are two items we should discuss:
1. What should happen when we close the last tab/window:
    - explicitly close the browser and end the session (same as webdriver classic?)
    - leave the decision to the implementation
    - do nothing but expose another command to close the browser (Browser.close ?)
2. What should be the return value of browsingContext.close
    - Nothing
    - List or count of remaining top level browsing contexts
    - List or count of remaining windows

juliandescottes:
Another thing to consider: Puppeteer currently supports a `runBeforeUnload` parameter for Page.close: https://pptr.dev/#?product=Puppeteer&version=v1.20.0&show=api-pagecloseoptions

Internally it will switch between CDP's Page.close and Target.closeTarget depending on the value of this parameter, but maybe this is something we should handle directly in the BrowsingContext.close command for BiDi?

juliandescottes:
Quick note about CDP: testing CDP's Page.close, at least on Mac calling Page.close on the last tab doesn't close the Chrome application itself, although the behavior is probably different on Linux/Windows.

juliandescottes:
No strong opinions for the return value (but some options for closing the last tab/window might make use of the return value to communicate the behavior to the consumer).
Regarding onBeforeUnload, no strong opinions either, could be done differently if needed (re-reading the transcript, maybe it was not clear that the proposal was to have a flag to skip before unload promps, not to run code at beforeunload).

Most of the discussion revolved around the behavior when closing the last tab/window.

[...]

PR #172, opened 2022-02-04
Define the browsingContext.close command

juliandescottes:
Fixes https://github.com/w3c/webdriver-bidi/issues/170.

This adds the definition for browsingContext.close. I copied and modified existing browsingContext commands and tried to match https://w3c.github.io/webdriver/#close-window.

Opening the PR to get a first round of feedback, thanks in advance!
cc @jgraham  @whimboo

***
<a href="https://pr-preview.s3.amazonaws.com/juliandescottes/webdriver-bidi/pull/172.html" title="Last updated on Feb 23, 2022, 5:44 PM UTC (7e73690)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/172/8b12b43...juliandescottes:7e73690.html" title="Last updated on Feb 23, 2022, 5:44 PM UTC (7e73690)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -122,6 +122,7 @@ spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
 spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
   type: dfn
     text: a browsing context is discarded; url: window-object.html#a-browsing-context-is-discarded
+    text: close; url: window-object.html#close-a-browsing-context
     text: create a classic script; url: webappapis.html#creating-a-classic-script
     text: create a new browsing context; url: browsers.html#creating-a-new-browsing-context
     text: default classic script fetch options; url: webappapis.html#default-classic-script-fetch-options
@@ -2270,6 +2271,7 @@ navigation status</dfn> struct, which has the following items:
 <pre class="cddl remote-cddl">
 
 BrowsingContextCommand = (
+    BrowsingContextCloseCommand //
     BrowsingContextCreateCommand //
     BrowsingContextGetTreeCommand //
     BrowsingContextNavigateCommand //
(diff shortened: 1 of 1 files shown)


juliandescottes:
Also for now I left out the question about what to do when closing the last window.
I would first like to clean up the "straightforward" part of the PR from obvious mistakes before diving into this question.

index.bs
```diff
@@ -2536,6 +2539,59 @@ To <dfn>get the navigation info</dfn>, given |context| and |navigation status|:
...
+
+  1. Let |context id| be the value of the <code>context</code> field of
+     |command parameters|.
+
+  1. Let |context| be the result of [=trying=] to [=get a browsing context=]
+     with |context id|.
+
+  1. Assert: |context| is not null.
+
+  1. Assert: |context| is a [=top-level browsing context=]
```
jgraham, line 2582: You can't use "assert" like this; assert in this case means that there's a spec bug if you ever hit that condition. You need to explicitly handle the case by returning an error with crror code invalid argument.

index.bs
```diff
@@ -2536,6 +2539,59 @@ To <dfn>get the navigation info</dfn>, given |context| and |navigation status|:
...
+
+  <!-- XXX: This is step 2 from https://w3c.github.io/webdriver/#close-window,
```
jgraham, line 2584: BiDi doesn't have automatic prompt handling, so you can remove this check (I think in all modern implementations it should be fine to close a tab even if it has an alert showing).

index.bs
```diff
@@ -2536,6 +2539,59 @@ To <dfn>get the navigation info</dfn>, given |context| and |navigation status|:
...
+       but not sure if it needs more adaptation here? -->
+  1. [=Handle any user prompts=] and return its value if it is an [=error=].
+
+  1. [=Close=] |context|.
+
+  1. Return the result of running the [=remote end steps=] for the
```
jgraham, line 2590: Arguably we don't need to do this since a client could just use the relevant events to keep track of which contexts still exist.

jgraham (changes requested):

index.bs
juliandescottes: I see! Let me know if the new version works here.

index.bs
juliandescottes: Works for me, removed.

index.bs
juliandescottes: I just removed any return value, is that ok?

juliandescottes:
Thanks for getting to the review so quickly @jgraham!

index.bs
```diff
@@ -2536,6 +2539,45 @@ To <dfn>get the navigation info</dfn>, given |context| and |navigation status|:
...
+<div algorithm="remote end steps for browsingContext.close">
+The [=remote end steps=] with |command parameters| are:
+
+  1. Let |context id| be the value of the <code>context</code> field of
+     |command parameters|.
+
+  1. Let |context| be the result of [=trying=] to [=get a browsing context=]
+     with |context id|.
+
+  1. Assert: |context| is not null.
```
whimboo, line 2572: I can still see an assert here. I assume you missed that one to remove?

whimboo (changes requested):
I had a quick look and in general it looks fine. Still two comments inline which would require an update.

index.bs
juliandescottes: No I kept it intentionally. 

`Assert: |context| is not null.` is also used in other commands in the same file, eg `browsingContext.navigate`, so I assumed the comment was only about the second assert. I can remove it if it is unneeded, but will wait for confirmation.

whimboo: Maybe I also don't understand `Assert` then, but if we cannot get a browsing context from the `id` it's still an `invalid argument`, and should not trigger a bug in the spec.

jgraham: The use of "trying" and the fact that https://w3c.github.io/webdriver-bidi/#get-a-browsing-context returns `error` in the case that the context doesn't exist, means that we really can't get to this point without having a valid context. So the assert is fine.

juliandescottes:
> > Also for now I left out the question about what to do when closing the last window. I would first like to clean up the "straightforward" part of the PR from obvious mistakes before diving into this question.
> 
> Thanks for the updates @juliandescottes. So I assume this is left to do then.
> 
> Btw. this fixes issue #170.

Yes, I'll bring it up during our meeting and will update here afterwards. Also added the reference to #170 in the description, thanks for the reminder!

foolip (approved):

juliandescottes:
FWIW, `Issue(w3c/webdriver-bidi#170)` seems to generate a valid link

juliandescottes:
@foolip can you re-review since you reviewed the latest version before the update :) Thanks!

foolip (approved):

foolip:
@jgraham do you want to review again?

jgraham (approved):

PR #175, opened 2022-02-09
Add an input module with performActions and releaseActions commands

jgraham:
This adds support for user input, exactly following the actions model
from classic WebDriver. The user sends a list of input devices, and
the state change (if any) of each device in each time period
("tick"). The implementation then converts this into a list of ticks
and performs the relevant state updates in each time period.

By supporting multiple devices acting simultaneously it's possible to
model complex multi-device interactions e.g. multitouch gestures
combined with different key input states.

Compared with classic WebDriver, the main difference is that it's
possible to target the actions at any browsing context, not just the
active browsing context.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -239,6 +239,8 @@ Command = {
 CommandData = (
   SessionCommand //
   BrowsingContextCommand
+  ScriptCommand //
+  InputCommand
 )
 
 EmptyParams = { *text }
@@ -4284,6 +4286,243 @@ The [=remote end subscribe steps=], with [=subscribe priority=] 10, given
 
         1. If [=log event buffer=] contains |other context id|, remove |event|
            from [=log event buffer=][|other context id|].
+
+</div>
+
+## Input ## {#module-input}
+
+The <dfn export for=modules>input</dfn> module contains functionality for
+simulated user input.
+
+### Definition ### {#module-input-definition}
+
+[=remote end definition=]
+
+<pre class="cddl remote-cddl">
+
+InputCommand = (
+  InputPerformActionsCommand //
+  InputReleaseActionsCommand
+)
+</pre>
+
+### Commands ### {#module-input-commands}
+
+#### The input.performActions Command ####  {#command-input-performActions}
+
+The <dfn export for=commands>input.performActions</dfn> command performs a
+specified sequence of user input actions.
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl remote-cddl">
+      InputPerformActionsCommand = {
+        method: "input.performActions",
+        params: InputPerformActionsParameters
+      };
+
+      InputPerformActionsParameters = {
+        context: BrowsingContext,
+        actions: [* InputSourceActions]
+      };
+
+      InputSourceActions = (
+        NoneSourceActions //
+        KeySourceActions //
+        PointerSourceActions //
+        WheelSourceActions
+      );
+
+      NoneSourceActions = {
+        type: "none",
+        actions: [* NoneSourceAction]
+      };
+
+      NoneSourceAction = PauseAction;
+
+      KeySourceActions = {
+        type: "key",
+        actions: [* KeySourceAction]
+      };
+
+      KeySourceAction = (
+        PauseAction //
+        KeyDownAction //
+        KeyUpAction
+      );
+
+      PointerSourceActions = {
+        type: "pointer",
+        ? parameters: PointerParameters,
+        actions: [* PointerSourceAction]
+      };
+
+      PointerParameters = {
+        ? pointerType: "mouse" / "pen" / "touch"
+      };
+
+      PointerSourceAction = (
+        PauseAction //
+        PointerDownAction //
+        PointerUpAction //
+        PointerMoveAction
+      );
+
+      WheelSourceActions = {
+        type: "wheel",
+        actions: [* WheelSourceAction]
+      };
+
+      WheelSourceActions = (
+        PauseAction //
+        WheelScrollAction
+      );
+
+      PauseAction = {
+        type: "pause",
+        ? duration: uint
+      };
+
+      KeyDownAction = {
+        type: "keyDown",
+        value: text
+      };
+
+      KeyUpAction = {
+        type: "keyUp",
+        value: text
+      };
+
+      PointerUpAction = {
+        type: "pointerUp",
+        button: uint,
+        PointerCommonProperties,
+      };
+
+      PointerDownAction = {
+        type: "pointerUp",
+        button: uint,
+        PointerCommonProperties
+      };
+
+      PointerMoveAction = {
+        type: "pointerMove",
+        x: uint,
+        y: uint,
+        ? duration: uint,
+        ? origin: Origin,
+        PointerCommonProperties,
+      };
+
+      WheelScrollAction = {
+        type: "scroll",
+        x: uint,
+        y: unint,
+        deltaX: int,
+        deltaY: int,
+        ? duration: uint,
+        ? origin: Origin,
+        PointerCommonProperties,
+      };
+
+      PointerCommonProperties = (
+        ? width: uint,
+        ? height: uint,
+        ? pressure: float,
+        ? tangentialPressure: float,
+        ? tiltX: int,
+        ? tiltY: int,
+        ? twist: uint,
+        ? altitudeAngle: float,
+        ? azimuthAngle: float,
+      )
+
+      Origin = "viewport" / "pointer" / ElementOrigin;
+
+      //TODO ElementOrigin
+    </pre>
+   </dd>
+   <dt>Return Type</dt>
+   <dd>
+    <pre class="cddl local-cddl">
+        EmptyResult
+    </pre>
+   </dd>
+</dl>
+
+<div algorithm="remote end steps for input.performActions">
+
+The [=remote end steps=] with |session| and |command parameters| are:
+
+1. Let |context id| be the value of the <code>context</code> field of
+   |command parameters|.
+
+1. Let |context| be the result of [=trying=] to [=get a browsing context=]
+   with |context id|.
+
+1. If |context| has an active [=user prompt=], return [=error=] with [=error
+   code=] [=unexpected alert open=].
+
+1. Let |actions by tick| be the result of trying to [=extract an action
+   sequence=] with |command parameters|.
+
+1. [=Try=] to [=dispatch actions=] with argument |actions by tick|.
+
+1. Return [=success=] with data null.
+
+</div>
+
+The <dfn export for=commands>input.releaseActions</dfn> command resets the input
+state associated with the current session.
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl remote-cddl">
+      InputReleaseActionsCommand = {
+        method: "input.releaseActions",
+        params: ReleaseActionsParameters
+      };
+
+      InputReleaseActionsParameters = {
+        context: BrowsingContext,
+      }
+    </pre>
+   </dd>
+   <dt>Return Type</dt>
+   <dd>
+    <pre class="cddl local-cddl">
+        EmptyResult
+    </pre>
+   </dd>
+</dl>
+
+<div algorithm="remote end steps for input.releaseActions">
+
+The [=remote end steps=] given <var ignore>session</var>, and <var
+ignore>command parameters</var> are:
+
+1. Let |context id| be the value of the <code>context</code> field of
+   |command parameters|.
+
+1. Let |context| be the result of [=trying=] to [=get a browsing context=]
+   with |context id|.
+
+1. If |context| has an active [=user prompt=], return [=error=] with [=error
+   code=] [=unexpected alert open=].
+
+1. Let |undo actions| be equal to the |session|’s [=input cancel list=] in reverse order.
+
+1. [=Dispatch tick actions=] with arguments |undo actions| and duration 0.
+
+1. Let |session|’s [=input cancel list=] be an empty List.
+
+1. Let the |session|’s [=input state table=] be an empty Map.
+
+1. Let |session|’s [=active input sources=] be an empty List.
+
+1. Return [=success=] with data null.
 
 </div>
 


jgraham:
This assumes the right model for input events is basically what's in WebDriver, rather than what's in e.g. CDP.

Assuming there's agreement on that point, it also depends on making some changes to the WebDriver spec to allow the existing actions dispatch steps take a target context rather than using the WebDriver "current browsing context". In addition the following questions are open:

* How to handle element origins (for e.g. `pointerMove`). For classic we do something like `{origin: {element_key: element_id}}` where `element_key` is the UUID-based key used to identify an object containing an element reference. But for BiDi it might make more sense to just allow `{origin: {type: "element", value: RemoteObjectReference}}` or something. But that makes the specification a bit more complex since we need to distinguish the BiDi case from the classic case when deserializing an origin. Alternatively we could just directly resuse the setup in WebDriver, but it would be quite ugly to have this magic key which is totally unnecessary and (unlike in classic) not consistent with the rest of the protocol.
* Are we happy with the alert handing? This is the first thing in BiDi that just fails if the context has an alert open. But I don't think it makes a lot of sense to try to dispatch actions in a context where the event loop is blocked by a modal. But we could, for example, just dispatch the actions after the alert has been clearerd.
* Is there any concern with focus? WebDriver supposedly allows dispatching actions to any context, but one could arrange for the current browsing context to always get focus, or at least be the active tab. That's not really possible with BiDi since you could send inputs to multiple tabs at a time. We are also sharing the input state across all navigables. That makes it possible to do crazy racy things like simultaneously send pointerMove events for the same pointer to multiple tabs, with different targets. Those will then fight each other for the actual pointer position.

sadym-chromium:
Current Chromium implementation sends input events to the active tab, so it would be great if BiDi keeps the same concept.

jgraham:
> Current Chromium implementation sends input events to the active tab

What does "active tab" mean, exactly? Does it mean that it has to have os-level focus, or just be the top tab in a window? How does it interact with headless mode? It would seem pretty weird if you can send events to any tab in headless mode, but only to a single tab in non-headless mode.

sadym-chromium:
~~Current Chromium implementation sends input events to the active tab, so it would be great if BiDi keeps the same concept.~~
UPD: It was a bug in CDP and it was fixed. Now events can be sent to any tab by providing the corresponding target.

jgraham:
Some questions for discussion:
* Do we scope actions state per (top-level) browsing context or per session? Should it be possible to e.g. have the mouse pointer in different positions per context?
* At whatever scope actions are constrained to, do we need a lock to serialize actions? Without this if two actions commands are sent one after another, the second set of commands can run before the first set is complete, which would interfere with the state in an unpredictable way.

bwalderman:
> * Are we happy with the alert handing? This is the first thing in BiDi that just fails if the context has an alert open. But I don't think it makes a lot of sense to try to dispatch actions in a context where the event loop is blocked by a modal. But we could, for example, just dispatch the actions after the alert has been clearerd.

The new BiDi actions commands can run the [handle any user prompts](https://w3c.github.io/webdriver/#dfn-handle-any-user-prompts) steps like the classic commands do. Since we're considering queueing actions commands I think we would want to specify that user prompts are handled just before the actions are dispatched; not when the command is first processed.

bwalderman:
> Some questions for discussion:
> 
> * Do we scope actions state per (top-level) browsing context or per session? Should it be possible to e.g. have the mouse pointer in different positions per context?

> * At whatever scope actions are constrained to, do we need a lock to serialize actions? Without this if two actions commands are sent one after another, the second set of commands can run before the first set is complete, which would interfere with the state in an unpredictable way.

A queue seems like a good idea, so an actions command doesn't dispatch actions until all earlier actions commands have completed.

Actions state per session would more accurately represent how a real user device works. In the real world, input sources are associated with the device and not with individual browser tabs. I think it might lead to some complications though. For this to work, I believe we would need to either add a concept of active/focused top-level browsing context or let users specify the context as a parameter of the actions command. Assuming the client can switch contexts, what happens if the client presses down some keys in context A, switches to context B, and then releases the keys? Should context A get the keyup events so the page's state remains consistent? Should switching the active context implicitly release actions so this doesn't happen?

It might actually be simpler from an implementation perspective (at least for Chromium) to scope actions state per top-level browsing context. We'd still need to add a parameter to let the client choose a target context. However it would be more difficult for a client to put some context's page in an inconsistent state because each context would have its own set of input sources. This would also be the more powerful of the two approaches. Frameworks/tools that want to present a single session-wide set of input sources to the end user can still build that view on top of this approach.

jgraham:
I created https://github.com/w3c/webdriver/pull/1653 to provide a solution to the problem of how to represent element-offset origins in BiDi without weirdly reintroducing the web element identifier for this one case.

jgraham:
https://github.com/w3c/webdriver/pull/1654 to make the state per top-level browsing context (and per session, see the commit message).

jgraham:
Updated this PR under the assumption that the webdriver PR will land.

index.bs
```diff
@@ -4437,6 +4439,252 @@ The [=remote end subscribe steps=], with [=subscribe priority=] 10, given
...
+   </dd>
+   <dt>Return Type</dt>
+   <dd>
+    <pre class="cddl local-cddl">
+        EmptyResult
+    </pre>
+   </dd>
+</dl>
+
+A map matching the <code>ElementOrigin</code> production [=represents a web
```
foolip, line 4607: The "represents a web element" term ends up not linked to anything. I guess it should be https://w3c.github.io/webdriver/#dfn-represents-a-web-element, but that doesn't quite make sense to me.

Should this say that an `Element`-relative origin is represented by `ElementOrigin`? (I don't think we need to say "matching a ... production" except where creating these structures before sending them in a response.)

index.bs
```diff
@@ -4437,6 +4439,252 @@ The [=remote end subscribe steps=], with [=subscribe priority=] 10, given
...
+
+      InputSourceActions = (
+        NoneSourceActions //
+        KeySourceActions //
+        PointerSourceActions //
+        WheelSourceActions
+      );
+
+      NoneSourceActions = {
+        type: "none",
```
foolip, line 4488: What are all the "none" bits used for? They only appear in the CDDL so far. Related to https://w3c.github.io/webdriver/#null-input-source?

index.bs
```diff
@@ -4437,6 +4439,252 @@ The [=remote end subscribe steps=], with [=subscribe priority=] 10, given
...
+    </pre>
+   </dd>
+</dl>
+
+A map matching the <code>ElementOrigin</code> production [=represents a web
+element=].
+
+<div algorithm>
+
+To <dfn>get a known element</a> for an <code>ElementOrigin</code>:
```
foolip, line 4612: Add a TODO here so it doesn't look like the following remote end steps are related?

index.bs
```diff
@@ -4437,6 +4439,252 @@ The [=remote end subscribe steps=], with [=subscribe priority=] 10, given
...
+1. Let |actions by tick| be the result of trying to [=extract an action
+   sequence=] with |input state| and |command parameters|.
+
+1. [=Try=] to [=dispatch actions=] with argument |actions by tick| and |context|.
+
+1. Return [=success=] with data null.
+
+</div>
+
+The <dfn export for=commands>input.releaseActions</dfn> command resets the input
```
foolip, line 4638: A new section for input.releaseActions?

index.bs
```diff
@@ -4437,6 +4439,252 @@ The [=remote end subscribe steps=], with [=subscribe priority=] 10, given
...
+1. Let |context id| be the value of the <code>context</code> field of
+   |command parameters|.
+
+1. Let |context| be the result of [=trying=] to [=get a browsing context=]
+   with |context id|.
+
+1. Let |input state| be [=get the input state=] with |session| and |context|'s
+   [=top-level browsing context=].
+
+1. Let |actions by tick| be the result of trying to [=extract an action
```
foolip, line 4629: "extract an action sequence" isn't linked here and is doing a lot of work.

index.bs
```diff
@@ -4437,6 +4439,252 @@ The [=remote end subscribe steps=], with [=subscribe priority=] 10, given
...
+   sequence=] with |input state| and |command parameters|.
+
+1. [=Try=] to [=dispatch actions=] with argument |actions by tick| and |context|.
```
foolip, line 4632: Like WebDriver classic, I guess this will wait until the action sequence is done? Is that the model we want here too, or is there any value in returning fast and using events to communicate progress?

index.bs
jgraham: The broken link notwithstanding, this is not needed if we merge https://github.com/w3c/webdriver/pull/1653. The "problem" is that we don't want to copy all the actions text from WebDriver, we want to reuse it. But WebDriver defines an object as "representing a web element" when it has the magic property  "<code>element-6066-11e4-a52e-4f735466cecf</code>". That's an ugly hack that we don't need in BiDi where we already have a uniform JS value representation. So the problem we need to solve is to allow the BiDi representation of an element without replicating the entire actions section just to change one definition. The approach here was to say that for a BiDi implementation anything matching `ElementOrigin` is also something that "represents a web element" (assuming you implement BiDi), but the approach in the WebDriver PR is just to make the BiDi representation always work for the `origin` property in actions, even if called from HTTP. The latter is somewhat simpler at the cost of (theoretically at least) adding new requirements on existing implementations.

index.bs
jgraham: All this CDDL is just representing the data format that's already used in WebDriver where it's checked manually.

index.bs
jgraham: I think we should wait. The client of course is free to not wait for the actions to complete if that's what they want.

jgraham:
This depends on changes in  https://github.com/w3c/webdriver/pull/1654 and https://github.com/w3c/webdriver/pull/1653

index.bs
```diff
@@ -4654,6 +4662,245 @@ The [=remote end subscribe steps=], with [=subscribe priority=] 10, given
...
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl remote-cddl">
+      InputReleaseActionsCommand = {
+        method: "input.releaseActions",
+        params: ReleaseActionsParameters
+      };
+
+      InputReleaseActionsParameters = {
```
sadym-chromium, line 4866:
```suggestion
      ReleaseActionsParameters = {
```

sadym-chromium (approved):
I like the fact WebDriver Classic can be referenced for the action specification.

index.bs
```diff
@@ -4654,6 +4662,245 @@ The [=remote end subscribe steps=], with [=subscribe priority=] 10, given
...
+      NoneSourceActions = {
+        type: "none",
+        actions: [* NoneSourceAction]
+      };
+
+      NoneSourceAction = PauseAction;
+
+      KeySourceActions = {
+        type: "key",
+        actions: [* KeySourceAction]
```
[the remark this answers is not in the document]
jgraham: I'm not sure I understand the question.

The model in WebDriver is that you have a set of virtual inputs, each of which is a given type. And in each actions sequence you define what each input is doing at each time (either "pause" which means its state remains unchanged or one of the type-specific actions).

The CDDL here should just be directly recreating the WebDriver model.

index.bs
```diff
@@ -4654,6 +4662,245 @@ The [=remote end subscribe steps=], with [=subscribe priority=] 10, given
...
+      PointerCommonProperties = (
+        ? width: uint,
+        ? height: uint,
+        ? pressure: float,
+        ? tangentialPressure: float,
+        ? tiltX: int,
+        ? tiltY: int,
+        ? twist: uint,
+        ? altitudeAngle: float,
+        ? azimuthAngle: float,
```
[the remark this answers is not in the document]
jgraham: That is a generic problem with input actions; we want the browser to behave as it does with real inputs, but depend on other specifications for how it will react to real inputs. So we can write tests that will make everyone happy as long as everyone agrees on what the behaviour of the browser given a specific input is. But we don't define that here.

index.bs
```diff
@@ -4654,6 +4662,245 @@ The [=remote end subscribe steps=], with [=subscribe priority=] 10, given
...
+The <dfn export for=commands>input.releaseActions</dfn> command resets the input
+state associated with the current session.
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl remote-cddl">
+      InputReleaseActionsCommand = {
+        method: "input.releaseActions",
+        params: ReleaseActionsParameters
```
jgraham, line 4863:
```suggestion
        params: InputReleaseActionsParameters
```

index.bs
jgraham: I think we're converging on prefixing CDDL productions with the module name.

index.bs
```diff
@@ -4654,6 +4662,245 @@ The [=remote end subscribe steps=], with [=subscribe priority=] 10, given
...
+      };
+
+      PointerUpAction = {
+        type: "pointerUp",
+        button: uint,
+        PointerCommonProperties,
+      };
+
+      PointerDownAction = {
+        type: "pointerUp",
```
jimevans, line 4777: Should this not be `pointerDown`?

jimevans (changes requested):

index.bs
jgraham: Good catch, thanks!

mathiasbynens:
In combination with https://github.com/w3c/webdriver/pull/1653, this would help us get WebDriver Classic-style input actions in BiDi.

whimboo:
@jgraham the CDDL validation failed. Maybe you can rebase and fix this issue?

whimboo (changes requested):

index.bs
```diff
@@ -5667,6 +5678,303 @@ The [=remote end subscribe steps=], with [=subscribe priority=] 10, given
...
+### Types ### {#module-input-types}
+
+#### input.ElementOrigin ### {#module-input-origin}
+
+Tjhe <code>input.ElementOrigin</code> type represents an {{Element}} that will
+be used as a coordinate origin.
+
+<pre class="cddl remote-cddl">
+input.ElementOrigin = {
+  "type": "node",
```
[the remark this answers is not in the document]
jgraham: We check that the value matches this specific production, which it can't if the `type` is missing. So I don't quite understand the concern?

index.bs
```diff
@@ -5667,6 +5678,303 @@ The [=remote end subscribe steps=], with [=subscribe priority=] 10, given
...
+        params: input.PerformActionsParameters
+      };
+
+      input.PerformActionsParameters = {
+        context: browsingContext.BrowsingContext,
+        actions: [*input.SourceActions]
+      }
+
+      input.SourceActions = (
+        input.NoneSourceActions /
```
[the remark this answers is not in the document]
jgraham: I think we need to generally do a pass over the entire spec CDDL, so I suggest we leave this for now.

index.bs
```diff
@@ -5667,6 +5678,303 @@ The [=remote end subscribe steps=], with [=subscribe priority=] 10, given
...
+
+      input.WheelScrollAction = {
+        type: "scroll",
+        x: js-uint,
+        y: js-uint,
+        deltaX: js-int,
+        deltaY: js-int,
+        ? duration: js-uint,
+        ? origin: input.Origin,
+        input.PointerCommonProperties,
```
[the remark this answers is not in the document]
jgraham: This was a copy/paste error.

ollie-iterators:
"LINE ~5757: The var 'session' (in algorithm 'get an Element from input.ElementOrigin') is only used once." is confusing, as line 5757 appears to be "   <dt>Command Type</dt>"

jgraham:
To fix the build failure (and a bug in the spec) I needed to pass the session to the `get element origin` steps. To do that without making another patch to the classic WebDriver spec I turned those steps into a closure that capture the session. This is perhaps a bit odd, but I think works and avoids another large "pass the session through everything" change.

index.bs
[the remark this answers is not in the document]
whimboo: For `script.callFunction` we can pass-in any object that only needs `sharedId` to have set and we deserialize it as a `Node`. So the type there (`SharedReference`) is optional while here it is required.

index.bs
```diff
@@ -5685,6 +5696,311 @@ The [=remote end subscribe steps=], with [=subscribe priority=] 10, given
...
+  input.PerformActions //
+  input.ReleaseActions
+)
+</pre>
+
+### Types ### {#module-input-types}
+
+#### input.ElementOrigin #### {#type-input-origin}
+
+Tjhe <code>input.ElementOrigin</code> type represents an {{Element}} that will
```
whimboo, line 5720:
```suggestion
The <code>input.ElementOrigin</code> type represents an {{Element}} that will
```

index.bs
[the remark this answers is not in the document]
whimboo: Oh, and this probably needs to be `Extensible`? A client should not have to modify a node reference that it got back from eg `script.evaluate` before it can be used for an action.

index.bs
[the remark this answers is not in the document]
whimboo: Even if we have a wrong usage for other CDDL definitions I think that it should not stop us from using the correct one here?

index.bs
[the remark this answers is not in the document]
jgraham: I want `type` here so that in the future we have the option to extend to other origin types, without having to inspect more than the type field.

I'd happily make the same change for `SharedReference`.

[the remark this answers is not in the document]
whimboo: But isn't this `type` a problem when we deserialize a `SharedReference` then? Right now it's useless. Wouldn't it make more sense instead to make this change together with the SharedReference one in a follow-up PR?

ollie-iterators:
Is this going to help people perform actions like zooming in on a browser?

index.bs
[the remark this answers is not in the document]
jgraham: If we just have a `sharedId` field and we want to add another kind of element-offset origin in the future (e.g. one related to the top left of the element rather than the centre point), what happens? We can't just add a `type` field because that will be ignored by existing clients (assuming this is `Extensible`). So in that case we'd have to use a different field name to represent the `sharedId`. But of course that would break the property that you can just directly reuse an element that you got as a response as the origin. So I think fundamentally that naming variants gives us a much more extensible protocol, at the cost of clients needing some explicit typecasts.

index.bs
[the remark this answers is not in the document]
jgraham: I honestly don't know if it is correct. In practice the library doesn't seem to care much, but it depends somewhat on whether you think that something inside `{}` is a type or a group. Anyway I changed it for consistency.

index.bs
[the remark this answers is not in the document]
whimboo: @sadym-chromium do you still have objections here or can we close this particular conversation?

index.bs
```diff
@@ -5681,6 +5692,313 @@ The [=remote end subscribe steps=], with [=subscribe priority=] 10, given
...
+
+  1. Let |environment settings| be the [=environment settings object=] whose
+     [=relevant global object=]'s <a>associated <code>Document</code></a> is
+     |document|.
+
+  1. Let |realm| be |environment settings|' [=realm execution context=]'s
+     Realm component.
+
+  1. Let |element| be the result of [=trying=] to [=deserialize remote reference=]
+     with |reference|, |realm|, and |session|.
```
whimboo, line 5757: Note that this addition will conflict with the changes on https://github.com/w3c/webdriver-bidi/pull/385 so please do not forget to update whether which of the PRs are merged first.

index.bs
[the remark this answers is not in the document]
whimboo: Thanks for the update James! With the wrapping this looks fine to me now. 

@sadym-chromium maybe you can have a look at the recent changes as well?

whimboo (approved):
Thanks! Looks fine to me now. Might still be good to get a final feedback from @sadym-chromium.

index.bs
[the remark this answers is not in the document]
sadym-chromium: solved

index.bs
whimboo: I was actually wrong. In PR #385 we only made modifications to the serialization part and that is not affected here.

sadym-chromium:
LGTM

sadym-chromium (approved):

PR #180, opened 2022-03-10
`SharedReference` serialization + deserialization

sadym-chromium:
* Defined `SharedReference`;
* Described RemoteReference deserialization based on `handle` first and `sharedId` afterwards. 
* Define `Node` serialization.

(edit @whimboo) No longer blocked by https://github.com/w3c/webdriver/pull/1707

***
<a href="https://pr-preview.s3.amazonaws.com/sadym-chromium/webdriver-bidi/pull/180.html" title="Last updated on Feb 1, 2023, 12:16 PM UTC (62b0fbd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/180/dfd69fc...sadym-chromium:62b0fbd.html" title="Last updated on Feb 1, 2023, 12:16 PM UTC (62b0fbd)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1636,7 +1636,7 @@ To get the <dfn>handle for an object</dfn> given |realm|, |ownership type| and
 </div>
 
 <div algorithm>
-To get <dfn>shared id for a node</dfn> given |realm| and |node|:
+To get <dfn>shared id for a node</dfn> given |session|, |realm| and |node|:
 
 1. Let |node| be [=unwrapped=] |node|.
 
@@ -1646,8 +1646,8 @@ To get <dfn>shared id for a node</dfn> given |realm| and |node|:
 
 1. If |browsing context| is <code>null</code>, return <code>null</code>.
 
-1. Return [=get or create a node reference=] with [=current session=],
-   |browsing context| and |node|.
+1. Return [=get or create a node reference=] with |session|, |browsing context| and
+   |node|.
 
 </div>
 
@@ -1874,8 +1874,8 @@ Issue: handle String / Number / etc. wrapper objects specially?
 
 <div algorithm>
 
-To <dfn>serialize as a remote value</dfn> given a |value|, a |max depth|,
-an |ownership type|, a |serialization internal map| and a |realm|:
+To <dfn>serialize as a remote value</dfn> given a |session|, a |value|,
+a |max depth|, an |ownership type|, a |serialization internal map| and a |realm|:
 
 1. Let |remote value| be a result of [=serialize primitive protocol value=]
    given a |value|.
@@ -1900,7 +1900,7 @@ an |ownership type|, a |serialization internal map| and a |realm|:
         property set to |handle id| if it's not null, or omitted otherwise.
 
     <dt>[=IsArray=](|value|)
-    <dd>Let |remote value| be [=serialize an Array-like=] with
+    <dd>Let |remote value| be [=serialize an Array-like=] with |session|,
         <code>ArrayRemoteValue</code>, |handle id|, |known object|, |value|,
         |max depth|, |ownership type|, |serialization internal map|, and |realm|.
 
@@ -1946,8 +1946,8 @@ an |ownership type|, a |serialization internal map| and a |realm|:
     1. If |known object| is <code>false</code>, and |max depth| is not 0, run
        the following steps:
 
-       1. Let |serialized| be the result of [=serialize as a mapping=] given
-          [=CreateMapIterator=](|value|, key+value), |max depth|,
+       1. Let |serialized| be the result of [=serialize as a mapping=] with
+          |session|, [=CreateMapIterator=](|value|, key+value), |max depth|,
           |ownership type|, |serialization internal map| and |realm|.
 
     1. If |serialized| is not null, set field <code>value</code> of |remote value| to
@@ -1968,8 +1968,8 @@ an |ownership type|, a |serialization internal map| and a |realm|:
     1. If |known object| is <code>false</code>, and |max depth| is not 0, run
        the following steps:
 
-       1. Let |serialized| be the result of [=serialize as a list=] given
-          [=CreateSetIterator=](|value|, value), |max depth|, |ownership type|,
+       1. Let |serialized| be the result of [=serialize as a list=] with
+          |session|, [=CreateSetIterator=](|value|, value), |max depth|, |ownership type|,
           |serialization internal map| and |realm|.
 
     1. If |serialized| is not null, set field <code>value</code> of |remote value| to
@@ -2018,7 +2018,7 @@ an |ownership type|, a |serialization internal map| and a |realm|:
 
     <dt>|value| is a [=platform object=] that implements [=Node=]
     <dd>
-      1. Let |shared id| be [=shared id for a node=] given |realm| and
+      1. Let |shared id| be [=shared id for a node=] given |session|, |realm| and
          |value|.
 
       1. Let |remote value| be a map matching the <code>NodeRemoteValue</code>
@@ -2060,7 +2060,7 @@ an |ownership type|, a |serialization internal map| and a |realm|:
             1. Let |child depth| be |max depth| - 1 if |max depth| is not null, or null otherwise.
 
             1. Let |serialized| be the result of [=serialize as a remote value=]
-               with |child|, |child depth|, |ownership type|,
+               with |session|, |child|, |child depth|, |ownership type|,
                |serialization internal map| and |realm|.
 
             1. Append |serialized| to |children|.
@@ -2090,7 +2090,7 @@ an |ownership type|, a |serialization internal map| and a |realm|:
                   null, or null otherwise.
 
                1. Let |serialized shadow| be the result of
-                  [=serialize as a remote value=] with |shadow root|, |child depth|,
+                  [=serialize as a remote value=] with |session|, |shadow root|, |child depth|,
                   false, |ownership type|, |serialization internal map| and |realm|.
 
                 Note: this means the <code>handle</code> for the shadow root
@@ -2133,8 +2133,8 @@ an |ownership type|, a |serialization internal map| and a |realm|:
     1. If |known object| is <code>false</code>, and |max depth| is not 0, run
        the following steps:
 
-       1. Let |serialized| be the result of [=serialize as a mapping=] given
-          [=EnumerableOwnPropertyNames=](|value|, key+value), |max depth|,
+       1. Let |serialized| be the result of [=serialize as a mapping=] with
+          |session|, [=EnumerableOwnPropertyNames=](|value|, key+value), |max depth|,
           |ownership type|, |serialization internal map| and
           |realm|.
 
@@ -2160,7 +2160,7 @@ remove optional flag from the field.
 </div>
 
 <div algorithm>
-To <dfn>serialize an Array-like</dfn> given |production|, |handle id|, |known object|,
+To <dfn>serialize an Array-like</dfn> given |session|, |production|, |handle id|, |known object|,
 |value|, |max depth|, |ownership type|, |serialization internal map|, and |realm|:
 
 1. Let |remote value| be a map matching |production|, with the
@@ -2184,7 +2184,7 @@ To <dfn>serialize an Array-like</dfn> given |production|, |handle id|, |known ob
 </div>
 
 <div algorithm>
-To <dfn>serialize as a list</dfn> given |iterable|, |max depth|, |ownership type|,
+To <dfn>serialize as a list</dfn> given |session|, |iterable|, |max depth|, |ownership type|,
 |serialization internal map| and |realm|:
 
   1. Let |serialized| be a new list.
@@ -2195,7 +2195,7 @@ To <dfn>serialize as a list</dfn> given |iterable|, |max depth|, |ownership type
        otherwise.
 
     1. Let |serialized child| be the result of [=serialize as a remote value=]
-       with arguments |child value|, |child depth|, |ownership type|,
+       with |session|, |child value|, |child depth|, |ownership type|,
        |serialization internal map| and |realm|.
 
     1. Append |serialized child| to |serialized|.
@@ -2206,7 +2206,7 @@ To <dfn>serialize as a list</dfn> given |iterable|, |max depth|, |ownership type
 Issue: this assumes for-in works on iterators
 
 <div algorithm>
-To <dfn>serialize as a mapping</dfn> given |iterable|, |max depth|,
+To <dfn>serialize as a mapping</dfn> given |session|, |iterable|, |max depth|,
 |ownership type|, |serialization internal map| and |realm|:
 
 1. Let |serialized| be a new list.
@@ -2226,11 +2226,11 @@ To <dfn>serialize as a mapping</dfn> given |iterable|, |max depth|,
 
   1. If [=Type=](|key|) is String, let |serialized key| be |child key|,
      otherwise let |serialized key| be the result of [=serialize as a remote value=]
-     with arguments |child key|, |child depth|, |ownership type|,
+     with |session|, |child key|, |child depth|, |ownership type|,
      |serialization internal map| and |realm|.
 
   1. Let |serialized value| be the result of [=serialize as a remote value=]
-     with arguments |value|, |child depth|, |ownership type|,
+     with |session|, |value|, |child depth|, |ownership type|,
      |serialization internal map| and |realm|.
 
   1. Let |serialized child| be («|serialized key|, |serialized value|»).
@@ -4052,8 +4052,8 @@ The <code>script.ExceptionDetails</code> type represents a JavaScript exception.
 
 <div algorithm>
 
-To <dfn>get exception details</dfn> given a |realm|, a [=completion record=]
-|record| and an |ownership type|:
+To <dfn>get exception details</dfn> given a |session|, a |realm|,
+a [=completion record=] |record| and an |ownership type|:
 
 1. Assert: |record|.\[[Type]] is <code>throw</code>.
 
@@ -4063,7 +4063,7 @@ To <dfn>get exception details</dfn> given a |realm|, a [=completion record=]
    TODO: Tighten up the requirements here; people will probably try to parse
    this data with regex or something equally bad.
 
-1. Let |exception| be the result of [=serialize as a remote value=] given
+1. Let |exception| be the result of [=serialize as a remote value=] with |session|,
    |record|.\[[Value]], <code>1</code> as max depth, |ownership type|,
    a new [=Map=] as serialization internal map and |realm|.
 
@@ -4796,7 +4796,7 @@ The [=remote end steps=] with |session| and |command parameters| are:
 1. If |function body evaluation status|.\[[Type]] is <code>throw</code>:
 
    1. Let |exception details| be the result of [=get exception details=] given
-      |realm|, |function body evaluation status| and |result ownership|.
+      |session|, |realm|, |function body evaluation status| and |result ownership|.
 
    1. Return a new map matching the <code>script.EvaluateResultException</code>
       production, with the <code>exceptionDetails</code> field set to
@@ -4824,7 +4824,7 @@ The [=remote end steps=] with |session| and |command parameters| are:
 1. If |evaluation status|.\[[Type]] is <code>throw</code>:
 
   1. Let |exception details| be the result of [=get exception details=] given
-     |realm|, |evaluation status| and |result ownership|.
+     |session|, |realm|, |evaluation status| and |result ownership|.
 
   1. Return a new map matching the <code>script.EvaluateResultException</code>
      production, with the <code>exceptionDetails</code> field set to
@@ -4832,8 +4832,8 @@ The [=remote end steps=] with |session| and |command parameters| are:
 
 1. Assert: |evaluation status|.\[[Type]] is <code>normal</code>.
 
-1. Let |result| be the result of [=serialize as a remote value=] given
-   |evaluation status|.\[[Value]], <code>1</code> as |max depth|, |result
+1. Let |result| be the result of [=serialize as a remote value=] with
+   |session|, |evaluation status|.\[[Value]], <code>1</code> as |max depth|, |result
    ownership|, <code>new map</code> as |serialization internal map| and |realm|.
 
 1. Return a new map matching the <code>script.EvaluateResultSuccess</code>
@@ -4880,7 +4880,7 @@ TODO: Add timeout argument. It's not totally clear how this ought to work; in
 Chrome it seems like the timeout doesn't apply to the promise resolve step, but
 that likely isn't what clients want.
 
-The [=remote end steps=] with |command parameters| are:
+The [=remote end steps=] with |session| and |command parameters| are:
 
 1. Let |realm| be the result of [=trying=] to [=get a realm from a target=]
    given the value of the <code>target</code> field of |command parameters|.
@@ -4920,7 +4920,7 @@ The [=remote end steps=] with |command parameters| are:
 1. If |evaluation status|.\[[Type]] is <code>throw</code>:
 
   1. Let |exception details| be the result of [=get exception details=] given
-     |realm|, |evaluation status| and |result ownership|.
+     |session|, |realm|, |evaluation status| and |result ownership|.
 
   1. Return a new map matching the <code>script.EvaluateResultException</code>
      production, with the <code>realm</code> field set to |realm id|, and the
@@ -4928,8 +4928,8 @@ The [=remote end steps=] with |command parameters| are:
 
 1. Assert: |evaluation status|.\[[Type]] is <code>normal</code>.
 
-1. Let |result| be the result of [=serialize as a remote value=] given
-   |evaluation status|.\[[Value]], <code>1</code> as |max depth|, |result
+1. Let |result| be the result of [=serialize as a remote value=] with
+   |session|, |evaluation status|.\[[Value]], <code>1</code> as |max depth|, |result
    ownership|, <code>new map</code> as |serialization internal map| and |realm|.
 
 1. Return a new map matching the <code>script.EvaluateResultSuccess</code>
@@ -5368,7 +5368,7 @@ The [=remote end event trigger=] is:
 
 <div algorithm="remote end event trigger for log.entryAdded">
 
-Define the following [=console steps=] with |method|, |args|, and <var
+Define the following [=console steps=] with |session|, |method|, |args|, and <var
 ignore>options</var>:
 
 1. If |method| is "<code>error</code>" or "<code>assert</code>", let |level| be
@@ -5399,8 +5399,8 @@ ignore>options</var>:
 1. Let |serialized args| be a new list.
 
 1. For each |arg| of |args|:
-   1. Let |serialized arg| be the result of [=serialize as a remote value=] given
-      |arg| as value, <code>1</code> as max depth, <code>none</code> as
+   1. Let |serialized arg| be the result of [=serialize as a remote value=] with
+      |session|, |arg| as value, <code>1</code> as max depth, <code>none</code> as
       ownership type, a new [=Map=] as serialization internal map, and
       |realm|.
 
@@ -5550,11 +5550,14 @@ Other specifications can define <dfn>console steps</dfn>. When any method of the
 {{console}} interface is called, with method name
 |method| and argument |args|:
 
-1. If that method does not call the [=Printer=] operation, call any [=console
-   steps=] defined in external specification with arguments |method|, |args|,
-   and undefined.
+1. For each |session| in [=active BiDI sessions=]:
 
-   Otherwise, at the point when the [=Printer=] operation is called with
-   arguments |name|, |printerArgs| and |options| (which is undefined if the
-   argument is not provided), call any [=console steps=] defined in
-   external specification with arguments |name|, |printerArgs|, and |options|.
+  1. If that method does not call the [=Printer=] operation, call any [=console
+     steps=] defined in external specification with arguments |session|, |method|,
+     |args|, and undefined.
+
+     Otherwise, at the point when the [=Printer=] operation is called with
+     arguments |name|, |printerArgs| and |options| (which is undefined if the
+     argument is not provided), call any [=console steps=] defined in
+     external specification with arguments |session|, |name|, |printerArgs|, and
+     |options|.
--- a/index.bs
+++ b/index.bs
@@ -70,6 +70,7 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
     text: session not created; url: dfn-session-not-created
     text: session; url: dfn-sessions
     text: set a property; url: dfn-set-a-property
+    text: shadow root reference; url: dfn-shadow-root-reference
     text: success; url: dfn-success
     text: try; url: dfn-try
     text: trying; url: dfn-try
--- a/index.bs
+++ b/index.bs
@@ -427,6 +427,16 @@ with the following additional codes:
   <dd>Tried to remove an unknown [=preload script=].
 </dl>
 
+<dl>
+  <dt><dfn>No such handle</dfn>
+  <dd>Tried to deserialize an unknown <code>RemoteObjectReference</code>.
+</dl>
+
+<dl>
+  <dt><dfn>No such node</dfn>
+  <dd>Tried to deserialize an unknown <code>SharedReference</code>.
+</dl>
+
 <pre class="cddl local-cddl">
 ErrorCode = ("invalid argument" /
              "invalid session id" /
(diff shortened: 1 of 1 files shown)


sadym-chromium:
WDYT about the deserialization approach?

index.bs
```diff
@@ -1026,26 +1026,66 @@ Issue: Define how this works.
 <dfn>RemoteReference</dfn> represents a remote reference to an exising ECMAScript object in
 [=object id map=].
 
-A [=BiDi session=] has an <dfn>object id map</dfn>. This is a weak map
-from objects to their corresponding id.
+A [=BiDi session=] has:
+ * <dfn>object id map</dfn>, a weak map from objects to their corresponding id, unique for each Sandbox.
+ * <dfn>dom id map</dfn>, a weak map from [=platform object=] to their corresponding id and shared among Sandboxes.
```
foolip, line 1031: This would mean everything that is defined using Web IDL, per https://webidl.spec.whatwg.org/#dfn-platform-object

Is that really what is intended?

sadym-chromium: Yes, any Web IDL object CAN be put there. But serialization will define, if the y WILL. I'd start with adding only Nodes to the map. This PR is intended to:
1. Proto an discuss the deserialization approach discussed on the W3C meeting.
2. Figure out the difference between ECMAScript wrappers and WebIDL on the spec level.

foolip: I see, that makes sense, we can make it such that this weak map can map any platform object to a stable ID, but only but `Node`s in the map for now. How to deal with `Window` then becomes a question of whether we put it in the map.

jgraham: I wonder if shared array buffers are another special case.

sadym-chromium: I made the reference independent of the platform objects, and rely only on `SandboxProxy`. It should make the approach scalable enough, assuming the `SandboxProxy` can be defined for **shared array buffers**.

foolip: @jgraham do you mean that we should have a stable id for shared array buffers between a "main" context and sandboxes, or for the "same" shared array buffer in a window and worker context?

foolip: Discussed in a chat with James:

@jgraham 

> "I don't know" is pretty much my answer there. It just occured to me that SAB might be another case where we have the same object in multiple contexts. For sharing DOM nodes between Window and Sandbox globals we have pretty strong use cases; I don't know if we have corresponding requirements for SAB or if it's always OK to treat them as seperate objects per realm.

@foolip 

> To get the "same" SAB in two places would require `postMessage` I guess. Unlike with `Node`s, it's not already there in the other sandbox. So this feels to me like if one really needs to keep track of which SABs are the same, one can `postMessage` a random ID together with it.
>
> Which is all to say I think it would be nice to stick to just platform objects unless we have strong use cases for other kinds of objects.

@jgraham 

> Yeah, I think I buy that argument. I guess the worst case here is we later serialize some kind of `arrayBufferId` that can then be used to look up the same SAB. So if there are use cases that come up in the futute we can probably deal with them without doing anything special now.

sadym-chromium: @jgraham do I use the `SandboxProxy` in the way it was intended to?

jgraham:
A couple of quick high-level comments; I still need to do a full review:
* I think we should try to make `SandboxProxy` transparent. It's basically the representation of a wrapper object, but in implementations you can't check if something "is" a wrapper object or not; it's just that all objects are always accessed through a wrapper. So I think we should say things like "If the object wrapped by a SandboxProxy implements interface X then the SandboxProxy also implements X". 
* I don't love `crossSandboxId` as a name. I guess the idea was to avoid tying it to DOM objects in particular in case we later need it for Shared Array Buffer or similar. If we want to avoid `domId` I think I'd even prefer something like `sharedId` or `uniqueId`.

sadym-chromium:
> * I think we should try to make `SandboxProxy` transparent. It's basically the representation of a wrapper object, but in implementations you can't check if something "is" a wrapper object or not; it's just that all objects are always accessed through a wrapper. So I think we should say things like "If the object wrapped by a SandboxProxy implements interface X then the SandboxProxy also implements X".

I'm not sure if I understand this. Do you mean we should not reference the wrapped object?

> * I don't love `crossSandboxId` as a name. I guess the idea was to avoid tying it to DOM objects in particular in case we later need it for Shared Array Buffer or similar. If we want to avoid `domId` I think I'd even prefer something like `sharedId` or `uniqueId`.

I changed name to `shared`: `sharedId`, `sharedReference` etc. We can use `domId`, but not sure if it makes sense. 
* `domId`: expresses scope and usage scenarios, which can make developers' lifes easier.
* `sharedId`: keeps flexibility in case of any other sharing scenario came up.

WDYT?

jgraham:
> I'm not sure if I understand this. Do you mean we should not reference the wrapped object?

I mean if we use the model that any DOM object is really a `JSWrapper` wrapping a platform `IDLObject` (that's implemented in C++ or whatever), then given e.g. a JS object that's a `JSWrapper` wrapping a `Node` `IDLObject` we're happy to say that the object itself "implements `Node`", even though at the lowest levels the implementation does have to make the distinction between the wrapper and the underlying DOM object. 

A `SandboxProxy` object is just intended to be a spec representation of the `JSWrapper`, but scoped just to sandboxes to avoid rewriting the whole world. So I think we want to avoid asking "is this object a `SandboxProxy`" whenever possible; we largely want to handle the `SandboxProxy` objects as if they are indistinguishable from the underlying object (or inditinguishable from a non-sandbox `JSWrapper`, because AIUI in e.g. Blink's implementation they actually are the same type, and in Gecko's implementation they're actually all different views of the same object).

I'm not sure if that clarifies things or not?

index.bs
```diff
@@ -1439,22 +1509,60 @@ object is discarded in the runtime, subsequent attempts to access it via the
...
    id map=].
 
 </div>
 
+<div algorithm>
+To get the <dfn>shared id for an object</dfn> given a |browsing context|,
+|realm| and |object|:
+
+1. If |object| is not a {{SandboxProxy}}, set |object| to a {{SandboxProxy}}
+   wrapping the |object| in |realm|.
```
sadym-chromium, line 1534: bug: not checking if the SandboxProxy already exists in realm causes multiple SandboxProxies in Realm

sadym-chromium:
To clarify: To ensure a `SandboxProxy` is a singleton for the given Platform object, we need to be able to map the given Platform object to something. In current approach, I use the term "object wrapped by the `SandboxProxy`".

index.bs
```diff
@@ -1439,22 +1509,60 @@ object is discarded in the runtime, subsequent attempts to access it via the
...
+To get the <dfn>object id for an object</dfn> given a |realm| and |object|:
 
-1. If |session|'s [=object id map=] does not contain |object|, run the
+1. If |realm|'s [=object id map=] does not contain |object|, run the
    following steps:
 
-  1. Let |object id| be a new, unique, string identifier for |object|. If |object| is an
-     [=/element=] this must be the [=web element reference=] for |object|; if it's a {{WindowProxy}}
-     object, this must be the [=browsing context id=] for |object|.
+  1. Let |object id| be a new, unique, string identifier for |object|. If
```
jgraham, line 1517: It seems more reasonable for the shared id to match the web element reference or window handle, rather than the object id.

index.bs
```diff
@@ -1023,36 +1023,106 @@ Issue: Define how this works.
...
 To <dfn>deserialize remote reference</dfn> given a |remote reference|:
 
-1. Let |object id| be the value of the <code>objectId</code> field of |remote reference|.
+1. Let |remote value| be undefined.
+
+1. If |remote reference| matches the <code>RemoteObjectReference</code>
+   production, set |remote value| to the result of
+   [=deserialize remote object reference=] given |remote reference|.
+
+1. If |remote value| is undefined and |remote reference| matches the
```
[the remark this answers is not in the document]
sadym-chromium: I wanted to avoid having `InvalidArgument` in case of sanding back the object received from  the backend, as we do with `remoteReference`.

whimboo:
Hi @sadym-chromium. As it looks like this pull request hasn't been updated for quite a while and it might be great to actually have the proposed changes landed. Would you be still interested to continue? Thanks!

index.bs
```diff
@@ -1023,36 +1023,106 @@ Issue: Define how this works.
...
+In [=BiDi session=]:
+ * Each realm has an <dfn>object id map</dfn>, a weak map from objects to their
+   corresponding id. The [=object id map=] is unique per realm.
+ * Each realm has a <dfn>shared id to proxy map</dfn>, a weak map
+   from shared id to their corresponding {{SandboxProxy}} in
+   the given realm. The [=shared id to proxy map=] is unique per
+   realm.
+ * Each browsing context has a <dfn>shared id map</dfn>, a weak map from
+   ECMAScript object to their corresponding shared id. The
+   [=shared id map=] is shared among all of the realms of the context's
```
whimboo, line 1045: I assume this shouldn't be shared between all the realms of the `context` but the `Window` so that it matches what we want to land via https://github.com/w3c/webdriver/pull/1667?

sadym-chromium:
Hi Henrik! Thanks for reminding, I'm going to continue working on it soon.

whimboo:
> Hi Henrik! Thanks for reminding, I'm going to continue working on it soon.

@sadym-chromium, just a friendly reminder... Did you had a chance to continue on this PR? Another three months passed by and it would be great to have a shared id for elements. Thanks.

whimboo:
> <jgraham> sadym: I expect to start working on this again in the next week or two.

@sadym-chromium do you have an updated timeline when you can continue on this PR? Thanks

sadym-chromium:
I'm starting working on this PR again.

sadym-chromium:
I reworked the approach, and this is the first draft. WindowProxy is still not implemeneted.

index.bs
sadym-chromium: Done. Restricted to the platform objects.

sadym-chromium:
If we want to extract a specific `Element` type, consider splitting `sharedId` to the `elementId`, `nodeId` etc.

index.bs
```diff
@@ -1078,38 +1095,150 @@ Issue: Define how this works.
...
-object in [=handle object map=] in the given [=Realm=].
+<dfn>RemoteReference</dfn> is either a <code>RemoteObjectReference</code>
+representing a remote reference to an existing ECMAScript object in
+[=handle object map=] in the given [=Realm=], or is a <code>SharedReference</code>
+representing a reference to a [=platform object=] in the given [=browsing context=]'s
+[=list of known elements=], [=list of known shadow roots=] on in the given
+{{Window}}'s [=platform objects map=].
 
 <pre class="cddl remote-cddl local-cddl">
+RemoteReference = {
```
jgraham, line 1113: Could we just have a single variant that always has a `handle` and optionally has a `sharedId`? That has a few advantages:
* Some client functionality can just unconditionally use `handle` and not care about `sharedId`.
* It makes it easier to add a sharedId to new DOM types in the future if we need to, without breaking working code.
* We don't have to introspect the remote reference object to figure out which variant we got. If we do keep this design we need to add a `type` field.

The main disadvantage I can think of is that we might really want to encourage people to always use `sharedId` if we have one.

sadym-chromium: I don't quite understand the suggestion. I understood it as "let's make the handle field mandatory", which basically means the `sharedId` is useless. Maybe there is a confusion between "RemoteReference" and "RemoteValue"? The first is passed to server by client, the last is sent by server to client.

index.bs
```diff
@@ -1523,6 +1652,51 @@ To get the <dfn>handle for an object</dfn> given |realm|, |ownership type| and
...
+
+  1. Let |browsing context| be the result of [=get the browsing context=] of the
+     |realm|.
+
+  1. If |browsing context| is <code>null</code>, return <code>null</code>.
+
+  1. Return the result of [=trying=] to [=get or create a shadow root reference=]
+     with |object| and |browsing context|.
+
+1. Let |window| be the result of [=get the window=] of the |realm|.
```
foolip, line 1682: Before this, should we handle the case that _object_ is a window object, using the WebDriver classic window handle format?

index.bs
```diff
@@ -1078,38 +1095,150 @@ Issue: Define how this works.
...
+
+SharedReference = {
+   sharedId: SharedId,
+   Extensible
+}
 </pre>
 
 Issue: handle "stale object reference" case.
 
+Note: if the provided reference has both <code>handle</code> and
```
jgraham, line 1136: This is a case where I especially don't love this strategy of allowing ambiguous input and just picking a processing order to resolve the ambiguity. In this case I'd almost prefer that we defined things like
```
RemoteObjectReference = {
   handle: Handle,
   ?sharedId: SharedId
   Extensible
}

SharedReference = {
   ?handle: Handle,
   sharedId: SharedId
   Extensible
}
```
and then if you get both fields check that they actually refer to the same object.

index.bs
```diff
@@ -1078,38 +1095,150 @@ Issue: Define how this works.
...
+
+1. Let |shared id| be the value of the <code>sharedId</code> field of
+   |shared reference|.
+
+1. If |browser context|'s [=list of known elements=] contains an |element|, which
+   [=web element reference=] equals to |shared id|, return the result of [=trying=]
+   to [=get a known connected element=] with argument |shared id| and
+   |browsing context|.
+
+1. If |browser context|'s [=list of known shadow roots=] contains an |element|, which
```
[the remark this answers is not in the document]
jgraham: Yeah, I think we should fix the WebDriver spec here to have a single cache for Nodes that can be reused by BiDi, and add the additional limitations that WebDriver classic commands can only return certain Node types into that spec.

jgraham (changes requested):
Note that I haven't reviewed this in detail yet.

index.bs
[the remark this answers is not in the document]

index.bs
sadym-chromium: > This is a case where I especially don't love this strategy of allowing ambiguous input and just picking a processing order to resolve the ambiguity. In this case I'd almost prefer that we defined things like.

Changed the signature.

> if you get both fields check that they actually refer to the same object

This approach would not allow re-using the serialized node in another sandbox, as the handle would not point to any value.

jgraham: > This approach would not allow re-using the serialized node in another sandbox, as the handle would not point to any value.

It would mean that you couldn't provide an unusable handle along with the sharedId, in case you were trying to use things in a different sandbox. So yes, in that case you'd have to remove the field. But this is why I don't really like the DWIM approach to API design where we allow ambiguous protocol messages and hope that we define the right resolution rules to work how people expect most of the time, vs just having an unambigous protocol but requiring some extra upfront work to make the payloads match the format.

sadym-chromium:
> Yeah, I think we should fix the WebDriver spec here to have a single cache for Nodes that can be reused by BiDi, and add the additional limitations that WebDriver classic commands can only return certain Node types into that spec.

Correct me if I'm mistaken, but the elements and shadow roots are stored in Classic in Browsing Context. I'm not sure if we want (and can) store all other potential fields there. We can change the storage form Browsing Context to Window (Navigable?), but I'd prefer to do it not in this PR.

jgraham:
> Correct me if I'm mistaken, but the elements and shadow roots are stored in Classic in Browsing Context. I'm not sure if we want (and can) store all other potential fields there. 

We can, and I have a PR for it, which is currently blocked on unbreaking the classic master for HTML changes. Should be up later today.

sadym-chromium:
> > Correct me if I'm mistaken, but the elements and shadow roots are stored in Classic in Browsing Context. I'm not sure if we want (and can) store all other potential fields there.
> 
> We can, and I have a PR for it, which is currently blocked on unbreaking the classic master for HTML changes. Should be up later today.

Does it mean I should wait with my PR for Classic with passing "browsing context" as an argument?

sadym-chromium:
the build is failing as the `get a node` and the `get or create a node reference` are not merged yet. https://github.com/w3c/webdriver/pull/1707

index.bs
sadym-chromium: We can do the following:
1. Move this question to another issue.
1. Ask for client's feedback.
1. Change the spec if required to something like

```
RemoteObjectReference = {
   type: "handle",
   handle: Handle,
}

SharedReference = {
   type: "sharedReference",
   sharedId: SharedId
}
```

WDYT?

index.bs
```diff
@@ -1088,34 +1104,111 @@ strong map from handle ids to their corresponding objects.
...
+   [=invalid argument=].
+
+1. Let |node| be result of [=trying=] to [=get a node=] given [=current session=],
+   |browsing context| and |shared id|.
+
+1. If |browsing context| is <code>null</code>, return [=error=] with [=error code=]
+   [=no such element=].
+
+1. If |node| [=is stale=], return [=error=] with [=error code=]
+   [=stale element reference=].
```
sadym-chromium, line 1201: As we want Classic to be implementable over BiDi, the `is stale` check is needed. I'm fine with throwing "no such element" instead.

index.bs
```diff
@@ -1088,34 +1104,111 @@ strong map from handle ids to their corresponding objects.
...
+   ?handle: Handle,
+   sharedId: SharedId
    Extensible
 }
 </pre>
 
 Issue: handle "stale object reference" case.
 
+Note: if the provided reference has both <code>handle</code> and
+<code>sharedId</code>, the algorithm will first try to resolve <code>handle</code>,
```
jgraham, line 1139: Didn't we at least decide that we should use `sharedId` if present, and if that fails then error? I don't think we have a use case for providing an incorrect `sharedId`, whereas sandboxes are one for providing an incorrect `handle` (if there's also a `sharedId`).

index.bs
```diff
@@ -1088,34 +1104,111 @@ strong map from handle ids to their corresponding objects.
...
+
+1. Let |browsing context| be [=get the browsing context=] of the |realm|.
+
+1. If |browsing context| is <code>null</code>, return [=error=] with [=error code=]
+   [=invalid argument=].
+
+1. Let |shared id| be the value of the <code>sharedId</code> field of
+   |shared reference|.
+
+1. If |shared id| is <code>null</code>, return [=error=] with [=error code=]
```
jgraham, line 1191: Can it be null here? Do we ever call this with something that doesn't match `SharedReference`? If it can we should move this check right to the top.

index.bs
```diff
@@ -1088,34 +1104,111 @@ strong map from handle ids to their corresponding objects.
...
+   [=invalid argument=].
+
+1. Let |node| be result of [=trying=] to [=get a node=] given [=current session=],
```
jgraham, line 1194: We can't use `[=current session=]` because it's not a well defined concept for BiDi sessions.

index.bs
```diff
@@ -1088,34 +1104,111 @@ strong map from handle ids to their corresponding objects.
...
+   |browsing context| and |shared id|.
+
+1. If |browsing context| is <code>null</code>, return [=error=] with [=error code=]
```
jgraham, line 1197: Do you mean `node` here?

index.bs
```diff
@@ -1088,34 +1104,111 @@ strong map from handle ids to their corresponding objects.
...
+   [=no such element=].
```
jgraham, line 1198: We should coin a new error for this case.

index.bs
jgraham: We shouldn't do either. We should put enough data on the serialized element that WebDriver classic implementations built on BiDi can implement this check.

However, what we probably do need to do here is check that `|node|`'s `ownerDocument`'s [origin](https://dom.spec.whatwg.org/#concept-document-origin) is [same origin-domain](https://html.spec.whatwg.org/#same-origin-domain) with `|context|`'s associated Document. 

@foolip to check the details here: what we're aiming for is that `|node|` is reachable from scripts running in `|realm|`.

index.bs
```diff
@@ -1529,6 +1622,23 @@ To get the <dfn>handle for an object</dfn> given |realm|, |ownership type| and
...
+
+1. Let |node| be [=unwrapped=] |node|.
+
+1. If |node| is not a [=node=], return <code>null</code>.
+
+1. Let |browsing context| be [=get the browsing context=] of the |realm|.
+
+1. If |browsing context| is <code>null</code>, return <code>null</code>.
+
+1. Return [=get or create a node reference=] with [=current session=],
```
jgraham, line 1636: Remove `current session`.

index.bs
```diff
@@ -3971,6 +4087,33 @@ Note: there's a 1:1 relationship between the <code>script.RealmInfo</code>
 
 The <code>script.RealmInfo</code> type represents the properties of a realm.
 
+<div algorithm>
+To <dfn>get the top level window</dfn> of the given |realm|:
+
+1. Let |global object| be the [=realm/global object=] of the |realm|.
+
+1. While |global object| is a {{Window}} object and has a non-null [=parent=], set
```
jgraham, line 4095: This reference to `[=parent=]` is wrong because it's for a navigable's parent, which isn't a property of a `{{Window}}` object. But also I don't know why you're trying to get the top-level window here? iframes have their own browsing context, so I think something in the model is wrong.

index.bs
```diff
@@ -1088,34 +1104,111 @@ strong map from handle ids to their corresponding objects.
...
+1. If |handle map| does not contain |handle id|, then return <code>null</code>.
 
 1. Return |handle map|[|handle id|].
 
 </div>
 
+<div algorithm>
+To <dfn>deserialize shared reference</dfn> given |realm| and |shared reference|:
+
+1. Let |browsing context| be [=get the browsing context=] of the |realm|.
```
jgraham, line 1183: Can we use "with" rather than "of the"? I think that it's nicer to generally try to call algorithms in a consistent way.

index.bs
sadym-chromium: Yeah, it was added before the change in the Classic, Fixed.

index.bs
```diff
@@ -1964,7 +2080,8 @@ an |ownership type|, a |serialization internal map| and a |realm|:
 
                1. Let |serialized shadow| be the result of
                   [=serialize as a remote value=] with |shadow root|, |child depth|,
-                  false, |child ownership|, |serialization internal map| and |realm|.
```
sadym-chromium, line 1967: I assume `false` is an artefact.

index.bs
```diff
@@ -1088,34 +1104,111 @@ strong map from handle ids to their corresponding objects.
...
+[=handle object map=] in the given [=Realm=], or is a <code>SharedReference</code>
+representing a reference to a [=node=] in the given [=Realm=].
 
 <pre class="cddl remote-cddl local-cddl">
 Handle = text;
+SharedId = text;
+</pre>
 
+<pre class="cddl remote-cddl">
 RemoteReference = {
```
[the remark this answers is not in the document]

index.bs
sadym-chromium: Implemented this approach.

index.bs

index.bs
```diff
@@ -5368,11 +5511,14 @@ Other specifications can define <dfn>console steps</dfn>. When any method of the
 {{console}} interface is called, with method name
 |method| and argument |args|:
 
-1. If that method does not call the [=Printer=] operation, call any [=console
-   steps=] defined in external specification with arguments |method|, |args|,
-   and undefined.
+1. For each |session| in [=active BiDI sessions=]:
```
jgraham, line 5514: This part can't really know about WebDriver, we need to refactor.

index.bs
```diff
@@ -1100,31 +1125,105 @@ strong map from handle ids to their corresponding objects.
...
+
+<div algorithm>
+To <dfn>deserialize shared reference</dfn> given |shared reference|, |realm| and
+|session|:
+
+1. Assert |shared reference| matches the <code>SharedReference</code> production.
+
+1. Let |browsing context| be [=get the browsing context=] with |realm|.
+
+1. If |browsing context| is <code>null</code>, return [=error=] with [=error code=]
```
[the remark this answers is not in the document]
jgraham: I changed this to return `no such node` since it's at least true that the node you're trying to access doesn't existing in the realm you're trying to access it in.

jgraham:
I've rebased this and fixed some problems. I think it's good to go now, but it would be useful for e.g. @whimboo and @foolip to check for errors and omissions.

index.bs
```diff
@@ -1944,8 +2082,8 @@ an |ownership type|, a |serialization internal map| and a |realm|:
             1. Let |child depth| be |max depth| - 1 if |max depth| is not null, or null otherwise.
 
             1. Let |serialized| be the result of [=serialize as a remote value=]
-               with |child|, |child depth|, |ownership type|,
-               |serialization internal map| and |realm|.
+               with |session|, |child|, |child depth|, |ownership type|,
```
whimboo, line 2085: `session` is listed twice here.

index.bs
```diff
@@ -1974,8 +2112,9 @@ an |ownership type|, a |serialization internal map| and a |realm|:
                   null, or null otherwise.
 
                1. Let |serialized shadow| be the result of
-                  [=serialize as a remote value=] with |shadow root|, |child depth|,
-                  false, |ownership type|, |serialization internal map| and |realm|.
+                  [=serialize as a remote value=] with |session|, |shadow root|, |child depth|,
```
whimboo, line 2115: `session` is listed twice here.

index.bs
```diff
@@ -5224,67 +5380,68 @@ The [=remote end event trigger=] is:
 
 <div algorithm="remote end event trigger for log.entryAdded">
 
-Define the following [=console steps=] with |method|, |args|, and <var
-ignore>options</var>:
+Define the following [=console steps=] with |method|, |args|, and
```
whimboo, line 5383: There is a lot of refactoring here for the console steps which would have been great to avoid and only include the calling convention for the object serialization. But now it's already here... so I'm ok.

index.bs
```diff
@@ -1101,31 +1127,116 @@ strong map from handle ids to their corresponding objects.
...
+RemoteReference = (
+  SharedReference //
+  RemoteObjectReference
+)
 
-RemoteReference = {
+SharedReference = {
+   sharedId: SharedId
+   <!-- Ensure that if we have a handle, it at least has the correct type -->
+   ?handle: Handle,
```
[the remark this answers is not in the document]
jgraham: It means that you can't write:
```
{
   sharedId: "foo",
   handle: ["some other type"]
}
```
which otherwise would match due to `Extensible`. Given that the proposed use case is "we want to allow people to directly send JSON data from responses" it makes sense to me to ensure that we can't extend this in a way that doesn't make sense (but honestly I still think we should care less about this use case and prefer nomnative typing over structural typing. However I don't seem to have much support for that position).

index.bs
```diff
@@ -1100,31 +1125,105 @@ strong map from handle ids to their corresponding objects.
...
 Issue: handle "stale object reference" case.
 
+Note: if the provided reference has both <code>handle</code> and
+<code>sharedId</code>, the algorithm will ignore <code>handle</code> and respect
+only <code>sharedId</code>.
+
+### Deserialization ### {#data-types-reference-deserialization}
+
+<div algorithm>
+To <dfn>deserialize remote reference</dfn> given |remote reference|, |realm| and
```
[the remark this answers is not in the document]
jgraham: I don't propose that we fix this in this issue; I want to move all this into the script module anyway, so we may as well refactor later.

index.bs
```diff
@@ -1101,31 +1127,116 @@ strong map from handle ids to their corresponding objects.
...
+
+1. Return [=success=] with data |handle map|[|handle id|].
+
+</div>
+
+<div algorithm>
+To <dfn>deserialize shared reference</dfn> given |shared reference|, |realm| and
+|session|:
+
+1. Assert |shared reference| matches the <code>SharedReference</code> production.
```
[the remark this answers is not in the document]
jgraham: We never "need" these asserts. But the idea is to clearly express invariants so that we're less likely to introduce breakage with future edits (we could probably be better at this in general e.g. giving each algorithm parameter an explifit spec type).

index.bs
```diff
@@ -1101,31 +1127,116 @@ strong map from handle ids to their corresponding objects.
...
+
+1. Let |browsing context| be [=get the browsing context=] with |realm|.
```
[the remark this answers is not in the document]
jgraham: No, I think, although the spec is wildly inconsistent and I find it confusing. We use `given` for declaring algorithm parameters and `with` for passing specific values for those parameters.

index.bs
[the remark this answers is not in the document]
jgraham: We could make up a new error, or we could say "well you tried to look up a node in this realm and it wasn't found (because this realm doesn't have any nodes in it)". I don't really have a good sense for when people prefer to construct new errors vs when they want to reuse existing ones.

index.bs
```diff
@@ -1361,7 +1472,8 @@ SetLocalValue = {
 
 <div algorithm>
 
-To <dfn>deserialize key-value list</dfn> given |realm| and |serialized key-value list|:
+To <dfn>deserialize key-value list</dfn> given |serialized key-value list|, |realm|
+and |session|:
```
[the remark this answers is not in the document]
jgraham: Let's defer that.

index.bs
```diff
@@ -1542,6 +1657,23 @@ To get the <dfn>handle for an object</dfn> given |realm|, |ownership type| and
 
 </div>
 
+<div algorithm>
+To get <dfn>shared id for a node</dfn> given |node|, |realm| and |session|:
+
+1. Let |node| be [=unwrapped=] |node|.
+
+1. If |node| does not implement {{Node}}, return null.
```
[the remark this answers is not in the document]
jgraham: That isn't my understanding.

index.bs
jgraham: I think GitHub's diff is just terrible and what's changed is that we need to know the session to do the serialization, so we move that loop to the outside.

index.bs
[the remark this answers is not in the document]
whimboo: Note that in WebDriver classic we raise a different error which is `no such window` if the related browsing context cannot be found. This is outside of the deserialization and serialization code but at least gives more details why the element / ShadowRoot wasn't found. 

With BiDi we basically operate on any browsing context and would already throw a `no such frame` error if eg. a script should be evaluated for a context / realm that doesn't exist. So maybe we are fine with keeping a single additional error then.

index.bs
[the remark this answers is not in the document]
whimboo: I see. That's fine as long as the calling conventions are correct.

index.bs
```diff
@@ -425,11 +428,23 @@ with the following additional codes:
...
 </dl>
 
+<dl>
+  <dt><dfn>No such handle</dfn>
+  <dd>Tried to deserialize an unknown <code>RemoteObjectReference</code>.
+</dl>
+
+<dl>
+  <dt><dfn>No such node</dfn>
+  <dd>Tried to deserialize an unknown <code>SharedReference</code>.
```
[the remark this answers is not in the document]
whimboo: We only want to use `sharedId` for a subset of [node types](https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType) but not all of them. So my comment is related to `no such node`.

index.bs
```diff
@@ -1101,31 +1127,116 @@ strong map from handle ids to their corresponding objects.
 OwnershipModel = "root" / "none";
 </pre>
 
-<dfn>RemoteReference</dfn> represents a remote reference to an existing ECMAScript
-object in [=handle object map=] in the given [=Realm=].
+<dfn>RemoteReference</dfn> is either a <code>RemoteObjectReference</code>
+representing a remote reference to an existing ECMAScript object in
+[=handle object map=] in the given [=Realm=], or is a <code>SharedReference</code>
+representing a reference to a [=node=].
```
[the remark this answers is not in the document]
whimboo: Was that the original intention? So it's basically that WebDriver classic only has the limitation for `Element` and `ShadowRoot`? If that's the case then we can close this issue.

index.bs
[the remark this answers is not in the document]
whimboo: I'll close this issue in favor of the other one to centralize the discussion.

index.bs
whimboo: Maybe. It was also more a comment so I'm going to close the issue.

index.bs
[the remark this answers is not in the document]
jgraham: Yes, BiDi doesn't have the same limitations as Classic here, and it should be possible to refer to any DOM object with a shared id.

index.bs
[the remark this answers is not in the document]
sadym-chromium: > So this actually happens if |realm| doesn't have a `Window` object as its global scope (e.g. it's a worker realm). So I think we need a better error type.

I guess if the |realm| doesn't have a browsing context for whatever reason, we can consider it doesn't have a `frame`, and `[=no such frame=]` seems to be a proper error code.

sadym-chromium:
@jgraham thanks for rebasing and fixing issues.

I reviewed the new changes, and it LGTM, but I cannot self-approve the PR. @jgraham feel free to merge.

Issue #178, opened 2022-03-09
browsingContext.getTree has to also include the parent context


whimboo:
As I can see the spec as currently written does not include the browsing context details for the parent context as given for [browsingContext.getTree](https://w3c.github.io/webdriver-bidi/#command-browsingContext-getTree). Instead of getting the [browsing context info](https://w3c.github.io/webdriver-bidi/#get-the-browsing-context-info) for the parent it directly calls into [get the descendent browsing contexts](https://w3c.github.io/webdriver-bidi/#get-the-descendent-browsing-contexts).

whimboo:
Also note that for both the `browsingContext.contextCreated` and `browsingContext.contextDestroyed` events we call `get the browsing context info` with the browsing context, while the method expects a browsing context id.

To avoid silly conversions between numbers and contexts we should always pass around the browsing contexts internally and only make use of numbers when returning data to the client.

PR #181, opened 2022-03-10
Let browsingContext.getTree also include the parent, and don't use context ids internally

whimboo:
Fixes issue #178.

Appropriate tests for this change are part of https://bugzilla.mozilla.org/show_bug.cgi?id=1694391 and will land soon.

***
<a href="https://pr-preview.s3.amazonaws.com/whimboo/webdriver-bidi/pull/181.html" title="Last updated on Mar 17, 2022, 10:02 AM UTC (7c317f5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/181/8b12b43...whimboo:7c317f5.html" title="Last updated on Mar 17, 2022, 10:02 AM UTC (7c317f5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2367,9 +2367,9 @@ To <dfn>get the browsing context info</dfn> given |context|,
 
    Note: This includes the fragment component of the URL.
 
-1. Let |child infos| be a list.
-
 1. If |max depth| is null, or |max depth| is greater than 0:
+
+  1. Let |child infos| be an empty list.
 
   1. Let |child contexts| be a list containing all [=/browsing contexts=]
      which are [=child browsing contexts=] of |parent|.
@@ -2379,9 +2379,11 @@ To <dfn>get the browsing context info</dfn> given |context|,
   1. For each |context| of |child contexts|:
 
     1. Let |info| be the result of [=get the browsing context info=] given
-       |context|, |child depth|, and <code>false</code>.
+       |context|, |child depth|, and false.
 
     1. Append |info| to |child infos|
+
+1. Otherwise let |child infos| be null.
 
 1. Let |context info| be a [=map=] matching the
    <code>BrowsingContextInfo</code> production with the <code>context</code>
@@ -2642,7 +2644,7 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
 1. For each |context| of |contexts|:
 
   1. Let |info| be the result of [=get the browsing context info=] given
-     |context|, |max depth|, and <code>true</code>.
+     |context|, |max depth|, and true.
 
   1. Append |info| to |contexts info|
 
@@ -2807,7 +2809,7 @@ To <dfn>Recursively emit context created events</dfn> given |session| and |conte
 To <dfn>Emit a context created event</dfn> given |session| and |context|:
 
 1. Let |params| be the result of [=get the browsing context info=] given
-   |context|, 0, and <code>true</code>.
+   |context|, 0, and true.
 
 1. Let |body| be a [=map=] matching the
    <code>BrowsingContextCreatedEvent</code> production, with the
@@ -2869,7 +2871,7 @@ Define the following [=browsing context tree discarded=] steps:
 1. Let |context| be the browsing context being discarded.
 
 1. Let |params| be the result of [=get the browsing context info=], given
-   |context|, 0, and <code>true</code>.
+   |context|, 0, and true.
 
 1. Let |body| be a [=map=] matching the
     <code>BrowsingContextDestroyedEvent</code> production, with the
--- a/index.bs
+++ b/index.bs
@@ -2368,7 +2368,7 @@ To <dfn>get the browsing context info</dfn> given |context|,
    Note: This includes the fragment component of the URL.
 
 1. Let |child info| be the result of [=get the descendent browsing contexts=]
-   given |context id|, |depth| + 1, and |max depth|.
+   given |context|, |depth| + 1, and |max depth|.
 
 1. Let |context info| be a [=map=] matching the
    <code>BrowsingContextInfo</code> production with the <code>context</code>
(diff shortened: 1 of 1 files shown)


whimboo:
@jgraham please let me know what you think.

index.bs
```diff
@@ -2381,17 +2381,15 @@ To <dfn>get the browsing context info</dfn> given |context|,
...
+To <dfn>get the descendent browsing contexts</dfn> given |parent|, |depth|
 and |max depth|:
 
-1. If |max depth| is greater than zero, and |depth| is equal to |max depth|,
-   return null.
+1. Assert: |parent| is not null.
 
-1. Let |parent| be the result of [=trying=] to [=get a browsing context=]
-   given |parent id|.
+1. If |max depth| is not null, and |depth| is greater than |max depth|,
```
jgraham, line 2389: I think in the current formulation there's an off-by-one error here, right? If depth is 0 and max depth is 0 we just want to include the parent context, so we should return.

index.bs
```diff
@@ -2381,17 +2381,15 @@ To <dfn>get the browsing context info</dfn> given |context|,
 </div>
 
 <div algorithm>
-To <dfn>get the descendent browsing contexts</dfn> given |parent id|, |depth|
+To <dfn>get the descendent browsing contexts</dfn> given |parent|, |depth|
```
jgraham, line 2384: I think it might be sensible to just have a single parameter like we do with serialization that counts down toward 0, rather than two parameters one of which is constant and one that is incremented.

index.bs
jgraham: Oh I guess there isn't because we only call this from `get the browsing context info`, which increments the depth counter.

index.bs
whimboo: While this might be fine I see a problem when we emit the reference to the parent browsing context. As given by the spec the `parent` should only be included for the top-most browsing context info. If we only pass-in the `maxDepth` there is no chance to actually retrieve that information anymore. Any suggestions?

jgraham: Two options I think: 1) Just always return the parent id, even when that's redundant information, or 2) have an `root` flag that's `True` for the initial call and `False` for recursive calls.

whimboo: I would choose the `root` flag to not change the data that we return and to not emit redundant information. I'll push an update soon. Thanks.

index.bs
```diff
@@ -2643,15 +2629,27 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
    |command parameters| if present, or null otherwise.
 
 1. Let |max depth| be the value of the <code>maxDepth</code> field of |command
-   parameters| if present, or 0 otherwise.
+   parameters| if present, or null otherwise.
 
 1. Let |depth| be 0.
```
juliandescottes, line 2634: Is |depth|  still used in the current version?

whimboo: No, it's a left-over and I'll remove. 

Note that this follows what we have for serialization and if we want to do the rename I would propose a different issue / PR.

index.bs
```diff
@@ -2367,43 +2367,29 @@ To <dfn>get the browsing context info</dfn> given |context|,
...
+  1. Let |child depth| be |max depth| - 1 if |max depth| is not null, or null otherwise.
 
-1. If |max depth| is greater than zero, and |depth| is equal to |max depth|,
-   return null.
+  1. For each |context| of |child contexts|:
 
-1. Let |parent| be the result of [=trying=] to [=get a browsing context=]
-   given |parent id|.
+    1. Let |info| be the result of [=get the browsing context info=] given
+       |context|, |child depth|, and <code>false</code>.
```
jgraham, line 2382: I think we've tried to move away from using `<code>` around boolean literals.

index.bs
```diff
@@ -2643,15 +2629,25 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
 
-1. Let |depth| be 0.
+1. Let |contexts info| be a list.
+
+1. For each |context| of |contexts|:
 
-1. Let |contexts| be the result of [=get the descendent browsing contexts=],
-   given |parent id|, |depth|, and |max depth|.
+  1. Let |info| be the result of [=get the browsing context info=] given
+     |context|, |max depth|, and <code>true</code>.
```
jgraham, line 2645: Also boolean literal here.

index.bs
```diff
@@ -2811,7 +2807,7 @@ To <dfn>Recursively emit context created events</dfn> given |session| and |conte
 To <dfn>Emit a context created event</dfn> given |session| and |context|:
 
 1. Let |params| be the result of [=get the browsing context info=] given
-   |context|, 0, and 1.
+   |context|, 0, and <code>true</code>.
```
jgraham, line 2810: And here

index.bs
```diff
@@ -2873,7 +2869,7 @@ Define the following [=browsing context tree discarded=] steps:
 1. Let |context| be the browsing context being discarded.
 
 1. Let |params| be the result of [=get the browsing context info=], given
-   |context|, 0, and 0.
+   |context|, 0, and <code>true</code>.
```
jgraham, line 2872: And here.

index.bs
```diff
@@ -2367,43 +2367,29 @@ To <dfn>get the browsing context info</dfn> given |context|,
 
    Note: This includes the fragment component of the URL.
 
-1. Let |child info| be the result of [=get the descendent browsing contexts=]
-   given |context id|, |depth| + 1, and |max depth|.
+1. Let |child infos| be a list.
```
jgraham, line 2370: This is wrong; if `|max depth|` is 0 we want to return `null` (or maybe only if there are actually child contexts?) The point is we only want to return the empty list if there actually aren't any children; we use `null` to signify that we reached the maximum dpeth.

jgraham (changes requested):

index.bs
whimboo: Oh yes! Good catch. In my implementation I actually use `null` but forgot to change it here. Will be fixed with the next update.

index.bs
```diff
@@ -2367,43 +2367,31 @@ To <dfn>get the browsing context info</dfn> given |context|,
...
-1. Let |contexts info| be a list.
+    1. Let |info| be the result of [=get the browsing context info=] given
+       |context|, |child depth|, and false.
 
-1. For each |context| of |child contexts|:
+    1. Append |info| to |child infos|
 
-  1. Let |info| be the result of [=get the browsing context info=] given
-     |context|, |depth|, and |max depth|.
+1. Otherwise let |child infos| be null.
```
jgraham, line 2386: We're a bit inconsistent in whether "otherwise" clauses should be part of the same list item as the corresponding "if" or not; I tend to think they probably should. But in this case it would perhaps be clearer to do:
```
1. Let |child infos| be null.

1. If |max depth| is null, or |max depth| is greater than 0:

 1. Set |child infos| to an empty list
 […]
```

whimboo: Will be done in the next update that i'm going to push in a minute.

jgraham (approved):

PR #183, opened 2022-03-16
Add simple support for browsing context screenshots

jgraham:
This adds support for capturing the screenshot of a browsing
context. Unlike WebDriver classic it isn't limited to a top-level
browsing context. For now there isn't any support for full-page
screenshots, output formats other than PNG, or screenshots of specific
elements.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/183.html" title="Last updated on May 25, 2022, 10:58 AM UTC (164fa8e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/183/60005be...164fa8e.html" title="Last updated on May 25, 2022, 10:58 AM UTC (164fa8e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -40,6 +40,8 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
     text: additional capability deserialization algorithm; url: dfn-additional-capability-deserialization-algorithm
     text: capability name; url: dfn-capability-name
     text: create a session; url: dfn-create-a-session
+    text: draw a bounding box from the framebuffer; url: dfn-draw-a-bounding-box-from-the-framebuffer
+    text: encode a canvas as Base64; url: dfn-encoding-a-canvas-as-base64
     text: endpoint node; url: dfn-endpoint-node
     text: error code; url: dfn-error-code
     text: error; url: dfn-errors
@@ -66,6 +68,7 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
     text: trying; url: dfn-try
     text: unknown command; url: dfn-unknown-command
     text: unknown error; url: dfn-unknown-error
+    text: unsupported operation; url: dfn-unsupported-operation
     text: web element reference; url: dfn-web-element-reference
     text: webdriver-active flag; url: dfn-webdriver-active-flag
     text: window handle; url: dfn-window-handle
@@ -119,6 +122,9 @@ spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
     text: realm; url: sec-code-realms
     text: running execution context; url: running-execution-context
     text: time value; url: sec-time-values-and-time-range
+spec: GEOMETRY; urlPrefix: https://drafts.fxtf.org/geometry/
+  type: dfn
+    text: rectangle; url: rectangle
 spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
   type: dfn
     text: a browsing context is discarded; url: window-object.html#a-browsing-context-is-discarded
@@ -132,6 +138,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
     text: navigation id; url: browsing-the-web.html#navigation-id
     text: report an error; url: webappapis.html#report-the-error
     text: remove a browsing context; url: browsers.html#bcg-remove
+    text: run the animation frame callbacks; url: imagebitmap-and-animations.html#run-the-animation-frame-callbacks
     text: session history; url: history.html#session-history
     text: set up a window environment settings object; url: window-object.html#set-up-a-window-environment-settings-object
     text: set up a worker environment settings object; url: workers.html#set-up-a-worker-environment-settings-object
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2623,6 +2631,69 @@ To <dfn>get the navigation info</dfn>, given |context| and |navigation status|:
...
+   <dd>
+    <pre class="cddl local-cddl">
+        BrowsingContextCaptureScreenshotResult = {
+          data: text
+        }
+    </pre>
+   </dd>
+</dl>
+
+TODO: Full page screenshots, and multiple output formats
```
whimboo, line 2665: I assume we also want to allow to capture screenshots for specific elements only?

jgraham: Right, I was imagining that as part of a different module since it would be about elements. But we could of course divide things up in more than one way e.g. we could have a screenshots module with a command that would either take a context or an element.

In any case that TODO was only supposed to be for things that are unambiguously in the remit of the browsing context screenshot feature.

bwalderman (approved):

index.bs
whimboo: Would you mind filing a new issue for that and reference it here?

index.bs
```diff
@@ -2623,6 +2631,69 @@ To <dfn>get the navigation info</dfn>, given |context| and |navigation status|:
...
+1. Let |context| be the result of [=trying=] to [=get a browsing context=]
+   with |context id|.
+
+1. Let |document| be |context|'s [=active document=].
+
+1. Immediately after the next invocation of the [=run the animation frame
+   callbacks=] algorithm for |document|:
+
+   Issue: This ought to be integrated into the update rendering algorithm
+   algorithm in some more explicit way.
```
juliandescottes, line 2681:
```suggestion
   Issue: This ought to be integrated into the update rendering algorithm
   in some more explicit way.
```

jgraham:
@sadym-chromium @foolip I think you weren't at the meeting where this was discussed. Any concerns if I go ahead and merge this?

index.bs
```diff
@@ -2623,6 +2631,69 @@ To <dfn>get the navigation info</dfn>, given |context| and |navigation status|:
...
+
+1. Let |context id| be the value of the <code>context</code> field of
+   |command parameters|.
+
+1. Let |context| be the result of [=trying=] to [=get a browsing context=]
+   with |context id|.
+
+1. Let |document| be |context|'s [=active document=].
+
+1. Immediately after the next invocation of the [=run the animation frame
```
sadym-chromium, line 2677: NIT: discussed recently, and underline this again. This line breaks make search way harder. Can we avoid having them by manually fixing? It doesn't seem to be too much overhead.

sadym-chromium:
> Unlike WebDriver classic it isn't limited to a top-level browsing context.

Unfortunately, due to the Blink specific, there is no way to make a screenshot of a nested context (iFrame) in CDP. Can we limit it to the top-level at least for now?

sadym-chromium (changes requested):
Main concern is in screenshots of nested contexts.

jgraham:
If we think there's a use case for screenshots of iframes it seems reasonable to make the spec handle that case; then once Blink has fixed its support clients can use it immediately rather than needing to also wait for a spec change. I can add a general step to handle the possibility that the command can't be used in a specific context.

whimboo:
@sadym-chromium once you are back maybe you could have a look at this PR and @jgraham's last comment again? Thanks.

index.bs
```diff
@@ -2698,6 +2707,72 @@ To <dfn>get the navigation info</dfn>, given |context| and |navigation status|:
...
+   <dd>
+    <pre class="cddl local-cddl">
+        BrowsingContextCaptureScreenshotResult = {
+          data: text
+        }
+    </pre>
+   </dd>
+</dl>
+
+TODO: Full page screenshots, and multiple output formats
```
sadym-chromium, line 2741: This comment is confusing. I understand "document element's rectangle" as the full-page. Do I misunderstand "full-page" or "document element's rectangle"?

index.bs
```diff
@@ -2698,6 +2707,72 @@ To <dfn>get the navigation info</dfn>, given |context| and |navigation status|:
...
+   <dt>Command Type</dt>
+   <dd>
+      <pre class="cddl remote-cddl">
+      BrowsingContextCaptureScreenshotCommand = {
+        method: "browsingContext.captureScreenshot",
+        params: BrowsingContextCaptureScreenshotParameters
+      }
+
+      BrowsingContextCaptureScreenshotParameters = {
+        context: BrowsingContext
```
sadym-chromium, line 2727: Additional parameters are needed, but we can add them later.

sadym-chromium (approved):
1. I'd avoiding breaking lines in references.
2. I'm a bit confused by the TODO about full-page screenshot

Both concerns are not blockers, so LGTM.

index.bs
jgraham: https://w3c.github.io/webdriver/#dfn-draw-a-bounding-box-from-the-framebuffer is limiting the screenshot to the viewport, irrespective of the rectangle dimensions.

PR #192, opened 2022-04-14
Editorial: Correct references to "result type"

jugglinmike:
"Commands" are generally described to include a "result type," but none
of the commands defined actually include such a value. They do include a
"return type," as defined by WEBIDL. Since WEBIDL's "return type" does
not seem particularly relevant and since the terms "result type" and
"return type" have similar connotations in this context, this appears to
be an editorial oversight.

Update the command definitions to specify a "result type," and update
the explanation of the commands' "remote end steps" to return an
instance of the "result type."

***
<a href="https://pr-preview.s3.amazonaws.com/bocoup/webdriver-bidi/pull/192.html" title="Last updated on Nov 17, 2022, 3:38 AM UTC (f5a6864)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/192/60b640b...bocoup:f5a6864.html" title="Last updated on Nov 17, 2022, 3:38 AM UTC (f5a6864)">Diff</a>


jugglinmike:
Hi @whimboo! Hi @jgraham! This branch conflicted with some recent changes to the main branch, so I've resolved them. Mind taking a look?

whimboo (approved):
Discussed with @jgraham and there are no strong opinions from our side. As mentioned both definitions can be used, but we should stay in sync for all of the usages so this PR is fine. Thanks.

jugglinmike:
Sure thing!

PR #194, opened 2022-04-21
Switch to Rust CDDL implementation

jgraham:
This has a few advantages:
* It prints a more useful error message when it fails, so it's easier
to identify the point of failure
* It supports validation, so we can write tests against the CDDL
defined in the spec.
* It seems to be more actively maintained

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,3 +19,8 @@ jobs:
           toolchain: stable
       - name: Validate CDDL files
         run: ./scripts/test.sh --install
+      - name: Archive CDDL files
+        uses: actions/upload-artifact@v3
+        with:
+          name: cddl
+          path: "*.cddl"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,15 +14,8 @@ jobs:
       - uses: actions/setup-node@v1
         with:
           node-version: 14.x
-      - name: Install Dependencies
-        run: npm install cddl parse5
-      - name: Extract CDDL content from spec into files
-        run: ./scripts/cddl/generate.js
-      - name: Print remote.cddl
-        run: cat remote.cddl
-      - name: Print local.cddl
-        run: cat local.cddl
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
       - name: Validate CDDL files
-        run: |
-          npx cddl validate local.cddl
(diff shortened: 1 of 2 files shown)


scripts/test.sh
```diff
@@ -0,0 +1,14 @@
+#! /bin/bash
+set -ex
+
+SCRIPT_DIR=$(cd $(dirname "$0") && pwd -P)
+ROOT=$SCRIPT_DIR/..
```
whimboo, line 5:
```suggestion
ROOT=$(dirname $SCRIPT_DIR)
```

.github/workflows/test.yml
```diff
@@ -14,15 +14,13 @@ jobs:
...
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
       - name: Validate CDDL files
-        run: |
-          npx cddl validate local.cddl
-          npx cddl validate remote.cddl
+        run: ./scripts/test.sh --install
+      - name: Archive CDDL files
+        uses: actions/upload-artifact@v3
```
whimboo, line 23: 👍🏻 for uploading the generated files as artifacts. Now it could be used by consumers to grab the validated files pretty easily.

scripts/test.sh
```diff
@@ -0,0 +1,14 @@
...
+set -ex
+
+SCRIPT_DIR=$(cd $(dirname "$0") && pwd -P)
+ROOT=$(dirname $SCRIPT_DIR)
+
+if [ "$1" = "--install" ]; then
+    cargo install cddl
+    npm install parse5
+fi
+$ROOT/scripts/cddl/generate.js
```
whimboo, line 11: Mind adding the comment for `Extract CDDL content from spec into files` which was previously in `test.yml`?

whimboo (approved):
It looks all fine. I would be interested to actually see a case when the CDDL validation actually fails. In case you have such a CI job at hand.

mathiasbynens:
In case anyone else is wondering where these artifacts are uploaded or how they can be accessed by users, this is answered here: https://github.com/actions/upload-artifact#where-does-the-upload-go

Instructions specific to this repo are:

1. Go to https://github.com/w3c/webdriver-bidi/actions/workflows/test.yml
2. Pick the latest workflow run
3. Scroll down to the Artifacts section
    image
4. Click the filename to download

PR #195, opened 2022-04-25
Initial support for handling user prompts

jgraham:
Add two events for user prompts: one when a prompt is opened and one
when it's closed. Also a command to respond to an existing user prompt.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -51,6 +51,7 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
     text: local end; url: dfn-local-ends
     text: matched capability serialization algorithm; url: dfn-matched-capability-serialization-algorithm
     text: maximum active sessions; url: dfn-maximum-active-sessions
+    text: no such alert; url: dfn-no-such-alert
     text: no such element; url: dfn-no-such-element
     text: no such frame; url: dfn-no-such-frame
     text: process capabilities; url: dfn-processing-capabilities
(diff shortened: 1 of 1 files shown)


jgraham:
This also depends on some changes in HTML to cause the events to be emitted at the correct time, but those are ~trivial, so I'll make them if there's broad agreement on the overall approach.

index.bs
```diff
@@ -2788,6 +2796,70 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+<div algorithm="remote end steps for browsingContext.handleUserPrompt">
+
+1. Let |context id| be the value of the <code>context</code> field of
+   |command parameters|.
+
+1. Let |context| be the result of [=trying=] to [=get a browsing context=]
+   with |context id|.
+
+1. Let |accept| be the value of the <code>accept</code> field of |command
+   parameters| if present, or true otherwise.
```
foolip, line 2837: Defaulting to true is a bit unusual, at least in Web IDL, but I don't see a problem with doing it here.

index.bs
```diff
@@ -2788,6 +2796,70 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+1. If |context| is currently showing a simple dialog from a call to [=alert=] then
+   acknowledge the prompt.
+
+   Otherwise if |context| is currently showing a simple dialog from a call to
+   [=confirm=], then respond positively if |accept| is true, or respond
+   negatively if |accept| is false.
+
+   Otherwise if |context| is currently showing a simple dialog from a call to
+   [=prompt=], then respond with the string value |text| if |accept| is
+   true, or abort if  |accept| is false.
```
foolip, line 2851:
```suggestion
   true, or abort if |accept| is false.
```

index.bs
```diff
@@ -3296,6 +3368,92 @@ failed</dfn> steps given |context| and |navigation status|:
...
+   <dt>Event Type</dt>
+   <dd>
+      <pre class="cddl local-cddl">
+        BrowsingContextUserPromptClosedEvent = {
+         method: "browsingContext.userPromptClosed",
+         params: UserPromptInfo
+       }
+       UserPromptResult = {
+         context: BrowsingContext,
+         confirmed: bool,
```
foolip, line 3383: Should we use the same terminology as in the command, `accepted`? Or is there a difference?

index.bs
```diff
@@ -3296,6 +3368,92 @@ failed</dfn> steps given |context| and |navigation status|:
...
+         ? userInput: text
```
foolip, line 3384:
```suggestion
         ? userText: text
```

index.bs
```diff
@@ -3296,6 +3368,92 @@ failed</dfn> steps given |context| and |navigation status|:
...
+<div algorithm>
+The [=remote end event trigger=] is the <dfn export>WebDriver BiDi user prompt
+closed</dfn> steps given |context|, |confirmed| and optional |user input|
+(default: null).
+
+1. Let |context id| be the [=browsing context id=] for |context|.
+
+1. Let |params| be a [=map=] matching the <code>UserPromptResult</code>
+   production with the <code>context</code> field set to |context id|, the
+   <code>confirmed</code> field set to |confirmed|, and the <code>userInput</code> field set
```
foolip, line 3399:
```suggestion
   <code>confirmed</code> field set to |confirmed|, and the <code>userText</code> field set
```

foolip (approved):

index.bs
jgraham: I just chose the terminology to match https://chromedevtools.github.io/devtools-protocol/tot/Page/#event-javascriptDialogClosed but I like your suggestion as well.

jgraham:
The HTML PR to integrate with this was merged, so now we just need to decide between using the field names that match the equivalent CDP feature and some slightly cleaned up names as suggested by @foolip.

whimboo:
@OrKoN would you mind to give some feedback from a Puppeteer perspective? Would the different naming as suggested by @foolip be fine or should it be close to CDP?

OrKoN:
I think the naming does not have to match CDP here, the BiDi implementation will be responsible for translating the BiDi commands into CDP commands so please choose a name that makes sense for the spec.

OrKoN:
cc @sadym-chromium

PR #200, opened 2022-05-04
Change Issue about the formatter operation into a Note (log.entryAdded)

### LINKED ISSUE #139: Clarify handling of formatted args in log entryAdded
Initial discussion from https://bugzilla.mozilla.org/show_bug.cgi?id=1731543

The current version of the specification for `log.entryAdded` says that the `text` property should be made from the "formatted" arguments, and that the `args` property should contain the original (ie non formatted) arguments. https://w3c.github.io/webdriver-bidi/#event-log-entryAdded

Typically for a console call like `console.log("Hello %s", "world");`:
- `text` should be `Hello world`
- `args` should be `["Hello %s", "world"]`

Looking at the current implementations, none of them seem to really do that.

**CDP**
It seems that currently CDP is not applying any formatting, meaning that for this example
- `text` is `Hello %s world`
- `args` is `["Hello %s", "world"]`

**Gecko**
Gecko's Console implementation applies formatting very early at the moment. Which means we have no access to the non-formatted string or to the original arguments. For the example above, our current BiDi implementation has:
- `text` is `Hello world`
- `args` is `["Hello world"]`

Of course we can change the platform's implementation if needed, for the initial version we just reused what was readily available. I doubt this is a model which would make sense for BiDi, mostly adding it here for completeness.

So the question here is whether we want to keep the current specification or reuse the behavior from CDP (ie skip any formatting). CDP's approach is probably easier to implement and maintain but pushes the work towards consumers. Another option is to propose both a `formattedText` and a `text`, although this might create performance issues when logging long strings? 

Notes:
- on top of string formatters, there is also %c for CSS which can't be applied
- I haven't tested how CDP's `text` behaved when mixed with more complex objects

juliandescottes:
Fixes #139

Inconsistencies caused by stringifying non-primitives are already mentioned in the specification. 
The only thing left here to close the issue is to update the current issue into a note, since there is no action on our side besides notifying users that the formatted `text` might differ across implementations.

***
<a href="https://pr-preview.s3.amazonaws.com/juliandescottes/webdriver-bidi/pull/200.html" title="Last updated on May 4, 2022, 1:02 PM UTC (3590979)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/200/3dbd65e...juliandescottes:3590979.html" title="Last updated on May 4, 2022, 1:02 PM UTC (3590979)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4388,8 +4388,8 @@ ignore>options</var>:
    specifier=], let |formatted args| be [=Formatter=](|args|). Otherwise let
    |formatted args| be |args|.
 
-   Issue: This is underdefined in the console spec, so it's unclar if we can get
-   interoperable behaviour here.
+   Note: The formatter operation is underdefined in the console specification,
+   formatting can be inconsistent between different implementations.
 
 1. For each |arg| in |formatted args|:
 


jgraham (approved):
LGTM, but someone from outside Mozilla should also approve this ;)

juliandescottes:
Thanks for taking a look @jgraham 
cc @foolip @shs96c for a non-mozilla POV

whimboo:
@sadym-chromium mind having a look if this clarification is fine with you?

PR #204, opened 2022-05-09
Add Network module

jgraham:
This is designed to support logging network requests. Important use
cases include:

* Determing if a specific network request occured
* Monitoring request performance
* Generating a HAR file for offline analysis of requests

This does not currently attempt to support network request
interception (i.e. mutable access to requests), but the intent is that
the same lifecycle events could be used in a blocking way.

It also currently only supports HTTP-type requests and not
e.g. WebSockets. Some support for integration with service workers is
also missing.

The typical order of events is

network.beforeRequestSent - before request is sent, would later be the
right point to change the request headers or body or prevent the
request entirely.

network.responseStarted - after response headers are received, but
before body. Would later be a point to override the response headers
or body.

network.responseCompleted - after the request is fully complete.

network.fetchError - After any error that will prevent the request
from completing.

Compared to CDP this is missing an event for data being
received. Compared to WebExtensions, this is missing an event after
the headers are sent but before the data is sent.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/204.html" title="Last updated on Mar 29, 2023, 3:44 PM UTC (057aa4c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/204/71abe1b...057aa4c.html" title="Last updated on Mar 29, 2023, 3:44 PM UTC (057aa4c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -170,10 +170,27 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
     text: shared worker; url: workers.html#shared-workers
     text: window open steps; url: window-object.html#window-open-steps
     text: worker event loop; url: webappapis.html#worker-event-loop-2
-    text: worklet global scopes; url: worklets.html#concept-document-worklet-global-scopes
+    text: worklet global scopes; url:worklets.html#concept-document-worklet-global-scopes
+spec: RESOURCE-TIMING; urlPrefix: https://w3c.github.io/resource-timing/
+  type: dfn
+    text: convert fetch timestamp; url: dfn-convert-fetch-timestamp
+spec: HR-TIME; urlPrefix: https://w3c.github.io/hr-time/
+  type: dfn
+    text: get time origin timestamp; url: dfn-get-time-origin-timestamp
 spec: RFC4648; urlPrefix: https://tools.ietf.org/html/rfc4648
   type: dfn
     text: Base64 Encode; url: section-4
+</pre>
+
+<pre class="biblio">
+{
+    "SAME-SITE-COOKIES": {
+        "authors": ["Mike West", "Mark Goodwin"],
+        "href": "https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-cookie-same-site",
+        "publisher": "IETF",
+        "title": "Same-Site Cookies"
+    }
+}
 </pre>
 
 <style>
@@ -271,6 +288,7 @@ Command = {
 CommandData = (
   BrowsingContextCommand //
   InputCommand //
+  NetworkCommand //
   ScriptCommand //
   SessionCommand
 )
@@ -304,10 +322,11 @@ ErrorResponse = {
 }
 
 ResultData = (
-  EmptyResult //
-  SessionResult //
   BrowsingContextResult //
-  ScriptResult
+  EmptyResult //
+  NetworkResult //
+  ScriptResult //
+  SessionResult
 )
 
 EmptyResult = {
@@ -321,8 +340,9 @@ Event = {
 
 EventData = (
   BrowsingContextEvent //
-  ScriptEvent //
-  LogEvent
+  LogEvent //
+  NetworkEvent //
+  ScriptEvent
 )
 </pre>
 
@@ -578,6 +598,19 @@ To <dfn>obtain a set of event names</dfn> given an |name|:
    |events|.
 
 1. Return [=success=] with data |events|.
+
+</div>
+
+<div algorithm>
+To <dfn>emit events</dfn> given |body| and |related browsing contexts|:
+
+1. [=Assert=]: |body| has [=map/size=] 2 and [=contains=] "<code>method</code>"
+   and "<code>params</code>".
+
+1. For each |session| in the [=set of sessions for which an event is enabled=]
+   given |body|["method"] and |related browsing contexts|:
+
+  1. [=Emit an event=] with |session| and |body|.
 
 </div>
 
@@ -3085,7 +3118,6 @@ The <dfn export for=commands>browsingContext.captureScreenshot</dfn> command
 captures an image of the given browsing context, and returns it as a
 Base64-encoded string.
 
-
 <dl>
    <dt>Command Type</dt>
    <dd>
@@ -4168,6 +4200,799 @@ opened</dfn> steps given |window|, |type| and |message|.
 
 </div>
 
+## The network Module ## {#module-network}
+
+The <dfn export for=modules>network</dfn> module contains commands and events
+relating to network requests.
+
+### Definition ### {#module-network-definition}
+
+[=remote end definition=]
+
+<pre class="cddl remote-cddl">
+
+NetworkCommand = (
+)
+
+</pre>
+
+[=local end definition=]
+
+<pre class="cddl local-cddl">
+
+NetworkResult = (
+)
+
+NetworkEvent = (
+    network.BeforeRequestSent //
+    network.FetchError //
+    network.ResponseStarted //
+    network.ResponseCompleted
+)
+
+</pre>
+
+A [=remote end=] has a <dfn>before request sent map</dfn> which is initially an
+empty map.  It's used to track the network events for which a
+<code>network.beforeRequestSent</code> event has already been sent.
+
+### Types ### {#module-network-types}
+
+#### The network.BaseParameters type #### {#type-network-BaseParameters}
+
+<pre class="cddl local-cddl">
+network.BaseParameters = {
+    context: BrowsingContext / null,
+    navigation: Navigation / null,
+    redirectCount: js-uint,
+    request: network.RequestData,
+    timestamp: js-uint,
+}
+</pre>
+
+The <code>network.BaseParameters</code> type is an abstract type representing
+the data that's common to all network events.
+
+Issue: Consider including the `sharedId` of the document node that initiated the
+request in addition to the context.
+
+<div algorithm>
+To <dfn>get the base network event data</dfn> given |request| and |redirect count|:
+
+1. Let |request data| be the result of [=get the request data=] with |request|.
+
+1. Let |navigation| be |request|'s [=request/navigation id=].
+
+1. Let |context id| be null.
+
+1. If |request|'s [=request/client=] is not null, let |related browsing contexts|
+   be the result of [=get related browsing contexts=] with |request|'s
+   [=request/client=]. Otherwise let |related browsing contexts| be an empty
+   set.
+
+1. If |request|'s [=request/window=] is an [=environment settings object=]:
+
+  1. Let |environment settings| be |request|'s [=request/window=]
+
+  1. If there is a [=/browsing context=] whose [=active window=] is |environment
+     settings|' [=environment settings object/global object=], let |context id|
+     be the [=browsing context id=] for that context.
+
+1. Let |timestamp| be a [=time value=] representing the current date and time in UTC.
+
+1. Let |params| be [=/map=] matching the <code>network.BaseParameters</code>
+   production, with the <code>request</code> field set to |request data|, the
+   |navigation| field set to <code>navigation</code>, the <code>context</code>
+   field set to |context id|, the <code>timestamp</code> field set to
+   |timestamp|, and the <code>redirectCount</code> field set to |redirect count|.
+
+1. Return (|related browsing contexts|, |params|)
+
+</div>
+
+#### The network.Cookie type #### {#type-network-Cookie}
+
+[=Remote end definition=] and [=local end definition=]
+
+<pre class="cddl local-cddl">
+network.Cookie = {
+    name: text,
+    ? value: text,
+    ? binaryValue: [ uint ]
+    domain: text,
+    path: text,
+    ? expires: js-uint,
+    size: js-uint,
+    httpOnly: bool,
+    secure: bool,
+    sameSite: "strict" / "lax" / "none",
+};
+</pre>
+
+The <code>network.Cookie</code> type represents a cookie.
+
+If the cookie value can be represented as a UTF-8 encoded string, the
+<code>value</code> field will be present. Otherwise the <code>binaryValue</code>
+field will be present and consist of an array of integers representing the bytes
+of the cookie value.
+
+<div algorithm>
+To <dfn>get a cookie</dfn> given |stored cookie|:
+
+Note: The definitions of |stored cookie|'s fields are from [[COOKIES]], except
+samesite-flag, which is from [[SAME-SITE-COOKIES]].
+
+1. Let |name| be the result of [=UTF-8 decode=] with |stored cookie|'s name field.
+
+1. Let |utf8 decoded value| be the result of [=UTF-8 decode without BOM or fail=]
+   with |stored cookie|'s value.
+
+1. If |utf8 decoded value| is failure, then:
+
+    1. Let |value| be null and |binary value| be an empty list.
+
+    1. For each |byte| in |stored cookie|'s value:
+
+      1. Append the [=byte/value=] of |byte| to |binary value|.
+
+   Otherwise: Let |value| be |utf8 decoded value| and |binary value| be null.
+
+1. Let |domain| be |stored cookie|'s domain field.
+
+1. Let |path| be |stored cookie|'s path field.
+
+1. Let |expires| be |stored cookie|'s expires field represented as a unix
+   timestamp, if set, or null otherwise.
+
+1. Let |size| be the byte length of the result of serializing |stored cookie|
+   as it would be represented in a <code>Cookie</code> header.
+
+1. Let |http only| be true if |stored cookie|'s http-only-flag is true, or false
+   otherwise.
+
+1. Let |secure| be true if |stored cookie|'s secure-only-flag is true, or false
+   otherwise.
+
+1. Let |same site| be "<code>none</code>" if |stored cookie|'s samesite-flag is
+   "<code>None</code>", "<code>lax</code>" if it is "<code>Lax</code>, or
+   "<code>strict</code>" if it is "<code>Strict</code>".
+
+1. Return a map matching the <code>network.Cookie</code> production,
+   with the <code>name</code> field set to |name|, the <code>value</code> field
+   set to |value| if it's not null or omitted otherwise, the
+   <code>binaryValue</code> field set to |binary value| if it's not null or
+   omitted otherwise, the <code>domain</code> field set to |domain|, the
+   <code>path</code> field set to |path|, the <code>expires</code> field set to
+   |expires| if it's not null, or omitted otherwise, the <code>size</code> field
+   set to |size|, the <code>httpOnly</code> field set to |http only|, the
+   <code>secure</code> field set to |secure|, and the <code>sameSite</code>
+   field set to |same site|.
+
+</div>
+
+#### The network.FetchTimingInfo type #### {#type-network-FetchTimingInfo}
+
+[=Remote end definition=] and [=local end definition=]
+
+<pre class="cddl local-cddl">
+network.FetchTimingInfo = {
+    timeOrigin: float,
+    requestTime: float,
+    redirectStart: float,
+    redirectEnd: float,
+    fetchStart: float,
+    dnsStart: float,
+    dnsEnd: float,
+    connectStart: float,
+    connectEnd: float,
+    tlsStart: float,
+    <!-- tlsEnd: float this should be the same as connectEnd -->
+    requestStart: float,
+    responseStart: float,
+    <!-- TODO responseHeadersEnd: float: Not sure quite what to use for this -->
+    responseEnd: float,
+};
+</pre>
+
+The <code>network.FetchTimingInfo</code> type represents the time of each part
+of the request, relative to the <a spec=html>time origin</a> of the [=/request=]'s
+[=request/client=].
+
+<div algorithm>
+To <dfn>get the fetch timings</dfn> given |request|:
+
+1. Let |global| be |request|'s [=request/client=].
+
+1. If |global| is null, return a map matching the
+   <code>network.FetchTimingInfo</code> production, with all fields set to 0.
+
+1. Let |time origin| be [=get time origin timestamp=] with |global|.
+
+1. Let |timings| be |request|'s [=fetch timing info=].
+
+1. Let |connection timing| be |timings|' [=final connection timing info=] if
+   it's not null, or a new [=connection timing info=] otherwise.
+
+1. Let |request time| be [=convert fetch timestamp=] given |timings|' <a spec=fetch>start time</a> and |global|.
+
+1. Let |redirect start| be [=convert fetch timestamp=] given |timings|'
+   [=redirect start time=] and |global|.
+
+1. Let |redirect end| be [=convert fetch timestamp=] given |timings|' [=redirect
+   end time=] and |global|.
+
+1. Let |fetch start| be [=convert fetch timestamp=] given |timings|'
+   [=post-redirect start time=] and |global|.
+
+1. Let |DNS start| be [=convert fetch timestamp=] given |connection timing|'s
+   [=domain lookup start time=] and |global|.
+
+1. Let |DNS end| be [=convert fetch timestamp=] given |connection timing|'s
+   [=domain lookup end time=] and |global|.
+
+1. Let |TLS start| be [=convert fetch timestamp=] given |connection timing|'s
+   [=secure connection start time=] and |global|.
+
+1. Let |connect start| be [=convert fetch timestamp=] given |connection
+   timing|'s [=connection start time=] and |global|.
+
+1. Let |connect end| be [=convert fetch timestamp=] given |connection timing|'s
+   [=connection end time=] and |global|.
+
+1. Let |request start| be [=convert fetch timestamp=] given |timings|' [=final
+   network-request start time=] and |global|.
+
+1. Let |response start| be [=convert fetch timestamp=] given |timings|' [=final
+   network-response start time=] and |global|.
+
+1. Let |response end| be [=convert fetch timestamp=] given |timings|'
+   <a spec=fetch>end time</a> and |global|.
+
+1. Return a [=/map=] matching the <code>network.FetchTimingInfo</code> production
+    with the <code>timeOrigin</code> field set to |time origin|, the
+    <code>requestTime</code> field set to |request time|, the
+    <code>redirectStart</code> field set to |redirect start|, the
+    <code>redirectEnd</code> field set to |redirect end|, the
+    <code>fetchStart</code> field set to |fetch start|, the
+    <code>dnsStart</code> field set to |DNS start|, the <code>dnsEnd</code>
+    field set to |DNS end|, the <code>connectStart</code> field set to |connect
+    start|, the <code>connectEnd</code> field set to |connect end|, the
+    <code>tlsStart</code> field set to |TLS start|, the
+    <code>requestStart</code> field set to |request start|, the
+    <code>responseStart</code> field set to |response start|, and the
+    <code>responseEnd</code> field set to |response end|.
+
+</div>
+
+TODO: Add service worker fields
+
+#### The network.Header type #### {#type-network-Header}
+
+[=Remote end definition=] and [=local end definition=]
+
+<pre class="cddl local-cddl">
+network.Header = {
+    name: text,
+    ? value: text,
+    ? binaryValue: [ uint ]
+};
+</pre>
+
+The <code>network.Header</code> type represents a single request header.
+
+If the header value can be represented as a UTF-8 encoded string, the
+<code>value</code> field will be present. Otherwise the <code>binaryValue</code>
+field will be present and consist of an array of integers representing the bytes
+of the header.
+
+<div algorithm>
+To <dfn>get a header</dfn> given |name bytes| and |value bytes|:
+
+1. Let |name| be the result of [=UTF-8 decode=] with |name bytes|.
+
+   Assert: Since header names are constrained to be ASCII-only this cannot fail.
+
+1. Let |utf8 decoded value| be the result of [=UTF-8 decode without BOM or fail=]
+   with |value bytes|.
+
+1. If |utf8 decoded value| is failure, then:
+
+    1. Let |value| be null and |binary value| be an empty list.
+
+    1. For each |byte| in |value bytes|:
+
+      1. Append the [=byte/value=] of |byte| to |binary value|.
+
+   Otherwise: let |value| be |utf8 decoded value| and let |binary value|
+   be null.
+
+1. Return a map matching the <code>network.Header</code> production, with the
+   <code>name</code> field set to |name|, the <code>value</code> field
+   set to |value| if it's not null, or omitted otherwise, and the
+   <code>binaryValue</code> field set to |binary value| if it's not null, or
+   omitted otherwise.
+
+</div>
+
+#### The network.Initiator type #### {#type-network-Initiator}
+
+[=Remote end definition=] and [=local end definition=]
+
+<pre class="cddl local-cddl">
+network.Initiator = {
+    type: "parser" / "script" / "preflight" / "other",
+    ?columnNumber: js-uint,
+    ?lineNumber: js-uint,
+    ?stackTrace: script.StackTrace,
+    ?request: network.Request
+};
+</pre>
+
+The <code>network.Initiatior</code> type represents the source of a network
+request.
+
+<div algorithm>
+To <dfn>get the initiator</dfn> given |request|:
+
+1. Let |request id| be |request|'s [=request id=].
+
+1. Let |type| be "<code>other</code>".
+
+1. If |request| is a [=CORS-Preflight Request=], set |type| to
+   "<code>preflight</code>".
+
+1. TODO: Get the |type|. It's not quite clear how this ought to work; the CDP
+   data depends on whether the navigation was kicked off by the parser or by
+   script (so e.g. inserting an image from script causes the initiator to be
+   "<code>script</code>"), but that doesn't correspond to anything in Fetch.
+
+1. If |request|'s [=request/initiator type=] is "<code>fetch</code>" or
+   "<code>xmlhttprequest</code>":
+
+  1. Let |stack trace| be the [=current stack trace=].
+
+  1. If |stack trace| has size of 1 or greater, let |line number| be value of the
+     <code>lineNumber</code> field in |stack trace|[0], and let |column number| be
+     the value of the <code>columnNumber</code> field |stack trace|[0]. Otherwise
+     let |line number| and |column number| be 0.
+
+  Otherwise, let |stack trace|, |column number|, and |line number| all be null.
+
+  TODO: Chrome includes the current parser position as column number / line
+  number for parser-inserted resources.
+
+1. Return a [=/map=] matching the <code>network.Initiator</code> production, with
+   the <code>type</code> field set to |type|, the <code>columnNumber</code>
+   field set to |column number| if it's not null, or omitted otherwise, the
+   <code>lineNumber</code> field set to |line number| if it's not null, or
+   omitted otherwise, the <code>stackTrace</code> field set to |stack trace| if
+   it's not null, or omitted otherwise, and the <code>request</code> field set
+   to |request id|.
+
+</div>
+
+#### The network.Request type #### {#type-network-Request}
+
+[=Remote end definition=] and [=local end definition=]
+
+<pre class="cddl local-cddl remote-cddl">
+network.Request = text;
+</pre>
+
+Each network request has an associated <dfn export>request id</dfn>, which is a
+string uniquely identifying that request. The identifier for a request resulting from a
+redirect matches that of the request that initiated it.
+
+#### The network.RequestData type #### {#type-network-RequestData}
+
+[=Remote end definition=] and [=local end definition=]
+
+<pre class="cddl local-cddl">
+network.RequestData = {
+    request: network.Request,
+    url: text,
+    method: text,
+    headers: [*network.Header],
+    cookies: [*network.Cookie],
+    headersSize: js-uint,
+    bodySize: js-uint / null,
+    timings: network.FetchTimingInfo,
+};
+</pre>
+
+The <code>network.RequestData</code> type represents an ongoing network request.
+
+<div algorithm>
+
+To <dfn>get the request data</dfn> given |request|:
+
+1. Let |request id| be request's [=request id=].
+
+1. Let |url| be the result of running the [=URL serializer=] with |request|'s
+   [=request/URL=].
+
+1. Let |method| be |request|'s [=request/method=].
+
+1. Let |body size| be null.
+
+1. Let |body| be request's [=request/body=].
+
+1. If |body| is a [=byte sequence=], let |body size| be the length of that
+   sequence. Otherwise, if |body| is a [=/body=] then let |body size| be that
+   body's <a spec=fetch>length</a>.
+
+1. Let |headers size| be the size in bytes of |request|'s [=request/headers list=] when
+   serialized as mandated by [[HTTP11]].
+
+   Note: For protocols which allow header compression, this is the compressed
+   size of the headers, as sent over the network.
+
+1. Let |headers| be an empty list.
+
+1. Let |cookies| be an empty list.
+
+1. For each (|name|, |value|) in |request|'s [=request/headers list=]:
+
+  1. Append the result of [=get a header=] with |name| and |value| to |headers|.
+
+  1. If |name| is a [=byte-case-insensitive=] match for "<code>Cookie</code>" then:
+
+    1. For each |cookie| in the user agent's cookie store that are included in
+       |request|:
+
+       Note: [[COOKIES]] defines some baseline requirements for which cookies in
+       the store can be included in a request, but user agents are free to
+       impose additional constraints.
+
+       1. Append the result of [=get a cookie=] given |cookie| to |cookies|.
+
+1. Let |timings| be [=get the fetch timings=] with |request|.
+
+1. Return a map matching the <code>network.RequestData</code> production, with
+   the <code>request</code> field set to |request id|, <code>url</code> field
+   set to |url|, the <code>method</code> field set to |method|, the
+   <code>headers</code> field set to |headers|, the |cookies| field set to
+   |cookies|, and the <code>headersSize</code> field set to |headers size|,
+   the <code>bodySize</code> field set to |body size|, and the
+   <code>timings</code> field set to |timings|.
+
+</div>
+
+#### The network.ResponseContent type #### {#type-network-ResponseContent}
+
+[=Remote end definition=] and [=local end definition=]
+
+<pre class="cddl local-cddl">
+network.ResponseContent = {
+    size: js-uint
+};
+</pre>
+
+The <code>network.ResponseContent</code> type represents the decoded response to
+a network request.
+
+<!-- Not sure this is worthwhile if it only ends up with a single field, but it
+would be natural to add a field here if we have a way to return the body -->
+
+<div algorithm>
+To <dfn>get the response content info</dfn> given |response|.
+
+1. Return a new map matching the <code>network.ResponseContent</code>
+   production, with the <code>size</code> field set to |response|'s [=response body
+   info=]'s [=decoded size=]
+
+</div>
+
+#### The network.ResponseData type #### {#type-network-ResponseData}
+
+[=Remote end definition=] and [=local end definition=]
+
+<pre class="cddl local-cddl">
+network.ResponseData = {
+    url: text,
+    protocol: text,
+    status: js-uint,
+    statusText: text,
+    fromCache: bool,
+    headers: [*network.Header],
+    mimeType: text,
+    bytesReceived: js-uint,
+    headersSize: js-uint / null,
+    bodySize: js-uint / null,
+    content: network.ResponseContent
+};
+</pre>
+
+The <code>network.ResponseData</code> type represents the response to a network
+request.
+
+<div algorithm>
+
+To <dfn>get the protocol</dfn> given |response|:
+
+1. Let |protocol| be the empty string.
+
+1. If |response|'s [=final connection timing info=] is not null, set |protocol|
+   to |response|'s [=final connection timing info=]'s [=ALPN negotiated
+   protocol=].
+
+1. If |protocol| is the empty string, or is equal to "<code>unknown</code>":
+
+  1. Set |protocol| to |response|'s [=response/url=]'s [=url/scheme=]
+
+  1. If |protocol| is equal to either "<code>http</code>" or
+     "<code>https</code>" and |response| has an associated HTTP Response.
+
+     Note: [[FETCH]] isn't clear about the relation between a HTTP network response and
+     a response object.
+
+     <!-- It would be better to move this into Fetch itself. -->
+
+    1. Let |http version| be the HTTP Response's Status line's HTTP-version [[HTTP11]].
+
+    1. If |http version| starts with "<code>HTTP/</code>":
+
+      1. Let |version| be the [=code unit substring=] of |http version| from 5
+         to |http version|'s [=string/length=].
+
+      1. If |version| is "<code>0.9</code>", set |protocol| to
+         "<code>http/0.9</code>", otherwise if |version| is "<code>1.0</code>",
+         set |protocol| to "<code>http/1.0</code>", otherwise if |version| is
+         "<code>1.1</code>", set |protocol| to "<code>http/1.1</code>".
+
+1. Return |protocol|.
+
+</div>
+
+<div algorithm>
+To <dfn>get the response data</dfn> given |response|:
+
+1. Let |url| be the result of running the [=URL serializer=] with |response|'s
+   [=response/URL=].
+
+1. Set |protocol| to [=get the protocol=] given |response|.
+
+1. Let |status| be |response|'s [=response/status=].
+
+1. Let |status text| be |response|'s [=status message=].
+
+1. If |response|'s [=cache state=] is "<code>local</code>", let |from cache| be
+   true, otherwise let it be false.
+
+1. Let |headers| be an empty list.
+
+1. Let |mime type| be the [=essence=] of the [=computed mime type=] for |response|.
+
+   Note: this is whatever MIME type the browser is actually using, even if it
+   isn't following the exact algorithm in the [[MIMESNIFF]] specification.
+
+1. For each (|name|, |value|) in |response|'s [=response/headers list=]:
+
+  1. Append the result of [=get a header=] with |name| and |value| to |headers|.
+
+1. Let |bytes received| be the total number of bytes transmitted as part of the
+   HTTP response associated with |response|.
+
+1. Let |headers size| be the number of bytes transmitted as part of the header
+   fields section of the HTTP response.
+
+1. Let |body size| be |response|'s [=response body info=]'s [=encoded size=].
+
+1. Let |content| be the result of [=get the response content info=] with |response|.
+
+1. Return a [=/map=] matching the <code>network.ResponseData</code> production,
+   with the <code>url</code> field set to |url|, the <code>protocol</code> field
+   set to |protocol|, the <code>status</code> field set to |status|, the
+   <code>statusText</code> field set to |status text|, the
+   <code>fromCache</code> field set to |from cache|, the <code>headers</code>
+   field set to |headers|, the <code>mimeType</code> field set to |mime type|, the
+   <code>bytesReceived</code> field set to |bytes received|, the
+   <code>headersSize</code> field set to |headers size|, the
+   <code>bodySize</code> field set to |body size|, and the <code>content</code>
+   field set to |content|.
+
+</div>
+
+### Events ### {#module-network-event}
+
+#### The network.beforeRequestSent Event #### {#event-network-beforeSendRequest}
+<dl>
+   <dt>Event Type</dt>
+   <dd>
+      <pre class="cddl local-cddl">
+        network.BeforeRequestSent = {
+         method: "network.beforeRequestSent",
+         params: network.BeforeRequestSentParameters
+       }
+
+       network.BeforeRequestSentParameters = {
+         network.BaseParameters,
+         initiator: network.Initiator,
+       }
+      </pre>
+   </dd>
+</dl>
+
+This event is emitted before a request is sent (either over the network or
+before it's handled by a serviceworker or a local cache).
+
+<div algorithm>
+The [=remote end event trigger=] is the <dfn export>WebDriver BiDi before
+request sent</dfn> steps given |request|:
+
+1. If [=before request sent map=] does not contain |request|, set [=before
+   request sent map=][|request|] to a new set.
+
+1. Let |redirect count| be |request|'s [=redirect count=].
+
+1. Add |redirect count| to [=before request sent map=][|request|].
+
+1. Let (|related browsing contexts|, |params|) be the result of [=get the base
+   network event data=] with |request|, and |redirect count|.
+
+1. Let |initiator| be the result of [=get the initiator=] with |request|.
+
+1. Set the <code>initiator</code> field of |params| to |initiator|.
+
+1. Assert: |params| matches the <code>network.BeforeRequestSentParameters</code>
+   production.
+
+1. Let |body| be a map matching the <code>network.BeforeRequestSent</code>
+   production, with the <code>params</code> field set to |params|.
+
+1. [=Emit events=] with |body| and |related browsing contexts|.
+
+</div>
+
+#### The network.fetchError Event #### {#event-network-fetchError}
+
+<dl>
+   <dt>Event Type</dt>
+   <dd>
+      <pre class="cddl local-cddl">
+        network.FetchError = {
+         method: "network.fetchError",
+         params: network.FetchErrorParameters
+       }
+
+       network.FetchErrorParameters = {
+         network.BaseParameters,
+         errorText: text,
+       }
+      </pre>
+   </dd>
+</dl>
+
+This event is emitted when a network request ends in an error.
+
+<div algorithm>
+
+The [=remote end event trigger=] is the <dfn export>WebDriver BiDi fetch
+error</dfn> steps given |request|:
+
+1. If [=before request sent map=][|request|] does not contain |request|'s
+   [=redirect count=], then run the [=WebDriver BiDi before request sent=] steps
+   with |request|.
+
+   Note: This ensures that a <code>network.beforeRequestSent</code> can
+   always be emitted before a <code>network.fetchError</code>, without the
+   caller needing to explicitly invoke the [=WebDriver BiDi before request
+   sent=] steps on every error path.
+
+1. Let (|related browsing contexts|, |params|) be the result of [=get the
+   base network event data=] given |request|.
+
+1. TODO: Set the <code>errorText</code> field of |params|.
+
+1. Assert: |params| matches the <code>network.FetchErrorParameters</code>
+   production.
+
+1. Let |body| be a map matching the <code>network.FetchError</code>
+   production, with the <code>params</code> field set to |params|.
+
+1. [=Emit events=] with |body| and |related browsing contexts|.
+
+</div>
+
+#### The network.responseCompleted Event #### {#event-network-responseCompleted}
+
+<dl>
+   <dt>Event Type</dt>
+   <dd>
+      <pre class="cddl local-cddl">
+        network.ResponseCompleted = {
+         method: "network.responseCompleted",
+         params: network.ResponseCompletedParameters
+       }
+
+       network.ResponseCompletedParameters = {
+         network.BaseParameters,
+         response: network.ResponseData,
+       }
+      </pre>
+   </dd>
+</dl>
+
+This event is emitted after the full response body is received.
+
+<div algorithm>
+The [=remote end event trigger=] is the <dfn export>WebDriver BiDi response
+completed</dfn> steps given |request| and |response|:
+
+1. Let |redirect count| be |request|'s [=redirect count=].
+
+1. Assert: [=before request sent map=][|request|] contains |redirect count|.
+
+   Note: This implies that every caller needs to ensure that the [=WebDriver BiDi
+   before request sent=] steps are invoked with |request| before these steps.
+
+1. Let (|related browsing contexts|, |params|) be the result of [=get the
+   base network event data=] given |request| and |redirect count|.
+
+1. Let |response data| be the result of [=get the response data=] with |response|.
+
+1. Set the <code>response</code> field of |params| to |response data|.
+
+1. Assert: |params| matches the <code>network.ResponseCompletedParameters</code>
+   production.
+
+1. Let |body| be a map matching the <code>network.ResponseCompleted</code>
+   production, with the <code>params</code> field set to |params|.
+
+1. [=Emit events=] with |body| and |related browsing contexts|.
+
+</div>
+
+#### The network.responseStarted Event #### {#event-network-responseStarted}
+
+<dl>
+   <dt>Event Type</dt>
+   <dd>
+      <pre class="cddl local-cddl">
+        network.ResponseStarted = {
+         method: "network.responseStarted",
+         params: network.ResponseStartedParameters
+       }
+
+       network.ResponseStartedParameters = {
+         network.BaseParameters,
+         response: network.ResponseData,
+       }
+      </pre>
+   </dd>
+</dl>
+
+This event is emitted after the response headers are received but before the
+body is complete.
+
+<div algorithm>
+The [=remote end event trigger=] is the <dfn export>WebDriver BiDi response
+started</dfn> steps given |request| and |response|:
+
+1. Let |redirect count| be |request|'s [=redirect count=].
+
+1. Assert: [=before request sent map=][|request|] is equal to |redirect count|.
+
+   Note: This implies that every caller needs to ensure that the [=WebDriver BiDi
+   before request sent=] steps are invoked with |request| before these steps.
+
+1. Let (|related browsing contexts|, |params|) be the result of [=get the base
+   network event data=] with |request| and |redirect count|.
+
+1. Let |response data| be the result of [=get the response data=] with |response|.
+
+1. Set the <code>response</code> field of |params| to |response data|.
+
+1. Assert: |params| matches the <code>network.ResponseStartedParameters</code>
+   production.
+
+1. Let |body| be a map matching the <code>network.ResponseStarted</code>
+   production, with the <code>params</code> field set to |params|.
+
+1. [=Emit events=] with |body| and |related browsing contexts|.
+
+</div>
+
 ## The script Module ## {#module-script}
 
 The <dfn export for=modules>script</dfn> module contains commands and events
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3296,6 +3300,270 @@ failed</dfn> steps given |context| and |navigation status|:
...
+<code>value</code> field will be present. Otherwise the <code>binaryValue</code>
+field will be present and consist of an array of integers representing the bytes
+of the cookie value.
+
+#### The network.FetchTimingInfo type #### {#type-network-FetchTimingInfo}
+
+[=Remote end definition=] and [=local end definition=]
+
+<pre class="cddl local-cddl">
+NetworkFetchTimingInfo = {
```
foolip, line 3367: This is in parts similar to https://w3c.github.io/navigation-timing/#the-performancetiming-interface, is the intention that the definitions be taken from that spec?

jgraham: Good question! So in my ideal world this would be exactly like performance timing, since that will be easy to specify and implement. But I think performance timing is a bit different from what's in CDP (https://chromedevtools.github.io/devtools-protocol/tot/Network/#type-ResourceTiming) so some input on what's actually required here would be very helpful. @janodvarko might also have some input on what's required here e.g. for HAR export.

janodvarko: Performance timing spec is missing "requestEnd" attribute that could be used to calculate time needed to send a request: `requestEnd` - `requestStart` == `wait` in HAR spec. (this might be especially useful for requests sending POST data).

CDP has both `sendStart` and `sendEnd`, so from this perspective CDP timing is closer to HAR (i.e. it has all HAR needs)

HAR also needs `onLoad` and `domContentLoaded` timings. Those are defined by "Navigation Timing interface" https://w3c.github.io/navigation-timing/#dom-performancenavigationtiming

Does CDP has an interface for "NavigationTiming"?

HAR spec request timings: http://www.softwareishard.com/blog/har-12-spec/#timings
HAR spec page timings: http://www.softwareishard.com/blog/har-12-spec/#pageTimings

jgraham: CDP has some events in the Page domain that allow getting timestamps for load-related events e.g. https://chromedevtools.github.io/devtools-protocol/tot/Page/#event-domContentEventFired . I think we could add timestamps to similar events in `browsingContext` for that part of the use case.

jgraham:
On redirects: it seems like what CDP does is just sends the `network.responseReceivedExtraInfo` event, and not the `network.responseReceived` event when a response will be a redirect. Presumably the advantage of this is that the client that doesn't care about the individual redirects can just listen for `network.responseReceived` and ignore the intermediate `netowrk.responseReceivedExtraInfo` events. You also get a `network.requestWillBeSent[ExtraInfo]` pair per request, with the `redirectResponse` property of the event set to the `network.Response` data you would have got from a `network.ResponseReceived` event if there had been one.

jgraham:
So, for redirects I have two possible suggestions: we either have a property on events so that they're identified as a redirect, or we have different event names for response/request pairs caused by redirects.

In the former case, clients that just want to know when a request is complete and don't care about following the full chain have to listen for all the events, but can filter out any with (e.g.) an `isRedirect` flag set.

In the latter case we end up with `redirectResponseStarted` and `beforeSendRedirectRequest` events (or some similar names) and it has the advantage that clients can just not subscribe to those events if they don't care about redirects, but the disadvantage that we're making things more complex for clients that do care about following the redirect chain.

index.bs
```diff
@@ -3296,6 +3300,270 @@ failed</dfn> steps given |context| and |navigation status|:
...
+        NetworkBeforeRequestSentEvent = {
+         method: "network.beforeRequestSent",
+         params: NetworkBeforeRequestSentParams
+       }
+
+       NetworkBeforeRequestSentParams = {
+        request: Request,
+        navigation: Navigation / null,
+        context: BrowsingContext / null,
+        requestData: RequestData,
```
juliandescottes, line 3492: There's no definition for (Network?)RequestData but we have an unused NetworkRequestInfo. If that's what this is supposed to be, then we should have request cookies as well in NetworkRequestInfo?

jgraham: Thanks! I renamed `NetworkRequestInfo` to `NetworkRequestData` and added the cookies. I also cleaned up a few other misnamed things.

index.bs
```diff
@@ -3631,6 +3635,271 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+<pre class="cddl local-cddl">
+NetworkResponseData = {
+    url: text,
+    protocol: text,
+    status: unit,
+    statusText: text,
+    headers: [ *NetworkHeader ],
+    cookies: [ *NetworkCookie ],
+    mimeType: text,
+    bytesReceived: uint,
```
juliandescottes, line 3806: To build the `content` information of an entry we also need the content's `size`, which can differ from the bytesReceived depending on the compression. Actual content `text` and `compression` are optional though. 

> 
    size [number] - Length of the returned content in bytes. Should be equal to response.bodySize if there is no compression and bigger when the content has been compressed.
    compression [number, optional] - Number of bytes saved. Leave out this field if the information is not available.
    mimeType [string] - MIME type of the response text (value of the Content-Type response header). The charset attribute of the MIME type is included (if available).
    text [string, optional] - Response body sent from the server or loaded from the browser cache. This field is populated with textual content only. The text field is either HTTP decoded text or a encoded (e.g. "base64") representation of the response body. Leave out this field if the information is not available.
    encoding [string, optional] (new in 1.2) - Encoding used for response text field e.g "base64". Leave out this field if the text field is HTTP decoded (decompressed & unchunked), than trans-coded from its original character set into UTF-8.
    comment [string, optional] (new in 1.2) - A comment provided by the user or the application.

(https://w3c.github.io/web-performance/specs/HAR/Overview.html#content)

juliandescottes (commented):
Adding some additional fly-by comments based on a prototype building HAR files using the current events. 

There's also the `cache` entry which does not correspond to anything currently defined in the events, but I think all fields are marked as optional (https://w3c.github.io/web-performance/specs/HAR/Overview.html#cache).

index.bs
```diff
@@ -3631,6 +3635,271 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+      <pre class="cddl local-cddl">
+        NetworkBeforeRequestSentEvent = {
+         method: "network.beforeRequestSent",
+         params: NetworkBeforeRequestSentParams
+       }
+
+       NetworkBeforeRequestSentParams = {
+        request: NetworkRequest,
+        navigation: Navigation / null,
+        context: BrowsingContext / null,
```
juliandescottes, line 3826: For HAR generation, we normally have to map each request to a specific "page" (https://w3c.github.io/web-performance/specs/HAR/Overview.html#pages)

I am wondering if navigation + context will be enough information to match a request with a specific page. Imagine a tab navigates from pageA to pageB. For a late (non-navigation related) request triggered in pageA, we would not get a `navigation` information, only `context`. And since the `context` id remains identical for a top-level browsing context across navigations, we might have to "guess" the page based on when we process this event. This might work but feels racy and I know it's already a complaint users have for HAR creation in Firefox.

Exposing an id unique to the window global could help here, although that's something we would need to also expose in browsing context events?

foolip:
@jgraham this PR was mentioned (by @whimboo) in Berlin. It's a draft, which is mentioned in comments above, but would you like review to get it out of draft, or do you want to do some more work first?

jgraham:
OK, I've marked this as ready for review, although it depends on an integration in the fetch spec that currently doesn't exist. In the meantime I'd appreciate review on the API design and featureset, in the hope that we can have a pretty solid idea of what we need in Fetch before getting review there.

index.bs
```diff
@@ -3631,6 +3635,271 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+TODO: Body is actually bytes, not clear how to handle it as text
+
+#### The network.ResponseData type #### {#type-network-ResponseData}
+
+[=Remote end definition=] and [=local end definition=]
+
+<pre class="cddl local-cddl">
+NetworkResponseData = {
+    url: text,
+    protocol: text,
```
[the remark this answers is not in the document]
jgraham: OK, that's what `protocol` is, but CDP only has it in the response. In `beforeRequestSent` I think we might not know the protocol version that will actually be used yet, because even if the application layer says HTTP/1.1 we can negotiate an upgrade to HTTP/2 as part of the initial TLS handshake.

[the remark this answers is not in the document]
juliandescottes: Ah ok interesting, in my prototype I mapped the `protocol` here to the `protocolVersion` we have in devtools network data (TLSv1, TLSv1.1, TLSv1.2, TLSv1.3), that's why I was expecting another property. Thanks for the correction

index.bs
jgraham: Hmm, yeah, this is a good point. I wonder if we should defer this to after we have support for returning Window objects from scripts; I think that will end up with some kind of per-Window id along with the context id, and maybe we could reuse that here. Or we could use the id of the realm in which the request originated (although arguably that wouldn't make sense for a non-scripted request orginating in a document).

index.bs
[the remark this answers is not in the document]
jgraham: I don't _think_ CDP is exposing the low-level TLS details; do we have a use case for also doing that?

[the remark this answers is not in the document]
juliandescottes: FWIW, this information is not exposed in HAR, the only usage I know of is the DevTools Network Monitor security tab. I don't think Chrome DevTools show a similar security info for their requests.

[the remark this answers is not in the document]
jgraham: My proposal is that we consider this out of scope for now; it's easy enough to add `tlsVersion` or something similar later if required.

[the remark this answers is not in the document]
juliandescottes: sounds good to me! Regarding the absence of protocol in beforeRequestSent, I don't think this would be a blocker for HAR. We can still pickup the information from the response events corresponding to the same request.

index.bs
juliandescottes: We can leave this for later as well. In practice we should be able to infer the page from the time of the request for now (or we can leave the page information out in a first iteration as well)

sadym-chromium (commented):
I like the fact WebDriver Classic can be referenced for the action specification.

index.bs
whimboo: I just filed https://github.com/w3c/webdriver-bidi/issues/271 to add timestamps to `browsingContext` related load events.

foolip:
@whimboo is the needs-discussion label still needed?

whimboo:
> @whimboo is the needs-discussion label still needed?

Yes, this PR still needs to be discussed and we put it onto the agenda for TPAC.

whimboo:
As per action items from the W3C meeting @foolip wanted to review the PR.

@jgraham mind resolving the conflicts that meanwhile exist? Thanks.

index.bs
```diff
@@ -3631,6 +3634,377 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+
+<pre class="cddl local-cddl">
+NetworkCookie = {
+    name: text,
+    ? value: text,
+    ? binaryValue: [ uint ]
+    domain: text,
+    path: text,
+    expires: uint,
+    size: uint,
```
foolip, line 3681: This doesn't seem to be described or set anywhere. Is it needed, or is it redundant with `value` / `binaryValue`?

index.bs
```diff
@@ -3631,6 +3634,377 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+    httpOnly: boolean,
```
foolip, line 3682: This and other bits are also not referenced anywhere, but this one I can guess what it is from the name.

index.bs
```diff
@@ -3631,6 +3634,377 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+    tlsStart: uint,
+    tlsEnd: uint,
+    requestStart: uint,
+    responseStart: uint,
+    responseHeadersEnd: uint,
+    responseEnd: uint,
+};
+</pre>
+
+The <code>FetchTimingInfo</code> type represents the time of each part of
```
foolip, line 3719:
```suggestion
The <code>NetworkFetchTimingInfo</code> type represents the time of each part of
```

index.bs
```diff
@@ -3631,6 +3634,377 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+
+<pre class="cddl local-cddl">
+NetworkHeader = {
+    name: text,
+    ? value: text,
+    ? binaryValue: [ uint ]
+};
+</pre>
+
+The <code>Header</code> type represents a single request header.
```
foolip, line 3736:
```suggestion
The <code>NetworkHeader</code> type represents a single request header.
```

index.bs
```diff
@@ -3631,6 +3634,377 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+NetworkInitiator = {
+    type: "parser" / "script" / "preflight" / "other",
+    ?columnNumber: uint,
+    ?lineNumber: uint,
+    ?stackTrace: StackTrace,
+    ?request: NetworkRequest
+};
+</pre>
+
+The <code>Initiatior</code> type represents the source of a network request.
```
foolip, line 3785:
```suggestion
The <code>NetworkInitiator</code> type represents the source of a network request.
```

index.bs
```diff
@@ -3631,6 +3634,377 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+    url: text,
+    method: text,
+    headers: [ *NetworkHeader ],
+    cookies: [ *NetworkCookie ],
+    headersSize: uint,
+    bodySize: uint,
+};
+</pre>
+
+The <code>RequestData</code> type represents an ongoing network request.
```
foolip, line 3815:
```suggestion
The <code>NetworkRequestData</code> type represents an ongoing network request.
```

index.bs
```diff
@@ -3631,6 +3634,377 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+
+[=Remote end definition=] and [=local end definition=]
+
+<pre class="cddl local-cddl">
+NetworkResponseContent = {
+    size: unit / null
+};
+</pre>
+
+The <code>ResponseContent</code> type represents the decoded response to a
```
foolip, line 3877:
```suggestion
The <code>NetworkResponseContent</code> type represents the decoded response to a
```

index.bs
```diff
@@ -3631,6 +3634,377 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+    mimeType: text,
+    headersSize: uint / null,
+    bodySize: uint / null,
+    bytesReceived: uint,
+    content: NetworkResponseContent,
+    timings: NetworkFetchTimingInfo,
+};
+</pre>
+
+The <code>RequestData</code> type represents the response to a netowrk request.
```
foolip, line 3904:
```suggestion
The <code>NetworkResponseData</code> type represents the response to a network request.
```

Note that this isn't just prepending `Network`, also fixing typos.

index.bs
```diff
@@ -3631,6 +3634,377 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+The <code>Header</code> type represents a single request header.
+
+If the header value can be represented as a UTF-8 encoded string, the
+<code>value</code> field will be present. Otherwise the <code>binaryValue</code>
+field will be present and consist of an array of integers representing the bytes
+of the header.
+
+<div algorithm>
+
+To <dfn>get a Header</dfn> given |name bytes| and |value bytes|:
```
foolip, line 3745:
```suggestion
To <dfn>get a header</dfn> given |name bytes| and |value bytes|:
```

Or:

```suggestion
To <dfn>get a NetworkHeader</dfn> given |name bytes| and |value bytes|:
```

Or maybe something like "serialize a header", although this it not yet producing a string, but a map destined for a serializer...

index.bs
```diff
@@ -3631,6 +3634,377 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+1. Let |name| be the result of [=UTF-8 decode=] with |name bytes|.
+
+   Assert: Since header names are constrained to be ASCII-only this cannot fail.
+
+1. Let |utf8 decoded value| be the result of [=UTF-8 decode without BOM or fail=]
+   with |value byes|.
+
+1. If |utf8 decoded value| is failure, then:
+
+    1. let |value| be null and |binary value| be a list.
```
foolip, line 3756:
```suggestion
    1. Let |value| be null and |binary value| be an empty list.
```

index.bs
```diff
@@ -3631,6 +3634,377 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+    ?columnNumber: uint,
+    ?lineNumber: uint,
+    ?stackTrace: StackTrace,
+    ?request: NetworkRequest
+};
+</pre>
+
+The <code>Initiatior</code> type represents the source of a network request.
+
+TODO: Align more closely with Fetch here?
```
foolip, line 3787: What details in particular? The request type/source?

index.bs
```diff
@@ -3631,6 +3634,377 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+
+#### The network.Request type #### {#type-network-RequestId}
+
+[=Remote end definition=] and [=local end definition=]
+
+<pre class="cddl local-cddl remote-cddl">
+NetworkRequest = text;
+</pre>
+
+Each network request has an associated <dfn export>request id</dfn>, which is a string
```
foolip, line 3797: Do you want to say that the `NetworkRequest` type represents this request id?

index.bs
```diff
@@ -3631,6 +3634,377 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+uniquely identifying that request.
+
+#### The network.RequestData type #### {#type-network-RequestData}
+
+[=Remote end definition=] and [=local end definition=]
+
+<pre class="cddl local-cddl">
+NetworkRequestData = {
+    url: text,
+    method: text,
```
foolip, line 3807: Will this be upper case or lower case?

index.bs
```diff
@@ -3631,6 +3634,377 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+    headersSize: uint,
+    bodySize: uint,
+};
+</pre>
+
+The <code>RequestData</code> type represents an ongoing network request.
+
+<div algorithm>
+
+To <dfn>get the request data</div> given |request|:
```
foolip, line 3819:
```suggestion
To <dfn>get the request data</dfn> given |request|:
```

index.bs
```diff
@@ -3631,6 +3634,377 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+
+1. Let |url| be the result of running the [=URL serializer=] with |request|'s
+   [=URL=].
+
+1. Let |method| be |request|'s [=method=].
+
+1. let |body size| be 0.
```
foolip, line 3826:
```suggestion
1. Let |body size| be 0.
```

index.bs
```diff
@@ -3631,6 +3634,377 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+
+1. Let |body| be request's [=body=].
+
+1. If |body| is a [=byte sequence=], let |body size| be the length of that
+   sequence. Otherwise if |body| is a [=body=] then let |body size| be that
```
foolip, line 3831:
```suggestion
   sequence. Otherwise, if |body| is a [=body=] then let |body size| be that
```

index.bs
```diff
@@ -3631,6 +3634,377 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+   body's [=length=] if it's not null, or 0 otherwise.
+
+   TODO: Should we handle the case where the body length is not yet known?
```
foolip, line 3834: Do you mean when https://fetch.spec.whatwg.org/#concept-body is null? It would make sense to put null in the protcol for that case too I think.

index.bs
```diff
@@ -3631,6 +3634,377 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+
+1. Let |headers size| be the size in bytes of |request|'s [=headers list=] when
+   serialized as mandated by [=HTTP=].
```
foolip, line 3837: Does this effectively serializing the headers to strip out extra whitespace and stuff? That seems inefficient, do existing protocols work like this?

index.bs
```diff
@@ -3631,6 +3634,377 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+
+   Note: for protocols which allow header compression, this is the compressed
```
foolip, line 3839: I don't think this note follows from the above. If not for this note I'd assume that one should use a canonical and minimal representation, and then compress it.

Capitalize:

```suggestion
   Note: For protocols which allow header compression, this is the compressed
```

index.bs
```diff
@@ -3631,6 +3634,377 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+   size of the headers, as sent over the network.
+
+1. Let |headers| be an empty list.
+
+1. Let |cookies| be an empty list.
+
+1. For each (|name|, |value|) in |request|'s [=headers list=]:
+
+  1. Append the result of [=get a Header=] with |name| and |value| to |headers|.
```
foolip, line 3848:
```suggestion
  1. Append the result of [=get a header=] with |name| and |value| to |headers|.
```

index.bs
```diff
@@ -3631,6 +3634,377 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+<pre class="cddl local-cddl">
+NetworkResponseContent = {
+    size: unit / null
+};
+</pre>
+
+The <code>ResponseContent</code> type represents the decoded response to a
+network request.
+
+<!-- Not sure this is worthwhile if it only ends up with a single field, but it
```
foolip, line 3880: I think it's fairly strange to have a `content` field in the protocol that doesn't actually have the content. We should only do that if we're going to add it before anyone ships this I think. Is that the plan?

index.bs
```diff
@@ -3631,6 +3634,377 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+
+<pre class="cddl local-cddl">
+NetworkResponseData = {
+    url: text,
+    protocol: text,
+    status: uint,
+    statusText: text,
+    headers: [ *NetworkHeader ],
+    cookies: [ *NetworkCookie ],
+    mimeType: text,
```
foolip, line 3895: Note `contentType` to match the name of the header?

index.bs
```diff
@@ -3631,6 +3634,377 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+    headersSize: uint / null,
+    bodySize: uint / null,
+    bytesReceived: uint,
```
foolip, line 3898: Can this be different from `headersSize` + `bodySize`? A difference in the case of compression?

index.bs
```diff
@@ -3631,6 +3634,377 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+        requestData: NetworkRequestData,
+        initiator: NetworkInitiator
+        timestamp: number,
+       }
+      </pre>
+   </dd>
+</dl>
+
+The [=remote end event trigger=] is the <dfn export>WebDriver BiDi before
+    request send</dfn> steps given |request|:
```
foolip, line 3931:
```suggestion
request sent</dfn> steps given |request|:
```

And a missing TODO for the actual steps?

index.bs
```diff
@@ -3631,6 +3634,377 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+   <dd>
+      <pre class="cddl local-cddl">
+        NetworkFetchErrorEvent = {
+         method: "network.fetchError",
+         params: NetworkFetchErrorParams
+       }
+
+       NetworkFetchErrorParams = {
+        request: NetworkRequest,
+        errorText: text,
```
foolip, line 3947: Hmm, no code or anything? Is this text going to be implementation defined?

index.bs
```diff
@@ -3631,6 +3634,377 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+        timestamp: number,
```
foolip, line 3948: What kind of timestamp?

index.bs
```diff
@@ -3631,6 +3634,377 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+        navigaton: Navigation / null,
+        context: BrowsingContext / null,
+        responseData: NetworkResponseData,
+        timestamp: number,
+       }
+      </pre>
+   </dd>
+</dl>
+
+After the response headers are received but before the body.
```
foolip, line 4000: Similar to above.

index.bs
```diff
@@ -3631,6 +3634,377 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+        navigation: Navigation / null,
+        context: BrowsingContext / null,
+        responseData: NetworkResponseData,
+        timestamp: number,
+       }
+      </pre>
+   </dd>
+</dl>
+
+After the full response body is received.
```
foolip, line 3976: Something like this?

```suggestion
The event is fired after the full response body is received.
```

But is this sentence normative?

index.bs
```diff
@@ -3631,6 +3634,377 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+      </pre>
+   </dd>
+</dl>
+
+After the response headers are received but before the body.
+
+The [=remote end event trigger=] is the <dfn export>WebDriver BiDi before
+response started</dfn> steps given |request| and |response|:
+
+1. Let |request data| be the result of [=get the request data=] given |request|.
```
foolip, line 4005: TODO for the rest?

index.bs
jgraham: It's uppercase, since it matches what's in Fetch.

index.bs
jgraham: I think the question is this: If the body is a stream, do we want the length to be the number of bytes that have been written so far, or 0, or null? I can't actually tell what behaviour the spec currently requires because it seems like updating the body length is (probably) defered to a caller-defined [process request body chunk length](https://fetch.spec.whatwg.org/#fetch-params-process-request-body) algorithm, but I have no idea if that's always set to something non-null (I can't actually find it _ever_ being set, so presumably I'm missing something).

But this is definitely "you're in a maze of twisy algorithms, all alike" territory for me, so maybe the length field is updated somewhere I'm not seeing.

index.bs
jgraham: Well it gives you information about the content :)

For request interception, we for-sure need some way to get the content, so if we want to reuse the same types/events for that I think it makes sense to design for the use case where we will indeed have a way to get at the response content.

index.bs
jgraham: Well this matches CDP, and is supposed to be the post-sniffing MIME type rather than a serialization of the content type header.

jgraham:
I updated the naming to match https://github.com/w3c/webdriver-bidi/pull/319 so I've resolved all the issues relating to that, assuming that PR will land before this one.

jgraham:
Current known/open issues:
* Do we need a separate timestamp for `responseHeadersEnd`? Navigation timing doesn't currently expose that, but perhaps the timestamp on the `network.responseStarted` event is enough?
* Initiator: CDP exposes the diffrence between subresources loaded by the parser and subresources loaded via a script (e.g. the difference between `` encountered during parsing and `var img = document.createElement("img"); img.src = "foo.png"; document.body.appendChild(img)`). That isn't currently exposed anywhere else in the platform. What are the use cases for that feature: is it important to e.g. Puppeteer users?
* How to handle the `bodySize` when the body length is not yet known
* Cookies: do we include all the cookies that could have been sent (like CDP), and add reasons where some weren't sent, or is it enough for now just to include the cookies that are actually sent?
* How to get the protocol used for the connection / final MIME type (mostly spec questions).
* What level of detail do we want on errors? Should we have certain predefined error reasons (e.g. for requests that are blocked by CSP).
* What invariants do we want if you subscribe in the middle of a request? Should you always get a `beforeRequestSent` or if you "missed" that event, do you just get a later lifecycle event first?

jgraham:
https://github.com/whatwg/fetch/pull/1540 is a PR in fetch for the integration here

index.bs
```diff
@@ -3701,6 +3727,792 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+
+  1. Let |environment settings| be |request|'s [=request/window=]
+
+  1. If there is a [=/browsing context=] whose [=active window=] is |environment
+     settings|' [=environment settings object/global object=], let |context id|
+     be the [=browsing context id=] for that context.
+
+1. Let |timestamp| be a [=time value=] representing the current date and time in UTC.
+
+1. Let |params| be [=map=] matching the <div>network.BaseParameters</div> production, with
```
juliandescottes, line 3805:
```suggestion
1. Let |params| be a [=map=] matching the <div>network.BaseParameters</div> production, with
```

index.bs
```diff
@@ -3701,6 +3727,792 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+
+    1. If |http version| starts with "<code>HTTP/</code>":
+
+      1. Let |version| be the [=code unit substring=] of |HTTP version| from 5
+         to |http version|'s [=string/length=].
+
+     1. If |version| is "<code>0.9</code>", set |protocol| to
+        "<code>http/0.9</code>", otherwise if |version| is "<code>1.0</code>",
+        set |protocol| to "<code>http/1.0</code>", otherwise if |version| is
+        "<code>1.1</code>", set |protocol| to "<code>http/1.1</code>".
```
juliandescottes, line 4258:
```suggestion
      1. If |version| is "<code>0.9</code>", set |protocol| to
         "<code>http/0.9</code>", otherwise if |version| is "<code>1.0</code>",
         set |protocol| to "<code>http/1.0</code>", otherwise if |version| is
         "<code>1.1</code>", set |protocol| to "<code>http/1.1</code>".
```

wrong indentation

juliandescottes (commented):
Thanks for the detailed PR James, spotted a few nits while trying to implement the last version.

index.bs
```diff
@@ -3701,6 +3727,792 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+1. Let |navigation| be |request|'s [=request/navigation id=].
+
+1. Let |context id| be null.
+
+1. If |request|'s [=request/client=] is not null, let |related browsing
+   contexts| be the result of [=get related browsing contexts=] with
+   |request|'s [=request/client=]. Otherwise let |related browsing contexts| be
+   an empty set.
+
+1. If |request|'s [=request/window=] is an [=environment settings object=]:
```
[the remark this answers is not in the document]
juliandescottes: Not sure why I added this comment, maybe I mixed up two things anyway please ignore.

index.bs
```diff
@@ -3701,6 +3727,792 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+
+  1. Let |environment settings| be |request|'s [=request/window=]
+
+  1. If there is a [=/browsing context=] whose [=active window=] is |environment
+     settings|' [=environment settings object/global object=], let |context id|
+     be the [=browsing context id=] for that context.
+
+1. Let |timestamp| be a [=time value=] representing the current date and time in UTC.
+
+1. Let |params| be a [=map=] matching the <div>network.BaseParameters</div> production, with
```
juliandescottes, line 3805:
```suggestion
1. Let |params| be a [=map=] matching the <code>network.BaseParameters</code> production, with
```

index.bs
foolip: I see, so if CDP has this I'll default to assuming it serves some purpose, and not push to simplify away things I don't understand.

index.bs
```diff
@@ -3701,6 +3727,792 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+field will be present and consist of an array of integers representing the bytes
+of the cookie value.
+
+<div algorithm>
+To <dfn>get a cookie</dfn> given |stored cookie|:
+
+Note: The definitions of |stored cookie|'s fields are from [[COOKIES]], except
+samesite-flag, which is from [[SAME-SITE-COOKIES]].
+
+1. Let |name| be the result of [=UTF-8 decode=] with |stored cookie|'s name field.
```
foolip, line 3847: Double checked that this can't fail, because the default error mode is replacement.

index.bs
```diff
@@ -3701,6 +3727,792 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+
+1. Let |http only| be true if |stored cookie|'s http-only-flag is true, or false
+   otherwise.
+
+1. Let |secure| be true if |stored cookie|'s secure-only-flag is true, or false
+   otherwise.
+
+1. Let |same site| be "<code>none</code>" if |stored cookie|'s samesite-flag is
+   "<code>lax</code>" if it is "<code>Lax</code>, or "<code>strict</code>" if it
+   is "<code>Strict</code>".
```
foolip, line 3880:
```suggestion
1. Let |same site| be "<code>none</code>" if |stored cookie|'s samesite-flag is
   "<code>Lax</code>", or "<code>strict</code>" if it is "<code>Strict</code>".
```

index.bs
```diff
@@ -3701,6 +3727,792 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+
+      1. Append the [=byte/value=] of |byte| to |binary value|.
+
+   Otherwise: let |value| be |utf8 decoded value| and let |binary value|
+   be null.
+
+1. Return a map matching the <code>network.Header</code> production, with the
+   <code>name</code> field set to |name|, the <code>value</code> field
+   set to |value| if it's not null, or omitted otherwise, and the
+   <code>binaryValue</code> field set to |binary value| if it's not null, or
```
foolip, line 4030: _binary value_ might not have been initialized here. Maybe let _value_ and _binary value_ be null first, and only set one of them?

index.bs
```diff
@@ -3701,6 +3727,792 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+1. Let |secure| be true if |stored cookie|'s secure-only-flag is true, or false
+   otherwise.
+
+1. Let |same site| be "<code>none</code>" if |stored cookie|'s samesite-flag is
+   "<code>lax</code>" if it is "<code>Lax</code>, or "<code>strict</code>" if it
+   is "<code>Strict</code>".
+
+1. Return a map matching the <code>network.Cookie</code> production,
+   with the <code>name</code> field set to |name|, the <code>value</code> field
+   set to |value| if it's not null or omitted otherwise, the
```
foolip, line 3884: _value_ is always null or uninitialized here. I think an "otherwise" branch is missing above.

index.bs
```diff
@@ -3701,6 +3727,792 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+   <code>binaryValue</code> field set to |value bytes| if it's not null or
```
foolip, line 3885:
```suggestion
   <code>binaryValue</code> field set to |binary value| if it's not null or
```

index.bs
```diff
@@ -3701,6 +3727,792 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+1. Let |binary value| be null.
+
+1. Let |utf8 decoded value| be the result of [=UTF-8 decode without BOM or fail=]
+   with |stored cookie|'s value.
+
+1. If |utf8 decoded value| is failure, then:
+
+    1. Let |value| be null and |binary value| be an empty list.
+
+    1. For each |byte| in |value bytes|:
```
foolip, line 3858:
```suggestion
    1. For each |byte| in |stored cookie|'s value:
```

I think?

index.bs
```diff
@@ -3701,6 +3727,792 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+   [=URL=].
+
+1. Let |method| be |request|'s [=request/method=].
+
+1. Let |body size| be 0.
+
+1. Let |body| be request's [=request/body=].
+
+1. If |body| is a [=byte sequence=], let |body size| be the length of that
+   sequence. Otherwise if |body| is a [=/body=] then let |body size| be that
```
foolip, line 4136:
```suggestion
   sequence. Otherwise, if |body| is a [=/body=] then let |body size| be that
```

index.bs
foolip: Is there a minimum change we can make to omit the `bodySize` field for streams, and leave a TODO for handling it? I think now it will be 0, which seems un-great if actually implemented and shipped.

@sadym-chromium can you dig into what CDP does with a case like this to see if there's a model we can copy?

index.bs
foolip: Aha, didn't think about pre/post-sniffing.

foolip (commented):
I haven't reviewed all changes, but the bits I commented on last time and parts around utf8 decoding.

index.bs
```diff
@@ -3701,6 +3727,792 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+1. If |global| is null, return a map matching the
+   <code>network.FetchTimingInfo</code> production, with all fields set to 0.
+
+1. Let |timings| be |request|'s [=fetch timing info=].
+
+1. Let |connection timing| be |timings|' [=final connection timing info=] if
+   it's not null, or a new [=connection timing info=] otherwise.
+
+1. Let |request time| be [=convert fetch timestamp=] given [=convert fetch
+   timestamp=] given |timings|' <a spec=fetch>start time</a> and |global|.
```
juliandescottes, line 3935: Looking at https://www.w3.org/TR/navigation-timing-2/#processing-model I feel like `start time` is only set for navigation requests (and preceeds the unload event). Is this something we should omit for non-navigation requests?

index.bs
```diff
@@ -3701,6 +3727,792 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+1. If |request| is a [=CORS-Preflight Request=], set |type| to
+   "<code>preflight</code>".
+
+1. TODO: Get the |type|. It's not quite clear how this ought to work; the CDP
+   data depends on whether the navigation was kicked off by the parser or by
+   script (so e.g. inserting an image from script causes the initiator to be
+   "<code>script</code>"), but that doesn't correspond to anything in Fetch.
+
+1. If |request|'s [=request/initiator type=] is "<code>fetch</code>" or
+   "<code>xmlhttprequest</code>":
```
juliandescottes, line 4066: For the record I don't know how straightforward this will be to implement in Firefox. We collect most network data in the parent process, but stacktraces are captured in content processes. In devtools typically our network events only contain a flag to know if there is a stacktrace, and we retrieve it separately.

So I'm slightly concerned that if I need to add stacktrace information by default in the event, I might have to delay emitting them. I probably need to try and implement that part first, but I wanted to mention it already.

index.bs
juliandescottes, line 3935:
```suggestion
1. Let |request time| be [=convert fetch timestamp=] given |timings|' <a spec=fetch>start time</a> and |global|.
```

index.bs
```diff
@@ -3701,6 +3727,792 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+
+1. If |global| is null, return a map matching the
+   <code>network.FetchTimingInfo</code> production, with all fields set to 0.
+
+1. Let |timings| be |request|'s [=fetch timing info=].
+
+1. Let |connection timing| be |timings|' [=final connection timing info=] if
+   it's not null, or a new [=connection timing info=] otherwise.
+
+1. Let |request time| be [=convert fetch timestamp=] given [=convert fetch
```
juliandescottes, line 3934: Do we know how `convert fetch timestamp` is supposed to be implemented? An earlier section mentions 
> The <code>network.FetchTimingInfo</code> type represents the time of each part of
the request, relative to <code>requestTime</code>.

I understand that I must compute the difference between each timing and `requestTime`. If that's correct, a few questions remain:
- what about requestTime itself? should we expose the diff as well (so that would always be 0?). I guess it's more valuable to keep the actual time for this one?
- is the unit supposed to be https://w3c.github.io/hr-time/#dom-domhighrestimestamp ?
- what about timings which don't have any value and are set to 0. Should we keep them to 0 or set them to -1 (0 could indicate it happened at the same time as `requestTime` and be misleading)

index.bs
```diff
@@ -3701,6 +3727,792 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+empty map.  It's used to track the network events for which a
+<code>network.beforeRequest</code> event has already been sent.
+
+### Types ### {#module-network-types}
+
+#### The network.BaseEventParams type #### {#type-network-BasEventParams}
+
+<pre class="cddl local-cddl">
+network.BaseParameters = {
+    request: network.RequestData,
```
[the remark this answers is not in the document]
jgraham: Yes, I wondered a bit about that. I think if we did that we'd also want to make `response` into `responseData`, and I think it's a bit different to CDP which has `requestId` and `request`, but we have the precedent that ids are given the name of the thing they're an ID for (e.g. `context` rather than `contextId`).

index.bs
jgraham: So the short answer is "it's supposed to work exactly the same as the `network.performanceTImeline` interface. So in the ideal case we could get the values directly from that API (but I don't know if that will work in practice).

As written, the zero point of the timestamp isn't the request time, it's the zero point of the timestamp for the associated global object (which I guess corresponds to when it was created). Which is wrong in the descriptive text, I'll fix that.

index.bs
jgraham: https://w3c.github.io/resource-timing/ applies to all requests.

index.bs
juliandescottes: Thanks, in that case it should map directly to the timings retrieved from nsITimedChannel on Firefox' side.

juliandescottes: @jgraham In the last version, am I reading correctly that all the timings from network.FetchTimingInfo are relative timings (relative to the global's creation, which would be window.performance.timeOrigin)? How can a consumer convert them to absolute times?

index.bs
```diff
@@ -3781,6 +3810,802 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+1. Return a [=map=] matching the <code>network.ResponseData</code> production,
+   with the <code>url</code> field set to |url|, the <code>protocol</code> field
+   set to |protocol|, the <code>status</code> field set to |status|, the
+   <code>statusText</code> field set to |status text|, the
+   <code>fromCache</code> field set to |from cache|, the <code>headers</code>
+   field set to |headers|, the <code>mimeType</code> field set to |mime type|,
+   the <code>headers</code> field set to |headers|, the
+   <code>bytesReceived</code> field set to |bytes received|, the
+   <code>headersSize</code> field set to |headers size|, the
+   <code>bodySize</code> field set to |body size|, and the <code>content</code>
```
juliandescottes, line 4402:
```suggestion
   <code>fromCache</code> field set to |from cache|, the <code>headers</code>
   field set to |headers|, the <code>mimeType</code> field set to |mime type|, the
   <code>bytesReceived</code> field set to |bytes received|, the
   <code>headersSize</code> field set to |headers size|, the
   <code>bodySize</code> field set to |body size|, and the <code>content</code>
```

`the headers field set to headers` is duplicated

index.bs
```diff
@@ -3781,6 +3810,802 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+
+<pre class="cddl local-cddl">
+network.RequestData = {
+    request: network.Request,
+    url: text,
+    method: text,
+    headers: [*network.Header],
+    cookies: [*network.Cookie],
+    headersSize: uint,
+    bodySize: uint / null,
```
juliandescottes, line 4206: Based on the steps below, bodySize is either 0 or body's length, so it can't be null?

index.bs
```diff
@@ -3830,6 +3859,802 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+<pre class="cddl local-cddl">
+
+NetworkResult = (
+)
+
+NetworkEvent = (
+    network.BeforeRequestSent //
+    network.ResponseStarted //
+    network.ResponseCompleted //
+    network.FetchError
```
whimboo, line 3887:
```suggestion
    network.BeforeRequestSent //
    network.FetchError //
    network.ResponseStarted //
    network.ResponseCompleted
```

index.bs
```diff
@@ -3830,6 +3859,802 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+
+#### The network.BaseParameters type #### {#type-network-BaseParameters}
+
+<pre class="cddl local-cddl">
+network.BaseParameters = {
+    request: network.RequestData,
+    navigation: Navigation / null,
+    context: BrowsingContext / null,
+    timestamp: int,
+    isRedirect: bool,
```
whimboo, line 3906: Could we try to keep the alphabetical order of fields for each and every CDDL type you added in this PR? That would keep us in sync with already existing ones.

```suggestion
    context: BrowsingContext / null,
    isRedirect: bool,
    navigation: Navigation / null,
    request: network.RequestData,
    timestamp: int,
```

index.bs
```diff
@@ -3830,6 +3859,802 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+[=Remote end definition=] and [=local end definition=]
+
+<pre class="cddl local-cddl">
+network.RequestData = {
+    request: network.Request,
+    url: text,
+    method: text,
+    headers: [*network.Header],
+    cookies: [*network.Cookie],
+    headersSize: js-uint,
```
whimboo, line 4254: In `ResponseData` the `headerSize` can also be `null`. If that is true we probably also have to `allow` here?

jimevans (changes requested):

index.bs
```diff
@@ -3830,6 +3859,802 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+
+1. If |response|'s [=cache state=] is "<code>local</code>", let |from cache| be
+   true, otherwise let it be false.
+
+1. Let |headers| be an empty list.
+
+1. Let |mime type| be the [=computed mime type=] for |response|.
+
+   Note: this is whatever MIME type the browser is actually using, even if it
+   isn't following the exact algorithm in the [[MIMESNIFF]] specification.
```
juliandescottes, line 4426: Cf discussion at https://phabricator.services.mozilla.com/D166373#inline-918466 about including or not the (optional) charset in the `mime type`. Following the `computed mime type` algorithm, the charset should in theory be part of the `mime type` but since CDP does not include it, we might do the same.

I guess what we want here is the https://mimesniff.spec.whatwg.org/#mime-type-essence ?

index.bs
jgraham: We can, but there's not loads of precedent here. The current spec is less consistent than you think. CDP always puts optional fields at the end but otherwise seems to use "importance" order (maybe the non-JSON serializations want new fields added after existing ones for compatibility reasons?). Other web specs like DOM group IDL blocks by function rather than in strictly alphabetical order.

index.bs
jgraham: I think by the time we emit any event we already know request headers, so I don't think it's necessary to allow null.

index.bs
```diff
@@ -3830,6 +3862,804 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+
+1. Let |headers| be an empty list.
+
+1. Let |cookies| be an empty list.
+
+1. For each (|name|, |value|) in |request|'s [=request/headers list=]:
+
+  1. Append the result of [=get a header=] with |name| and |value| to |headers|.
+
+  1. If |name| is a [=byte-case-insensitive=] match for "<code>Cookie</code>" then:
```
foolip, line 4300: If it's intentional that the cookie data will be both in the header list (previous step) and in the cookie list (following steps) can you add a note about why that makes sense?

index.bs
```diff
@@ -3830,6 +3862,804 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+
+</div>
+
+#### The network.ResponseContent type #### {#type-network-ResponseContent}
+
+[=Remote end definition=] and [=local end definition=]
+
+<pre class="cddl local-cddl">
+network.ResponseContent = {
+    size: js-unit / null
```
foolip, line 4329: Can `size` actually be null? https://fetch.spec.whatwg.org/#fetch-timing-info-decoded-body-size defaults to 0 so it doesn't seem like it.

index.bs
```diff
@@ -3830,6 +3862,804 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+   isn't following the exact algorithm in the [[MIMESNIFF]] specification.
+
+1. For each (|name|, |value|) in |response|'s [=response/headers list=]:
+
+  1. Append the result of [=get a header=] with |name| and |value| to |headers|.
+
+1. Let |bytes received| be the total number of bytes transmitted as part of the
+   HTTP response associated with |response|.
+
+1. Let |headers size| be the number of bytes transmitted as part of the header
```
foolip, line 4438: When we get to testing this, is it clear enough exactly which bytes are included? In particular, how many of the newlines before the body are included, if any?

Presumably we'll converge on something through testing, but if it's already clear what the right behavior is, spelling it out would be good.

index.bs
```diff
@@ -3830,6 +3862,804 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+This event is emitted after the full response body is received.
+
+<div algorithm>
+The [=remote end event trigger=] is the <dfn export>WebDriver BiDi response
+completed</dfn> steps given |request| and |response|:
+
+1. Assert: [=before request sent map=][|request|] contains |request|'s
+   [=redirect count=].
+
+   This implies that every caller needs to ensure that the [=WebDriver BiDi
```
foolip, line 4588:
```suggestion
   Note: This implies that every caller needs to ensure that the [=WebDriver BiDi
```

index.bs
```diff
@@ -3830,6 +3862,804 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+body is complete.
+
+<div algorithm>
+The [=remote end event trigger=] is the <dfn export>WebDriver BiDi response
+started</dfn> steps given |request| and |response|:
+
+1. Assert: [=before request sent map=][|request|] is equal to |request|'s
+   [=redirect count=].
+
+   This implies that every caller needs to ensure that the [=WebDriver BiDi
```
foolip, line 4640:
```suggestion
   Note: This implies that every caller needs to ensure that the [=WebDriver BiDi
```

foolip (approved):
A bunch of comments, but I don't need to review again before this is merged. (I'm out next week.)

index.bs
```diff
@@ -3830,6 +3862,804 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+     the value of the <code>columnNumber</code> field |stack trace|[0]. Otherwise
+     let |line number| and |column number| be 0.
+
+  Otherwise, let |stack trace|, |column number|, and |line number| all be null.
+
+  TODO: Chrome includes the current parser position as column number / line
+  number for parser-inserted resources.
+
+1. Return a [=map=] matching the <code>network.Initiator</code> production, with
+   the <code>type</code> field set to |type|, the <code>columnNumber</code>
```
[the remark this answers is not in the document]
jgraham: I don't like that; I think we should always populate type fields since they're typically used to distinguish between different variants, and forcing people to also handle the missing value case (other than as "this wasn't a recognised response") is harder for little gain.

index.bs
jgraham: The reason is just "so clients don't have to manually parse the Cookie header, but can share code with the (future) get_cookies command". I can obviously just write that directly in a note, but it seems a bit odd to have that level of justification.

index.bs
jgraham: I suggest that https://datatracker.ietf.org/doc/html/rfc7230#section-3 makes it clear that this doesn't include the CRLF separating the headers and the body (in HTTP 1.1). I don't know if that's what people expect.

index.bs
```diff
@@ -3830,6 +3862,807 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+### Types ### {#module-network-types}
+
+#### The network.BaseParameters type #### {#type-network-BaseParameters}
+
+<pre class="cddl local-cddl">
+network.BaseParameters = {
+    context: BrowsingContext / null,
+    isRedirect: bool,
+    redirectCount: js-uint,
+    navigation: Navigation / null,
```
juliandescottes, line 3910: Maybe keep those alphabetically sorted. Not mandatory.

index.bs
```diff
@@ -4115,6 +4147,807 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+
+</div>
+
+#### The network.ResponseContent type #### {#type-network-ResponseContent}
+
+[=Remote end definition=] and [=local end definition=]
+
+<pre class="cddl local-cddl">
+network.ResponseContent = {
+    size: js-unit
```
jimevans, line 4617: Should be `js-uint`, not `js-unit`.

jimevans (changes requested):

index.bs
```diff
@@ -4115,6 +4147,810 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+[=Remote end definition=] and [=local end definition=]
+
+<pre class="cddl local-cddl">
+network.RequestData = {
+    request: network.Request,
+    url: text,
+    method: text,
+    headers: [*network.Header],
+    cookies: [*network.Cookie],
+    headersSize: js-uint,
```
sadym-chromium, line 4547: Do we need the size having the headers themselves? We don't have it in CDP.

index.bs
```diff
@@ -4115,6 +4147,810 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+
+1. If [=before request sent map=] does not contain |request|, set [=before
+   request sent map=][|request|] to a new set.
+
+1. Let |redirect count| be |request|'s [=redirect count=].
+
+1. Add |redirect count| to [=before request sent map=][|request|].
+
+1. Let |is redirect| be true if |request|'s [=redirect count=] is greater than
+   0 or false otherwise.
```
sadym-chromium, line 4781: Does it make sense to have `isRedirect`, if it's always `redirectCount>0`?

index.bs
```diff
@@ -4115,6 +4147,810 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+      </pre>
+   </dd>
+</dl>
+
+This event is emitted before a request is sent (either over the network or
+before it's handled by a serviceworker or a local cache).
+
+<div algorithm>
+The [=remote end event trigger=] is the <dfn export>WebDriver BiDi before
+request sent</dfn> steps given |request|:
```
sadym-chromium, line 4771: This looks self-recursive.

index.bs
jgraham: I believe that HAR wants it, and computing the byte size from the deserialized headers seems complex (you need to re-serialize the headers, and I don't know how that works in the face of header compression).

index.bs
jgraham: I don't mind. @juliandescottes wdyt?

index.bs
sadym-chromium: In Chrome the HAR's `headersSize` is always  `-1`, as it [relies](https://source.chromium.org/chromium/chromium/src/+/main:third_party/devtools-frontend/src/front_end/models/har/Log.ts;l=205?q=headersSize&ss=chromium%2Fchromium%2Fsrc:third_party%2Fdevtools-frontend%2Fsrc%2Ffront_end%2F) on [deprecated](https://chromedevtools.github.io/devtools-protocol/tot/Network/#event-responseReceivedExtraInfo) field `headersText`.

index.bs
juliandescottes: Sounds good, let's keep only `redirectCount`

index.bs
juliandescottes: (sorry misclicked on resolve...)

In Firefox DevTools HAR, we do provide the headers size, and yes it would not be trivial to recompute it from deserialized headers. 

And it seems at least one known consumer of HAR is re-implementing this for Chrome to workaround the -1 value: https://github.com/sitespeedio/chrome-har/blob/d23393967e1766fa78c001faec244b6b6378c52d/lib/headers.js#L6-L37

So it seems worth including?

index.bs
```diff
@@ -4115,6 +4147,810 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+1. Let |connect end| be [=convert fetch timestamp=] given |connection timing|'s
+   [=connection end time=] and |global|.
+
+1. Let |request start| be [=convert fetch timestamp=] given |timings|' [=final
+   network-request start time=] and |global|.
+
+1. Let |response start| be [=convert fetch timestamp=] given |timings|' [=final
+   network-response start time=] and |global|.
+
+1. Let |response end| be [=convert fetch timestamp=] given |timings|'
```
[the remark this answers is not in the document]
juliandescottes: FWIW this is how I implemented it for Firefox. To me it makes sense, the alternative would be to return the (converted) current time instead? How would the consumer differentiate between the "definitive" response end and an intermediary one?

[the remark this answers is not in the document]
sadym-chromium: I guess we should provide only values which makes sense (>0 in our case), otherwise it would be a bit odd.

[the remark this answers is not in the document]
juliandescottes: I think using 0 aligns with the default values used in the  PerformanceNavigationTiming spec

index.bs
whimboo: I assume we can now close this issue given that the changes have already been applied.

index.bs
whimboo: @jgraham this is still a problem given that the validation fails in CI.

index.bs
jgraham: I don't think it is; this defines the steps that are run; it's the entry point for external specifications when this event is going to be sent. That's different to the concept of the event.

jgraham:
I've added a small hack so that the build no longer depends on changes that are pending in fetch. That avoids a circular dependency, so landing this will allow me to complete that PR.

Apart from that I think this is at the point of being good enough for a first cut, and we should aim to land it soon.

index.bs
sadym-chromium: Oh, then I'm missing the part where the event is initiated.

jgraham: That happens in external specifications which actually create network events (principally fetch). See https://github.com/whatwg/fetch/pull/1540 but note there's a dependency on this PR landing before that one can.

sadym-chromium: Thanks for the link!

sadym-chromium:
How do we deal with `favicon`s? E.g. this test can be (and is in our case) flaky because of `favicon.ico` request.
https://github.com/web-platform-tests/wpt/blob/a933765a927ba5392b93edd47c69696d7b016658/webdriver/tests/bidi/network/before_request_sent/before_request_sent_tentative.py#L67

jgraham:
Re: favicons I don't know. @juliandescottes are we specifically not including those requests for now, or are we just lucky with timing that the test isn't intermittent? It seems sensible to me that all requests would be included, and we need to make the tests robust to the possibility of getting a browser-initiated request.

index.bs
```diff
@@ -4115,6 +4147,811 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+
+#### The network.BaseParameters type #### {#type-network-BaseParameters}
+
+<pre class="cddl local-cddl">
+network.BaseParameters = {
+    context: BrowsingContext / null,
+    isRedirect: bool,
+    navigation: Navigation / null,
+    redirectCount: js-uint,
+    request: network.RequestData,
```
sadym-chromium, line 4196: `BaseParameters` always contains `request`. Does it make sense in case of response events like `network.responseStarted` or `network.responseCompleted`?

jgraham: Yes, although other designs are possible.

The constraint is that we at least need to provide the request id, so you know which response this is a request for.

In the current design we also provide the timing data in the `RequestData` since it applies to both request and response phases.

The tradeoff of the current setup is that if you subscribe to more than one event you get duplicate data. But it also means you aren't required to subscribe to all events e.g. it would be possible to just subscribe to `network.responseCompleted` to find out which requests occurred.

juliandescottes:
> Re: favicons I don't know. @juliandescottes are we specifically not including those requests for now, or are we just lucky with timing that the test isn't intermittent? It seems sensible to me that all requests would be included, and we need to make the tests robust to the possibility of getting a browser-initiated request.

Checking this today, I suspect we might miss those requests if they hit the image cache, because that's handled in the content process and we don't process those yet. Otherwise I agree we should include them. So I would say feel free to add those requests to the impacted tests.

jgraham:
I think the problem is that the timing of favicon requests is unpredictable. So we might need to update the tests to filter out any unexpected requests, or something.

juliandescottes:
> I think the problem is that the timing of favicon requests is unpredictable. So we might need to update the tests to filter out any unexpected requests, or something.

Right. In the test mentioned by Maksim, we monitor network events during a page load, so it seems logical that a favicon request will happen, and it should be feasible to consistently wait for it.

But it's true that if they also start to intermittently appear in tests where we don't perform a navigation, we will rather have to use a filtering approach.

juliandescottes:
I was wrong, we normally capture favicon requests just fine. However we set `browser.chrome.guess_favicon` to false in our test environment, which means we don't try to use default/fallback favicons. 

@sadym-chromium not sure if Chrome would have a similar mode or if that's something we need to stop doing on Firefox side?

index.bs
sadym-chromium: I may be wrong here, but isn't `headersSize` used to read the proper data fields from the raw request? And if saw, it is useless in our case, as long as the headers are sent parsed.

index.bs
sadym-chromium: Sounds reasonable.

index.bs
juliandescottes: I don't think headersSize is expected to be used to process data from the request, it's rather a size/performance metric.

If we want to challenge this further we need inputs from consumers for HAR files to know if they could do without the data.

index.bs
```diff
@@ -4115,6 +4147,811 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+    url: text,
+    protocol: text,
+    status: js-uint,
+    statusText: text,
+    fromCache: bool,
+    headers: [*network.Header],
+    mimeType: text,
+    bytesReceived: js-uint,
+    headersSize: js-uint / null,
+    bodySize: js-uint / null,
```
sadym-chromium, line 4652: Note: It's reasonable to have the fields, but they are not implemented yet in CDP (the data length can be taken only with the data itself), so it would be nice to make the WPT tests not relate on this field. Same thing for the `ResponseData.content`.

juliandescottes: - `ResponseData.headersSize` and `ResponseData.bodySize` are optional, so we allow them to be null in the tests.
- `ResponseContent.size` is not supposed to be optional, but it seems we still allow it to be null in tests. 
- `RequestData.headersSize` is the only one expected by the tests to be non optional, so I imagine that's where you will have frequent issues

I feel like it's better to make progress here, so if it helps I would be in favor of having all those fields marked as optional for now so that we can land this PR, and then we can discuss each of them in separate issues. Then if we say that we keep a field in the spec, all vendors should try to support it

sadym-chromium: > I feel like it's better to make progress here, so if it helps I would be in favor of having all those fields marked as optional for now so that we can land this PR, and then we can discuss each of them in separate issues. Then if we say that we keep a field in the spec, all vendors should try to support it

SGTM

sadym-chromium: > If we want to challenge this further we need inputs from consumers for HAR files to know if they could do without the data.

My biggest concern is that those fields are not really needed or avoidable by users. If so, we are not going to implement them, which makes the spec not implementable. 

I hate blocking the progress of this PR, but that is what stopped me from approving it. After all, I guess the risk from our side is acceptable, so LGTM.

sadym-chromium (approved):

index.bs
```diff
@@ -4168,6 +4200,811 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+<code>network.beforeRequestSent</code> event has already been sent.
+
+### Types ### {#module-network-types}
+
+#### The network.BaseParameters type #### {#type-network-BaseParameters}
+
+<pre class="cddl local-cddl">
+network.BaseParameters = {
+    context: BrowsingContext / null,
+    isRedirect: bool,
```
juliandescottes, line 4246: I thought the intent was to remove `isRedirect` in favor of `redirectCount` because it was redundant, cf https://github.com/w3c/webdriver-bidi/pull/204#discussion_r1126441624

index.bs
```diff
@@ -4168,6 +4200,811 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+1. Let |expires| be |stored cookie|'s expires field represented as a unix
+   timestamp, if set, or null otherwise.
+
+1. Let |size| be the byte length of the result of serializing |stored cookie|
+   as it would be represented in a <code>Cookie</code> header.
+
+1. Let |http only| be true if |stored cookie|'s http-only-flag is true, or false
+   otherwise.
+
+1. Let |secure| be true if |stored cookie|'s secure-only-flag is true, or false
```
juliandescottes, line 4357:
```suggestion
1. Let |secure| be true if |stored cookie|'s secure-flag is true, or false
```

secure-only-flag -> secure-flag?

index.bs
```diff
@@ -4168,6 +4200,811 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+
+1. Let |bytes received| be the total number of bytes transmitted as part of the
+   HTTP response associated with |response|.
+
+1. Let |headers size| be the number of bytes transmitted as part of the header
+   fields section of the HTTP response.
+
+1. Let |body size| be |response|'s [=response body info=]'s [=encoded size=].
+
+1. Let |content| be |response|'s [=get the response content info=] with |response|.
```
juliandescottes, line 4785:
```suggestion
1. Let |content| be the result of [=get the response content info=] with |response|.
```

nit

juliandescottes (approved):

index.bs
jgraham: The RFC seems to use `secure-only-flag` for the data model.

index.bs
```diff
@@ -4168,6 +4200,800 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+    requestTime: float,
+    redirectStart: float,
+    redirectEnd: float,
+    fetchStart: float,
+    dnsStart: float,
+    dnsEnd: float,
+    connectStart: float,
+    connectEnd: float,
+    tlsStart: float,
+    <!-- tlsEnd: float this should be the same as connectEnd -->
```
thiagowfx, line 4390: There's a [WPT test](https://github.com/web-platform-tests/wpt/blob/707500a0474977b3eea2873288d9c395cd123626/webdriver/tests/bidi/network/__init__.py#L48) that checks for `tlsEnd`, however it is [not currently](https://w3c.github.io/webdriver-bidi/#type-network-FetchTimingInfo) part of the spec.

We should fix either side.

What is the correct one @jgraham ?

thiagowfx: I am going to remove it from WPT: https://github.com/web-platform-tests/wpt/pull/40223

jgraham:
@juliandescottes

index.bs
whimboo: Given that the PR got merged it looks like this is done now.

thiagowfx: Yes, resolving. Thanks

thiagowfx:
Also cc @OrKoN to confirm the above statement from the CDP side.

OrKoN:
To the extend of my knowledge, this is right: it does not look like there is an equivalent in CDP (responseReceivedExtraInfo is not exactly that and cannot be used for interception if that is the plan for the event).

whimboo:
Could we please discuss that on a new issue? Using a PR which has been closed already is not a good place. Thanks.

thiagowfx:
> Could we please discuss that on a new issue? Using a PR which has been closed already is not a good place. Thanks.

Sure! I'll just open a new issue and link to the existing PR then, sorry.

christian-bromann:
> It also currently only supports HTTP-type requests and not
> e.g. WebSockets. Some support for integration with service workers is
> also missing.

@jgraham do we have an issue for this? Should we start tracking this?

whimboo:
> > It also currently only supports HTTP-type requests and not
> > e.g. WebSockets. Some support for integration with service workers is
> > also missing.
> 
> @jgraham do we have an issue for this? Should we start tracking this?

@christian-bromann I would suggest that you file new issues. I cannot find any existing ones covering those topics.

Issue #90, opened 2021-02-19
Defined releaseObject


sadym-chromium:
To prevent memory leakage, there should be a way to release the received remote object.
This can be implemented like in CDP [`Runtime.releaseObject`](https://chromedevtools.github.io/devtools-protocol/tot/Runtime/#method-releaseObject)

jgraham:
What's the model here? Does serializing automatically create a reference to the serialized object? What's the lifetime of that reference? Does it apply to all object types? The WebDriver/HTTP model (for elements and windows) is that serialization doesn't affect the lifetime of the value at all, so dereferencing the remote object can fail. I can see that model might be more problematic with a pattern like `return [{foo: document.body}]` where there are a bunch of temporary objects that aren't otherwise kept alive. But at least with the current model when we don't hit serialization limits the contained values will be directly accessible to the client, so the problem is less bad than in CDP.

sadym-chromium:
I'm implementing the `ArrayValue` serialisation. I cannot understand what is the lifecycle of the object in case of this being evaluated: `return [{foo: 'bar'}]`.
According to the spec, the result should have an `objectId` along with the key-values. The question is when the GC can collect the object. 

I agree that exposing garbage collector logic `releaseObject` doesn't seems ergonomic, but I don't see other options of avoiding memory leakage.

jgraham:
So with `return [{foo: "bar"}]`, in the absence of autoref, the objects would likely be gc'd, making any attempt to use the handle fail. But that wouldn't necessarily be a problem in this case because the object serialization would contain all the same data as the original object, so there would be much less need to access the handles.

That said, I think I'm convinced that we need something that automatically holds a reference to objects. But there are a number of questions that need to be decided:

* Which objects so we hold (strong) references to? Anything that's sent over the wire? Just the root objects returned (e.g. the array in the example above, which will then keep the inner object alive)?
* Do DOM objects get special handling? It's much less likely that a node will be GC'd than a radom js object since it's typically owned by the DOM tree.
* What's the lifetime of the references? It seems likely we don't want to extend the lifetime of the relevant global. So we should probably arrange things so that the object map is destroyed immediately before the global would otherwise be destroyed (e.g. after non-bfcache navigation or when a worker is going to terminate).

whimboo:
[...]

PR #205, opened 2022-05-10
Make `objectId` realm-specific

sadym-chromium:
Step 1 towards addressing #90 and implementing the [proposal](https://github.com/w3c/webdriver-bidi/issues/90#issuecomment-1111052638).

* Make `ObjectId` realm-specific.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1019,13 +1019,11 @@ Issue: Define how this works.
 
 ## Reference ## {#data-types-reference}
 
-<dfn>RemoteReference</dfn> represents a remote reference to an exising ECMAScript object in
-[=object id map=].
-
-A [=BiDi session=] has an <dfn>object id map</dfn>. This is a weak map
+Each EMCAScript [=Realm=] has a corresponding <dfn>object id map</dfn>. This is a strong map
 from objects to their corresponding id.
 
-Issue: Should this be explicitly per realm?
+<dfn>RemoteReference</dfn> represents a remote reference to an exising ECMAScript object in
+[=object id map=] in the given [=Realm=].
 
 ```
 ObjectId = text;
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1019,13 +1019,11 @@ Issue: Define how this works.
 
 ## Reference ## {#data-types-reference}
 
-<dfn>RemoteReference</dfn> represents a remote reference to an exising ECMAScript object in
-[=object id map=].
-
-A [=BiDi session=] has an <dfn>object id map</dfn>. This is a weak map
+Each EMCAScript [=Realm=] has a corresponding <dfn>object id map</dfn>. This is a strong map
```
jgraham, line 1022: Should this be a strong map? That feels like it would be part of a later change, not this one. We don't want to leave the spec in a state where we're creating a strong reference to every serialized object.

index.bs
```diff
@@ -1039,11 +1037,11 @@ RemoteReference = {
 Issue: handle "stale object reference" case.
 
 <div algorithm>
-To <dfn>deserialize remote reference</dfn> given a |remote reference|:
+To <dfn>deserialize remote reference</dfn> given a |remote reference| in |realm|:
```
jgraham, line 1040: This is probably fine, but I've typically used a less prose, more function call, style where this would be written `To <dfn>deserialize remote reference</dfn> given |realm|, and |remote reference|`.

index.bs
```diff
@@ -1439,19 +1437,18 @@ object is discarded in the runtime, subsequent attempts to access it via the
 protocol will result in an error.
 
 <div algorithm>
-To get the <dfn>object id for an object</dfn> given a |session| and |object|:
+To get the <dfn>object id for an object</dfn> given an |object| in a given |realm|:
```
jgraham, line 1440: similar comment here, I'd say `given |realm| and |object|`

jgraham (changes requested):

index.bs
sadym-chromium: move back **weak** instead of **strong**.

index.bs

index.bs

sadym-chromium:
Closed in favor of https://github.com/w3c/webdriver-bidi/pull/207, to be able to create a nested PR, which should ease PR review.

PR #206, opened 2022-05-10
Strong unique `Handle` in serialization

sadym-chromium:
Step 2 towards addressing #90 and implementing the [proposal](https://github.com/w3c/webdriver-bidi/issues/90#issuecomment-1111052638).

* Add `ownership` logic into serialization with default `root`: https://github.com/w3c/webdriver-bidi/issues/90#issuecomment-1111052638.
* Rename `objectId` -> `handle`.
* Make `handle map` a strong reference.
* Make `handle`-`object` a many-to-one relation.
* Cyclic reference serialization via `internalId` field.
* `script.disown` command.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/206.html" title="Last updated on May 25, 2022, 10:05 AM UTC (94e7cbf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/206/f11e1fd...94e7cbf.html" title="Last updated on May 25, 2022, 10:05 AM UTC (94e7cbf)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -337,8 +337,8 @@ An implementation may define <dfn>extension modules</dfn>. These must have a
 [=module name=] that contains a single colon "<code>:</code>" character. The
 part before the colon is the prefix; this is typically the same for all
 extension modules specific to a given implementation and should be unique for a
-given implementation. Such modules extend the [=local end definition=] and
-[=remote end definition=] providing additional groups as choices for the defined
+given implementation. Such modules extend the [=local end definition=] and [=remote
+end definition=] providing additional groups as choices for the defined
 [=commands=] and [=events=].
 
 ## Commands ## {#commands}
@@ -351,17 +351,16 @@ long-running. As a consequence, commands can finish out-of-order.
 
 Each [=command=] is defined by:
 
-- A <dfn export for=command>command type</dfn> which is defined by a
-  [=remote end definition=] fragment containing a group. Each such group has two
-  fields:
+- A <dfn export for=command>command type</dfn> which is defined by a [=remote
+   end definition=] fragment containing a group. Each such group has two fields:
     - <code>method</code> which is a string literal of the form <code>[module
       name].[method name]</code>. This is the <dfn export for=command>command
       name</dfn>.
     - <code>params</code> which defines a mapping containing data that to be passed into
       the command. The populated value of this map is the
       <dfn export for=command>command parameters</dfn>.
-- A <dfn export for=command>result type</dfn>, which is defined by a
-  [=local end definition=] fragment.
+- A <dfn export for=command>result type</dfn>, which is defined by a [=local
+  end definition=] fragment.
 - A set of [=remote end steps=] which define the actions to take for a command
   given a [=BiDi session=] and [=command parameters=] and return an
   instance of the command [=return type=].
@@ -376,21 +375,21 @@ command. From the point of view of the [=remote end=] this identifier is opaque
 and cannot be used internally to identify the command.
 
 Note: This is because the command id is entirely controlled by the [=local end=]
-and isn't necessarily unique over the course of a session. For example a
-[=local end=] which ignores all responses could use the same command id for each
-command.
+and isn't necessarily unique over the course of a session. For example a [=local
+end=] which ignores all responses could use the same command id for each command.
 
 The <dfn export for=command>set of all command names</dfn> is a set containing
-all the defined [=command names=], including any belonging to [=extension modules=].
+all the defined [=command names=], including any belonging to [=extension
+modules=].
 
 ## Events ## {#events}
 
-An <dfn export>event</dfn> is a notification, sent by the [=remote end=] to the
-[=local end=], signaling that something of interest has occurred on the
-[=remote end=].
+An <dfn export>event</dfn> is a notification, sent by the [=remote
+end=] to the [=local end=], signaling that something of interest has
+occurred on the [=remote end=].
 
- - An <dfn export for=event>event type</dfn> is defined by a [=local end definition=]
-   fragment containing a group. Each such group has two fields:
+ - An <dfn export for=event>event type</dfn> is defined by a [=local
+   end definition=] fragment containing a group. Each such group has two fields:
     - <code>method</code> which is a string literal of the form <code>[module
       name].[event name]</code>. This is the <dfn export for=event>event
       name</dfn>.
@@ -509,10 +508,11 @@ Message transport is provided using the WebSocket protocol.
 Note: In the terms of the WebSocket protocol, the [=local end=] is the
 client and the [=remote end=] is the server / remote host.
 
-Note: The encoding of [=commands=] and [=events=] as messages is similar to JSON-RPC,
-but this specification does not normatively reference it. [[JSON-RPC]] The normative
-requirements on [=remote ends=] are instead given as a precise processing model,
-while no normative requirements are given for [=local ends=].
+Note: The encoding of [=commands=] and [=events=] as messages is
+similar to JSON-RPC, but this specification does not normatively
+reference it. [[JSON-RPC]] The normative requirements on [=remote
+ends=] are instead given as a precise processing model, while no
+normative requirements are given for [=local ends=].
 
 A <dfn>WebSocket listener</dfn> is a network endpoint that is able
 to accept incoming [[!RFC6455|WebSocket]] connections.
@@ -543,19 +543,22 @@ initially empty.
 A [=BiDi session=] |session| is <dfn>associated with connection</dfn>
 |connection| if |session|'s [=session WebSocket connections=] contains |connection|.
 
-Note: Each [=WebSocket connection=] is associated with at most one [=BiDi session=].
+Note: Each [=WebSocket connection=] is associated with at most one [=BiDi
+session=].
 
 <div>
 
-When a client [=establishes a WebSocket connection=] |connection| by connecting to
-one of the set of [=active listeners=] |listener|, the implementation must proceed
-according to the WebSocket [=server-side requirements=], with the following steps run
-when deciding whether to accept the incoming connection:
+When a client [=establishes a WebSocket connection=] |connection| by
+connecting to one of the set of [=active listeners=] |listener|, the
+implementation must proceed according to the WebSocket [=server-side
+requirements=], with the following steps run when deciding whether to
+accept the incoming connection:
 
-1. Let |resource name| be the resource name from
-   [=reading the client's opening handshake=]. If |resource name| is not in
-   |listener|'s [=list of WebSocket resources=], then stop running these steps and
-   act as if the requested service is not available.
+1. Let |resource name| be the resource name from [=reading the
+   client's opening handshake=]. If |resource name| is not in
+   |listener|'s [=list of WebSocket resources=], then stop
+   running these steps and act as if the requested service is not
+   available.
 
 1. If |resource name| is the byte string "<code>/session</code>",
    and the implementation [=supports BiDi-only sessions=]:
@@ -564,8 +567,8 @@ when deciding whether to accept the incoming connection:
        connection should be accepted, and if it is not stop running these
        steps and act as if the requested service is not available.
 
-    1. Add the connection to the set of
-       [=WebSocket connections not associated with a session=].
+    1. Add the connection to the set of [=WebSocket connections not associated
+       with a session=].
 
     1. Return.
 
@@ -583,21 +586,23 @@ when deciding whether to accept the incoming connection:
    connection should be accepted, and if it is not stop running these
    steps and act as if the requested service is not available.
 
-1. Otherwise append |connection| to |session|'s [=session WebSocket connections=],
-   and proceed with the WebSocket [=server-side requirements=] when a server chooses
-   to accept an incoming connection.
+1. Otherwise append |connection| to |session|'s [=session WebSocket
+   connections=], and proceed with the WebSocket [=server-side requirements=]
+   when a server chooses to accept an incoming connection.
 
 Issue: Do we support > 1 connection for a single session?
 
 </div>
 
-When [=a WebSocket message has been received=] for a [=WebSocket connection=]
-|connection| with type |type| and data |data|, a [=remote end=] must
-[=handle an incoming message=] given |connection|, |type| and |data|.
+When [=a WebSocket message has been received=] for a [=WebSocket
+connection=] |connection| with type |type| and data |data|, a [=remote
+end=] must [=handle an incoming message=] given |connection|, |type|
+and |data|.
 
-When [=the WebSocket closing handshake is started=] or when
-[=the WebSocket connection is closed=] for a [=WebSocket connection=] |connection|, a
-[=remote end=] must [=handle a connection closing=] given |connection|.
+When [=the WebSocket closing handshake is started=] or when [=the
+WebSocket connection is closed=] for a [=WebSocket connection=]
+|connection|, a [=remote end=] must [=handle a connection closing=]
+given |connection|.
 
 Note: Both conditions are needed because it is possible for a
 WebSocket connection to be closed without a closing handshake.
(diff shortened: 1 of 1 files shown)


jgraham (changes requested):
So I think this is intended to make `objectId` represent an optional strong reference to the object. But it doesn't restore the change that's removed from the base PR to make the map a strong map, and doesn't implement what I understood to be the current consensus that serilization-specific strong references are called `handle`.

sadym-chromium:
> So I think this is intended to make `objectId` represent an optional strong reference to the object. But it doesn't restore the change that's removed from the base PR to make the map a strong map, and doesn't implement what I understood to be the current consensus that serilization-specific strong references are called `handle`.

I wanted to make it in another iteration to keep the reviewing simple. `script.disown` and cyclic reference handling are still not implemented. Do you think it should be done in 1 PR?

sadym-chromium:
@jgraham @foolip PTAL

sadym-chromium:
So all the required changes are in this PR:
1. Strong unique reference.
2. Handling cycles.
3. `script.disown`.

Please let me know, if you want me to split it into several PRs.

index.bs
```diff
@@ -1437,18 +1441,38 @@ object is discarded in the runtime, subsequent attempts to access it via the
...
+1. Let |handle id| be a new, unique, string handle for |object|.
+
+1. Set the value of |handle id| in |realm|'s [=handle object map=] to |object|.
+
+1. Return |handle id| as a result.
+
+</div>
+
+<div algorithm>
+To <dfn>set internal ids if needed</dfn> given |object remote value map|,
```
jgraham, line 1458: I find the name "object remote value map" quite confusing. Couldn't it be "internal id map" or something similar?

index.bs
```diff
@@ -3736,9 +3801,55 @@ To <dfn>get a realm from a target</dfn> given |target|:
...
+
+The <code>SerializationOwnershipModel</code> specifies how the serialized value
+ownership should be treated.
+
 
 ### Commands ### {#module-script-commands}
 
+#### The script.disown Command ####  {#command-script-disown}
+
+The <dfn export for=commands>script.disown</dfn> command disowns the given handles.
```
jgraham, line 3818: Let's use "garbage collected" rather than "disposed".

index.bs
```diff
@@ -1610,7 +1655,8 @@ Issue: handle String / Number / etc. wrapper objects specially?
 <div algorithm>
 
 To <dfn>serialize as a remote value</dfn> given a |value|, a |max depth|,
-|node details|, a |set of known objects| and a |realm|:
+a |node details|, an |ownership|, a |set of known objects|, an
```
jgraham, line 1658: We don't really need |set of known objects| and |internal object remote value map|, we can just have a single map that either a) gives each object a random id when its inserted or b) sets the id to null unless the object is later referenced as part of a cycle.

index.bs
```diff
@@ -1477,102 +1501,122 @@ RemoteValue = {
   WindowProxyRemoteValue //
 }
 
+InternalId = String;
```
jgraham, line 1504: We could make this `ObjectId` in the hope of later reusing it for weakrefs if we add that feature. If we made that optional it would be a bit odd that objects which are part of a cycle always get an objectId. But it would also be odd to have an `internalId` that isn't the same as the `objectId` in case we do take that approach later on.

jgraham (changes requested):
Looks like a good start, I have a few questions about the design choices and spec details.

index.bs
```diff
@@ -1477,102 +1501,122 @@ RemoteValue = {
...
 ListRemoteValue = [*RemoteValue];
 
 MappingRemoteValue = [*[(RemoteValue / text), RemoteValue]];
 
 SymbolRemoteValue = {
-   type: "symbol",
-   objectId: ObjectId,
+  type: "symbol",
+  ?handle: Handle,
+  ?internalId: InternalId,
```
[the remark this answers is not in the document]
sadym-chromium: I guess having string provides more flexibility. When we switch to a general weakref, I imagine we want to keep some additional internal info in the ID like it's done in CDP's `objectId`. On the implementation side you still can keep it as number, and make `.toString()` to save some bites.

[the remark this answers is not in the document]
jgraham: A string that isn't just random but encodes additional information sounds like a problem, not a goal. If some implementations are encoding data into the element ids then clients are likely to start trying to parse the ids instead of using  them as opaque identifiers. Then we either end up having to specify the id format (which isn't designed to be compatible with multiple implementations and so might be difficult) or we end up with clients that are tied to specific implementations.

index.bs
sadym-chromium: done

index.bs
sadym-chromium: `|serialization internal map|`

index.bs
sadym-chromium: The problem is we don't have remote value at the moment we want to serialize the value. I can either:
1. Create RemoteValue before serializing each value.
2. Put `null` into `serialization internal map`
3. Use `set of known objects`.

I don't like option `2`, but option `1` seems reasonable. I'm working on fixing that.

index.bs
sadym-chromium: I find using `ObjectId` as `internalId` confusing. From the implementation perspective, if we implement weak `objectId` map, it will be different from the `internalId`, as it will unlikely live in V8.

index.bs
sadym-chromium: option `1` "Create RemoteValue before serializing each value" is implemented.

index.bs
```diff
@@ -1039,14 +1039,12 @@ Issue: handle "stale object reference" case.
...
 
-1. For each |object| → |id value| of the given |realm|'s [=object id map=]:
+1. If the given |realm|'s [=handle object map=] does not contain |handle value|,
+   return [=error=] with [=error code=] [=invalid argument=].
 
-  1. If |id value| is equal to |object id|, return [=success=] with data
-     |object|
-
-1. Return [=error=] with [=error code=] [=invalid argument=].
+1. Return the |handle id| value of the given |realm|'s [=handle object map=].
```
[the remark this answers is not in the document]

index.bs
```diff
@@ -1417,15 +1415,21 @@ object, specified as <code>RemoteValue</code>. The value's type is specified in
...
-further <code>RemoteValue</code> instances. The <code>value</code> property can
-be null if there is a duplicate object i.e. the object has already been
-serialized in the current <code>RemoteValue</code>, perhaps as part of a
-cycle, or otherwise when the maximum serialization depth is reached.
+string representation of the value.
+For non-primitive objects, the <code>handle</code> property contains a string id that
+provides a unique handle to the object, valid for realm's lifetime, unless the handle
+is disowned by the <code>script.disown</code> command. The handles are unique,
+meaning serialization of the same ECMAScript object twice will have 2 different
+handles. The object will not be disposed until all the handles are disowned.
```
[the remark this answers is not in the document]
sadym-chromium: Nice wording! Done.

index.bs
```diff
@@ -1417,15 +1415,21 @@ object, specified as <code>RemoteValue</code>. The value's type is specified in
...
+For some non-primitive types, the <code>value</code> property contains a
+representation of the data in the ECMAScript object; for container types this can
+contain further <code>RemoteValue</code> instances. The <code>value</code> property
+can be null or omitted if there is a duplicate object i.e. the object has already
+been serialized in the current <code>RemoteValue</code>, perhaps as part of a cycle,
+or otherwise when the maximum serialization depth is reached.
+In case of duplicated objects in the same <code>RemoteValue</code>, the value is
```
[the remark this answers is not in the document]
sadym-chromium: "for a given serialization" of cause. Fixed. Thanks!

index.bs
```diff
@@ -1437,18 +1441,38 @@ object is discarded in the runtime, subsequent attempts to access it via the
...
 <div algorithm>
-To get the <dfn>object id for an object</dfn> given |realm| and |object|:
+To get the <dfn>handle for an object</dfn> given |realm|, |ownership|
+and |object|:
+
+1. If |ownership| is equal <code>none</code>, return <code>null</code>.
+
+1. Let |handle id| be a new, unique, string handle for |object|.
+
+1. Set the value of |handle id| in |realm|'s [=handle object map=] to |object|.
```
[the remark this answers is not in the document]

index.bs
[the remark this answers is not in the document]
sadym-chromium: My suggestion is to keep string IDs to provide flexibility on the implementation side.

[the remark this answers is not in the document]
jgraham: Honestly if this is just an internal id that's fully set during serialization I wonder if we should just specify it as a counter i.e. the first object involved in a cycle gets internalId 0, the next one gets internal id 1, etc. That minimises the chance that people expect the id to work outside the specific serialization.

I'm not exactly sure what flexibility you're looking for on the implementation side, but if the plan is to expose the CDP objectId directly that would seem to be an interop hazard since a client might come to depend on that id being usable in other parts of the API. Generally I think we should either a) make the id here purely something that the local end (client) can use to deserialize the specific object graph, and useless outside of that or b) revisit the possibility of just using `objectId` and having a future upgrade path to `objectId` references that act as global identifiers for the object but don't express ownership.

index.bs
```diff
@@ -1610,7 +1660,7 @@ Issue: handle String / Number / etc. wrapper objects specially?
 <div algorithm>
 
 To <dfn>serialize as a remote value</dfn> given a |value|, a |max depth|,
-|node details|, a |set of known objects| and a |realm|:
+a |node details|, an |ownership|, a |serialization internal map| and a |realm|:
```
jgraham, line 1663: Maybe using `|ownership type|` or similar everywhere would be clearer? Minor nit though.

index.bs
```diff
@@ -1620,28 +1670,41 @@ To <dfn>serialize as a remote value</dfn> given a |value|, a |max depth|,
 1. In the following list of conditions and associated steps, run the first set
    of steps for which the associated condition is true:
 
+1. Let |handle id| be the [=handle for an object=] given |realm|, |ownership|
+   and |value|.
+
+1. Let |known object| be <code>true</code>, if |value| is in the
+   |serialization internal map|, otherwise <code>false</code>.
+
+1. Let |child ownership| be <code>none</code>.
```
jgraham, line 1679: We usually write string literals like `"<code>none</code>"`.

index.bs
```diff
@@ -3783,6 +3915,9 @@ argument doesn't affect function's <code>this</code> binding.
 
 Issue: TODO: Add timeout argument as described in the script.evaluate.
 
+Issue: TODO: Add ownership model parameter.
```
jgraham, line 3918: Why don't we just fix this? I'm not a fan of leaving a TODO to actually expose the feature we've got the backend implementation for, and it's just a matter of adding a single optional parameter to the CDDL and a single step to the algorithm.

index.bs
```diff
@@ -3931,6 +4067,9 @@ case the resolved value of the promise is returned.
    </dd>
 </dl>
 
+Issue: TODO: Add ownership model parameter.
```
jgraham, line 4070: Also here.

jgraham (changes requested):
One question I have is how you expect objects in events to be handled e.g. `log.entryAdded`. I assume we don't want to take ownership of any object that ends up in a log event, but in the current setup we don't have any way to access the original object, only the serialised copy.

index.bs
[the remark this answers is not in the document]
sadym-chromium: I made an `internalId` a unique per serialization number. I guess it prevents uses from relying on them in any other then an expected way.

index.bs
sadym-chromium: `|ownership type|` seems more readable, done, thanks!

index.bs

index.bs

sadym-chromium: As [discussed](https://github.com/w3c/webdriver-bidi/issues/90#issuecomment-1110841620), I made `none` a default to prevent potential memory leaks.

index.bs

index.bs
```diff
@@ -3362,8 +3450,9 @@ The <code>ExceptionDetails</code> type represents a JavaScript exception.
    this data with regex or something equally bad.
 
 1. Let |exception| be the result of [=serialize as a remote value=] given
-   |record|.\[[Value]], <code>1</code> as |max depth|, <code>true</code> as |node details|,
-   a <code>new set</code> as |set of known objects| and |realm|.
+   |record|.\[[Value]], <code>1</code> as |max depth|, <code>root</code> as
```
jgraham, line 3453: This shouldn't be hardcoding `ownership type` (or `max depth`, but that's an existing bug); instead it should be passed from the caller and match the command parameters.

jgraham (changes requested):

index.bs
```diff
@@ -1437,18 +1448,44 @@ object is discarded in the runtime, subsequent attempts to access it via the
...
-     object, this must be the [=browsing context id=] for |object|.
+    1. Let |previously serialized remote value| be
+       |serialization internal map|[|object|].
 
-  1. Set the value of |object| in |realm|'s [=object id map=] to |object id|.
+    1. If |previously serialized remote value| does not have a field
+       <code>internalId</code>, run the following steps:
 
-1. Return the result of getting the value for |object| in |realm|'s [=object id map=].
+      1. Let |internal id| be a unique across the <code>internalId</code> fields of
```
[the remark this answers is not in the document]
sadym-chromium: Having `internalId` a number seems enough to prevent uses from trying to parse it or rely on their structure. Requiring them to be a consequent numbers doesn't seem to benefit users, but can complicate implementation. E.g. reserving the `internalId` and the moment of putting the object to the set vs using counter at the moment of duplication found produces different results.

index.bs
sadym-chromium: valid point. Fixed.

index.bs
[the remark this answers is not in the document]
jgraham: OK, if you feel like it's going to complicate your implementation we can make it arbitary.

sadym-chromium:
> One question I have is how you expect objects in events to be handled e.g. `log.entryAdded`. I assume we don't want to take ownership of any object that ends up in a log event, but in the current setup we don't have any way to access the original object, only the serialised copy.

I can think of the following options with decreasing priority:
1. Add an optional `ownership` param to the `session.subscribe` command.
1. (least preferable) keep default ownership model `none` and delegate the keeping alive logic to user (least preferable).

index.bs
```diff
@@ -1477,102 +1514,122 @@ RemoteValue = {
   WindowProxyRemoteValue //
 }
 
+InternalId = Number;
```
jgraham, line 1517: Capital-N Number isn't a type. Also I'd strongly prefer that we use an integer here; allowing the identifier to be a float is just asking for problems.

jgraham (changes requested):

index.bs
sadym-chromium: My bad. Done

foolip:
> Add an optional `ownership` param to the `session.subscribe` command.

This seems like the better choice to me, since it avoids taking ownership (and requiring release) by default. One thing we'll need to work out when adding ownership to event subscription is what `session.subscribe("log")` + `session.subscribe("log.entryAdded")` does when called with different ownership settings and different orders.

index.bs
```diff
@@ -1514,7 +1514,7 @@ RemoteValue = {
   WindowProxyRemoteValue //
 }
 
-InternalId = Number;
+InternalId = integer;
```
jgraham, line 1517: I think you want `uint` (`integer` also allows BigInt values).

jgraham:
I think we should file a seperate issue for objects returned in events. It's possible some configuration system is required which allows attaching configuration data to specific modules/contexts.

One thing I'd like to understand is how this works in CDP. It looks like if I call `console.log([])` CDP is returning an objectId that is later used e.g. if I expand the array in devtools. So does CDP by default leak all logged objects? Is the devtools frontend actually performing some kind of cleanup to prevent unbounded memory usage?

index.bs
sadym-chromium: Exactly! Thanks )

sadym-chromium:
Issue to handle `log.entryAdded` ownership model: https://github.com/w3c/webdriver-bidi/issues/214

jgraham (approved):

PR #207, opened 2022-05-10
Make `objectId` realm-specific

sadym-chromium:
Step 1 towards addressing #90 and implementing the [proposal](https://github.com/w3c/webdriver-bidi/issues/90#issuecomment-1111052638).

* Make `ObjectId` realm-specific.

_Recreated [PR #205](https://github.com/w3c/webdriver-bidi/pull/205). It was done to move the PR from fork to the brunch, so that [the next PR](https://github.com/w3c/webdriver-bidi/pull/206) can be based on this one._

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/207.html" title="Last updated on May 10, 2022, 3:44 PM UTC (d69c01d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/207/3dbd65e...d69c01d.html" title="Last updated on May 10, 2022, 3:44 PM UTC (d69c01d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1019,13 +1019,11 @@ Issue: Define how this works.
 
 ## Reference ## {#data-types-reference}
 
-<dfn>RemoteReference</dfn> represents a remote reference to an exising ECMAScript object in
-[=object id map=].
-
-A [=BiDi session=] has an <dfn>object id map</dfn>. This is a weak map
+Each EMCAScript [=Realm=] has a corresponding <dfn>object id map</dfn>. This is a strong map
 from objects to their corresponding id.
 
-Issue: Should this be explicitly per realm?
+<dfn>RemoteReference</dfn> represents a remote reference to an exising ECMAScript object in
+[=object id map=] in the given [=Realm=].
 
 ```
 ObjectId = text;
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1299,12 +1297,12 @@ To <dfn>deserialize key-value list</dfn> given a |serialized key-value list|:
...
    1. Otherwise let |deserialized key| be result of [=trying=] to [=deserialize local value=]
-      with |serialized key|.
+      given |realm| and |serialized key|.
 
    1. Let |serialized value| be |serialized key-value|[1].
 
-   1. Let |deserialized value| be result of [=trying=] to [=deserialize local value=] with
-      |serialized value|.
+   1. Let |deserialized value| be result of [=trying=] to [=deserialize local value=]
+      given |realm| and |serialized value|.
```
jgraham, line 1305: I regularly get confused here, but I think the convention is to use `given` when defining an algorithm and `with` when calling it. So if we could update all the callers to use `with |realm| and |serialized value|` (or equivalent) that would help with consistency.

index.bs
```diff
@@ -1613,7 +1610,7 @@ Issue: handle String / Number / etc. wrapper objects specially?
 <div algorithm>
 
 To <dfn>serialize as a remote value</dfn> given a |value|, a |max depth|,
```
jgraham, line 1612: I would have made |realm| the first argument here.

jgraham (approved):
Basically this is fine, just a couple of editorial suggestions.

PR #208, opened 2022-05-11
Don't break lines in some blocks

sadym-chromium:
To ease text search, added check verifying:
1. No break lines in `[=...=]` blocks.
2. No break lines in `[...]` blocks.
3. No break lines in `|...|` blocks.
4. No break lines in `<var>...</var>` blocks.
5. No break lines in `<dfn>...</dfn>` blocks.

And fixed existing issues in code.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/208.html" title="Last updated on Dec 5, 2022, 6:58 PM UTC (ed7ebf9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/208/b3e6a24...ed7ebf9.html" title="Last updated on Dec 5, 2022, 6:58 PM UTC (ed7ebf9)">Diff</a>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,7 +43,7 @@ jobs:
         with:
           toolchain: stable
       - name: Validate CDDL files
-        run: ./scripts/test.sh
+        run: ./scripts/cddl/test.sh
       - name: Archive CDDL files
         uses: actions/upload-artifact@v3
         with:
--- a/scripts/cddl/test.sh
+++ b/scripts/cddl/test.sh
@@ -1,0 +1,21 @@
+#! /bin/bash
+set -ex
+
+SCRIPT_DIR=$(cd $(dirname "$0") && pwd -P)
+ROOT=$(dirname $SCRIPT_DIR)
+
+if ! [ -x "$(command -v cddl)" ] || [ "$1" = "--upgrade" ]; then
+  echo 'Installing cddl'
+  cargo install cddl
+fi
+
+if [[ "$(npm list parse5)" =~ "empty" ]] || [ "$1" = "--upgrade" ]; then
+  echo 'Installing npm package parse5'
+  npm install parse5
+fi
+
+# Extract CDDL content from spec into files
+$ROOT/cddl/generate.js
+
+cddl compile-cddl --cddl local.cddl
+cddl compile-cddl --cddl remote.cddl
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -4,18 +4,5 @@ set -ex
 SCRIPT_DIR=$(cd $(dirname "$0") && pwd -P)
 ROOT=$(dirname $SCRIPT_DIR)
 
-if ! [ -x "$(command -v cddl)" ] || [ "$1" = "--upgrade" ]; then
-  echo 'Installing cddl'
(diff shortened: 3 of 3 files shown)


jgraham:
I'm inclined not to merge this kind of change unless we have some way to maintain the formatting going forward. Otherwise it's going to regress again pretty quickly.

sadym-chromium:
> I'm inclined not to merge this kind of change unless we have some way to maintain the formatting going forward. Otherwise it's going to regress again pretty quickly.

I don't know an automatic way to handle this, so I don't have a batter idea rather then rely on authors and reviewers.

sadym-chromium:
I added an agenda to today's meeting.

sadym-chromium:
Not sure if we should keep `<dfn>...</dfn>` like in [`dfn>local end \n definition</dfn>`](https://github.com/w3c/webdriver-bidi/blob/20e65050a258c5541080c69103ac97af6158fa5b/index.bs#L219-L220) in the same line though.

sadym-chromium:
I misread your message as "inclined to merge", so continued working on that. 

From one side, checking string length manually doesn't seem convinient. From the other, the overhead doesn't seem to be that big comparing to the advantage of being able to search without tricky regexps.

juliandescottes:
This sounds easy to test with a small script, which could be triggered automatically via github actions. Don't know if we want to add a package.json here, but the following should do the trick?

```javascript
const fs = require('fs');

const lines = fs.readFileSync('index.bs').toString().split("\n");
let failed;
for(let i = 0; i < lines.length; i++) {
    const line = lines[i];
    if (line.lastIndexOf("[=") > line.lastIndexOf("=]")) {
        console.log(`Unclosed [= =] at line ${i + 1}`)
        console.log(`L${i}: ${line}`);
        failed = true;
    }

    if ((line.split("|") - 1).length % 2) {
        console.log(`Unclosed | ... | at line ${i + 1}`)
        console.log(`L${i}: ${line}`);
        failed = true;
    }
}

if (failed) {
    throw new Error("Unclosed '[= ... =]' or '|...|'");
}
```

sadym-chromium:
I added the checks and reformatted the code.

sadym-chromium:
WDYT @jgraham @whimboo ?

whimboo:
I think that this looks good and the formatter detects line breaks correctly. I wonder if we should also add this check to the `test.sh` script so that it is not always necessary to push the PR. This would save a lot of resources.

sadym-chromium:
> I think that this looks good and the formatter detects line breaks correctly. I wonder if we should also add this check to the `test.sh` script so that it is not always necessary to push the PR. This would save a lot of resources.

done

whimboo:
Thanks Maksim! @jgraham given that you had concerns regarding missing verifications for upcoming changes can you please do the final review?

jgraham:
I wonder if using https://github.com/domfarolino/specfmt is a better approach here (I haven't tested it yet).

jgraham:
Apparently that doesn't try to enforce this kind of condition yet (and therefore this PR would make us incompatible with using that tool). Bleh. Would be really nice to have some proper autoformatting for specs that could enforce this kind of property.

PR #216, opened 2022-05-25
Add a CI job to check for warnings

jgraham:
This attempts a build and fails the job if there are any warnings.


jgraham:
Note that this depends on https://github.com/w3c/webdriver-bidi/pull/194

foolip (approved):
If CI passes after rebase, then go!

PR #232, opened 2022-06-09
Fix result of deserialization for type date

### LINKED ISSUE #219: Shouldn't deserialized type `date` return `Date` object?
Right now [deserialization](https://w3c.github.io/webdriver-bidi/#deserialize-local-value) specifies for type `date` that we should return `Date.parse(value)` which would mean returning in the end a timestamp. Shouldn't it maybe be something more like `Date(Date.parse(value))` to get a Date object?

lutien:
Closes #219

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -87,6 +87,7 @@ spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
     text: CreateListFromArrayLike; url: sec-createlistfromarraylike
     text: CreateMapIterator; url: sec-createmapiterator
     text: CreateSetIterator; url: sec-createsetiterator
+    text: Date; url: sec-date
     text: Date Time String Format; url: sec-date-time-string-format
     text: Date.ToDateString; url: sec-todatestring
     text: Date.parse; url: sec-date.parse
@@ -1367,7 +1368,7 @@ To <dfn>deserialize local value</dfn> given |realm| and |local protocol value|:
       1. If |value| does not match [=Date Time String Format=], return [=error=] with
          [=error code=] [=invalid argument=].
 
-      1. Return [=success=] with data [=Date.parse=](|value|).
+      1. Return [=success=] with data [=Date=]([=Date.parse=](|value|)).
 
     <dt>|type| is the string "<code>map</code>"
     <dd>


index.bs
```diff
@@ -1367,7 +1368,7 @@ To <dfn>deserialize local value</dfn> given |realm| and |local protocol value|:
       1. If |value| does not match [=Date Time String Format=], return [=error=] with
          [=error code=] [=invalid argument=].
 
-      1. Return [=success=] with data [=Date.parse=](|value|).
+      1. Return [=success=] with data [=Date=]([=Date.parse=](|value|)).
```
whimboo, line 1371: So this would call `Date()` as function and would return a string representation of the date. I'm sure that we actually want to have a Date object instead, which would require the usage of `new`. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#constructor

whimboo: Actually the `Date` constructor should hereby also work identical to the static `Date.parse()` method so that we may be able to actually get rid of `Date.parse()`.

lutien: Ok, I just thought it's the way we write it in the specification, for example [here](https://w3c.github.io/webdriver-bidi/#deserialize-local-value) for the `Map` it's `Map(iterable)` instead of `new Map(iterable)`. I agree that's for `Date` it's confusing since it can be called as a function and constructor.
With `Date.parse`, yes, it's true, it's not really required, I'll remove it

whimboo: @jgraham looks like we need feedback in regards the usage of `new`. Could you tell us how should be used here?

jgraham: Well for the same operation HTML write `Otherwise, if serialized.[[Type]] is "Date", then set value to a new Date object in targetRealm whose [[DateValue]] internal slot value is serialized.[[DateValue]].`, which sidesteps the problem.

Since `Date` works differently when called as a function and as a constructor, I wonder if the spec-wise correct way to invoke it is using https://tc39.es/ecma262/#sec-construct and an assert that it returns a normal completion type.

lutien: So then would it be something like 
```
2. Let "date_result" be "Construct(Date, [value])". If "date_result" has no a [[DateValue]] internal slot, return error with error code "invalid argument".
3. Return "success" with data "date_result".
```

index.bs
```diff
@@ -1372,7 +1372,7 @@ To <dfn>deserialize local value</dfn> given |realm| and |local protocol value|:
 
       1. Let |date result| be [=Construct=]([=Date=], |value|).
 
-      1. Assert: [=ReturnIfAbrupt=](|date result|) is not a [=abrupt completion=].
+      1. Assert: [=ReturnIfAbrupt=](|date result|) is not an [=abrupt completion=].
```
jgraham, line 1375:
```suggestion
      1. Assert: |date result| is not an [=abrupt completion=].
```

Apologies for the confusion, I didn't mean that you need to call `ReturnIfAbrupt`, I just meant that ES is already calling that so the return type is either a value or an abrupt completion.

Issue #238, opened 2022-06-22
Consider changing the default value of `awaitPromise` to false


juliandescottes:
The `awaitPromise` parameter used in [script.evaluate](https://w3c.github.io/webdriver-bidi/#command-script-evaluate) and [script.callFunction](https://w3c.github.io/webdriver-bidi/#command-script-callFunction) is a boolean which defaults to true. 

It sounds like a sensible default behavior, but optional boolean parameters which default to true can be confusing. Developers commonly assume that if a flag is not provided, it will be falsy.

Could we make it default to false? Or make it mandatory? 

cc @sadym-chromium

jgraham:
I tend to agree that we should make it a general rule that missing booleans default to false (and there are probably other similar rules for simple types), and should update the existing spec to match these rules.

sadym-chromium:
The default behaviour should hide unnecessary complexity from the users don't need them. I'd assume `awaitPromise=false` is more extravagant use-case then `awaitPromise=true`, so keeping the behaviour `awaitPromise=true` by default makes more sense for my mind. Flipping the name to something like `donNotWaitForPromise` seems quite confusing as well.

juliandescottes:
Thanks for the feedback! I agree switching the default behavior is not great. Making it mandatory would partially address that but it's making the API more verbose for the sake of clarity. We could find a shorter name, `noAwait`, `sync`, ...?

If we are saying that for this parameter the default behavior is really what most of the consumers should be happy with, then making the parameter name slightly longer / obscure is not necessarily a bad thing?

sadym-chromium:
Solved by making the field obligated

PR #239, opened 2022-06-22
Make `awaitPromise` param obligatory.

sadym-chromium:
https://github.com/w3c/webdriver-bidi/issues/238

Make `awaitPromise` obligatory to avoid confusion by default `true` value.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/239.html" title="Last updated on Jun 22, 2022, 11:41 AM UTC (d18d927)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/239/0825f8b...d18d927.html" title="Last updated on Jun 22, 2022, 11:41 AM UTC (d18d927)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1578,16 +1578,16 @@ DateRemoteValue = {
 
 MapRemoteValue = {
   type: "map",
+  value: MappingRemoteValue,
   ?handle: Handle,
   ?internalId: InternalId,
-  value: MappingRemoteValue,
 }
 
 SetRemoteValue = {
   type: "set",
+  value: ListRemoteValue
   ?handle: Handle,
   ?internalId: InternalId,
-  value: ListRemoteValue
 }
 
 WeakMapRemoteValue = {
@@ -1654,9 +1654,9 @@ NodeRemoteValue = {
 NodeProperties = {
   nodeType: uint,
   nodeValue: text,
+  childNodeCount: uint,
   ?localName: text,
   ?namespaceURI: text,
-  childNodeCount: uint,
   ?children: [*NodeRemoteValue],
   ?attributes: {*text => text},
   ?shadowRoot: NodeRemoteValue / null,
@@ -2540,9 +2540,9 @@ BrowsingContextInfoList = [* BrowsingContextInfo]
 
 BrowsingContextInfo = {
   context: BrowsingContext,
-  ?parent: BrowsingContext / null,
   url: text,
   children: BrowsingContextInfoList / null
+  ?parent: BrowsingContext / null,
 }
 
 </pre>
@@ -4185,9 +4185,9 @@ Note: In case of an arrow function in <code>functionDeclaration</code>, the
       ScriptCallFunctionParameters = {
         functionDeclaration: text;
         awaitPromise: bool;
+        target: Target;
         ?arguments: [ArgumentValue];
         ?this: ArgumentValue;
-        target: Target;
         ?resultOwnership: OwnershipModel;
       }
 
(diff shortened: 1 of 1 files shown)


jgraham:
I don't think `sync` is nearly as clear in intent as `awaitPromise`.

sadym-chromium:
> I don't think `sync` is nearly as clear in intent as `awaitPromise`.

`syncCall` ?

jgraham:
The problem is that the operation is called `await`, and that's directly reflected in the ES syntax, and so anything using that term is clear and anything not using it is less clear. If you wanted to go down this path I think `noAwaitPromise` would be a clear name, but it's not especially obvious to me that adding a negation in the parameter name just to make the missing value default correspond to our perception of the common case is worthwhile.

jgraham:
Thinking a bit more about this, I think one could argue that the closest thing to the opposite of `await` is `return` and so `returnPromise` is an OK name in the sense that it's getting you to opt in to just returning a promise rather than first awaiting it. But I also think that explains why all these names feel a bit off; just returning the evaluated value is the default behaviour. This parameter is carving out a special behaviour for promises. Making the special behaviour the default feels more confusing to people who understand the model. Of course, many end users won't understand the model. But they are using clients that can put a more friendly API on top of this (e.g. puppeteer unconditionally awaits promises). So really the protocol API surface is targeted at client authors who need to understand the detailed model anyway. In that case it makes more sense to make the uniform behaviour the default and require opting in to the special behaviour, even if we expect most uses to opt in.

sadym-chromium:
> Thinking a bit more about this, I think one could argue that the closest thing to the opposite of `await` is `return` and so `returnPromise` is an OK name in the sense that it's getting you to opt in to just returning a promise rather than first awaiting it. But I also think that explains why all these names feel a bit off; just returning the evaluated value is the default behaviour. This parameter is carving out a special behaviour for promises. Making the special behaviour the default feels more confusing to people who understand the model. Of course, many end users won't understand the model. But they are using clients that can put a more friendly API on top of this (e.g. puppeteer unconditionally awaits promises). So really the protocol API surface is targeted at client authors who need to understand the detailed model anyway. In that case it makes more sense to make the uniform behaviour the default and require opting in to the special behaviour, even if we expect most uses to opt in.

To remove room of misinterpretation, let's simply make the param obligatory.

jgraham (approved):
I think this is a reasonable compromise.

juliandescottes (approved):
Thanks for the PR! I agree this is a good compromise

index.bs
```diff
@@ -4184,9 +4184,9 @@ Note: In case of an arrow function in <code>functionDeclaration</code>, the
 
       ScriptCallFunctionParameters = {
         functionDeclaration: text;
+        awaitPromise: bool;
         ?arguments: [ArgumentValue];
         ?this: ArgumentValue;
-        ?awaitPromise: bool;
         target: Target;
```
whimboo, line 4190: If the reason to move it up is that we have all required arguments first, we probably should do the same with `target` as well. Maybe it could be done now given that you modify the surrounding code in your PR.

sadym-chromium: Moved all the optional params to the end.

sadym-chromium:
> @sadym-chromium I'm adding support for awaitPromise in Firefox in https://phabricator.services.mozilla.com/D149412
> 
> Do you mind if I also update the tests to match the spec changes from this PR? I would do it as an additional patch in my stack.

Sure, please, go ahead!

PR #248, opened 2022-06-24
Ensure realm ids are different from handles or context ids

jgraham:
This is to prevent some implementations making e.g. the realm id for a
Window object equal to the handle for that object, and clients
starting to rely on those things always being the same.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/248.html" title="Last updated on Jun 24, 2022, 1:45 PM UTC (91094ec)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/248/3ca8428...91094ec.html" title="Last updated on Jun 24, 2022, 1:45 PM UTC (91094ec)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3730,6 +3730,13 @@ Each [=realm=] has an associated <dfn export>realm id</dfn>, which is a string
 uniquely identifying that realm. This is implicitly set when the realm is
 created.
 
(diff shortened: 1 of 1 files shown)


whimboo (approved):
Thank you @sadym-chromium! We have those different as well. As such lets get this change landed.

PR #250, opened 2022-06-28
Always add `parent` to `BrowsingContextInfo`

sadym-chromium:
Suggestion to always provide `parent` when serialising `BrowsingContextInfo`. Current approach can be confusing, and doesn't allow to process contexts received by `getTree` recursively, and requires to pass `parent` to process each child.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/250.html" title="Last updated on Jun 28, 2022, 11:56 AM UTC (87b4792)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/250/007295e...87b4792.html" title="Last updated on Jun 28, 2022, 11:56 AM UTC (87b4792)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2547,7 +2547,7 @@ BrowsingContextInfo = {
   context: BrowsingContext,
   url: text,
   children: BrowsingContextInfoList / null
-  ?parent: BrowsingContext / null,
+  parent: BrowsingContext / null,
 }
 
 </pre>
@@ -2556,8 +2556,7 @@ The <code>BrowsingContextInfo</code> type represents the properties of a
 browsing context.
 
 <div algorithm>
-To <dfn>get the browsing context info</dfn> given |context|,
-|max depth| and |is root|:
+To <dfn>get the browsing context info</dfn> given |context| and |max depth|:
 
 1. Let |context id| be the [=browsing context id=] for |context|.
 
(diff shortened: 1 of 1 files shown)


foolip (approved):
Seems OK to me, to have both a missing `parent` field and null as options seems unhelpful to me. Second opinion, @jgraham?

jgraham:
Missing and null meant quite different things though: null meant "this is a top level browsing context", missing meant "we didn't provide parent info", the latter typically happening in `getTree` where you already know the parent because you have the whole tree.

I don't really understand the argument that it prevents processing `getTree` recursively; it seenms pretty straightforward to write a recursive implementation that passes down the parent id rather than duplicating it in each node. I think the argument that it makes processing `getTree` different to `contextCreated` is a bit stronger, but in any case you need to handle the root which will always have an explicit `parent`. But I'm still not sure that adding duplicate information is a good tradeoff.

foolip:
@sadym-chromium can you elaborate on the implementation challenge (I guess?) that motivated this?

sadym-chromium:
It's not challenging to implement, it just seems a bit odd. It saves some bytes, but as James mentioned, having different way of processing the same thing (`BrowsingContextInfo`) seems a bit odd. I'm fine abandoning this PR if we decide that users convenience (testing frameworks developers) is overweighted by (quite tiny for my mind) traffic overhead.

jgraham:
My main concern isn't really overhead, it's having multiple sources of truth for the same information.

sadym-chromium:
> My main concern isn't really overhead, it's having multiple sources of truth for the same information.

Good point. Closed.

PR #254, opened 2022-07-07
Align serialization format with EcmaScript

sadym-chromium:
It's a bit late, as we have already WPT tests for this part landed, but it's not too late, as it's still in development. Maybe we can align serialization with the ECMAScript:

1. As ECMAScript doesn't have `+Infinity`, I suggest to use `Infinity` instead of `+Infinity` in the serialization as well.
~~2. (Discussible, no strong opinion here) add suffix `n` at the end of BigInt's `value`, so that it can be evaluated directly.~~

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/254.html" title="Last updated on Jul 7, 2022, 1:11 PM UTC (d9e5515)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/254/007295e...d9e5515.html" title="Last updated on Jul 7, 2022, 1:11 PM UTC (d9e5515)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1099,7 +1099,7 @@ StringValue = {
   value: text,
 }
 
-SpecialNumber = "NaN" / "-0" / "+Infinity" / "-Infinity";
+SpecialNumber = "NaN" / "-0" / "Infinity" / "-Infinity";
 
 NumberValue = {
   type: "number",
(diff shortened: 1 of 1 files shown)


whimboo:
Personally I would be in agreement with your proposal, but would leave the decision up to @jgraham. I agree if we want to change that we should do it as soon as possible.

jgraham:
Can you rebase this so that it doesn't have the `Always add parent to BrowsingContextInfo` commit?

I think the `+Infinity`→`Infinity` change is straightforwardly good. The `"1"` →`"1n"` `BigInt` change arguably is adopting a peculiarity of ECMAScript into the protocol (a client in another language is going to have to strip out the `n` in order to parse the numeric value; even an ECMAScript implementation can just call `BigInt("1")`, but not `BigInt("1n")`). So I think I'd prefer not to take that one. But I don't feel very strongly either way.

sadym-chromium:
Created an issue #255 to discuss if we should add suffix to `BigInt`.

foolip:
Only the `Infinity` change remains, so I'll merge this.

sadym-chromium:
Sure!

sadym-chromium:
> @sadym-chromium would you mind creating a PR to update the existing wpt tests? Thanks.

https://github.com/web-platform-tests/wpt/pull/34754

PR #265, opened 2022-08-29
Add roadmap document

mathiasbynens:
During the 2022-08 WG meeting in Berlin, we agreed to formulate a shared roadmap based on real-world end-to-end user scenarios.

**Consider this PR an early draft to kick-start the discussion.** Does this format make sense? Which other roadmap items should be added? Should things be re-ordered? Etc.


roadmap.md
```diff
@@ -0,0 +1,44 @@
...
+This document presents an overview of real-world end-to-end user scenarios we aim to enable via the WebDriver BiDi protocol. Each scenario requires one or more WebDriver BiDi commands and events to be specified, tested, and implemented across browser engines.
+
+### A. Extracting headlines
+
+This scenario loads a web page and uses script evaluation to extract the headlines on the page. In spec terms, this involves:
+
+- [x] [handling sessions](https://w3c.github.io/webdriver-bidi/#module-session)
+- [x] [creating a new `browsingContext`](https://w3c.github.io/webdriver-bidi/#command-browsingContext-create)
+- [x] [navigating to a URL](https://w3c.github.io/webdriver-bidi/#command-browsingContext-navigate)
+- [x] [evaluating JavaScript in the page context](https://w3c.github.io/webdriver-bidi/#command-script-evaluate)
```
OrKoN, line 14: plot twist: you evaluate JavaScript to `document.querySelector` elements but the page author re-defined/deleted `document.querySelector` => the need for sandboxes

roadmap.md
```diff
@@ -0,0 +1,44 @@
...
+- [ ] [capturing a screenshot via a command](https://w3c.github.io/webdriver-bidi/#command-browsingContext-captureScreenshot)
+
+### D. Printing to PDF
+
+This scenario loads a web page and prints it as a PDF. In spec terms, this involves:
+
+- [x] (some items from the previous milestones)
+- [ ] [printing to PDF via a command](https://github.com/w3c/webdriver-bidi/issues/210)
+
+### E. Blocking images
```
OrKoN, line 39: HAR file generation => network logging?

roadmap.md
```diff
@@ -0,0 +1,44 @@
+# WebDriver BiDi roadmap
+
+## Real-world end-to-end user scenarios
+
+This document presents an overview of real-world end-to-end user scenarios we aim to enable via the WebDriver BiDi protocol. Each scenario requires one or more WebDriver BiDi commands and events to be specified, tested, and implemented across browser engines.
+
+### A. Extracting headlines
```
whimboo, line 7: I would suggest to use numbers given that we might have more than 24 scenarios in the end. Do we want to make use of `Milestones`? If not maybe we want to finally transfer all these scenarios to real milestones eventually. Would be good to also get @jgraham's opinion given that he wasn't present at the WG meeting.

Note that the current list of milestones doesn't match with those that we finally decided on to work. You may want to have a look at the meeting notes again.

Eg. the logging feature was the API that we supported and I feel as such it should come first. Also it's the first thing the Selenium project implemented for BiDi support.

mathiasbynens: > I would suggest to use numbers given that we might have more than ~24~ _26_ scenarios in the end.

Happy to use whatever numbering scheme. I think running out of digits/letters might not really become a problem in practice — especially since one the suggestions during the meeting was to keep the list of e2e scenarios relatively small: https://www.w3.org/2022/08/24-webdriver-minutes.html#t02:~:text=simonstewart%3A%20my%20preference%20would%20be%20define%20top%203%2C%20and%20when%20one%20is%20done%2C%20add%20another

> Note that the current list of milestones doesn't match with those that we finally decided on to work. You may want to have a look at the meeting notes again.

To clarify, the current patch is a simplified version with some example e2e scenarios based on our own roadmap, in a format I think might be useful here. I do not intend for this to be the final version! With this PR I merely wanted to kick off the discussion.

If this format looks good to everyone, we can start thinking of e2e scenarios that capture the remaining pieces mentioned in the meeting notes: https://www.w3.org/2022/08/24-webdriver-minutes.html#t02:~:text=Updated%20list%3A,*%20printing I’m hoping for the WG’s input to help come up with clear & concise e2e use cases. E.g. @whimboo, what would be an easy-to-understand e2e example involving logging?

roadmap.md
mathiasbynens: I like that suggestion; sandboxes nicely fit into this e2e use case (or any of the following ones) as a requirement for robustness against mutated built-ins.

roadmap.md
jgraham: Yes, agreed. For the spec, and implementations not based on CDP, network request logging is a necessary precursor to network request interception, and opens up clear use cases which don't require interception; notably simple correctness checks around ensuring that a request completes as expected, and request timing / HAR generation as part of performance logging. We've had specific requests for the latter of these from (iirc) BrowserTime.

roadmap.md
```diff
@@ -0,0 +1,54 @@
...
+### A. Extracting content
+
+This scenario loads a web page and uses script evaluation to extract content on the page (e.g. the headlines). In spec terms, this involves:
+
+- [x] [handling sessions](https://w3c.github.io/webdriver-bidi/#module-session)
+- [x] [creating a new `browsingContext`](https://w3c.github.io/webdriver-bidi/#command-browsingContext-create)
+- [x] [navigating to a URL](https://w3c.github.io/webdriver-bidi/#command-browsingContext-navigate)
+- [x] [evaluating JavaScript in the page context](https://w3c.github.io/webdriver-bidi/#command-script-evaluate)
+- [x] [serialization/deserialization of JavaScript values](https://w3c.github.io/webdriver-bidi/#data-types-protocolValue)
+- [ ] [sandboxed script execution](https://github.com/w3c/webdriver-bidi/issues/144)
```
jgraham, line 16: Although I agree sandboxed script exection is a high priority feature, I'd say the use case it blocks is more like "enable writing robust generic clients". Although running script directly in the context of the page is bad, for simple cases like implementing element selection it's almost always OK to just run `document.querySelector()` since it has no side effects. You only need to worry about sandboxing to deal with more edge case scenarios like pages which overwrite `querySelector` or where you're doing complex things that require storing significant browser-side state that should be invisible to the page.

roadmap.md
```diff
@@ -0,0 +1,54 @@
...
+
+### B. Submitting forms
+
+This scenario loads a web page, enter text into a form field via the keyboard, and submits the form via a mouse click before extracting the results from the page. In spec terms, this involves:
+
+- [x] (everything from the previous milestone)
+- [ ] [emulating keyboard input](https://github.com/w3c/webdriver-bidi/pull/175)
```
jgraham, line 23: In practice the plan is to just port the actions API to BiDi directly, so we should get all the input types at once. However I understand that from an implementation point of view it might not be that simple, and that mouse/keyboard are the lowest hanging fruit here.

roadmap.md
```diff
@@ -0,0 +1,54 @@
...
+- [ ] [emulating mouse input](https://github.com/w3c/webdriver-bidi/pull/175)
+
+### C. Capturing screenshots
+
+This scenario loads a web page and captures a screenshot. In spec terms, this involves:
+
+- [x] (some items from the previous milestones)
+- [ ] [capturing a screenshot via a command](https://w3c.github.io/webdriver-bidi/#command-browsingContext-captureScreenshot)
+
+### D. Printing to PDF
```
jgraham, line 33: I don't think this will be terribly difficult to implement, but I haven't seen how it came so high in the prority list compared to other features like network request logging/interception. In particular from a "value add" point of view this is already available through classic WebDriver and doesn't really benefit from the BiDi nature of the protocol. From a "basic automation features" point of view it's not nearly so widely used as screenshots, so should likely be further down the roadmap (insofar as it can be read linearly).

roadmap.md
```diff
@@ -0,0 +1,54 @@
...
+- [ ] [intercepting requests](https://github.com/w3c/webdriver-bidi/issues/66)
+
+### F. Measuring performance
+
+This scenario measures the performance of a cold load of a web page (without any cookies, storage, or anything in the cache).
+
+- [x] (some items from the previous milestones)
+- [ ] manipulating cookies
+- [ ] managing the browser cache
+- [ ] [installing bootstrap scripts](https://github.com/w3c/webdriver-bidi/blob/master/proposals/bootstrap-scripts.md#record-navigation-performance)
```
jgraham, line 54: It's not really clear to me why bootstrap scripts are really for the performance-mesaurement usecase? Is this the only thing for which they're currently used in Puppeteer? As I understand from the meeting notes, these are a relatively high priority from library authors because having scripts that you know run before any page scripts or other injected scripts unlocks the possibility to implement features that are otherwise difficult or impossible to get right in a race-free way.

roadmap.md
```diff
@@ -0,0 +1,54 @@
...
+
+- [x] (some items from the previous milestones)
+- [ ] [intercepting requests](https://github.com/w3c/webdriver-bidi/issues/66)
+
+### F. Measuring performance
+
+This scenario measures the performance of a cold load of a web page (without any cookies, storage, or anything in the cache).
+
+- [x] (some items from the previous milestones)
+- [ ] manipulating cookies
```
jgraham, line 52: Given that cookies are already available via WebDriver, but performance measurement isn't, I think that read/write access to cookies has non-performance use cases (e.g. check a specific cookie is set as expected after some action).

jgraham (changes requested):
Overall I'm happy with the idea of a shared roadmap, although I think basing it entirely around specific end-user cases makes it harder to justify infrastructure-level work like bootstrap scripts or sandboxing. 

I also don't think the details match up with the priorities we have, or what I could understand from the minutes of the recent meeting.

roadmap.md
mathiasbynens: Absolutely!

My goal was simply to come up with end-to-end scenarios or use cases that capture the different building blocks in useful ways, to make it easier to explain to folks unfamiliar with the spec bits. I make no claims that these are the only use cases, and am open to suggestions for better examples.

roadmap.md
mathiasbynens: Similar to https://github.com/w3c/webdriver-bidi/pull/265#discussion_r962766000, I didn’t mean to suggest that this is the only use case.

roadmap.md
mathiasbynens: Happy to move this down.

mathiasbynens:
> I'm happy with the idea of a shared roadmap, although I think basing it entirely around specific end-user cases makes it harder to justify infrastructure-level work like bootstrap scripts or sandboxing.

WDYT about explicitly including infra-level work in the first use case? Sandboxed scripts are already in there, but we could explicitly work bootstrap scripts in as well.

whimboo:
> > I'm happy with the idea of a shared roadmap, although I think basing it entirely around specific end-user cases makes it harder to justify infrastructure-level work like bootstrap scripts or sandboxing.
> 
> WDYT about explicitly including infra-level work in the first use case? Sandboxed scripts are already in there, but we could explicitly work bootstrap scripts in as well.

Sorry for the delay! We should definitely [take the outcome from the last W3C meeting](https://www.w3.org/2022/08/24-webdriver-minutes.html#t02) and add the identified top 3 scenarios first. Maybe we should also add a milestone for logging given that this feature triggered the BiDi spec initially and is mostly done for Chrome as well.

roadmap.md
mathiasbynens: Agreed; sandboxing is only required in cases the web page is non-cooperative/hostile.

Do you have a preference:

- move sandboxing from this scenario to another one (if so, where?)
- keep sandboxing here, but clarify in the scenario description that it’s only required for robustness

roadmap.md
mathiasbynens: Ack.

mathiasbynens: Ack. Did you want me to change anything here, or did you just mean to provide that context?

roadmap.md
mathiasbynens: Moved this down.

roadmap.md
mathiasbynens: Are you okay with this particular e2e scenario? If not, please suggest a concrete alternate e2e scenario that depends on bootstrap scripts. I went through the examples in [your excellent explainer doc](https://github.com/w3c/webdriver-bidi/blob/master/proposals/bootstrap-scripts.md#record-navigation-performance) and the performance-related scenario intuitively felt like the most common one.

roadmap.md
mathiasbynens: Did you mean to suggest we replace this e2e scenario with “check a specific cookie is set as expected after some action”, or rather add it as an separate, new scenario?

mathiasbynens:
I did a pass with some changes based on feedback. PTAL and keep making concrete suggestions :) Thanks!

roadmap.md
jgraham: I'd add a scenario specifically about "generic clients" or something that contains this and botostrap scripts.

mathiasbynens: Starting point:

```
### X. Enabling robust generic clients

This scenario enables end-to-end user scenarios for web pages whose environments aren’t guaranteed to be pristine (e.g. the page’s `document.querySelector` and `Object.prototype.*` have been overwritten), or scenarios that require running JavaScript before the page’s own scripts.

- [x] (everything from the previous milestone)
- [ ] [sandboxed script execution](https://github.com/w3c/webdriver-bidi/issues/144)
- [ ] [installing bootstrap scripts](https://github.com/w3c/webdriver-bidi/blob/master/proposals/bootstrap-scripts.md#record-navigation-performance)
```

Any suggestions on how to make this a proper end-to-end user scenario? It would be nice to have a specific scenario that can be turned into a runnable script, like all the other entries on the current roadmap draft.

whimboo:
@mathiasbynens I'll have a look at latest early next week. There is some other important work that I need to finish first. Sorry for the delay...

roadmap.md
jgraham: Well any of the content extraction ones, but you get to supply a URL rather than it being for a specfic URL where you happen to know that the injected scripts don't clash with the page-defined scripts. Or maybe a specific URL where they would clash, except for the sandboxing / use of bootstrap scripts to inject early.

whimboo:
@mathiasbynens I had a quick look at the individual roadmap entries and when I compare those to the initial comment on this PR I see differences:

* Where did network logging end-up? I cannot find that anymore and it's in the works right now, and only necessary for the performance measurement. We should separate it out from interception and make it a milestone before form submission?
* Bootstrap scripts (pre-load scripts) are very high on the list from August. Now it's in measuring performance. Is that the only foreseeable use case that we have here? I feel we should move it out from milestone E?

I could help with proposing changes that you can easily apply but before I would like to discuss the above questions. Thanks

mathiasbynens:
> Where did network logging end-up? I cannot find that anymore and it's in the works right now, and only necessary for the performance measurement. We should separate it out from interception and make it a milestone before form submission?

Sounds good to me. Do you want to propose an edit?

> Bootstrap scripts (pre-load scripts) are very high on the list from August. Now it's in measuring performance. Is that the only foreseeable use case that we have here? I feel we should move it out from milestone E?

This roadmap doc is not meant to be an exhaustive list of use cases for each spec feature. Rather, I attempted to list at least one example use case for every spec feature. I’d be happy to change the examples to more realistic / better ones, or to add additional examples.

> I could help with proposing changes that you can easily apply but before I would like to discuss the above questions.

Yes, please go ahead and propose concrete changes. Thanks for the input!

roadmap.md
```diff
@@ -0,0 +1,61 @@
...
+- [x] [evaluating JavaScript in the page context](https://w3c.github.io/webdriver-bidi/#command-script-evaluate)
+- [x] [serialization/deserialization of JavaScript values](https://w3c.github.io/webdriver-bidi/#data-types-protocolValue)
+- [ ] [sandboxed script execution](https://github.com/w3c/webdriver-bidi/issues/144)
+
+### B. Logging
+
+This scenario loads a web page and uses BiDi event subscription to efficiently get notified about each `console.log` entry. **This is not possible with classic WebDriver.** In spec terms, this involves:
+
+- [x] (some items from the previous milestone)
+- [x] [the `log` module](https://w3c.github.io/webdriver-bidi/#module-log)
```
diemol, line 23: Are scenarios related to performance or network logs included in a different section?

roadmap.md
```diff
@@ -0,0 +1,61 @@
...
+
+### C. Submitting forms
+
+This scenario loads a web page, enter text into a form field via the keyboard, and submits the form via a mouse click before extracting the results from the page. In spec terms, this involves:
```
diemol, line 27: What happens if the form also allows to be submitted by pressing the `Enter` key? Or in case of a touch screen, a tap. Would this scenario consider that as well?

roadmap.md
```diff
@@ -0,0 +1,61 @@
...
+
+- [x] (everything from the previous milestone)
+- [ ] [emulating keyboard input](https://github.com/w3c/webdriver-bidi/pull/175)
+- [ ] [emulating mouse input](https://github.com/w3c/webdriver-bidi/pull/175)
+
+### D. Blocking images
+
+This scenario loads a web page and uses request interception to block any images from loading. In spec terms, this involves:
```
diemol, line 35: Is "blocking resources" a better name? I assume we can block more than images.

diemol (commented):
Thank you for this, @mathiasbynens!

Left a couple of comments, more in the shape of questions/doubts. I appreciate the time you spent on this.

roadmap.md
mathiasbynens: > I assume we can block more than images.

Absolutely.

The scenarios in this doc are just examples, hand-picked in an attempt for them to be easy-to-understand and somewhat common/realistic examples. The underlying spec features required for each scenario certainly enable many more use cases!

roadmap.md
mathiasbynens: “emulating keyboard input” would support the Enter key as well. A tap is not explicitly part of this particular scenario but it would be supported through the same Input module.

The scenarios in this doc are just examples, hand-picked in an attempt for them to be easy-to-understand and somewhat common/realistic examples. They’re not meant to be exhaustive demonstrations of what’s possible.

roadmap.md
mathiasbynens: There is currently no scenario involving those types of logs. Feel free to propose one. We could potentially discuss this as a follow-up PR, once this one lands — WDYT?

diemol: Makes sense. We can do that later.

roadmap.md
diemol: Thank you for clarifying. I also just read another comment where the actions API would be present in BiDi, so ideally that covers it.

roadmap.md
diemol: Thank you.

roadmap.md
whimboo: I'm currently working on improvements for this document and should have something later today. I think that those scenarios that we already did should be all present as well and that in an order we initially decided on.

roadmap.md
whimboo: @OrKoN does that answer your question? Can this issue be closed?

roadmap.md
whimboo: As discussed I was working on enhancements for this PR locally. I'll have an update soon that I can share. Closing this issue for now.

roadmap.md
whimboo: I wonder if this should actually be a e2e example or not. It's clearly something for the milestone list but not sure if it fits here given that we want to keep the list short.

roadmap.md
whimboo: @jgraham what do you think? On the other side we could also use an example for a MutationObserver which would not such much overlap with performance measurements via network and load events.

roadmap.md
whimboo: Given that network logging is a feature that we have kinda high on the priority list since the beginning we should include it. With my proposal that I can share soon I have an e2e scenario included.

roadmap.md
OrKoN: My suggestion was to mention the HAR generation use case and link to the network logging feature. It can be done in a follow-up though.

mathiasbynens: Thanks. Let’s pursue this a follow-up then.

whimboo: Note that I'll have a network logging example in my upcoming proposal. But it will exclude the HAR file generation given that this is not part of the BiDi spec.

OrKoN: I don't think to be able to mark this as resolved: please consider this thread resolved.

roadmap.md
jgraham: Yeah, I think a scenario that involves bootstrap scripts + mutation observers to see an element as soon as it appears would be better. I think the requirement to poll for elements is one of the key weaknesses of WebDriver and one of the actual reasons that Selenium and others are interested in preload scripts.

roadmap.md
```diff
@@ -0,0 +1,61 @@
...
+
+- [x] (some items from the previous milestones)
+- [ ] [intercepting requests](https://github.com/w3c/webdriver-bidi/issues/66)
+
+### E. Measuring performance
+
+This scenario measures the performance of a cold load of a web page (without any cookies, storage, or anything in the cache).
```
jgraham, line 42: I think per what @whimboo says higher up, it would make more sense to have a "measuring performance" scenario that involves network logging events (that's a real end to end usecase we've built — at least as a prototype — on top of our current implementation.

jgraham (commented):
There are at least two use cases we know are important to people that are missing:

* Accessing resources behind HTTP Auth.

* Replacing API responses with known test data.

Those both depend on network request interception, which is probably next on the list after the stuff here.

roadmap.md
jgraham: I don't especially mind, but for this kind of performance testing you're probably using a fresh browser profile without existing cookies or storage, so although it's possible to claim that clearing cookies is a requirement for this use case, if "use a new profile for each test" isn't possible, a better feature would be having a way to get a guaranteed clean slate (e.g. by exposing containers or equivalent in the protocol) rather than relying on the test author to correctly reset each individual bit of browser state.

So I think a scenario in which it's specifically necessary to set or read a cookie would be better. Maybe something around testing that logging in sets a session cookie and that the session cookie is cleared after some logout action.

roadmap.md
whimboo: Alright, I'll go ahead and modify the example appropriately. Thanks.

roadmap.md
whimboo: Yes, I already modified that locally so that performance measurement uses the network events. And the client can use the received data to built-up a HAR file. I think that we can close this issue.

whimboo:
https://github.com/w3c/webdriver-bidi/pull/362 has the proposed updates. It's ready to be merged.

mathiasbynens:
> #362 has the proposed updates. It's ready to be merged.

I manually merged the changes to `ROADMAP.md` here: 9d6786211ba3bec7c2658fc912e924b0c73b904e

roadmap.md
```diff
@@ -0,0 +1,92 @@
...
+- [x] [Creating a new `browsingContext`](https://w3c.github.io/webdriver-bidi/#command-browsingContext-create)
+- [x] [Evaluating JavaScript in the page context](https://w3c.github.io/webdriver-bidi/#command-script-evaluate)
+- [x] [Closing the `browsingContext`](https://w3c.github.io/webdriver-bidi/#command-browsingContext-close)
+
+### Network events for measuring page load performance
+
+This scenario sets up handlers for network events and then navigates to a web page. The provided timing information from the emitted events can be used to measure the page load performance by storing the relevant data eg. in a HAR file. In spec terms, this involves:
+
+- [x] Some items from the previous scenario
+- [] [Network events for the request to be sent, and response started and completed](https://github.com/w3c/webdriver-bidi/pull/204)
```
mathiasbynens, line 36:
```suggestion
- [ ] [Network events for the request to be sent, and response started and completed](https://github.com/w3c/webdriver-bidi/pull/204)
```

roadmap.md
```diff
@@ -0,0 +1,92 @@
...
+- [x] Some items from the previous scenarios
+- [ ] [Intercepting network requests](https://github.com/w3c/webdriver-bidi/issues/66)
+
+### HTTP authentication
+
+This scenario loads a web page that is protected behind user credentials. In spec terms, this involves:
+
+- [x] Some items from the previous scenarios
+- [] [The event for a HTTP auth challenge](https://github.com/w3c/webdriver-bidi/issues/66)
+- [] [The command to provide the authentication response](https://github.com/w3c/webdriver-bidi/issues/66)
```
mathiasbynens, line 69:
```suggestion
- [ ] [The event for a HTTP auth challenge](https://github.com/w3c/webdriver-bidi/issues/66)
- [ ] [The command to provide the authentication response](https://github.com/w3c/webdriver-bidi/issues/66)
```

roadmap.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Handling onbeforeunload prompts
+
+This scenario loads a web page with a registered `beforeunload` event handler. After updating the value of some form input elements it should be checked that navigating away opens the beforeonload prompt. In spec terms, this involves:
+
+- [x] Some items from the previous scenarios
+- [] [The event when a user prompt opens](https://w3c.github.io/webdriver-bidi/#webdriver-bidi-user-prompt-opened)
+- [] [Handling the beforeunload prompt](https://w3c.github.io/webdriver-bidi/#command-browsingContext-handleUserPrompt)
+- [] [The event when a user prompt closes](https://w3c.github.io/webdriver-bidi/#webdriver-bidi-user-prompt-closed)
```
mathiasbynens, line 78:
```suggestion
- [ ] [The event when a user prompt opens](https://w3c.github.io/webdriver-bidi/#webdriver-bidi-user-prompt-opened)
- [ ] [Handling the beforeunload prompt](https://w3c.github.io/webdriver-bidi/#command-browsingContext-handleUserPrompt)
- [ ] [The event when a user prompt closes](https://w3c.github.io/webdriver-bidi/#webdriver-bidi-user-prompt-closed)
```

mathiasbynens:
After much online and offline discussion and concrete suggestions from many reviewers, I believe this patch is now in a good enough state to land as an initial version. We can of course still iterate on the roadmap as follow-up patches.

I’m re-requesting review bits in the hope of getting this landed soon.

OrKoN (approved):

explainer.md
```diff
@@ -31,7 +31,9 @@ The protocol is designed with the following goals in mind:
     - Simple for browser vendors to implement and maintain.
     - Possible for clients to enhance their WebDriver automation with browser-specific devtools protocol features.
 
-This document doesn't attempt to dive into the any of the new feature scenarios identified above, but rather tries to provide a solid foundation and the necessary primitives to build these features on. The document does walk through an example of an existing WebDriver feature (unhandled prompts) being updated for a bidirectional world.
+This document doesn't attempt to dive into the any of the new feature scenarios identified above, but rather tries to provide a solid foundation and the necessary primitives to build these features on. See [the roadmap](roadmap.md) for some example real-world user scenarios we aim to enable.
+
+The document does walk through an example of an existing WebDriver feature (unhandled prompts) being updated for a bidirectional world.
```
whimboo, line 36: I'm fairly sure that I removed that paragraph in my update given that it doesn't really apply.

whimboo (approved):
Looks fine to me beside a small issue. 

I would also like that @jgraham finally checks it from our side.

explainer.md
mathiasbynens, line 36:
```suggestion
```

mathiasbynens: @whimboo’s suggestion

mathiasbynens: Fixed

roadmap.md
```diff
@@ -0,0 +1,92 @@
...
+### Handling onbeforeunload prompts
+
+This scenario loads a web page with a registered `beforeunload` event handler. After updating the value of some form input elements it should be checked that navigating away opens the beforeonload prompt. In spec terms, this involves:
+
+- [x] Some items from the previous scenarios
+- [ ] [The event when a user prompt opens](https://w3c.github.io/webdriver-bidi/#webdriver-bidi-user-prompt-opened)
+- [ ] [Handling the beforeunload prompt](https://w3c.github.io/webdriver-bidi/#command-browsingContext-handleUserPrompt)
+- [ ] [The event when a user prompt closes](https://w3c.github.io/webdriver-bidi/#webdriver-bidi-user-prompt-closed)
+
+### Capturing screenshots
```
jgraham, line 80: I'd move this to after "submitting forms", since it's basically a prerequisite for many kinds of testing.

roadmap.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Observing changes being made to the DOM tree
+
+In this scenario a MutationObserver will be installed by a bootstrap script as early as the document gets created. It watches for changes made to the DOM tree and sends the relevant updates to the client. In spec terms, this involves:
+
+- [x] Some items from the previous scenarios
+- [ ] [Adding a bootstrap script](https://w3c.github.io/webdriver-bidi/#command-script-addPreloadScript)
+- [ ] [Installing the bootstrap script](https://w3c.github.io/webdriver-bidi/#preload-scripts)
+- [ ] [Back channel for communicating with the client](https://github.com/w3c/webdriver-bidi/pull/361)
+- [ ] [Removing a bootstrap script][https://w3c.github.io/webdriver-bidi/#command-script-removePreloadScript]
```
jgraham, line 54:
```suggestion
- [ ] [Removing a preload script](https://w3c.github.io/webdriver-bidi/#command-script-removePreloadScript)
```

roadmap.md
```diff
@@ -0,0 +1,92 @@
...
+- [ ] [Emulating keyboard input](https://github.com/w3c/webdriver-bidi/pull/175)
+- [ ] [Emulating mouse input](https://github.com/w3c/webdriver-bidi/pull/175)
+
+### Observing changes being made to the DOM tree
+
+In this scenario a MutationObserver will be installed by a bootstrap script as early as the document gets created. It watches for changes made to the DOM tree and sends the relevant updates to the client. In spec terms, this involves:
+
+- [x] Some items from the previous scenarios
+- [ ] [Adding a bootstrap script](https://w3c.github.io/webdriver-bidi/#command-script-addPreloadScript)
+- [ ] [Installing the bootstrap script](https://w3c.github.io/webdriver-bidi/#preload-scripts)
```
jgraham, line 52:
```suggestion
- [ ] [Adding a preload script](https://w3c.github.io/webdriver-bidi/#command-script-addPreloadScript)
- [ ] [Installing the preload script](https://w3c.github.io/webdriver-bidi/#preload-scripts)
```

jgraham (approved):
Looks basically fine to me, just a couple of small suggestions and typo fix.

mathiasbynens:
Thanks! Feedback addressed. I think this is good to go now.

foolip (approved):
Looks great, thanks! I read the whole thing and didn't find any typos, which is my usual contribution 😆

PR #270, opened 2022-09-06
Handle sandbox realms in getRealms command.

jgraham:
We want to return all the Window realms with the name of the sandbox, if any.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/270.html" title="Last updated on Oct 15, 2022, 2:11 PM UTC (d5e49a3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/270/582eed0...d5e49a3.html" title="Last updated on Oct 15, 2022, 2:11 PM UTC (d5e49a3)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3818,6 +3818,9 @@ WorkletRealmInfo = {
 }
 </pre>
 
+Note: there's a 1:1 relationship between the <code>RealmInfo</code> variants and values of
+      <a href="#type-script-RealmType"><code>RealmType</code></a>.
+
 The <code>RealmInfo</code> type represents the properties of a realm.
 
 <div algorithm>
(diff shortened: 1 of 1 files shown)


juliandescottes:
@sadym-chromium Hi Maksim! For information this came up [during our implementation of getRealms](https://phabricator.services.mozilla.com/D156049), and for now we are moving forward based on the proposal in this PR. If you have any early feedback about this change, we're eager to know :)

sadym-chromium:
We faced the same question. The question is what to do with the sandboxes created on the initial `about:blank` page. It looks like the specification implies the sandboxes created there should be transferred to the new page because of the ["that both the initial `about:blank` Document, and the new Document that is about to be created, will share the same Window object"](https://html.spec.whatwg.org/multipage/browsers.html#navigating-across-documents:is-initial-about:blank).

sadym-chromium (approved):
LGTM

whimboo:
@jgraham do you also want to have a look at these proposed changes?

Btw the PR needs an update given that it has conflicts.

index.bs
```diff
@@ -3828,7 +3844,8 @@ BaseRealmInfo = {
 WindowRealmInfo = {
   BaseRealmInfo,
   type: "window",
-  context: BrowsingContext
+  context: BrowsingContext,
+  ?sandbox: Text
```
whimboo, line 3848: There is a CDDL failure.
```suggestion
  ?sandbox: text
```

PR #276, opened 2022-09-08
Allow parsing of pre tags for multiple CDDL names

### LINKED ISSUE #275: Inconsistent generation of local and remote CDDL files (missing definitions)
When generating the remote.cddl file I miss references to `CapabilitiesRequest` even through [this type is marked for both local and remote](https://w3c.github.io/webdriver-bidi/#type-session-capabilitiesRequest).

```
CapabilitiesRequestParameters = {
  ?alwaysMatch: CapabilitiesRequest,
  ?firstMatch: [*CapabilitiesRequest]
}
```

Instead it's only listed in `local.cddl`, which isn't actually needed there given that the client won't have to parse that anyway.

whimboo:
Fix #275.

Until now the parsing of the individual `pre` tags did not allow multiple CDDL names to be listed. This resulted in writing these entries only to the file which name was listed last in the `pre` class name.

Also I removed filtering out of lines that miss a declaration. This actually drops lines unexpectedly. Instead we should get the validation step fail and report it appropriately.

***
<a href="https://pr-preview.s3.amazonaws.com/whimboo/webdriver-bidi/pull/276.html" title="Last updated on Sep 9, 2022, 10:34 AM UTC (5fbe1aa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/276/22530a8...whimboo:5fbe1aa.html" title="Last updated on Sep 9, 2022, 10:34 AM UTC (5fbe1aa)">Diff</a>

### DIFF
--- a/scripts/cddl/generate.js
+++ b/scripts/cddl/generate.js
@@ -10,57 +10,41 @@ const spec = path.resolve(__dirname, '..', '..', 'index.bs')
 const specContent = fs.readFileSync(spec, 'utf-8')
 const specParsed = parse5.parse(specContent)
 
-/**
- * format the result and sort pre content to their corresponding classes
- */
-const cddlContent = getCDDLNodes([specParsed]).reduce((val, node) => {
-    if (!val[node.nodeClass]) {
-        val[node.nodeClass] = []
-    }
-    val[node.nodeClass].push(node.content)
-    return val
-}, {})
+// Extract local and remote CDDL definitions from pre nodes of the parsed DOM.
+const cddl = getCDDLNodes([specParsed]);
 
 /**
  * iterate over all pre cddl classes found in the document, sanitize the
  * content and write them to file.
  */
-Object.entries(cddlContent).forEach(([cddlClass, content]) => {
-    const fileContent = content
-        /**
-         * remove obsolete whitespace
-         */
-        .map((entry) => {
-            const preceedingSpace = entry.split('\n').reduce((prev, line) => {
-                if (line.length === 0) {
-                    return prev
-                }
+Object.entries(cddl).forEach(([cddlName, entries]) => {
+  const fileContent = entries
+    /**
+     * remove obsolete whitespace
+     */
+    .map((entry) => {
+      const preceedingSpace = entry.split('\n').reduce((prev, line) => {
+        if (line.length === 0) {
+          return prev
+        }
 
-                const spaces = line.match(/^\s+/)
-                if (spaces) {
-                    return Math.min(prev, spaces[0].length)
-                }
+        const spaces = line.match(/^\s+/)
+        if (spaces) {
+          return Math.min(prev, spaces[0].length)
+        }
 
-                return 0
-            }, Infinity)
-            return entry
-                .split('\n')
-                .map((line) => line.slice(preceedingSpace))
-        })
-        /**
-         * drop all cddl snippets without declarations, e.g.
-         * ```
-         * EmptyResult
-         * ```
-         */
-        .filter((pre) => {
-            return pre.filter(Boolean).length > 1
-        })
-        /**
-         * join code lines
-         */
-        .map((pre) => pre.join('\n'))
-        .join('\n')
+        return 0
+      }, Infinity);
 
-    fs.writeFileSync(path.join(process.cwd(), `${cddlClass.split('-').shift()}.cddl`), fileContent)
-})
+      return entry
+        .split('\n')
+        .map((line) => line.slice(preceedingSpace))
+    })
+    /**
+     * join code lines
+     */
+    .map((pre) => pre.join('\n'))
+    .join('\n');
+
+  fs.writeFileSync(path.join(process.cwd(), `${cddlName}.cddl`), fileContent);
+});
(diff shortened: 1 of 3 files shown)


whimboo:
> Also I removed filtering out of lines that miss a declaration. This actually drops lines unexpectedly. Instead we should get the validation step fail and report it appropriately.

Please drop that. We cannot remove it given that in some cases we simply want to reference an existing type only.

scripts/cddl/generate.js
```diff
@@ -13,54 +13,55 @@ const specParsed = parse5.parse(specContent)
...
+        .map((line) => line.slice(preceedingSpace))
+    })
+    /**
+     * drop all cddl snippets without declarations, e.g.
+      * ```
+      * EmptyResult
+      * ```
+      */
+    .filter((pre) => {
+      return pre.filter((line) => line.includes("=")).length > 0
```
whimboo, line 58: This change fixes the issue that single line declarations are dropped. Basically each definition should include an equal sign, so it should be easy this way to filter out types only listed to indicate what gets returned.

christian-bromann (approved):
LGTM 👍

scripts/cddl/generate.js
```diff
@@ -13,54 +13,55 @@ const specParsed = parse5.parse(specContent)
...
  */
-const cddlContent = getCDDLNodes([specParsed]).reduce((val, node) => {
-    if (!val[node.nodeClass]) {
-        val[node.nodeClass] = []
-    }
-    val[node.nodeClass].push(node.content)
-    return val
-}, {})
+const cddlNodes = getCDDLNodes([specParsed]).reduce((val, node) => {
+  if (!val[node.nodeClass]) {
```
jgraham, line 17: As a note, adding whitespace or other reformatting changes to this kind of PR is unnecessary and makes it harder to review.

scripts/cddl/generate.js
jgraham: I think this makes it better in the short term, but is the wrong fix. We should give the sections a different class name in the spec so that they don't get selected as part of the schema (probably just by removing the `cddl-local` and `cddl-remote` classes).

scripts/cddl/utils.js
```diff
@@ -2,35 +2,37 @@
...
-    return nodes.map((node) => {
-        if (node.nodeName === 'pre') {
-            const nodeClass = node.attrs.find((attr) => attr.name === 'class')
-            const content = node.childNodes.map((child) => child.value).join('')
+function getCDDLNodes(nodes) {
+  const data = nodes.map((node) => {
+    if (node.nodeName === 'pre') {
+      const nodeClass = node.attrs.find((attr) => attr.name === 'class')
+      if (nodeClass && nodeClass.value.includes('cddl')) {
+        const content = node.childNodes.map((child) => child.value).join('')
```
jgraham, line 18: Is this different to `node.textContent`?

scripts/cddl/utils.js
```diff
@@ -2,35 +2,37 @@
...
-function getCDDLNodes (nodes) {
-    return nodes.map((node) => {
-        if (node.nodeName === 'pre') {
-            const nodeClass = node.attrs.find((attr) => attr.name === 'class')
-            const content = node.childNodes.map((child) => child.value).join('')
+function getCDDLNodes(nodes) {
+  const data = nodes.map((node) => {
+    if (node.nodeName === 'pre') {
+      const nodeClass = node.attrs.find((attr) => attr.name === 'class')
+      if (nodeClass && nodeClass.value.includes('cddl')) {
```
jgraham, line 17: All the string manipulation feels like it's being far too clever. Since there are only two cases `cddl-remote` and `cddl-local` I'd just check for those explicitly, and return the nodes for each as a object like `{local: [], remote: []}`, rather than a flat list that later has to be processed again.

jgraham (changes requested):

scripts/cddl/generate.js
whimboo: Sorry, you are right and I should have done that in a separate commit. Let me know if you want to see it reverted / splitted.

scripts/cddl/utils.js
whimboo: Indeed. Changing that would make the code all the way easier to understand and faster. I'll update that.

whimboo:
@jgraham PR should be fine for the next review. Thanks.

scripts/cddl/utils.js
```diff
@@ -2,35 +2,41 @@
...
-        if (!node.childNodes) {
-            return
+        if (nodeClass.value.includes("remote-cddl")) {
+          entries.remote.push(content);
         }
+      }
+    }
+
+    if (node.childNodes) {
+      const { local, remote } = getCDDLNodes(node.childNodes);
```
jgraham, line 33: If we cared about performance we shouldn't do all these object allocations, but I strongly suspect we don't and it's fine ;)

jgraham (approved):

scripts/cddl/utils.js
whimboo: I concur here. The major bottle neck right now is really installing, which included building) the `cddl` tool for each and every CI job! I will file an issue for that.

PR #282, opened 2022-09-13
Enable Rust cache for cddl test job

### LINKED ISSUE #281: Use cache to stop building the `cddl` binary for each and every verification job
Currently for the `cddl` test job we are building the `cddl` binary each time. This takes quite a while and we would be better of in caching the rust related files. It should actually be pretty simple as the following example shows:

https://github.com/webcompat/knowledge-base/blob/main/.github/workflows/check.yml#L14-L23

whimboo:
Fix #281.


whimboo:
The Rust cache works now so that we no longer have to build the `cddl` binary for each and every Ci job.

whimboo:
@jgraham or @foolip could you please have a look? Thanks.

.github/workflows/test.yml
```diff
@@ -23,11 +23,26 @@ jobs:
       - uses: actions/setup-node@v1
         with:
           node-version: 14.x
+      - name: Get cddl version
+        run: curl -s https://crates.io/api/v1/crates/cddl | python3 -c "import sys, json; print(json.load(sys.stdin)['crate']['max_stable_version'])" > cddl.txt && echo cddl.txt
```
jgraham, line 27: You can also pipe the output into `tee cddl.txt` to both write to stdut and to a file.

scripts/build.sh
```diff
@@ -4,7 +4,8 @@ set -ex
 SCRIPT_DIR=$(cd $(dirname "$0") && pwd -P)
 ROOT=$(dirname $SCRIPT_DIR)
 
-if [ "$1" = "--install" ]; then
+if ! [ -x "$(command -v bikeshed)" ] || [ "$1" = "--upgrade" ]; then
```
jgraham, line 7: I don't love installing software by default, but I guess the next easiest option involves writing a proper python script or something, so this is probably fine.

jgraham (approved):

.github/workflows/test.yml
whimboo: Ah good to know. Will make the change which is much nicer.

scripts/build.sh
jgraham: I'm happy to just land this for now, but generally it seems better not to write outside the project without at least getting am opt-in.

PR #284, opened 2022-09-14
Empty lines should not affect space removal

OrKoN:
For example, previously some commands were formatted with extra whitespace in the CDL file:

```
  ScriptCallFunctionCommand = {
    method: "script.callFunction",
    params: ScriptCallFunctionParameters
  }
```

With this change, it becomes:

```
ScriptCallFunctionCommand = {
  method: "script.callFunction",
  params: ScriptCallFunctionParameters
}
```

I am not sure where those lines containing only spaces come from though.

### DIFF
--- a/scripts/cddl/generate.js
+++ b/scripts/cddl/generate.js
@@ -24,7 +24,7 @@ Object.entries(cddl).forEach(([cddlName, entries]) => {
      */
     .map((entry) => {
       const preceedingSpace = entry.split('\n').reduce((prev, line) => {
-        if (line.length === 0) {
+        if (line.trim().length === 0) {
           return prev
         }
 


whimboo:
It's because of the indendation of the `pre` node as you can see here:
https://github.com/w3c/webdriver-bidi/blob/6f755a2d2c9109871656acfb03864e0405eb63ab/index.bs#L4190-L4194

I've spotted that as well when I refactored the CDDL generation and wanted to do that in a follow-up PR. But thanks that you already took care of it!

I've approved the test jobs and when inspecting the CDDL artifacts it looks all fine and is way better to read!

I'll leave it for @jgraham for the final review.

whimboo:
I wonder if we should also still fix the indentation in the spec file?

jgraham (approved):
It's probably a reasonable idea to fix the spec, but I don't think we need to block on that.

whimboo:
> It's probably a reasonable idea to fix the spec, but I don't think we need to block on that.

I can take care of in my next PR for CDDL. Just triggered the tests and if all passes I'll go ahead and merge.

PR #286, opened 2022-09-16
Add browser.close command

jgraham:
This specifies shutting down all active sessions and closing the browser, but leaves it open what exactly closing the browser implies.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/286.html" title="Last updated on Apr 13, 2023, 10:45 AM UTC (b3a20ff)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/286/6e40aba...b3a20ff.html" title="Last updated on Apr 13, 2023, 10:45 AM UTC (b3a20ff)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -24,6 +24,7 @@ spec: RFC6455; urlPrefix: https://tools.ietf.org/html/rfc6455
     text: %x1 denotes a text frame; url: section-5.2
     text: Send a WebSocket Message; url: section-6.1
     text: A WebSocket Message Has Been Received; url: section-6.2
+    text: Start The WebSocket Closing Handshake; url: section-7.1.2
     text: The WebSocket Closing Handshake is Started; url: section-7.1.3
     text: The WebSocket Connection is Closed; url: section-7.1.4
     text: Fail the WebSocket Connection; url: section-7.1.7
@@ -40,6 +41,7 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
     text: additional WebDriver capability; url: dfn-additional-webdriver-capability
     text: additional capability deserialization algorithm; url: dfn-additional-capability-deserialization-algorithm
     text: capability name; url: dfn-capability-name
+    text: close the session; url: dfn-close-the-session
     text: create a session; url: dfn-create-a-session
     text: dispatch actions; url: dfn-dispatch-actions
     text: dispatch tick actions; url: dfn-dispatch-tick-actions
@@ -287,6 +289,7 @@ Command = {
 }
 
 CommandData = (
+  BrowserCommand //
   BrowsingContextCommand //
   InputCommand //
   NetworkCommand //
@@ -470,6 +473,12 @@ with the following additional codes:
   <dd>Tried to deserialize an unknown <code>SharedReference</code>.
 </dl>
 
+<dl>
+  <dt><dfn>unable to close browser</dfn>
+  <dd>Tried to close the browser, but failed to do so.
+</dl>
+
+
 <pre class="cddl local-cddl">
 ErrorCode = ("invalid argument" /
              "invalid session id" /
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2441,6 +2443,75 @@ The [=remote end steps=] with |session| and |command parameters| are:
...
+<div algorithm="remote end steps for browser.close">
+The [=remote end steps=] with |session| and <var ignore>command parameters</var> are:
+
+1. For each |active session| in [=active sessions=]:
+
+  1. [=Close the session=] |active session|.
+
+  Note: Since this terminates |session|, the [=local end=] will not receive any
+  events that would be generated during subsequent steps
+  (e.g. [=browsingContext.close=]).
```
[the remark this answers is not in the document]
whimboo: Note that we have such a response for Firefox/Marionette when `Marionette:Quit` is requested. With a response we could at least return if something blocks a shutdown. So having an `EmptyResult` here which would be extensible sounds like a good idea to me.

jgraham:
Another option here would be to say that a clean shutdown of the browser should always emit an event on any open BiDi session (that's not mutually exclusive to also providing a command response of course).

sadym-chromium (approved):

index.bs
```diff
@@ -2772,6 +2865,102 @@ The [=remote end steps=] with |session| and |command parameters| are:
...
+1. For each |active session| in [=active sessions=]:
+
+  1. [=End the session=] |active session|.
+
+  1. [=Cleanup the session=] with |active session|
+
+1. Return [=success=] with data null, and run the following steps [=in parallel=].
+
+  1. Wait until the [=Send a WebSocket message=] steps have been called with the
+     response to this command.
```
whimboo, line 2944: What kind of response do you actually expect to see here for the other open sessions that did not trigger the shutdown? Note that the `id` of the response will not match with any command as sent such a client. Maybe we should send an event like `browser.closing` instead?

index.bs
```diff
@@ -2772,6 +2865,102 @@ The [=remote end steps=] with |session| and |command parameters| are:
...
+
+     Note: This implicitly only affects browsing contexts that were under
+     automation in the first place.
+
+  1. Perform implementation defined steps to clean up resources associated with
+     the [=remote end=] under automation.
+
+     Note: For example this might include cleanly shutting down any OS-level
+     processes associated with the browser under automation, removing temporary
+     state, such as user profile data, created by the [=remote end=] while under
```
whimboo, line 2958: Not sure if it would be possible at all to expect a remote end to be able to clean-up created user profile data. I could only see this working when we would support the BrowserContext (identities) in the BiDi spec, but even here not all the data it tight to this specific browser context.

index.bs
jgraham: None. I  just expect their WebSocket connections to close. Since clients have to cope with the websocket connection closing at an arbitrary time, a clean close at the websocket protocol level seems like enough of a signal that something closed the browser. It's of course also possible that something just sends it SIGKILL and we get an unclean shutdown.

Given that, I'm not sure an extra event adds much, but we could always add one later if needed. The only reason to respond to the command at all really is to ensure that we can provide an error in the case that shutdown is blocked.

index.bs
jgraham: Firefox probably won't do this if you launch it directly, but Geckodriver probably will do this if you create the session via that. So I think it's reasonable to add as a non-normative example of something that might happen.

whimboo: Sure. Sounds fine to me.

index.bs
whimboo: Yes, adding the event at a later time sounds good. I was only worried about other clients being able to distinct between a clean and expected shutdown, or a browser crash or WebSocket disconnect.

whimboo (approved):
Thanks James and sorry for the delay in reviewing.

Issue #65, opened 2020-10-27
Add a bootstrap scripts feature


jgraham:
Initial proposal at https://github.com/w3c/webdriver-bidi/blob/master/proposals/bootstrap-scripts.md

sadym-chromium:
I believe this issue can be closed as implemented: https://w3c.github.io/webdriver-bidi/#command-script-addPreloadScript

PR #292, opened 2022-09-20
Add support for preload scripts

jgraham:
These are scripts which are run when a realm (initially only for realms with a Window global) is created. They allow a test harness to inject functionality that's guaranteed to be available for any content scripts that are subsequently loaded, and before any later scripts that WebDriver injects into the context.

Registering a preload script happens with the script.addPreloadScript command. This returns a handle to the added script.

Load scripts can later be removed by passing the handle to script.removePreloadScript.

Load scripts can either be run directly in the newly created Window, or in a named sandbox.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/292.html" title="Last updated on Dec 21, 2022, 9:26 PM UTC (8681af4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/292/14731db...8681af4.html" title="Last updated on Dec 21, 2022, 9:26 PM UTC (8681af4)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3792,12 +3792,12 @@ relating to script realms and execution.
 
 <pre class="cddl remote-cddl">
 ScriptCommand = (
-  script.AddLoadScriptCommand //
+  script.AddPreloadScriptCommand //
   script.CallFunction //
   script.Disown //
   script.Evaluate //
   script.GetRealms
-  script.RemoveLoadScriptCommand
+  script.RemovePreloadScriptCommand
 )
 </pre>
 
@@ -3805,7 +3805,7 @@ ScriptCommand = (
 
 <pre class="cddl local-cddl">
 ScriptResult = (
-  script.AddLoadScriptResult //
+  script.AddPreloadScriptResult //
   script.EvaluateResult //
   script.GetRealmsResult
 )
@@ -3816,20 +3816,20 @@ ScriptEvent = (
 )
 </pre>
 
-### Load Scripts ### {#load-scripts}
+### Preload Scripts ### {#preload-scripts}
 
-A <dfn>load script</dfn> is one which runs on creation of a new {{Window}},
+A <dfn>Preload script</dfn> is one which runs on creation of a new {{Window}},
 before any author-defined script have run.
 
 TODO: Extend this to scripts in other kinds of realms.
 
-A [=BiDi session=] has a <dfn>load script map</dfn> which is a map in which the
+A [=BiDi session=] has a <dfn>preload script map</dfn> which is a map in which the
 keys are [[!RFC4122|UUID]]s, and the values are [=structs=] with an <a
 for=struct>item</a> named <code>expression</code>, which is a string, and an
 item named <code>sandbox</code> which is a string or null.
 
 <div algorithm>
-To <dfn export>run WebDriver BiDi load scripts</dfn> given |environment settings|:
+To <dfn export>run WebDriver BiDi preload scripts</dfn> given |environment settings|:
 
 1. Let |document| be |environment settings|' [=relevant global object=]'s
    <a>associated <code>Document</code></a>.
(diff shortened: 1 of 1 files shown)


jgraham:
This is a very initial cut at https://github.com/w3c/webdriver-bidi/issues/65

I changed the name from "bootstrap scripts" to "load scripts" because I thought "bootstrap" might be a bit obscure, but maybe "load" is suggestive of the wrong point in the lifecycle.

Compared to the original propsal, this only supports window-type globals and doesn't have the URL pattern matching. URL pattern matching can be implemented in the script itself (with some care; navigating from the initial about:bank reuses the same Window). This featureset currently matches what's in CDP which suggests it's enough to enable existing client features.

jgraham:
https://github.com/whatwg/html/pull/8300 for the HTML side of this change.

index.bs
```diff
@@ -4542,6 +4655,53 @@ Issue: We might want to have a more sophisticated filter system than just a
...
+
+1. Let |script| be the value of the "<code>script</code>" field in |command
+   parameters|.
+
+1. Let |load script map| be |session|'s [=load script map=].
+
+1. If |load script map| does not <a for=map>contain</a> |script|, return
+   [=error=] with [=error code=] [=invalid argument=].
+
+   TODO: Should this be a "no such script" error?
```
whimboo, line 4696: We would raise invalid argument for not supported types, so maybe having a dedicated error like for invalid browsing context would be good to have.

How do we want to return the failure given that the CDDL expects an `EmptyResult` (this actually also applies to other commands)?

whimboo (commented):
Thanks James! As I can see the PR only contains the details for the two methods but not the trigger for `run WebDriver BiDi load scripts`. Do I miss something or did you forget to add that to the browsing context / realm created sections?

Also I agree that we should have a discussion for a proper name.

index.bs
jgraham: Protocol-level errors like this have a totally different return type compared to commands that are run successfully; that's already handled in the transport section. Script errors are different because that error is itself a response to the command, not something that prevented the command running.

jgraham (commented):
>  As I can see the PR only contains the details for the two methods but not the trigger for run WebDriver BiDi load scripts

That happens in the linked HTML change. Since there's no event here there's no ordering question between `contextCreated` and running the scripts (although maybe things will be more complex once we have custom events).

jimevans:
@jgraham Nitpicky comment: If we think "load script" might be confusing because of the cognitive issue surrounding when "load" occurs, and "bootstrap" being obscure, is there any mileage in the term "preload scripts?"

jgraham:
I like "preload scripts". Unless people prefer "bootstrap scripts" I'll change to that for the next revision.

index.bs
```diff
@@ -3664,6 +3667,54 @@ ScriptEvent = (
 )
 </pre>
 
+### Load Scripts ### {#load-scripts}
+
+A <dfn>load script</dfn> is one which runs on creation of a new {{Window}},
+before any author-defined script have run.
+
+TODO: Extend this to scripts in other kinds of realms.
```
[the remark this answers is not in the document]
whimboo: Loaded scripts are actually global and are shared with newly created realms for window globals. As such it is indeed shared. Yes, we can close this issue.

index.bs
```diff
@@ -3813,6 +3828,55 @@ ScriptEvent = (
...
+
+A [=BiDi session=] has a <dfn>preload script map</dfn> which is a map in which the
+keys are [[!RFC4122|UUID]]s, and the values are [=structs=] with an <a
+for=struct>item</a> named <code>expression</code>, which is a string, and an
+item named <code>sandbox</code> which is a string or null.
+
+<div algorithm>
+To <dfn export>run WebDriver BiDi preload scripts</dfn> given |environment settings|:
```
sadym-chromium, line 3844: Do we have to specify when this algo should be triggered?

sadym-chromium (commented):
I like the name "PreloadScript".

index.bs
jgraham: That's specified in HTML: https://github.com/whatwg/html/pull/8300

sadym-chromium (approved):

jgraham:
I'm going to merge this on the assumption that the HTML PR will merge without requiring substantive changes on the WebDriver side.

PR #305, opened 2022-10-04
Use JavaScript compatible range for uint type

### LINKED ISSUE #177: Missing definition for upper boundary of uint type
As discussed with @jgraham today we miss to define the upper bound for parameters of type `uint`. In CDDL there is no limit compared to JavaScript which has the [`SafeInteger`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isSafeInteger) range.

whimboo:
Fix #177.

***
<a href="https://pr-preview.s3.amazonaws.com/whimboo/webdriver-bidi/pull/305.html" title="Last updated on Oct 5, 2022, 2:38 PM UTC (f3f079f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/305/01c8d75...whimboo:f3f079f.html" title="Last updated on Oct 5, 2022, 2:38 PM UTC (f3f079f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -246,7 +246,7 @@ the remainder of the specification.
 
 <pre class="cddl remote-cddl">
 Command = {
-  id: uint,
+  id: js-uint,
   CommandData,
   *text => any,
 }
@@ -270,13 +270,13 @@ Message = (
 )
 
 CommandResponse = {
-  id: uint,
+  id: js-uint,
   result: ResultData,
   *text => any
 }
 
 ErrorResponse = {
-  id: uint / null,
+  id: js-uint / null,
   error: ErrorCode,
   message: text,
   ?stacktrace: text,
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2197,7 +2203,7 @@ CapabilitiesRequest = {
     ?noProxy: [*text],
     ?sslProxy: text,
     ?socksProxy: text,
-    ?socksVersion: int,
+    ?socksVersion: 0..255,
```
whimboo, line 2206: Per WebDriver classic spec this is the correct range.

index.bs
```diff
@@ -3937,9 +3943,9 @@ The <code>RealmType</code> type represents the different types of Realm.
 
 <pre class="cddl remote-cddl local-cddl">
 StackFrame = {
-  columnNumber: int,
+  columnNumber: JSON-uint,
```
whimboo, line 3946: Any idea why we used `int` here before?

index.bs
```diff
@@ -4814,7 +4820,7 @@ BaseLogEntry = {
   level: LogLevel,
   source: Source,
   text: text / null,
-  timestamp: int,
+  timestamp: JSON-uint,
```
whimboo, line 4823: Same question here with `int` which doesn't seem to be appropriate.

jgraham (approved):

PR #320, opened 2022-11-04
Pin CI to Python 3.10

jgraham:
This avoid an issue where Bikeshed doesn't work with 3.11

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v3
         with:
-          python-version: 3.x
+          python-version: "3.10"
       - name: Build spec without warnings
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v3
         with:
-          python-version: 3.x
+          python-version: "3.10"
```
whimboo, line 16: Are the quotes needed? If not please remove them. Otherwise I may want to add a comment to why we pin to 3.10 and add a bug reference for bikeshed?

jgraham: Yes the quotes are needed, otherwise YAML interprets it as the floating point number `3.1`…

I added a comment.

whimboo (approved):

PR #333, opened 2022-11-23
Add serialization support for NodeList and HTMLCollection

### LINKED ISSUE #318: Handle NodeList and similar Array-like platform objects in serialization
Currently if you try to serialize a `NodeList` you're just going to end up with a handle to the remote object, rather than a helpful `value` property containing an array of the nodes. We should fix this case so that authors don't need to remember to write `Array.from(document.querySelectorAll(selector))` and clients don't need a roundtrip to work with common array-like platform objects.

jgraham:
Fixes #318.

These should serialize like Arrays, in particular for the `document.querySelectorAll` use case where you want to directly get the list of elements back.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/333.html" title="Last updated on Dec 20, 2022, 8:52 AM UTC (ab70ad6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/333/14731db...ab70ad6.html" title="Last updated on Dec 20, 2022, 8:52 AM UTC (ab70ad6)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1888,7 +1888,7 @@ an |ownership type|, a |serialization internal map| and a |realm|:
 
     <dt>|value| is a [=platform object=] that implements {{HTMLCollection}}
     <dd>Let |remote value| be [=serialize an Array-like=] with
-        <code>HTMLCollectionRemoteValue</code>, |value|, |handle id|, |value|, |max
+        <code>HTMLCollectionRemoteValue</code>, |handle id|, |value|, |max
         depth|, |ownership type|, |known object|, |serialization internal map|, and
         |realm|.
 
--- a/index.bs
+++ b/index.bs
@@ -1748,10 +1748,10 @@ an |ownership type|, a |serialization internal map| and a |realm|:
 1. Let |handle id| be the [=handle for an object=] given |realm|, |ownership type|
    and |value|.
 
+1. Set |ownership type| to "<code>none</code>".
+
 1. Let |known object| be <code>true</code>, if |value| is in the
    |serialization internal map|, otherwise <code>false</code>.
-
-1. Let |child ownership| be "<code>none</code>".
 
   <dl>
     <dt>[=Type=](|value|) is Symbol
@@ -1825,7 +1825,7 @@ an |ownership type|, a |serialization internal map| and a |realm|:
 
        1. Let |serialized| be the result of [=serialize as a mapping=] given
           [=CreateMapIterator=](|value|, key+value), |max depth|,
-          |child ownership|, |serialization internal map| and |realm|.
+          |ownership type|, |serialization internal map| and |realm|.
 
     1. If |serialized| is not null, set field <code>value</code> of |remote value| to
        |serialized|.
@@ -1846,7 +1846,7 @@ an |ownership type|, a |serialization internal map| and a |realm|:
        the following steps:
 
        1. Let |serialized| be the result of [=serialize as a list=] given
-          [=CreateSetIterator=](|value|, value), |max depth|, |child ownership|,
+          [=CreateSetIterator=](|value|, value), |max depth|, |ownership type|,
           |serialization internal map| and |realm|.
 
     1. If |serialized| is not null, set field <code>value</code> of |remote value| to
@@ -1921,7 +1921,7 @@ an |ownership type|, a |serialization internal map| and a |realm|:
             1. Let |child depth| be |max depth| - 1 if |max depth| is not null, or null otherwise.
 
             1. Let |serialized| be the result of [=serialize as a remote value=]
-               with |child|, |child depth|, |child ownership|,
+               with |child|, |child depth|, |ownership type|,
                |serialization internal map| and |realm|.
 
             1. Append |serialized| to |children|.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2011,6 +2018,31 @@ remove optional flag from the field.
 
 </div>
 
+<div algorithm>
+To <dfn>serialize an Array-like</a> given |production|, |value|, |handle id|,
```
jugglinmike, line 2022:
```suggestion
To <dfn>serialize an Array-like</dfn> given |production|, |value|, |handle id|,
```

whimboo (changes requested):

index.bs
whimboo: That issue was fixed by the last update.

index.bs
```diff
@@ -2011,6 +2018,31 @@ remove optional flag from the field.
...
+|known object|, and |max depth|:
```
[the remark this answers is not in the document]
whimboo: @jgraham are you going to update your PR based on my feedback as well?

index.bs
```diff
@@ -1687,6 +1689,18 @@ ArrayBufferRemoteValue = {
   ?internalId: InternalId,
 }
 
+NodeListRemoteValue = {
```
jimevans, line 1692: Is it intentional that `NodeListRemoteValue` and HTMLCollectionRemoteValue` omit the `type` property? How should a local end detect these constructs when encountered?

jgraham: Haha, no that's just incompetence. I'll fix.

index.bs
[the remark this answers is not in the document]
whimboo: @jgraham this issue is still unresolved.

jimevans (approved):

index.bs
```diff
@@ -2021,7 +2031,31 @@ remove optional flag from the field.
 </div>
 
 <div algorithm>
-To <dfn>serialize as a list</dfn> given |iterable|, |max depth|, |child ownership|,
+To <dfn>serialize an Array-like</dfn> given |production|, |value|, |handle id|,
+|max depth|, |ownership type|, |known object|, |serialization internal map|, and |realm|:
```
whimboo, line 2035: I would suggest the following change so that we keep the order of the general arguments similar to the other helper definitions and custom ones only affecting this helper at the front.

```suggestion
To <dfn>serialize an Array-like</dfn> given |production|, |handle id|, |value|,
|max depth|, |ownership type|, |known object|, |serialization internal map|, and |realm|:
```

index.bs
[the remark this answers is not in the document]
whimboo: Looks like this has been addressed now. Will do a final review now. Thanks @jgraham.

index.bs
```diff
@@ -1751,7 +1767,7 @@ an |ownership type|, a |serialization internal map| and a |realm|:
 1. Let |known object| be <code>true</code>, if |value| is in the
    |serialization internal map|, otherwise <code>false</code>.
 
-1. Let |child ownership| be "<code>none</code>".
+1. Let |ownership type| be "<code>none</code>".
```
whimboo, line 1770: I don't think this change is wanted and it wasn't here last time. Note that it will override the `ownership type` argument of the `serialize as a remote value` method.

whimboo (changes requested):

index.bs
jgraham: I found it confusing that we were calling the argument `|child ownership|` in the algorithms, when it's not related to any children. So I'd like to keep that part at least. Replacing the passed in value is technically correct: we use the original `ownership type` variable once in `handle for an object` and after that it's an error to ever use the original value again for future calls (in Rust parlance, we move that value into the function). But maybe this is also confusing.

whimboo: Could you maybe be fine with splitting off this particular change to a different PR? It's not really related to adding support for `NodeList` and `HTMLCollection`.

index.bs
```diff
@@ -2021,7 +2031,31 @@ remove optional flag from the field.
 </div>
 
 <div algorithm>
-To <dfn>serialize as a list</dfn> given |iterable|, |max depth|, |child ownership|,
+To <dfn>serialize an Array-like</dfn> given |production|, |handle id|, |value|,
+|max depth|, |ownership type|, |known object|, |serialization internal map|, and |realm|:
```
whimboo, line 2035: Nit: Given that only that helper requires a reference to the known object we may want to move it as well close to the special arguments.
```suggestion
To <dfn>serialize an Array-like</dfn> given |production|, |handle id|, |known object|, |value|, |max depth|, |ownership type|, |serialization internal map|, and |realm|:
```

whimboo (approved):
There are two remaining issues whereby only one is blocking the landing. Please update.

Otherwise I would like to see a review from @sadym-chromium or @foolip as well.

whimboo:
Pinging @sadym-chromium and @foolip. Could you please review this PR? Thanks.

index.bs
```diff
@@ -2021,7 +2031,31 @@ remove optional flag from the field.
 </div>
 
 <div algorithm>
-To <dfn>serialize as a list</dfn> given |iterable|, |max depth|, |child ownership|,
+To <dfn>serialize an Array-like</dfn> given |production|, |handle id|, |value|,
```
sadym-chromium, line 2034: If `|iterable|` is passed instead of the `|value|`, serialization of `set` can be implemented via `serialize an Array-like` as well. Feel free to ignore.

sadym-chromium (approved):

jgraham:
Since the remaining suggestions are just editorial, I'm going to merge; we can always make additional changes later on.

index.bs
whimboo: I don't think that this is possible given that we still run `set internal ids` based on the `remote value` within `serialize as Array-like`.

index.bs
whimboo: I'll push a follow-up PR for that proposal.

PR #334, opened 2022-11-24
Fix build for HTML navigables change

jgraham:
This is a minimal set of fixes required to make the build pass. It doesn't yet do the work of replacing the extensive use of "browsing context" in the spec with navigables (or, where appropriate, their active window or document).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/334.html" title="Last updated on Nov 24, 2022, 12:41 PM UTC (3fa8c25)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/334/1558000...3fa8c25.html" title="Last updated on Nov 24, 2022, 12:41 PM UTC (3fa8c25)">Diff</a>

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -9,4 +9,5 @@ if ! [ -x "$(command -v bikeshed)" ] || [ "$1" = "--upgrade" ]; then
     pip install bikeshed
 fi
 
+bikeshed update
 bikeshed --die-on=warning spec index.bs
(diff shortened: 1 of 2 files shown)


scripts/build.sh
```diff
@@ -9,4 +9,5 @@ if ! [ -x "$(command -v bikeshed)" ] || [ "$1" = "--upgrade" ]; then
     pip install bikeshed
 fi
 
+bikeshed update
```
foolip, line 12: Is this for testing or still required to pass CI? This is a slow step so it's not great to run locally every time.

foolip (approved):

scripts/build.sh
jgraham: Required to pass CI; the version of bikeshed we're installing doesn't define `navigable` at the moment.

I suggest that for local use the CI version of this script isn't very useful since you don't want to always update or die on warnings, so we should stop encouraging that and instead suggest e.g. `bikeshed serve`.

PR #335, opened 2022-11-25
Fix "nodeValue" and "childNodeCount" serialization for DOM nodes

whimboo:
While implementing the DOM node serialization in Firefox I noticed the following bug and a situation for improvement:

* `childNodeCount` is set based on the length of the still to be serialized `childNodes`.
* `nodeValue` is set for all kind of nodes even that it [is `null` for most of the known node types](https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeValue).

***
<a href="https://pr-preview.s3.amazonaws.com/whimboo/webdriver-bidi/pull/335.html" title="Last updated on Nov 25, 2022, 12:37 PM UTC (a7c7735)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/335/a676ae9...whimboo:a7c7735.html" title="Last updated on Nov 25, 2022, 12:37 PM UTC (a7c7735)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1700,12 +1700,12 @@ NodeRemoteValue = {
 
 NodeProperties = {
   nodeType: js-uint,
-  nodeValue: text,
   childNodeCount: js-uint,
+  ?attributes: {*text => text},
+  ?children: [*NodeRemoteValue],
   ?localName: text,
   ?namespaceURI: text,
-  ?children: [*NodeRemoteValue],
-  ?attributes: {*text => text},
+  ?nodeValue: text,
   ?shadowRoot: NodeRemoteValue / null,
 }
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1897,15 +1897,18 @@ an |ownership type|, a |serialization internal map| and a |realm|:
 
           1. Set |serialized|["<code>nodeType</code>"] to [=Get=](|value|, "nodeType").
 
-          1. Set |serialized|["<code>nodeValue</code>"] to [=Get=](|value|, "nodeValue")
+          1. Set |node value| to [=Get=](|value|, "nodeValue").
+
+          1. If |node value| is not undefined set
```
jgraham, line 1902: It can't be `undefined`, but can be `null`: https://dom.spec.whatwg.org/#dom-node-nodevalue

jgraham (changes requested):

whimboo:
I'll add required web-platform tests via https://bugzilla.mozilla.org/show_bug.cgi?id=1770731.

index.bs
whimboo: Ups, not sure why I used `undefined` here. Thanks for catching that.

jgraham (commented):
lgtm, but since the change to not include `nodeValue` unless it's set is non-editorial, we should wait for someone else to review.

sadym-chromium (approved):
LGTM, but as long as in our implementation the serialization is implemented in CDP, it would take some time and effort to implement the change. So we'd prefer to collect all the serialization changes and fix them in 1 go. It should not influence the specification though.

sadym-chromium:
@whimboo will be there a WPT PR afterwards?

whimboo:
> @whimboo will be there a WPT PR afterwards?

Please see my comment from above. Right now we do not have tests for those scenarios and I'll add cases to the above mentioned bug. I'll create a WPT PR for that yes.

jgraham (approved):

PR #340, opened 2022-12-01
Add type field to "ScriptEvaluateResult" for "success" or "exception"

### LINKED ISSUE #339: Add type field to "ScriptEvaluateResult" for "success" or "exception"
Found when discussing issue #274.

Right now the `ScriptEvaluateResult` could contain a `result` or an `exception`. The client can only differentiate between those depending on which field is set. There might be cases when no field or both fields are set.

To make it obvious for the client such a `type` field should be added to the `ScriptEvaluateResult` field.

whimboo:
Fixes #339.

This change needs an update of the wpt tests that I can work on if the change is fine.

***
<a href="https://pr-preview.s3.amazonaws.com/whimboo/webdriver-bidi/pull/340.html" title="Last updated on Dec 1, 2022, 3:59 PM UTC (827fa68)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/340/a056c3a...whimboo:827fa68.html" title="Last updated on Dec 1, 2022, 3:59 PM UTC (827fa68)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4162,12 +4162,14 @@ ScriptEvaluateResult = (
 )
 
 ScriptEvaluateResultSuccess = {
+  type: "success",
   result: RemoteValue,
   realm: Realm
 }
 
 ScriptEvaluateResultException = {
-  exceptionDetails: ExceptionDetails
+  type: "exception",
+  exceptionDetails: ExceptionDetails,
   realm: Realm
 }
 </pre>


index.bs
```diff
@@ -4162,12 +4162,14 @@ ScriptEvaluateResult = (
 )
 
 ScriptEvaluateResultSuccess = {
+  type: "result",
```
sadym-chromium, line 4165: Maybe `success`?

sadym-chromium (approved):
LGTM with  `type: "success"`

index.bs
whimboo: I fixed that.

whimboo:
@sadym-chromium given that [changing the behavior of the webdriver bidi client for wdspec tests](https://github.com/web-platform-tests/wpt/blob/d994afbbb6f37fc5d0ca722b621e8637cf6749a8/tools/webdriver/webdriver/bidi/modules/script.py#L101-L103) to correctly check for that type field would break all of the evaluation tests for Chrome do you think that you could add that field relatively soon, and we are going to land the test changes when that change hit the beta channel? Or what would be your suggestion?

sadym-chromium:
That's a good idea to land the implementation before the WPT. Let me do that.

whimboo:
@sadym-chromium I've landed the support for the `type` field in Firefox 109. Please let us know which Chrome release will have support for that as well. Thanks!

PR #345, opened 2022-12-13
Editorial: correct handling of session connections

jugglinmike:
This proposal defines the term "WebSocket connection" to describe a type. It defines the term "session WebSocket connections" as a property of the "session" primitive. Prior to this commit, the proposal twice referred to the type as though it were a property of a session.

Correct the references to the session property.


jugglinmike:
@jgraham @whimboo Any thoughts on this?

PR #351, opened 2023-01-04
Fix returning null from commands

jgraham:
We want returning null to map onto an empty result. We could do this explicitly per command, but it's less verbose to just allow null at the top level.


jugglinmike:
This supports the commands which return "success with data null" (as of fa580311f1533d9147a6dc1c2fdc98b7f2e7c31a, that's session.subscribe, session.unsubscribe, and browsingContext.handleUserPrompt). It does not support script.removePreloadScript since those steps currently specify "return null." Maybe this patch should update that last command, too.

Also, we'll have to update this part of the definition of "command":

>     - A set of [=remote end steps=] which define the actions to take for a command
>       given a [=BiDi session=] and [=command parameters=] and return an
>       instance of the command [=result type=].

Alternatively, I'll suggest a tweak to gh-350 which makes it more palatable.

PR #356, opened 2023-01-23
Fix map and set links

jgraham:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/356.html" title="Last updated on Feb 7, 2023, 12:28 PM UTC (a5a785b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/356/620a116...a5a785b.html" title="Last updated on Feb 7, 2023, 12:28 PM UTC (a5a785b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -105,7 +105,7 @@ spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
     text: IsPromise; url: sec-ispromise
     text: IsRegExp; url: sec-isregexp
     text: LengthOfArrayLike; url: sec-lengthofarraylike
-    text: Map; url: sec-map-objects
+    text: Map; url: #sec-map-iterable; for: constructor
     text: Number; url: sec-number-constructor
     text: null; url: sec-null-value
     text: Object; url: sec-object-objects
@@ -163,10 +163,6 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
     text: worklet global scopes; url: worklets.html#concept-document-worklet-global-scopes
 </pre>
 
-<pre class="link-defaults">
-spec:infra; type:dfn; for:/; text:set
-</pre>
-
 <style>
 var {
   color: #cd5c5c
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -806,7 +802,7 @@ To <dfn>handle an incoming message</dfn> given a [=WebSocket connection=]
        [=WebSocket connection=] be |connection|, and remove |connection| from
        the set of [=WebSocket connections not associated with a session=].
 
-    1. Let |response| be a new map matching the <code>CommandResponse</code>
+    1. Let |response| be a new [=/map=] matching the <code>CommandResponse</code>
```
whimboo, line 805: Some line below there is a `map` entry like `If parsed is a map ...`.  I assume this should also be linked? If yes, maybe there are other instances as well.

whimboo: I think that there is still one instance left which should be correctly linked.

`The return value of this algorithm is a map between event names and
contexts.`

whimboo: Similar for `set`, mostly when I search for ` set.`

`|browsing contexts| is a set because a [=shared worker=] can be associated with multiple contexts.`

`contains |connection|, remove |connection| from that set.`

`Let |event names| be an empty set.`

And others.

whimboo (changes requested):
There are a couple of more instances which you seem to have missed. See my inline comment. Please let me know if those shouldn't be linked.

index.bs
jgraham: FWIW I don't think we need to link text that's just informative e.g. `the return value of this alogrithm is a map`. But I'll fix the rest.

whimboo: Sounds fair. Thanks!

whimboo (approved):
As it looks like we still have quite a few remaining non-linked instances left to be done for both `map` and `set`. Here all the entries that I was able to find.

* `The populated value of this map is the command parameters`?
* `Let remote value be a map matching the NodeRemoteValue`
* `new map as serialization internal map, realm and session` (2x)
* `Let entry be a map matching the log.ConsoleLogEntry production`
* `Let body be a map matching the log.EntryAdded production`
* `Let targets be an empty map.`
* `Let contexts be an empty set.`
* `Let top-level browsing contexts be an empty set.`
* `Let events be an empty set.`
* `remove connection from that set.`
* `Let event names be an empty set.`
* `Let child browsing contexts be an empty set.`
* `Let related browsing contexts be an empty set.`

With all of these fixed r=me.

jgraham:
I don't think we need to link all the cases where we're not constructing a new value, but talking about an existing value. Anyway, I've updated a bit (thanks for the suggestions), and merged what I have to avoid this bitrotting.

PR #361, opened 2023-02-01
Add support for custom browser to client messages

jgraham:
This introduces a `script.Channel` type which deserializes to a function that when called sends a `script.message` event with its first argument.

It also changes preload scripts to work as functions rather than just being evaluated directly. This allows arguments to be passed in particuarly for the use of channels so that preload scripts can install event listeners that message back to the client.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/361.html" title="Last updated on Apr 12, 2023, 12:48 PM UTC (72ed86e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/361/9da823f...72ed86e.html" title="Last updated on Apr 12, 2023, 12:48 PM UTC (72ed86e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -101,6 +101,7 @@ spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
     text: Construct; url: sec-construct
     text: CreateArrayFromList; url: sec-createarrayfromlist
     text: CreateArrayIterator; url: sec-createarrayiterator
+    text: CreateBuiltinFunction; url: sec-createbuiltinfunction
     text: CreateListFromArrayLike; url: sec-createlistfromarraylike
     text: CreateMapIterator; url: sec-createmapiterator
     text: CreateSetIterator; url: sec-createsetiterator
@@ -1576,6 +1577,9 @@ To <dfn>deserialize local value</dfn> given |local protocol value|, |realm| and
 
 1. If |local protocol value| matches the [=PrimitiveProtocolValue=] production, return
    [=deserialize primitive protocol value=] with |local protocol value|.
+
+1. If |local protocol value| matches the <code>script.ChannelValue</code> production,
+   return [=create a channel=] with |session|, |realm| and |local protocol value|.
 
 1. Let |type| be the value of the <code>type</code> field of |local protocol value| or undefined if
    no such a field.
@@ -5064,6 +5068,17 @@ To <dfn export>run WebDriver BiDi preload scripts</dfn> given |environment setti
        |browsing context|. Otherwise let |realm| be |environment settings|'
        [=realm execution context=]'s Realm component.
 
+    1. Let |arguments| be |preload script|'s <code>arguments</code>.
+
+    1. Let |deserialized arguments| be an empty list.
+
+    1. For each |argument| in |arguments|:
+
+      1. Let |channel| be [=create a channel=] with |session|, |realm| and
+         |argument|.
+
+      1. Append |channel| to |deserialized arguments|.
+
     1. Let |base URL| be the [=API base URL=] of |environment settings|.
 
     1. Let |options| be the [=default classic script fetch options=].
@@ -5089,7 +5104,7 @@ To <dfn export>run WebDriver BiDi preload scripts</dfn> given |environment setti
     1. [=Prepare to run script=] with |environment settings|.
 
     1. Set |evaluation status| to
-       [=Call=](|function object|, null, «»).
+       [=Call=](|function object|, null, |deserialized arguments|).
 
     1. [=Clean up after running script=] with |environment settings|.
 
@@ -5099,6 +5114,49 @@ To <dfn export>run WebDriver BiDi preload scripts</dfn> given |environment setti
 </div>
 
 ### Types ### {#module-script-types}
+
+#### The script.Channel type #### {#type-script-Channel}
+<pre class="cddl remote-cddl local-cddl">
+script.Channel = text;
+</pre>
+
+The <code>script.Channel</code> type represents the id of a specific channel
+used to send custom messages from the [=remote end=] to the [=local end=].
+
+#### The script.ChannelValue type #### {#type-script-ChannelValue}
+
+<pre class="cddl remote-cddl">
+script.ChannelValue = {
+  type: "channel",
+  value: script.ChannelProperties,
+};
+
+script.ChannelProperties = {
+  channel: script.Channel,
+  ?serializationOptions: script.SerializationOptions,
+  ?ownership: script.ResultOwnership,
+}
+</pre>
+
+The <code>script.ChannelValue</code> type represents an
+<code>ArgumentValue</code> that can be deserialized into a function that sends
+messages from the [=remote end=] to the [=local end=].
+
+<div algorithm>
+
+To <dfn>create a channel</dfn> given |session|, |realm| and |protocol value|:
+
+1. Let |channel properties| be |protocol value|["<code>value</code>"].
+
+1. Let |steps| be the following steps given the argument |message|:
+
+  1. Let |current realm| be the [=current Realm Record=].
+
+  1. [=Emit a script message=] with |session|, |current realm|, |channel properties| and |message|.
+
+1. Return [=CreateBuiltinFunction=](|steps|, 1, "", « », |realm|).
+
+</div>
 
 #### The script.ExceptionDetails type #### {#type-script-ExceptionDetails}
 
@@ -5669,6 +5727,7 @@ script=].
 
       script.AddPreloadScriptParameters = {
         functionDeclaration: text,
+        ?arguments: [*script.ChannelValue],
         ?sandbox: text
       }
       </pre>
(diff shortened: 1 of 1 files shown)


jgraham:
There are a couple of open issues around the proposed change to make preload scripts work like `callFunction` rather than evaluate. This is required if we want to pass in a Channel as an argument when defining the preload scripts, but it comes with some tradeoffs:
* How to report an error when deserializing the arguments. We have to do this at the point at which the script is executed, so it can't be a response to the command. If the deserialization fails it's not really a javascript error that we can report through `report an exception` because it's purely about invalid data in the WebDriver protocol layer.
* Does it make sense to allow all argument types? In particular primitives and channels probably make sense, but remote references are going to be at best fragile.
* Is it OK to have a function for which you can't get the return value? I don't think we have a use case for the preload script function returning a value, but it still seems somewhat strange that it should be dropped entirely.

So I think there are a few options:

1. We could add an event which happens after a preload script, indicating either the result of the function, or an error. That would make it unambiguous that something went wrong with a specific preload script, but I don't know of other use cases for the event.
2. We could restrict preload scripts to getting primitives and channels as the arguments. In that case deserialization wouldn't depend on external factors (e.g. whether some node or window is accessible), so we could just return an "invalid argument" error to the `addPreloadScript` command.
3. We could model failure to deserialize an argument as some kind of JS-level error, even though it's not really.
4. We could revert to using evaluation for preload scripts, and allow channels (only) to be injected as some kind of script-scoped variable. 

I think I prefer Option 2, but it does restrict the use cases somewhat. For example you can't open a page, then install a preload script, with a `SharedReference` to a node on the first page as an argument, and then `window.open` a bunch of other browsing contexts which run the reload script and access the node on the original page. But since it would only work in cases where the Documents were mutually accessible anyway, there's probably a workaround.

whimboo:
@jgraham there is one other question that I have. In regards of the usage for Selenium I think that they also wanted to have the possibility to install such a script that then can be re-used to not have to send the specific atoms for each and every eg. element displayed check. How would this actually work with the `callFunction` like method? There has to be a way to trigger the script with the `SharedReference` as argument.

jgraham:
Something like:
```js
webdriver.installPreloadScript(
  () => {
    window.isDisplayed = element => {
      // Implementation
    }
  },
  sandbox="selenium-internal")
```
and then later
```js
webdriver.callFunction(elem => isDisplayed(elem), arguments=[element], context=contextId, sandbox="selenium-internal")
```

whimboo:
I see. So the `channel` would actually only be required for the pre-load script itself and in those cases when it is needed. Otherwise it can be left-out. Thanks.

index.bs
```diff
@@ -4606,7 +4664,8 @@ script=].
       }
 
       script.AddPreloadScriptParameters = {
-        expression: text;
+        arguments: [*script.ArgumentValue],
```
lutien, line 4667: Do we maybe want to have `arguments` parameter as optional, the same as in callFunction?

jgraham:
I updated this as follows:
* Only accept `script.Channel` as an argument to preload scripts
* Error reporting all happens with JS exceptions

One quality of life improvement we could make would be to return an error to the command if the script can't be evaluated as a function body. The question is where that evaluation happens e.g. if I pass in `window.foo = 1` as the preload script, we don't want `foo=1` to be set on some existing global, before noticing that the script is not in fact a function.

index.bs
```diff
@@ -4619,18 +4677,27 @@ script=].
...
+
 1. Let |script| be the string representation of a [[!RFC4122|UUID]].
 
 1. Let |sandbox| be the value of the "<code>sandbox</code>" field in |command
    parameters|, if present, or null otherwise.
 
+1. Issue: We could parse the script at this point so we can response with an
+   error if the script doesn't evaluate as a function. However it's not clear
+   which realm would be used for that evaluation (we don't want to actually add
+   the function to any globals).
```
whimboo, line 4694: Hi @sadym-chromium. Could you maybe help here and tell us if CDP parses the bootstrap script when it gets added, or is the parsing happening once it gets injected and an error is returned? If it parses / validates it directly it would be good to know how this is done.

index.bs
```diff
@@ -5080,6 +5157,58 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+   [=realm execution context=]'s Realm component is |realm|.
+
+1. Let |related browsing contexts| be the result of [=get related browsing
+   contexts=] given |environment settings|.
+
+1. If [=event is enabled=] given |session|, "<code>script.message</code>"
+   and |related browsing contexts|:
+
+  1. Let |data| be the result of [=serialize as a remote value=] given
+     |message|, <code>1</code> as max depth, <code>none</code> as ownership type,
```
whimboo, line 5194: I wonder if we never want to have ownership here. Given that there might be cases when a client wants to use a returned objects as argument for a successive `callFunction` call a handle would be needed. Right now we allow any kind of value returned via the channel.

index.bs
```diff
@@ -1529,6 +1530,9 @@ To <dfn>deserialize local value</dfn> given |local protocol value|, |realm| and
 1. If |local protocol value| matches the [=PrimitiveProtocolValue=] production, return
    [=deserialize primitive protocol value=] with |local protocol value|.
 
+1. If |local protocol value| matches the <code>script.Channel</code> production,
```
sadym-chromium, line 1533: Can we pick other naming, as `channel` [is already used](https://github.com/GoogleChromeLabs/chromium-bidi#field-channel) in out implementation-specific extension for other purpose?

index.bs
sadym-chromium: The script is parsed at the moment it is executed.

index.bs
```diff
@@ -5080,6 +5157,58 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+      <pre class="cddl local-cddl">
+        script.Message = {
+         method: "script.message",
+         params: script.MessageParameters
+       }
+
+       script.MessageParameters = {
+         channel: script.Channel,
+         data: RemoteValue
+         realm: script.Realm
```
sadym-chromium, line 5174: As a client, in order to understand from which BrowsingContext the event came, I have to track the map of realm-BrowsingContext by subscribing to and handling `script.realmCreated` + `script.realmDestroyed`. 

Consider adding field `context: browsingContext.BrowsingContext / null`. Nullable to support workers in future.

jgraham:
To help with reviewing, I've split the preload script changes out into https://github.com/w3c/webdriver-bidi/pull/381 and made this PR just about the addition of the message channel. This PR is based on that one, so it needs to merge first.

index.bs
```diff
@@ -4222,6 +4234,43 @@ To <dfn export>run WebDriver BiDi preload scripts</dfn> given |environment setti
...
+script.Channel = {
+  type: "channel",
+  value: script.ChannelProperties,
+};
+
+script.ChannelId = text;
+
+script.ChannelProperties = {
+  channel: script.ChannelId,
+  ?maxDepth: js-uint,
```
sadym-chromium, line 4249: We don't support `maxDepth > 1` yet, and the work on it is not planned yet, so let's either remove it and default to `1` like `log.entryAdded` and `script.callFunction`.

jgraham: It does default to 1 (not in the CDDL but in the prose).

sadym-chromium: Sorry, didn't finish the sentence. Let's either remove it, or restrict the value to be in `[0,1]`

index.bs
```diff
@@ -5259,6 +5314,66 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+   <dt>Event Type</dt>
+   <dd>
+      <pre class="cddl local-cddl">
+        script.Message = {
+         method: "script.message",
+         params: script.MessageParameters
+       }
+
+       script.MessageParameters = {
+         channel: script.Channel,
```
sadym-chromium, line 5329: Maybe just the `script.ChannelId`?
```suggestion
         channel: script.ChannelId,
```

jgraham: We don't use the `Id` suffix elsewhere (except `sharedId` / `internalId`, which maybe in retrospect we should have called `object` / `node` or something).

sadym-chromium: `script.Channel` refer to the structure `script.Channel = { type: "channel", value: script.ChannelProperties }`, and I assume we want only the channel name here, which is `script.ChannelId`.

jgraham: Ah, good catch. I fixed this by changing the `Channel` type to `ChannelValue` and `ChannelId` to `Channel`.

index.bs
sadym-chromium: Addressed https://github.com/w3c/webdriver-bidi/issues/390

sadym-chromium: Addressed https://github.com/w3c/webdriver-bidi/issues/390

jimevans:
@sadym-chromium Are you happy with the most recent changes here? I’d really like to get this merged, if there is consensus that it’s ready.

sadym-chromium (approved):

PR #363, opened 2023-02-07
Add browsingContext.print command

jgraham:
This is just like the WebDriver classic print endpoint, and produces a base64-encoded PDF representation of the document in the specified context.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/363.html" title="Last updated on Feb 21, 2023, 4:02 PM UTC (1930b1f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/363/620a116...1930b1f.html" title="Last updated on Feb 21, 2023, 4:02 PM UTC (1930b1f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3464,7 +3464,7 @@ PDF document represented as a Base64-encoded string.
         ?orientation: ("portrait" / "landscape") .default "portrait",
         ?page: browsingContext.PrintPageParameters,
         ?pageRanges: [*text],
-        ?scale: 0.1..2.0 .default 1,
+        ?scale: 0.1..2.0 .default 1.0,
         ?shrinkToFit: bool .default true,
       }
 
--- a/index.bs
+++ b/index.bs
@@ -3445,7 +3445,7 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
 
 The <dfn export for=commands>browsingContext.print</dfn> command
 creates a paginated representation of a document, and returns it as a
-Base64-encoded string.
+PDF document represented as a Base64-encoded string.
 
 
 <dl>
--- a/index.bs
+++ b/index.bs
@@ -3513,8 +3513,8 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
    <code>browsingContext.PrintPageParameters</code> with the fields set to
    their default values.
 
-1. Let |page ranges| be the value of hte <code>pageRanges</code> field of
-   |command parameters| is present or an empty [=/list=] otherwise.
+1. Let |page ranges| be the value of the <code>pageRanges</code> field of
+   |command parameters| if present or an empty [=/list=] otherwise.
 
 1. Let |document| be |context|'s [=active document=].
 
--- a/index.bs
+++ b/index.bs
@@ -3577,7 +3577,7 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
 
   1. Let |encoded data| be |encoding result|’s data.
 
-  1. Let |body| be a [=map=] matching the
+  1. Let |body| be a [=/map=] matching the
      <code>browsingContext.PrintResult</code> production, with the
      <code>data</code> field set to |encoded data|.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3435,6 +3441,150 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
 
 </div>
 
+#### The browsingContext.print Command ####  {#command-browsingContext-print}
+
+The <dfn export for=commands>browsingContext.print</dfn> command
+creates a paginated representation of a document, and returns it as a
```
whimboo, line 3447: Shall we mention that it actually `renders` the content into a paginated format similar to what we have in classic? Especially including that it is in PDF format would be helpful.

index.bs
```diff
@@ -3435,6 +3441,150 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+   <dd>
+      <pre class="cddl remote-cddl">
+      browsingContext.Print = {
+        method: "browsingContext.print",
+        params: browsingContext.PrintParameters
+      }
+
+      browsingContext.PrintParameters = {
+        context: browsingContext.BrowsingContext,
+        ?background: bool .default false,
```
whimboo, line 3462: Question: shall we start using defaults for other commands as well? I haven't seen that yet and it's nice. Happy to file a follow-up issue for that.

index.bs
```diff
@@ -3435,6 +3441,150 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+   parameters| if present, or otherwise a [=/map=] matching the
+   <code>browsingContext.PrintMarginParameters</code> with the fields set to
+   their default values.
+
+1. Let |page size| be the value of the <code>page</code> field of |command
+   parameters| if present, or otherwise a [=/map=] matching the
+   <code>browsingContext.PrintPageParameters</code> with the fields set to
+   their default values.
+
+1. Let |page ranges| be the value of hte <code>pageRanges</code> field of
```
whimboo, line 3516:
```suggestion
1. Let |page ranges| be the value of the <code>pageRanges</code> field of
```

index.bs
```diff
@@ -3435,6 +3441,150 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+   |command parameters| is present or an empty [=/list=] otherwise.
```
whimboo, line 3517:
```suggestion
   |command parameters| if present or an empty [=/list=] otherwise.
```

whimboo (changes requested):

index.bs
jgraham: Yeah, if people like it then I agree it's better than prose :)

index.bs
jgraham: I was considering that PDF is an implementation detail, and that we could support other formats in the future. But I guess if that happens we can update the text.

index.bs
```diff
@@ -3435,6 +3441,150 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+      }
+
+      browsingContext.PrintParameters = {
+        context: browsingContext.BrowsingContext,
+        ?background: bool .default false,
+        ?margin: browsingContext.PrintMarginParameters,
+        ?orientation: ("portrait" / "landscape") .default "portrait",
+        ?page: browsingContext.PrintPageParameters,
+        ?pageRanges: [*text],
+        ?scale: 0.1..2.0 .default 1,
```
whimboo, line 3467: Would this require a float type as well?

index.bs
whimboo: Great. I filed https://github.com/w3c/webdriver-bidi/issues/368.

index.bs
whimboo: @jgraham I actually meant more the following: `?scale: (int 0.1..2.0) .default 1.0,` or do we want to force `float` type only?

jgraham: There isn't a difference in JS/JSON between number types; if we allow floats we also allow integers i.e. there's no semantic difference in JSON between `1` and `1.0`. The integer check is basically post-processing of the form "does this JS number represent an exact integer".

foolip (approved):

index.bs
whimboo: Ok, so we can actually omit specifying the type when a default value or a range is given. So I'm fine with what we have here.

whimboo (approved):

whimboo:
FYI we are [going to land our implementation soon](https://bugzilla.mozilla.org/show_bug.cgi?id=1806810) and a basic set of tests will be included.

PR #381, opened 2023-03-09
Make preload scripts take a function declaration instead of expression

jgraham:
This is a prerequisite to adding support for arguments, which is required to allow passing in message channels.

This also cleans up a few bugs with evaluating functions.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/381.html" title="Last updated on Mar 30, 2023, 11:05 AM UTC (bc976d4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/381/99db2d4...bc976d4.html" title="Last updated on Mar 30, 2023, 11:05 AM UTC (bc976d4)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -92,7 +92,6 @@ spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
     text: Construct; url: sec-construct
     text: CreateArrayFromList; url: sec-createarrayfromlist
     text: CreateArrayIterator; url: sec-createarrayiterator
-    text: CreateBuiltinFunction; url: sec-createbuiltinfunction
     text: CreateListFromArrayLike; url: sec-createlistfromarraylike
     text: CreateMapIterator; url: sec-createmapiterator
     text: CreateSetIterator; url: sec-createsetiterator
@@ -4237,11 +4236,13 @@ To <dfn export>run WebDriver BiDi preload scripts</dfn> given |environment setti
 
     1. Let |function declaration| be |preload script|'s <code>function declaration</code>.
 
-    1. Let (|script|, |function body evaluation status|) be the result of [=evaluate function body=]
-       |function declaration|, |environment settings|, |base URL|, and |options|.
+    1. Let (|script|, |function body evaluation status|) be the result of
+       [=evaluate function body=] with |function declaration|,
+       |environment settings|, |base URL|, and |options|.
 
     1. If |function body evaluation status| is an [=abrupt completion=], then
-       [=report the exception=] given by |evaluation status|.\[[Value]] for |script|.
+       [=report the exception=] given by |function body evaluation status|.\[[Value]]
+       for |script|.
 
     1. Let |function object| be |function body evaluation status|.\[[Value]].
 
@@ -4258,7 +4259,7 @@ To <dfn export>run WebDriver BiDi preload scripts</dfn> given |environment setti
 
     1. [=Clean up after running script=] with |environment settings|.
 
-    1. If |function body evaluation status| is an [=abrupt completion=], then
+    1. If |evaluation status| is an [=abrupt completion=], then
        [=report the exception=] given by |evaluation status|.\[[Value]] for |script|.
 
 </div>
--- a/index.bs
+++ b/index.bs
@@ -4864,7 +4864,7 @@ The [=remote end steps=] given |session| and |command parameters| are:
 1. Let |preload script map| be |session|'s [=preload script map=].
 
 1. Set |preload script map|[|script|] to a struct with <code>function
-   declaration</code> |function declaration|, , and <code>sandbox</code> |sandbox|.
+   declaration</code> |function declaration|, and <code>sandbox</code> |sandbox|.
 
 1. Return a new [=/map=] matching the <code>script.AddPreloadScriptResult</code> with the
    <code>script</code> field set to |script|.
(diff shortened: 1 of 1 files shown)


sadym-chromium:
Handing over to @thiagowfx

whimboo:
FYI related test changes could be done via https://bugzilla.mozilla.org/show_bug.cgi?id=1822779

index.bs
```diff
@@ -4779,8 +4799,8 @@ The [=remote end steps=] given |session| and |command parameters| are:
 
 1. Let |preload script map| be |session|'s [=preload script map=].
 
-1. Set |preload script map|[|script|] to a struct with <code>expression</code>
-   |expression| and <code>sandbox</code> |sandbox|.
+1. Set |preload script map|[|script|] to a struct with <code>function
+   declaration</code> |function declaration|, , and <code>sandbox</code> |sandbox|.
```
whimboo, line 4803: I think we want to use the `<code>` variant here, and please also remove the left over comma.

whimboo (changes requested):
Thanks @jgraham for those changes! I had a first pass over the code and it looks fine to me in general. I have a couple of questions and comments through. Please have a look at those.

whimboo:
I think that we should get this out sooner than later before clients actually start using this API and would be confronted with backward incompatible changes.

index.bs
jgraham, line 4803:
```suggestion
   declaration</code> |function declaration|, and <code>sandbox</code> |sandbox|.
```

jgraham: I don't know what you mean by "the `<code>` variant".

index.bs
```diff
@@ -4769,7 +4789,7 @@ script=].
 <div algorithm="remote end steps for script.addPreloadScript">
 The [=remote end steps=] given |session| and |command parameters| are:
 
-1. Let |expression| be the <code>expression</code> field of |command
+1. Let |function declaration| be the <code>function declaration</code> field of |command
```
juliandescottes, line 4792:
```suggestion
1. Let |function declaration| be the <code>functionDeclaration</code> field of |command
```

nit: Should be `functionDeclaration` here?

index.bs
whimboo: I've read that wrongly. Please ignore.

thiagowfx:
> Handing over to @thiagowfx

Seems OK to me. Keep in mind I am just starting to get familiar with the spec though.

thiagowfx (approved):

index.bs
```diff
@@ -4225,24 +4231,36 @@ To <dfn export>run WebDriver BiDi preload scripts</dfn> given |environment setti
...
+       1. [=Report the exception=] |error|.
+
+    1. [=Prepare to run script=] with |environment settings|.
+
+    1. Set |evaluation status| to
+       [=Call=](|function object|, null, «»).
 
     1. [=Clean up after running script=] with |environment settings|.
 
+    1. If |function body evaluation status| is an [=abrupt completion=], then
```
whimboo, line 4261:
```suggestion
    1. If |evaluation status| is an [=abrupt completion=], then
```

whimboo (changes requested):

index.bs
```diff
@@ -4225,24 +4231,36 @@ To <dfn export>run WebDriver BiDi preload scripts</dfn> given |environment setti
...
+       1. Let |error| be a new [=TypeError=] object in |realm|.
 
-    1. If |evaluation status| is an [=abrupt completion=], then [=report the
-       exception=] given by |evaluation status|.\[[Value]] for |script|.
+       1. [=Report the exception=] |error|.
+
+    1. [=Prepare to run script=] with |environment settings|.
+
+    1. Set |evaluation status| to
+       [=Call=](|function object|, null, «»).
```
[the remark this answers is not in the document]
jgraham: It's an empty list, as used in Infra and ECMAScript. See https://infra.spec.whatwg.org/#lists. Undefined would be wrong.

index.bs
```diff
@@ -4225,24 +4230,38 @@ To <dfn export>run WebDriver BiDi preload scripts</dfn> given |environment setti
...
+    1. Let |function object| be |function body evaluation status|.\[[Value]].
+
+    1. If [=IsCallable=](|function object|) is <code>false</code>:
 
-    1. Let |evaluation status| be [=ScriptEvaluation=](|script|'s record).
+       1. Let |error| be a new [=TypeError=] object in |realm|.
 
-    1. If |evaluation status| is an [=abrupt completion=], then [=report the
-       exception=] given by |evaluation status|.\[[Value]] for |script|.
+       1. [=Report the exception=] |error|.
```
whimboo, line 4253: Oh, I missed to ask if this error should actually be a dedicated WebDriver specific error or if `TypeError` is fine here which later would be converted to a `UnknownError` most likely.

whimboo (approved):
Looks fine to me now. Thanks @jgraham! Just a nit inline you might want to have a look.

index.bs
jgraham: It doesn't ever get converted into a WebDriver error; there's no command in progress at the point at which preload scripts are run. Instead an exception is reported to the console and via `log.logEntryAdded` if the client is subscribed. We _could_ theoretically define some new JS error types for exceptions that happen in WebDriver scripts, but I don't think that's a good idea (UAs can already do things like replace the script filename by something that indicates a preload script in the stack trace).

PR #385, opened 2023-03-20
Make JS value serialization more flexible

jgraham:
This is required to allow one-shot serialisation of nested objects, which is important to avoid having to make multiple script calls to get the data from nested objects.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/385.html" title="Last updated on Mar 28, 2023, 8:23 AM UTC (d22d88d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/385/fff0819...d22d88d.html" title="Last updated on Mar 28, 2023, 8:23 AM UTC (d22d88d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2227,7 +2227,7 @@ and |session|:
 To <dfn>serialize as a list</dfn> given |iterable|, |serialization options|, |ownership type|,
 |serialization internal map|, |realm|, and |session|:
 
-1. Assert: serialization options|["<code>maxObjectDepth</code>"] is greater
+1. Assert: |serialization options|["<code>maxObjectDepth</code>"] is greater
    than 0.
 
 1. Let |serialized| be a new list.
@@ -2258,7 +2258,7 @@ Issue: this assumes for-in works on iterators
 To <dfn>serialize as a mapping</dfn> given |iterable|, |serialization options|,
 |ownership type|, |serialization internal map|, |realm|, and |session|:
 
-1. Assert: serialization options|["<code>maxObjectDepth</code>"] is greater
+1. Assert: |serialization options|["<code>maxObjectDepth</code>"] is greater
    than 0.
 
 1. Let |serialized| be a new list.
--- a/index.bs
+++ b/index.bs
@@ -4864,12 +4864,13 @@ Note: In case of an arrow function in <code>functionDeclaration</code>, the
       }
 
       script.CallFunctionParameters = {
-        functionDeclaration: text;
-        awaitPromise: bool;
-        target: script.Target;
-        ?arguments: [*script.ArgumentValue];
-        ?this: script.ArgumentValue;
-        ?resultOwnership: script.ResultOwnership;
+        functionDeclaration: text,
+        awaitPromise: bool,
+        target: script.Target,
+        ?arguments: [*script.ArgumentValue],
+        ?maxDepth: js-uint,
+        ?resultOwnership: script.ResultOwnership,
+        ?this: script.ArgumentValue,
       }
 
       script.ArgumentValue = (
@@ -4962,6 +4963,9 @@ The [=remote end steps=] with |session| and |command parameters| are:
 1. Let |await promise| be the value of the <code>awaitPromise</code> field of
    |command parameters|.
 
+1. Let |max depth| be the value of the <code>maxDepth</code> field of |command
+   parameters|, if present, or 1 otherwise.
+
 1. Let |result ownership| be the value of the <code>resultOwnership</code> field of
    |command parameters|, if present, or <code>none</code> otherwise.
 
@@ -5008,7 +5012,7 @@ The [=remote end steps=] with |session| and |command parameters| are:
 1. Assert: |evaluation status|.\[[Type]] is <code>normal</code>.
 
 1. Let |result| be the result of [=serialize as a remote value=] with
-   |evaluation status|.\[[Value]], <code>1</code> as |max depth|, |result ownership|,
+   |evaluation status|.\[[Value]], |max depth|, |result ownership|,
    a new [=/map=] as serialization internal map, |realm| and |session|.
 
 1. Return a new [=/map=] matching the <code>script.EvaluateResultSuccess</code>
@@ -5038,10 +5042,11 @@ the promise is returned.
       }
 
       script.EvaluateParameters = {
-        expression: text;
-        target: script.Target;
-        awaitPromise: bool;
-        ?resultOwnership: script.ResultOwnership;
+        expression: text,
+        target: script.Target,
+        awaitPromise: bool,
+        ?maxDepth: js-uint,
+        ?resultOwnership: script.ResultOwnership,
       }
       </pre>
    </dd>
@@ -5072,6 +5077,9 @@ The [=remote end steps=] with |session| and |command parameters| are:
 
 1. Let |await promise| be the value of the <code>awaitPromise</code> field of
    |command parameters|.
+
+1. Let |max depth| be the value of the <code>maxDepth</code> field of |command
+   parameters|, if present, or 1 otherwise.
 
 1. Let |result ownership| be the value of the <code>resultOwnership</code> field of
    |command parameters|, if present, or <code>none</code> otherwise.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -4864,12 +4864,13 @@ Note: In case of an arrow function in <code>functionDeclaration</code>, the
...
-        awaitPromise: bool;
-        target: script.Target;
-        ?arguments: [*script.ArgumentValue];
-        ?this: script.ArgumentValue;
-        ?resultOwnership: script.ResultOwnership;
+        functionDeclaration: text,
+        awaitPromise: bool,
+        target: script.Target,
+        ?arguments: [*script.ArgumentValue],
+        ?maxDepth: js-uint,
```
whimboo, line 4871: I assume that as long it is not clear if we can use defaults here you wont further use them? Otherwise we could set a default of `1` already here and won't need the prose.

whimboo (approved):
Just a single question. Otherwise this is a great idea.

@sadym-chromium can you please review as well?

sadym-chromium:
In the light of https://github.com/w3c/webdriver-bidi/issues/390 adding this parameter without changing the serialization logic is pushing users to make the potentially fatal mistake with using big-enough `maxDepth` by default.

sadym-chromium (changes requested):
In the light of https://github.com/w3c/webdriver-bidi/issues/390 adding this parameter without changing the serialization logic is pushing users to make the potentially fatal mistake with using big-enough `maxDepth` by default.

jgraham:
I think "potentially fatal" is wildly overstating the potential impact here. But sure, I'm happy to change this PR to add more finegrained control over serialization.

index.bs
```diff
@@ -2029,12 +2033,12 @@ an |ownership type|, a |serialization internal map|, a |realm| and a |session|:
     <dt>|value| is a [=platform object=] that implements {{NodeList}}
     <dd>Let |remote value| be [=serialize an Array-like=] with
         <code>NodeListRemoteValue</code>,|handle id|, |known object|, |value|,
-        |max depth|, |ownership type|, |serialization internal map|, and |realm|.
+        |serialization options|, |ownership type|, |serialization internal map|, and |realm|.
```
sadym-chromium, line 2036: Shouldn't we distinguish between `NodeList`+`HTMLCollection` and `Array`, and use  `maxDomDepth` / `maxObjectDepth` accordingly?

index.bs
```diff
@@ -2107,20 +2124,16 @@ an |ownership type|, a |serialization internal map|, a |realm| and a |session|:
...
                 Otherwise run the following substeps:
 
-               1.  Let |child depth| be |max depth| - 1 if |max depth| is not
-                  null, or null otherwise.
-
                1. Let |serialized shadow| be the result of
-                  [=serialize as a remote value=] with |shadow root|, |child depth|,
-                  false, |ownership type|, |serialization internal map|,
+                  [=serialize as a remote value=] with |shadow root|,
+                  |serialization options|, |ownership type|, |serialization internal map|,
```
sadym-chromium, line 2129: Shouldn't we decrease `maxDomDepth` here?

sadym-chromium (approved):

index.bs
jgraham: The fact that `HTMLCollection` and so on are platform objects rather than JS runtime objects seems less important than the fact that they're not part of the DOM.

index.bs
jgraham: See the commit message. The intent is that we always serialize the id of the shadow root, but whether we descend into the actual nodes in the shadow tree depends on the other options that are specified.

sadym-chromium:
LGTM

index.bs
```diff
@@ -2206,18 +2221,23 @@ To <dfn>serialize an Array-like</dfn> given |production|, |handle id|, |known ob
...
 
-    1. Let |child depth| be |max depth| - 1 if |max depth| is not null, or null
-       otherwise.
+    1. Let |child serialization options| be a [=map/clone=] of
+       |serialization options|.
+
+    1. If |child serialization options|["<code>maxObjectDepth</code>"] is
+       not null, set |child serialization
+       options|["<code>maxObjectDepth</code>"] to |child serialization
+       options|["<code>maxObjectDepth</code>"] - 1.
```
whimboo, line 2237: I'm fairly sure we can move the cloning out of the loop so that we do not have to do it for each and every child. Also when should happen when this method gets a `maxObjectDepth` of `0` passed in? Right now we end-up again with `-1`.

index.bs
```diff
@@ -2243,16 +2263,21 @@ To <dfn>serialize as a mapping</dfn> given |iterable|, |max depth|,
...
 
-  1. Let |child depth| be |max depth| - 1 if |max depth| is not null, or null
-     otherwise.
+  1. Let |child serialization options| be a [=map/clone=] of
+     |serialization options|.
+
+  1. If |child serialization options|["<code>maxObjectDepth</code>"] is
+     not null, set |child serialization
+     options|["<code>maxObjectDepth</code>"] to |child serialization
+     options|["<code>maxObjectDepth</code>"] - 1.
```
whimboo, line 2272: Same question here when `0` is passed in as value.

whimboo (changes requested):
Generally this looks like a great improvement and will cause quite a bit of refactoring including the tests. I found a couple of issues which you should have a look at. Thanks.

index.bs
```diff
@@ -2107,20 +2126,16 @@ an |ownership type|, a |serialization internal map|, a |realm| and a |session|:
...
-                  false, |ownership type|, |serialization internal map|,
+                  [=serialize as a remote value=] with |shadow root|,
+                  |serialization options|, |ownership type|, |serialization internal map|,
                   |realm|, and |session|.
 
-                Note: this means the <code>handle</code> for the shadow root
-                will be serialized irrespective of whether the shadow is open or closed,
-                but no properties of the node will be returned.
-
              1. Set |serialized|["<code>shadowRoot</code>"] to |serialized shadow|.
```
lutien, line 2134: Since you're on the topic, maybe you could fix [this issue](https://w3c.github.io/webdriver-bidi/#issue-e8085260)?
```suggestion
             1. If |serialized shadow| is not null, set |serialized|["<code>shadowRoot</code>"] to |serialized shadow|.
```

index.bs
```diff
@@ -2206,31 +2224,43 @@ To <dfn>serialize an Array-like</dfn> given |production|, |handle id|, |known ob
 </div>
 
 <div algorithm>
-To <dfn>serialize as a list</dfn> given |iterable|, |max depth|, |ownership type|,
+To <dfn>serialize as a list</dfn> given |iterable|, |serialization options|, |ownership type|,
 |serialization internal map|, |realm|, and |session|:
 
-  1. Let |serialized| be a new list.
+1. Assert: serialization options|["<code>maxObjectDepth</code>"] is greater
```
whimboo, line 2230:
```suggestion
1. Assert: |serialization options|["<code>maxObjectDepth</code>"] is greater
```

index.bs
```diff
@@ -2206,31 +2224,43 @@ To <dfn>serialize an Array-like</dfn> given |production|, |handle id|, |known ob
...
 </div>
 
 Issue: this assumes for-in works on iterators
 
 <div algorithm>
-To <dfn>serialize as a mapping</dfn> given |iterable|, |max depth|,
+To <dfn>serialize as a mapping</dfn> given |iterable|, |serialization options|,
 |ownership type|, |serialization internal map|, |realm|, and |session|:
 
+1. Assert: serialization options|["<code>maxObjectDepth</code>"] is greater
```
whimboo, line 2261:
```suggestion
1. Assert: |serialization options|["<code>maxObjectDepth</code>"] is greater
```

whimboo (changes requested):
I've seen that the PR got updated but not all the comments have been addressed or have been replied to. Maybe you could have a look at those? Otherwise two small extra nits.

index.bs
```diff
@@ -1867,6 +1867,7 @@ NodeProperties = {
   ?attributes: {*text => text},
   ?children: [*NodeRemoteValue],
   ?localName: text,
+  ?mode: "open" / "closed",
```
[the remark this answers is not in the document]
whimboo: Oh ok. I might have missed that then. If it's not part of a normal element then it's all fine. And indeed it makes it harder to spot if we do not have specific types defined. But not sure if that's really needed.

PR #386, opened 2023-03-21
Do not change maxDepth when serializing shadow root

lutien:
Since current assumption that we want always to return shadow root id, we shouldn't change maxDepth on this step.

***
<a href="https://pr-preview.s3.amazonaws.com/lutien/webdriver-bidi/pull/386.html" title="Last updated on Mar 21, 2023, 10:19 AM UTC (83b2f9a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/386/fff0819...lutien:83b2f9a.html" title="Last updated on Mar 21, 2023, 10:19 AM UTC (83b2f9a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2109,8 +2109,8 @@ an |ownership type|, a |serialization internal map|, a |realm| and a |session|:
 
                1. Let |serialized shadow| be the result of
                   [=serialize as a remote value=] with |shadow root|, |max depth|,
-                  false, |ownership type|, |serialization internal map|,
-                  |realm|, and |session|.
+                  |ownership type|, |serialization internal map|, |realm|, and
+                  |session|.
 
                 Note: this means the <code>handle</code> for the shadow root
                 will be serialized irrespective of whether the shadow is open or closed,
--- a/index.bs
+++ b/index.bs
@@ -2107,11 +2107,8 @@ an |ownership type|, a |serialization internal map|, a |realm| and a |session|:
              1. If |shadow root| is null, let |serialized shadow| be null.
                 Otherwise run the following substeps:
 
-               1.  Let |child depth| be |max depth| - 1 if |max depth| is not
-                  null, or null otherwise.
-
                1. Let |serialized shadow| be the result of
-                  [=serialize as a remote value=] with |shadow root|, |child depth|,
+                  [=serialize as a remote value=] with |shadow root|, |max depth|,
                   false, |ownership type|, |serialization internal map|,
                   |realm|, and |session|.
 


index.bs
```diff
@@ -2109,8 +2109,8 @@ an |ownership type|, a |serialization internal map|, a |realm| and a |session|:
 
                1. Let |serialized shadow| be the result of
                   [=serialize as a remote value=] with |shadow root|, |max depth|,
-                  false, |ownership type|, |serialization internal map|,
```
lutien, line 2115: Looks like `false` is not needed here, since [serialize as a remote value](https://w3c.github.io/webdriver-bidi/#serialize-as-a-remote-value) accepts only 6 arguments which match all the other mentioned here.

whimboo:
Hm, not sure that I fully understand. When `deserialize` is called with `maxDepth` of `1` which is the default right now we call `deserialize` again in step 5.10.5.2 and then with a depth of 0. This time we serialize a `ShadowRoot` and not an `Element`. As such in step 1 of "value is a [platform object](https://webidl.spec.whatwg.org/#dfn-platform-object) that implements [Node](https://dom.spec.whatwg.org/#node)" we immediately create the `sharedId` and assign it to the struct similar to `handleId` in step 2.

Changing that behavior here to not decrease by `1` would mean that we also try to serialize all the `children` for a `ShadowRoot` now. Is that really what we want?

lutien:
The question is what do we do if `maxDepth` is `0`. The way it would work now (with adding the check for `maxDepth` before subtracting), that we would not create `shareId` and would not serialize any data for shadow root. And since shadow root is not children, the assumption was that we always want to do it.

whimboo:
TLDR; `maxDepth` only affects the recursion into `children` but not `shadow root`. 

With a maxDepth of `0` we jump into `serializing a node` and directly will create a `sharedId` for the element to serialize. In step 8 `maxDepth` is used re assemble the `children`, but we do not care here. For us step 5.10.4 and 5.10.5 are interesting.  First we retrieve the `ShadowRoot` and if present we definitely call into `serialize a remote value` again. Here the same steps for `serializing a node` will be performed and the `sharedId` for the `ShadowRoot` is immediately created and as field of the production value (step 2). Now we ignore the children again as expected, but we still have the `sharedid` set and in step 6 we return a `{ type: "ShadowRoot", sharedId: "some-shadow-root-id" }` dict to the callee, where it is added to the final dict like: 

```
{
  nodeType: "Element", 
  sharedId: "some-element-id",
  value: {
    [..],
    shadowRoot: {
      nodeType: "ShadowRoot",
      sharedId: "some-shadow-root-id",
      value: {
        childNodeCount: X,
      }
    } 
  } 
}

```

whimboo (approved):
As discussed in our internal meeting today we were fine in getting this PR merged. Also because https://github.com/w3c/webdriver-bidi/pull/385 will follow-up pretty soon as well.

jgraham:
Closing in favour of https://github.com/w3c/webdriver-bidi/pull/385

PR #392, opened 2023-03-29
Remove the issue about optional shadowRoot field

lutien:
Removes the issue (see the deiscussion in this PR): https://w3c.github.io/webdriver-bidi/#issue-e8085260

***
<a href="https://pr-preview.s3.amazonaws.com/lutien/webdriver-bidi/pull/392.html" title="Last updated on Mar 29, 2023, 9:47 AM UTC (689d132)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/392/6fdbd49...lutien:689d132.html" title="Last updated on Mar 29, 2023, 9:47 AM UTC (689d132)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2203,9 +2203,6 @@ reference to <code>childNodes</code>, or it should be renamed to <code>childrenC
 Issue: <code>nodeValue</code> can be null. Omit the field in such a case, or adjust type to allow
 null value.
 
-Issue: <code>shadowRoot</code> field is optional. Omit it in case of |shadow root| is null, or
-remove optional flag from the field.
-
 </div>
 
 <div algorithm>


jgraham:
Yeah, I think that makes sense. In the serialization we're typically using `null` to mean "this thing is definitively null` and a missing property to mean "we` didn't try to serialize this" (e.g. because we reached the max depth). It's muddied a bit by the fact that we also don't serialize properties that don't make sense for a specific node type (this is one reason I'm wondering about having a specific `type` for each node type).

The other option would be to have a specific sentinel type to indicate that we omitted a property due to the serialization configuration (i.e. instead of not serializing `children` when we reach max node depth we'd serialize `{type: "unserialized"}` or similar). But that would also get messy e.g. if for normal JS objects we end up with configuration to include non-enumerable properties, because we wouldn't want to include them as "unserialized".

lutien:
Alright, thanks, James! Then instead I'll remove the issue in this PR.

whimboo:
> I think the right resolution is to always serialize for elements and never for other node types. That appears to be the existing behaviour of the spec, so I think we could just delete the issue?

Well we have cases like the following which is exactly only adding an e.g `value` field if it is not `none:

> If node value is not null set serialized["nodeValue"] to node value.

So why should `shadowRoot` be different?

jgraham:
That's already the case I believe.

jgraham (approved):

PR #406, opened 2023-04-18
Remove unused "OwnershipModel" definition

thiagowfx:
These definitions are exactly the same ("root" / "none"). It would be better to stick to a single one for consistency.

"ResultingOwnership" was picked arbitrarily. Either name would have been fine.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/406.html" title="Last updated on Apr 18, 2023, 9:13 AM UTC (af10fe1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/406/e123a41...af10fe1.html" title="Last updated on Apr 18, 2023, 9:13 AM UTC (af10fe1)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1167,10 +1167,6 @@ Issue: Define how this works.
 Each EMCAScript [=Realm=] has a corresponding <dfn>handle object map</dfn>. This is a
 strong [=/map=] from handle ids to their corresponding objects.
 
-<pre class="cddl remote-cddl">
-OwnershipModel = "root" / "none";
-</pre>
-
 <dfn>RemoteReference</dfn> is either a <code>RemoteObjectReference</code>
 representing a remote reference to an existing ECMAScript object in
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1168,7 +1168,7 @@ Each EMCAScript [=Realm=] has a corresponding <dfn>handle object map</dfn>. This
 strong [=/map=] from handle ids to their corresponding objects.
 
 <pre class="cddl remote-cddl">
-OwnershipModel = "root" / "none";
+ResultingOwnership = "root" / "none";
```
jgraham, line 1171: This definition appears to be unused and can be removed.

jgraham (changes requested):

index.bs
thiagowfx: That works too, will do.

jgraham (approved):

PR #407, opened 2023-04-18
Move "Common data types" to the script module 

jgraham:
Move "Common data types" to the script module
    
The Common Data Types section was a bit of a relic from the early organisation of the spec, before we clearly divided things into modules and namespaced all the value types.
    
Since all the types defined in that section are related to the representation of ECMAScript types in the protocol, it makes sense to move the definitions into the script module, and so ensure that only the top-level protocol definitions are unnamespaced.


jgraham:
This change is just editorial, but it would be nice to land it relatively quickly since this kind of refactor has a high chance of annoying merge conflicts.

jgraham:
(also the diff is relatively hard to read, but it's almost only moving stuff around, only a few small editorial changes to break out some of the concepts into their own sections).

index.bs
```diff
@@ -1160,4298 +1160,4345 @@ access to that data in a {{Window}} global.
...
-   true, otherwise let it be false.
+</div>
 
-1. Let |headers| be an empty list.
+<div algorithm>
+To <dfn>deserialize shared reference</dfn> given |shared reference|, |realm| and
+|session|:
 
-1. Let |mime type| be the [=essence=] of the [=computed mime type=] for |response|.
+1. Assert |shared reference| matches the <code>SharedReference</code> production.
```
juliandescottes, line 4763:
```suggestion
1. Assert |shared reference| matches the <code>script.SharedReference</code> production.
```

juliandescottes (commented):
Could not really review the diff, just some comments based on reading the spec preview. Otherwise this looks good to me. Leaving the review flag in case @whimboo or @sadym-chromium want to comment.

index.bs
```diff
@@ -1160,4298 +1160,4345 @@ access to that data in a {{Window}} global.
...
+       property set to |handle id| if it's not null, or omitted otherwise, and the
+       value set to |serialized|.
 
-    1. For each |argument| in |arguments|:
+    <dt>|value| has a \[[MapData]] [=internal slot=]
+    <dd>
 
-      1. Let |channel| be [=create a channel=] with |session|, |realm| and
-         |argument|.
+    1. Let |remote value| be a [=/ma
```
jgraham, line 5497:
```suggestion
is usable in any realm (including [=Sandbox Realms=]).
```

index.bs
```diff
@@ -1160,4298 +1160,4345 @@ access to that data in a {{Window}} global.
...
+   [=deserialize remote reference=] of given |local protocol value|, |realm| and
+   |session|.
 
-1. Let |request data| be the result of [=get the request data=] with |request|.
+1. If |local protocol value| matches the [=script.PrimitiveProtocolValue=] production, return
+   [=deserialize primitive protocol value=] with |local protocol value|.
 
-<!-- TODO: update this to "[=request/navigation id=] once the fetch parts land-->
-1. Let |navigation| be |request|'s [=navigation id=].
+1. If |local protocol value| matches the <code>script.ChannelValue</code> production,
```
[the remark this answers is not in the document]
jgraham: Yeah, I noticed that too. I think the problem is that using `<dfn>type</dfn>` causes an error if we don't ever reference `type` elsewhere, which quite commonly happens if it's only used as part of the CDDL rather than referred to in prose. I guess in an ideal world Bikeshed would parse the CDDL and automatically create linkable references for us. In the meantime I think it would be possible to turn all the type definitions into actual linkable definitions, and suppress the error if we don't actually link to it, but that seems like something we should do in a different PR.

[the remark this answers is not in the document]
juliandescottes: yes, definitely not for this PR, thanks for the input!

index.bs
```diff
@@ -1160,4298 +1160,4345 @@ access to that data in a {{Window}} global.
...
+  </dl>
 
-1. Let |domain| be |stored cookie|'s domain field.
+</div>
 
-1. Let |path| be |stored cookie|'s path field.
 
-1. Let |expires| be |stored cookie|'s expires field represented as a unix
-   timestamp, if set, or null otherwise.
+#### The script.PrimitiveProtocolValue Type #### {#type-script-PrimitiveProtocolValue}
```
juliandescottes, line 4282: nit: alphabetical order (if we care about this here...)
- script.PrimitiveProtocolValue should be after script.PreloadScript
- script.ResultOwnership should be before script.SharedId

juliandescottes (approved):
I imagine it's better to avoid too much bitrot for this patch so I'm approving now. @whimboo @sadym-chromium feel free to take a look.

sadym-chromium (approved):
`+3,061 −3,041 `. I guess we can trust you and raise concerns later if something is missed, as it's not feasible to review )

Issue #373, opened 2023-03-01
Add `browsingContext` parameter to `script.addPreloadScript`


sadym-chromium:
Currently, there is no way for the ["preload script"](https://w3c.github.io/webdriver-bidi/#preload-script) to be run only in specific BrowsingContext, and there is not way for the script to determine in which browsing context it is running, which makes some scenarios quite hard to implement. E.g. scenario "setup mutation observer on a page I'm about to navigate".

I suggest adding either `context: browsingContext.BrowsingContext` or `?contexts: [*browsingContext.BrowsingContext]` parameter to the [`script.addPreloadScript`](https://w3c.github.io/webdriver-bidi/#command-script-addPreloadScript) command.

juliandescottes:
I think we have the same constraint here as for events, meaning we can't set a preload script for non top-level contexts. The context id for a non top level browsing context should not survive navigation, so you can't schedule a preload script specifically for let's say an iframe.

With this in mind, if we pass a context (top-level) to `addPreloadScript`, should it also load the script in all the child contexts?

sadym-chromium:
Yes, all non-top-level browsing contexts inherits it from the top-level parent.

juliandescottes:
To clarify: the context id will be preserved if we navigate the frame only, but not if we navigate the top level document. I was mixing up the two scenarios. 

Nevertheless it seems practical to inherit preload scripts in non-top-level contexts, because top level navigations are the most common scenario.

sadym-chromium:
> To clarify: the context id will be preserved if we navigate the frame only, but not if we navigate the top level document. I was mixing up the two scenarios.
> 
> Nevertheless it seems practical to inherit preload scripts in non-top-level contexts, because top level navigations are the most common scenario.

My understanding is: Navigating top level document keeps the `browsingContext` (tab), but changes `navigable` (specific document).

jgraham:
No, the `navigable` is the top-level tab or iframe, the `Document` is the document and the `browsingContext` isn't quite either (it's close to navigable, but changes for some cross-origin navigations).

I think inheriting to descendents by default makes sense. I'm not sure if only allowing the context to be a top-level context makes sense; if you know you're going to navigate an iframe having to add the script to the top-level frame seems more complex (note that I don't think we have the same problem as we have for event subscriptions because we don't have a generic unsubscribe, but a handle to remove a specific preload script, which removes it from all the contexts in which it was added).

thiagowfx:
[...]

PR #411, opened 2023-04-24
Add browsing context to add preload script

thiagowfx:
Bug: #373

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/411.html" title="Last updated on Apr 26, 2023, 12:56 PM UTC (400dd7f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/411/a81ced6...400dd7f.html" title="Last updated on Apr 26, 2023, 12:56 PM UTC (400dd7f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4048,8 +4048,8 @@ TODO: Extend this to scripts in other kinds of realms.
 A [=BiDi session=] has a <dfn>preload script map</dfn> which is a [=/map=] in
 which the keys are [[!RFC4122|UUID]]s, and the values are [=structs=] with an <a
 for=struct>item</a> named <code>function declaration</code>, which is a string,
-<code>arguments</code>, and an item named <code>sandbox</code> which is a string
-or null.
+<code>arguments</code>, an item named <code>sandbox</code> which is a string
+or null, and an item named <code>context</code> which is a string or null.
 
 Note: If executing a [=preload script=] fails, either due to a syntax error, or
 a runtime exception, an [[ECMAScript]] exception is reported in the realm in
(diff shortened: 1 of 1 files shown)


thiagowfx:
@sadym-chromium Took a initial stab at this, but note this is my first spec contrib so there could be some gaps.

I believe the only part missing is to actually use the `context` in `preload script map`, but I am not sure how to word this. Could you help?

thiagowfx:
Also CC'ing @juliandescottes, giving participation in https://github.com/w3c/webdriver-bidi/issues/373.

index.bs
```diff
@@ -4090,6 +4090,13 @@ To <dfn export>run WebDriver BiDi preload scripts</dfn> given |environment setti
 
     1. Let |function declaration| be |preload script|'s <code>function declaration</code>.
 
+    1. Let |context id| be |preload script|'s <code>context</code>.
+
+    1. Let |context| be the result of [=trying=] to [=get a browsing context=]
+       with |context id|.
+
+    // TODO: Need to use |context| somewhere.
```
jgraham, line 4098: You need to check something like `if |context| is not null and |context|`s [=active window=] is not |environment setting|'s [=relevant global object=], then continue'. But you want to move this check up to before we try to deserialize arguments, so we aren't expected to throw an error in the case where argument deserialization would fail but the script wouldn't run anyway.

jgraham (changes requested):

index.bs
thiagowfx: I am not familiar enough with the precision of the spec at this point, so I will put this on hold. If someone wants to continue from here, feel free to take over this PR.

@sadym-chromium FYI.

thiagowfx:
Closing as there is no prioritization for this in the roadmap. Saving a back reference in https://github.com/w3c/webdriver-bidi/issues/373 in case someone picks this up in the future.

PR #414, opened 2023-04-26
make a few script improvements

thiagowfx:


jgraham:
Happy to take the script improvements. I'm pretty skeptical about the Makefile however; I tend to think they turn into a maintenance hazard as the few people familiar with Make decide it's a good place to add actual functionality which becomes difficult to fix when there's some other change.

In this specific case I also note that typically what you want locally is to run `bikeshed serve`, which isn't what the Makefile does (since it's reusing the CI script that doesn't want to run the server).

jgraham (approved):

Issue #403, opened 2023-04-12
Extend existing roadmap


sadym-chromium:
As we are quite close to specify all the scenarios from the roadmap, we have to extend it. This are scenarios we think makes sense to work on:
1. [Device emulation](https://github.com/w3c/webdriver-bidi/pull/420): https://github.com/w3c/webdriver-bidi/pull/420
2. [Sandbox session](https://github.com/w3c/webdriver-bidi/pull/422): https://github.com/w3c/webdriver-bidi/pull/422
3. ~bfcache/pre-rendering~ Can it be tested via network events?
4. [Support web extensions](https://github.com/w3c/webdriver-bidi/pull/421): https://github.com/w3c/webdriver-bidi/pull/421)
5. ~Support workers~ Can it be tested via [network interception?](https://github.com/w3c/webdriver-bidi/pull/429)
6. ~2FA~ subset of [HTTP authentication](https://github.com/w3c/webdriver-bidi/blob/master/roadmap.md#http-authentication)
7. [Cookies](https://github.com/w3c/webdriver-bidi/pull/423): https://github.com/w3c/webdriver-bidi/pull/423

Anything else we should consider adding there?

jgraham:
Would be nice to get input here from people using the protocol in the wild rather than just guessing. Obviously @mathiasbynens and the other Puppeteer people are good candidates, but @shs96c @jimevans and others from Selenium should pitch in, as well as other users like @CanadaHonk.

From a spec point of view I think workers are already supported; obviously implementation wise it's a bit different.

I agree that various kinds of device emulation and overrides (media queries, timezone, etc.) seem like obvious next steps, because I understand that those are widely used in existing clients. I think there are also a couple of more basic things missing from the existing roadmap:
* Sizing browser windows
* Cookies

Also for "incognito" mode, I think we need some clear thought around requirements. Is it just about opening a new incognito window? Firefox probably wants to also be able to expose containers, since they can provide storage isolation between different tests.

CanadaHonk:
Personally, #157 is a major blocker as otherwise there is no (easy) way to have bidirectional communication between a client and a page. Also #119 seems like something which might be often used / handy.

jimevans:
Window sizing/positioning is important for Selenium, as is cookie manipulation (the latter is more important to implement first than the former). These are the biggest things missing from the roadmap as it currently stands from the Selenium perspective.

[...]

PR #422, opened 2023-05-15
Incognito mode

sadym-chromium:
#403

### DIFF
--- a/roadmap.md
+++ b/roadmap.md
@@ -90,3 +90,10 @@ This scenario loads a web page and prints it as a PDF. In spec terms, this invol
 
 - [x] Some items from the previous milestones
 - [x] [Printing to PDF as Base64-encoded string](https://w3c.github.io/webdriver-bidi/#command-browsingContext-print)
+
+### Incognito mode
+
+This scenario creates a session with browser in incognito mode. In spec terms, this involves:
+
+- [x] Some items from the previous milestones
+- [ ] Create session with isolation and deleting of cookies, site data, and browsing history.


roadmap.md
```diff
@@ -90,3 +90,10 @@ This scenario loads a web page and prints it as a PDF. In spec terms, this invol
 
 - [x] Some items from the previous milestones
 - [x] [Printing to PDF as Base64-encoded string](https://w3c.github.io/webdriver-bidi/#command-browsingContext-print)
+
+### Incognito mode
+
+This scenario creates a session with browser in incognito mode. In spec terms, this involves:
```
jgraham, line 96: Note that "incognito mode" is a Chromism, and the equivalent feature may work slightly differently in different browsers. We need to be clear about which guarantees we are actually intending to offer here.

sadym-chromium: Renamed to "sandbox" mode to avoid vendor-specific naming.

whimboo: I don't think that `Sandbox` mode is good either given that this conflicts with the Sandbox definition of the script module.

sadym-chromium:
> We need to be clear about which guarantees we are actually intending to offer here.

The main observable difference between Chromium, Firefox and Safari is in cookies. Specifically in blocking/allowing third-party cookies. We can vaguely describe it as "other vendor-specific cookies limitations like can be applied".

roadmap.md
sadym-chromium: I'm open for any other name )

sadym-chromium: "Isolated" mode?

thiagowfx: How about "Extend roadmap: Incognito/Private mode"

(as per https://support.mozilla.org/en-US/kb/private-browsing-use-firefox-without-history, and Safari also uses the "Private" terminology)

PR #423, opened 2023-05-15
Extend roadmap: cookies

sadym-chromium:


jgraham:
The big practical question for cookies is how to handle double keyed cookies (maybe more generally multi-keyed cookies) i.e. situations where the cookies sent depend both on the origin of the request and the origin of the request source. Current WebDriver entirely punts on this issue and just says "return the entire content of the cookie store", without any way to tell which cookies will actually apply to a given request, or set the key for cookies.

sadym-chromium:
TODO: allow cross-origin cookies access

mathiasbynens:
> The big practical question for cookies is how to handle double keyed cookies (maybe more generally multi-keyed cookies) i.e. situations where the cookies sent depend both on the origin of the request and the origin of the request source. 

For my own understanding: is this only about `SameSite` cookies, or does double-keying apply to more scenarios?

@OrKoN How does CDP handle this, if at all? CDP’s `Network.getCookies` seems to only accept a list of URLs, returning all applicable cookies for those URLs — allowing to specify only one out of two “keys”. https://chromedevtools.github.io/devtools-protocol/tot/Network/#method-getCookies

OrKoN:
@mathiasbynens we should probably be using https://chromedevtools.github.io/devtools-protocol/tot/Storage/#method-getCookies which has partition keys (I assume that it is used for double-keying?)

jgraham:
See https://hacks.mozilla.org/2021/02/introducing-state-partitioning/ for example.

Basically it means that cookies are keyed not just on the destination origin (e.g. `example.org` only gets `example.org` cookies), but also on the source origin i.e. a request from `example.net` for an `example.org` resource doesn't see cookies for `example.org` that were set as part of a request from `example.com` (or any other origin).

Also, at least in Gecko, we have the concept of containers, which are used as an additional key for storage. It's unclear to me how we want this to work for WebDriver. For example if we want a feature to partition storage between different tests, using containers for that partitioning makes sense. But then should getting cookies return cookies for all containers, or require specifying which container to use? This might also apply to WebKit/Safari's profiles; I'm not sure (it doesn't apply to Firefox profiles because there's a 1:1 relationship between a browser process and a profile, but 1:many between a browser process and containers).

lolaodelola:
Hi @sadym-chromium, just wanted to know if you're waiting for anything in particular to get this merged? Since work on cookies is being discussed & definitely going to happen, it'd be good to have this in the main roadmap.

roadmap.md
```diff
@@ -90,3 +90,12 @@ This scenario loads a web page and prints it as a PDF. In spec terms, this invol
...
 - [x] [Printing to PDF as Base64-encoded string](https://w3c.github.io/webdriver-bidi/#command-browsingContext-print)
+
+### Cookies
+
+This scenario loads a web page and gets and sets its cookies. In spec terms, this involves:
+
+- [x] Some items from the previous milestones
+- [ ] Get cookies
+- [ ] Set cookies
+- [ ] Delete cookies
```
mathiasbynens, line 101:
```suggestion
- [ ] Getting cookies
- [ ] Setting cookies
- [ ] Deleting cookies
```

mathiasbynens (approved):

mathiasbynens:
I’ve marked the PR as ready for review now. I agree it would make sense to merge this _in some form_, since we already agreed that this is important and happening.

roadmap.md
```diff
@@ -90,3 +90,12 @@ This scenario loads a web page and prints it as a PDF. In spec terms, this invol
 
 - [x] Some items from the previous milestones
 - [x] [Printing to PDF as Base64-encoded string](https://w3c.github.io/webdriver-bidi/#command-browsingContext-print)
+
+### Cookies
+
+This scenario loads a web page and gets and sets its cookies. In spec terms, this involves:
```
OrKoN, line 96:
```suggestion
This scenario loads a web page, reads the cookies set by the server and the page, updates and deletes some of the cookies and triggers an action to verify that modified cookies are processed correctly. In spec terms, this involves:
```

OrKoN (approved):
LGTM % suggestion to make scenario a bit more specific.

roadmap.md
```diff
@@ -93,7 +93,7 @@ This scenario loads a web page and prints it as a PDF. In spec terms, this invol
 
 ### Cookies
 
-This scenario loads a web page and gets and sets its cookies. In spec terms, this involves:
+This scenario loads a web page, reads the cookies set by the server and the page, updates and deletes some of the cookies, and triggers an action to verify that modified cookies are processed correctly. In spec terms, this involves:
```
jgraham, line 96: This doesn't cover cookies for pages that aren't loaded. My understanding is that the need to load a page before interacting with its cookies is considered a significant limitation for WebDriver. It's also the source of much of the complexity here (since having a page loaded in a specific context answers a lot of the questions around keying, given that a particular browsing context will have a particular storage key).

jgraham (changes requested):

roadmap.md
```diff
@@ -90,3 +90,12 @@ This scenario loads a web page and prints it as a PDF. In spec terms, this invol
 
 - [x] Some items from the previous milestones
 - [x] [Printing to PDF as Base64-encoded string](https://w3c.github.io/webdriver-bidi/#command-browsingContext-print)
+
+### Cookies
+
+This scenario loads a web page, reads the cookies set by the server and the page, updates and deletes some of the cookies, and triggers an action to verify that modified cookies are processed correctly. In spec terms, this involves:
```
OrKoN, line 96:
```suggestion
This scenario starts the browser, reads cookies available in the browser's storage, updates and deletes some of the cookies, and navigates to a page to verify that cookies are processed correctly. In spec terms, this involves:
```

jgraham, line 96: Maybe something like
```suggestion
This scenario sets cookies that will be used when loading a web page, and its subresources, loads that page, and verifies that the correct cookies were sent with the requests. It also allows inspecting the cookies set by the responses, and clearing cookies to reset state. In spec terms, this involves:
```
although strictly that phrasing only requires _write_ access to the cookie jar when the page hasn't yet been loaded.

roadmap.md
mathiasbynens: Is this addressed by the edit that happened after this comment was posted?

jgraham (approved):

roadmap.md
```diff
@@ -90,3 +90,12 @@ This scenario loads a web page and prints it as a PDF. In spec terms, this invol
 
 - [x] Some items from the previous milestones
 - [x] [Printing to PDF as Base64-encoded string](https://w3c.github.io/webdriver-bidi/#command-browsingContext-print)
+
+### Cookies
+
+This scenario sets cookies that will be used when loading a web page, and its subresources, loads that page, and verifies that the correct cookies were sent with the requests. It also allows inspecting the cookies set by the responses, and clearing cookies to reset state. In spec terms, this involves:
```
mathiasbynens, line 96:
```suggestion
This scenario sets cookies to be used when loading a web page and its subresources, loads that page, and verifies that the correct cookies were sent with the requests. It also inspects the cookies set by the responses, and clears all cookies to reset the state. In spec terms, this involves:
```

PR #425, opened 2023-05-16
Add browsingContext.setViewport

OrKoN:
This PR proposes a command to emulate the viewport of the top-level browsing context.

***
<a href="https://pr-preview.s3.amazonaws.com/OrKoN/webdriver-bidi/pull/425.html" title="Last updated on Jun 15, 2023, 2:19 PM UTC (7b15d58)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/425/88b5c82...OrKoN:7b15d58.html" title="Last updated on Jun 15, 2023, 2:19 PM UTC (7b15d58)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -291,6 +291,7 @@ Command = {
 CommandData = (
   BrowserCommand //
   BrowsingContextCommand //
+  EmulationCommand //
   InputCommand //
   NetworkCommand //
   ScriptCommand //
@@ -3209,6 +3210,76 @@ opened</dfn> steps given |window|, |type| and |message|.
    given "<code>browsingContext.userPromptOpened</code>" and |related browsing contexts|:
 
   1. [=Emit an event=] with |session| and |body|.
+
+</div>
+
+## Emulation ## {#module-emulation}
+
+The <dfn export for=modules>emulation</dfn> module contains functionality for
+emulating device-dependent and system-dependent parameters.
+
+### Definition ### {#module-emulation-definition}
+
+[=remote end definition=]
+
+<pre class="cddl remote-cddl">
+
+EmulationCommand = (
+  emulation.SetViewport
+)
+</pre>
+
+### Types ### {#module-emulation-types}
+
+### Commands ### {#module-emulation-commands}
+
+#### The emulation.setViewport Command ####  {#command-emulation-setViewport}
+
+The <dfn export for=commands>emulation.setViewport</dfn> command turns the emulation of the specified viewport on.
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl remote-cddl">
+      emulation.SetViewport = {
+        method: "emulation.setViewport",
+        params: emulation.SetViewportParameters
+      };
+
+      emulation.SetViewportParameters = {
+        context: browsingContext.BrowsingContext,
+        ? viewport: emulation.Viewport
+      }
+
+      emulation.Viewport = {
+        ? width: js-uint,
+        ? height: js-uint,
+      }
+    </pre>
+   </dd>
+   <dt>Return Type</dt>
+   <dd>
+    <pre class="cddl">
+     EmptyResult
+    </pre>
+   </dd>
+</dl>
+
+<div algorithm="remote end steps for emulation.setViewport">
+
+The [=remote end steps=] with |command parameters| are:
+
+1. Let |context id| be the value of the <code>context</code> field of
+   |command parameters|.
+
+1. Let |context| be the result of [=trying=] to [=get a browsing context=]
+   with |context id|.
+
+1. Let |viewport| be value of the <code>viewport</code> field of |command parameters| in [[webdriver#dfn-css-pixels]].
+
+1. Apply |viewport| as the actual vieport as defined in [[css-viewport#the-viewport]] to the |context| and all current and future sub-contexts that have a concept of a viewport.
+
+1. Return [=success=] with data null.
 
 </div>
 
--- a/index.bs
+++ b/index.bs
@@ -3281,9 +3281,10 @@ The [=remote end steps=] with |command parameters| are:
     1. Let |width| be the <code>width</code> field of |viewport parameter| in CSS pixels.
     1. Let |height| be the <code>height</code> field of |viewport parameter| in CSS pixels.
     1. If the |width| or |height| is outside of the implementation-specific or runtime constraints, return an [=error=] with the [=error code=] [=unsupported operation=].
-    1. Change the layout viewport of the |context| to be |width| pixels wide and |height| pixels high and run the [[cssom-view-1#resizing-viewports]] steps.
+    1. Set the layout [[css2#viewport]] of the |context| to be |width| pixels wide and |height| pixels high.
 1. Else:
-    1. Turn off the viewport emulation on the |context|, reset it to the original value and run the [[cssom-view-1#resizing-viewports]] steps.
+    1. Set the layout [[css2#viewport]] of |context| to the implementation-defined default layout viewport for |context|.
+1. Run the [[cssom-view-1#resizing-viewports]] steps.
 1. Return [=success=] with data null.
 
 </div>
--- a/index.bs
+++ b/index.bs
@@ -291,7 +291,6 @@ Command = {
 CommandData = (
   BrowserCommand //
   BrowsingContextCommand //
-  EmulationCommand //
   InputCommand //
   NetworkCommand //
   ScriptCommand //
@@ -2730,6 +2729,62 @@ The [=remote end steps=] with |command parameters| are:
 
 </div>
 
+#### The browsingContext.setViewport Command ####  {#command-browsingContext-setViewport}
+
+The <dfn export for=commands>browsingContext.setViewport</dfn> command  specific viewport characteristics (e.g. viewport width and viewport height) on the given top-level browsing context.
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl remote-cddl">
+      browsingContext.SetViewport = {
+        method: "browsingContext.setViewport",
+        params: browsingContext.SetViewportParameters
+      };
+
+      browsingContext.SetViewportParameters = {
+        context: browsingContext.BrowsingContext,
+        viewport: browsingContext.Viewport / null,
+      }
+
+      browsingContext.Viewport = {
+        width: js-uint,
+        height: js-uint,
+      }
+    </pre>
+   </dd>
+   <dt>Result Type</dt>
+   <dd>
+    <pre class="cddl">
+     EmptyResult
+    </pre>
+   </dd>
+</dl>
+
+<div algorithm="remote end steps for browsingContext.setViewport">
+
+The [=remote end steps=] with |command parameters| are:
+
+1. Let |context id| be the value of the |command parameters|["<code>context</code>"] field.
+
+1. Let |context| be the result of [=trying=] to [=get a browsing context=] with |context id|.
+
+1. If |context| is not a [=top-level browsing context=], return [=error=] with [=error code=] [=invalid argument=].
+
+1. Let |viewport parameter| be the |command parameters|["<code>viewport</code>"] field.
+
+1. If |viewport parameter| is not null:
+    1. Let |width| be the |viewport parameter|'s["<code>width</code>""] field in CSS pixels.
+    1. Let |height| be the |viewport parameter|'s["<code>height</code>""] field in CSS pixels.
+    1. If the |width| or |height| is outside of the implementation-specific or runtime constraints, return an [=error=] with the [=error code=] [=unsupported operation=].
+    1. Set the layout [[css2#viewport]] of the |context| to be |width| pixels wide and |height| pixels high.
+1. Else:
+    1. Set the layout [[css2#viewport]] of |context| to the implementation-defined default layout viewport for |context|.
+1. Run the [[cssom-view-1#resizing-viewports]] steps.
+1. Return [=success=] with data null.
+
+</div>
+
 
 ### Events ### {#module-contexts-events}
 
@@ -3213,81 +3268,6 @@ opened</dfn> steps given |window|, |type| and |message|.
 
 </div>
 
-## Emulation ## {#module-emulation}
-
-The <dfn export for=modules>emulation</dfn> module contains functionality for
-emulating device-dependent and system-dependent parameters.
-
-### Definition ### {#module-emulation-definition}
-
-[=remote end definition=]
-
-<pre class="cddl remote-cddl">
-
-EmulationCommand = (
-  emulation.SetViewport
-)
-</pre>
-
-### Types ### {#module-emulation-types}
-
-### Commands ### {#module-emulation-commands}
-
-#### The emulation.setViewport Command ####  {#command-emulation-setViewport}
-
-The <dfn export for=commands>emulation.setViewport</dfn> command emulates specific viewport characteristics (e.g. viewport width and viewport height) on the given top-level browsing context.
-
-<dl>
-   <dt>Command Type</dt>
-   <dd>
-    <pre class="cddl remote-cddl">
-      emulation.SetViewport = {
-        method: "emulation.setViewport",
-        params: emulation.SetViewportParameters
-      };
-
-      emulation.SetViewportParameters = {
-        context: browsingContext.BrowsingContext,
-        viewport: emulation.Viewport / null,
-      }
-
-      emulation.Viewport = {
-        width: js-uint,
-        height: js-uint,
-      }
-    </pre>
-   </dd>
-   <dt>Result Type</dt>
-   <dd>
-    <pre class="cddl">
-     EmptyResult
-    </pre>
-   </dd>
-</dl>
-
-<div algorithm="remote end steps for emulation.setViewport">
-
-The [=remote end steps=] with |command parameters| are:
-
-1. Let |context id| be the value of the |command parameters|["<code>context</code>"] field.
-
-1. Let |context| be the result of [=trying=] to [=get a browsing context=] with |context id|.
-
-1. If |context| is not a [=top-level browsing context=], return [=error=] with [=error code=] [=invalid argument=].
-
-1. Let |viewport parameter| be the |command parameters|["<code>viewport</code>"] field.
-
-1. If |viewport parameter| is not null:
-    1. Let |width| be the |viewport parameter|'s["<code>width</code>""] field in CSS pixels.
-    1. Let |height| be the |viewport parameter|'s["<code>height</code>""] field in CSS pixels.
-    1. If the |width| or |height| is outside of the implementation-specific or runtime constraints, return an [=error=] with the [=error code=] [=unsupported operation=].
-    1. Set the layout [[css2#viewport]] of the |context| to be |width| pixels wide and |height| pixels high.
-1. Else:
-    1. Set the layout [[css2#viewport]] of |context| to the implementation-defined default layout viewport for |context|.
-1. Run the [[cssom-view-1#resizing-viewports]] steps.
-1. Return [=success=] with data null.
-
-</div>
 
 ## The network Module ## {#module-network}
 
--- a/index.bs
+++ b/index.bs
@@ -3274,7 +3274,7 @@ The [=remote end steps=] with |command parameters| are:
 
 1. Let |context| be the result of [=trying=] to [=get a browsing context=] with |context id|.
 
-1. If |context| is not a [=top-level browsing context=], return [=error=] with [=error code=] [=unsupported operation=].
+1. If |context| is not a [=top-level browsing context=], return [=error=] with [=error code=] [=invalid argument=].
 
 1. If |command parameters| [=map/contains=] "<code>viewport</code>", then:
     1. Let |viewport parameters| be the <code>viewport</code> attribute of |command parameters|.
--- a/index.bs
+++ b/index.bs
@@ -2774,8 +2774,8 @@ The [=remote end steps=] with |command parameters| are:
 1. Let |viewport parameter| be the |command parameters|["<code>viewport</code>"] field.
 
 1. If |viewport parameter| is not null:
-    1. Let |width| be the |viewport parameter|["<code>width</code>"]`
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -7260,6 +7260,76 @@ The [=remote end steps=] given |session|, and |command parameters| are:
...
+The <dfn export for=modules>emulation</dfn> module contains functionality for
+emulating device-dependent and system-dependent parameters.
+
+### Definition ### {#module-emulation-definition}
+
+[=remote end definition=]
+
+<pre class="cddl remote-cddl">
+
+EmulationCommand = (
```
jgraham, line 7274: If we're adding this as a new module, then you need to add it to the top-level `Command` definition.

index.bs
```diff
@@ -7260,6 +7260,76 @@ The [=remote end steps=] given |session|, and |command parameters| are:
...
+     EmptyResult
+    </pre>
+   </dd>
+</dl>
+
+<div algorithm="remote end steps for emulation.setViewport">
+
+The [=remote end steps=] with |command parameters| are:
+
+1. Let |context id| be the value of the <code>context</code> field of
```
jgraham, line 7319: Is this expected to work for non-top-level browsing contexts / navigables?

index.bs
```diff
@@ -7260,6 +7260,76 @@ The [=remote end steps=] given |session|, and |command parameters| are:
...
+   |command parameters|.
+
+1. Let |context| be the result of [=trying=] to [=get a browsing context=]
+   with |context id|.
+
+1. Let |viewport| be value of the <code>viewport</code> field of |command parameters| in [[webdriver#dfn-css-pixels]].
+
+1. Apply |viewport| as the actual vieport as defined in [[css-viewport#the-viewport]] to the |context| and all current and future sub-contexts that have a concept of a viewport.
```
jgraham, line 7327: So with the understanding that the viewport spec is still in flux; I don't think this is unambiguous enough to be confident of interoperable implementations. Which viewport (initial / actual / layout) is being adjusted? What does it mean to set it to a Infra Map (you probably want to set individual fields instead). What's a "sub-context"? It seems inevitable that this is inherited into iframes, but is the idea that it's also inherited into auxillary browsing contexts (e.g. from `window.open`)? What about random new browsing contexts (e.g. from the `browsingContext.create` command)? Should there be a way to set the viewport properties globally so it applies to all contexts?

index.bs
```diff
@@ -7260,6 +7260,76 @@ The [=remote end steps=] given |session|, and |command parameters| are:
...
+   <dd>
+    <pre class="cddl">
+     EmptyResult
+    </pre>
+   </dd>
+</dl>
+
+<div algorithm="remote end steps for emulation.setViewport">
+
+The [=remote end steps=] with |command parameters| are:
```
jgraham, line 7317: There needs to be a step here to return with unsupported operation in case the browser doesn't support changing the viewport dimensions (this seems plausible on mobile for example).

jgraham (changes requested):

index.bs
OrKoN: The PR is not ready for review yet but I can provide some answers now, thanks for the good questions and suggestions!

> Which viewport (initial / actual / layout) is being adjusted? 
Currently, I believe that actual viewport is the one that needs to adjusted.

> What does it mean to set it to a Infra Map (you probably want to set individual fields instead)

I didn't get the question about the infra map? what is an infra map?

> It seems inevitable that this is inherited into iframes, but is the idea that it's also inherited into auxillary browsing contexts (e.g. from window.open)? 

I think it should not apply to iframes because they don't really have a viewport concept (dimensions of the iframe are controlled by the layout). Auxillary browsing contexts should be excluded too. By sub-contexts that have a concept of a viewport I mean something like fenced frames, prerendered or bfcache contexts that the current context might navigate to but which exist in parallel to the current context. I am still searching for a good term and if such details really need to be defined.

> What about random new browsing contexts (e.g. from the browsingContext.create command)? Should there be a way to set the viewport properties globally so it applies to all contexts?

For now, I think there is no use case for that. Perhaps it would make sense to specify the default viewport via capabilities? For example, it can be done via CLI flag in Chrome.

ewinslow: > Perhaps it would make sense to specify the default viewport via capabilities? For example, it can be done via CLI flag in Chrome.

Can you clarify which cli flag you have in mind here? AFAIK there is no way to set viewport from cli (unless you mean window size, but that includes the browser's UI so IMHO it's not a good candidate for a cross browser API).

jgraham: Infra maps: https://infra.spec.whatwg.org/#maps It's a spec-internal key/value pair type which is how we represent command bodies in the spec prose. So in your case, you first need to check if `|command parameters| [=map/contains=] "<code>viewport</code>` and if it does `|command parameters["<code>viewport</code>"]` will be another map that (maybe) has properties named `"<code>width</code>"` and `"<code>height</code>"`. That's not an object you can work with directly; you'd need to extract the individual fields.

> By sub-contexts that have a concept of a viewport I mean something like fenced frames, prerendered or bfcache contexts that the current context might navigate to but which exist in parallel to the current context.

So this seems like a testable question: if I navigate to A, then navigate to B such that A ends up in bfcache, then set the viewport size, then go back in history to A, do I end up with the new viewport size or the previous one (I assume that for a normal navigation the viewport size won't change?). I think that's something that needs to be clear.

OrKoN: @ewinslow you are right, my memory failed me :) the flag does not exist and, in Puppeteer's case, it sets the default viewport after the page is created (`defaultViewport` is a launch argument).

index.bs
```diff
@@ -7260,6 +7260,76 @@ The [=remote end steps=] given |session|, and |command parameters| are:
...
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl remote-cddl">
+      emulation.SetViewport = {
+        method: "emulation.setViewport",
+        params: emulation.SetViewportParameters
+      };
+
+      emulation.SetViewportParameters = {
+        context: browsingContext.BrowsingContext,
```
whimboo, line 7297: As discussed last week the context might be optional so that the viewport emulation can directly be set when a new top-level browsing context opens.

index.bs
```diff
@@ -7260,6 +7260,76 @@ The [=remote end steps=] given |session|, and |command parameters| are:
...
+        ? viewport: emulation.Viewport
```
whimboo, line 7298: I cannot see that you use the `devicePixelRatio` yet. I assume you are still planning to get it added.

OrKoN: After some consideration, I think it would be better to add that separately because it does not seem to be strictly related to the viewport.

index.bs
```diff
@@ -7260,6 +7260,76 @@ The [=remote end steps=] given |session|, and |command parameters| are:
...
+)
+</pre>
+
+### Types ### {#module-emulation-types}
+
+### Commands ### {#module-emulation-commands}
+
+#### The emulation.setViewport Command ####  {#command-emulation-setViewport}
+
+The <dfn export for=commands>emulation.setViewport</dfn> command turns the emulation of the specified viewport on.
```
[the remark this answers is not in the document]
OrKoN: The current idea is that providing no values would reset the emulation. This is not captured in the algorithm yet.

index.bs
```diff
@@ -3209,6 +3210,83 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+1. Let |context| be the result of [=trying=] to [=get a browsing context=] with |context id|.
+
+1. If |context| is not a [=top-level browsing context=], return [=error=] with [=error code=] [=unsupported operation=].
+
+1. If |command parameters| [=map/contains=] "<code>viewport</code>", then:
+    1. Let |viewport parameters| be the <code>viewport</code> attribute of |command parameters|.
+    1. If |viewport parameters| [=map/contains=] "<code>width</code>" and |viewport parameters| [=map/contains=] "<code>height</code>", then:
+        1. Let |width| be the <code>width</code> attribute of |viewport parameters| in CSS pixels.
+        1. Let |height| be the <code>height</code> attribute of |viewport parameters| in CSS pixels.
+        1. Change the layout viewport of the |context| to be |width| pixels wide and |height| pixels high and run the [[cssom-view-1#resizing-viewports]] steps.
```
mathiasbynens, line 3281: I think we need to define what happens if this fails. Perhaps something like:

> If the implementation is unable to do this for any reason, then return error with error code unsupported operation.

OrKoN: Added a line about erroring if the width/height are outside of implementation-supported limits but let's see if we really need a catch-all statement.

thiagowfx (approved):

index.bs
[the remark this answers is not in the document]
thiagowfx: This is now captured in the algorithm.

index.bs
OrKoN: I have changed it to only work for top-level browsing contexts.

index.bs
OrKoN: I have updated the spec text which I believe should clarify the questions. 

> Should there be a way to set the viewport properties globally so it applies to all contexts?

Currently, we propose only changing the emulation per context.

index.bs
OrKoN: I have updated the spec: there is a mention of an error if the viewport is out of bounds (implementation-specific) but let's add a general clause for unsupported operations once we are able to confirm that some configurations cannot support it (it seems to work on mobile).

index.bs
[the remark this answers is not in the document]
OrKoN: Thanks, I have updated the spec: it gets disabled if called without the viewport parameter.

index.bs
OrKoN: For now we don't have a use case for setting the emulation for all new contexts so I have decided to keep the command accepting a single context.

index.bs
```diff
@@ -3209,6 +3210,84 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+1. Let |context| be the result of [=trying=] to [=get a browsing context=] with |context id|.
+
+1. If |context| is not a [=top-level browsing context=], return [=error=] with [=error code=] [=unsupported operation=].
+
+1. If |command parameters| [=map/contains=] "<code>viewport</code>", then:
+    1. Let |viewport parameters| be the <code>viewport</code> attribute of |command parameters|.
+    1. If |viewport parameters| [=map/contains=] "<code>width</code>" and |viewport parameters| [=map/contains=] "<code>height</code>", then:
+        1. Let |width| be the <code>width</code> attribute of |viewport parameters| in CSS pixels.
+        1. Let |height| be the <code>height</code> attribute of |viewport parameters| in CSS pixels.
+        1. If the |width| or |height| is outside of the implementation-specific bounds, return an [=error=] with the [=error code=] [=invalid argument=].
```
juliandescottes, line 3281: Given that this would only fail depending on the implementation (or maybe some runtime constraints), maybe this should be unsupported operation?

index.bs
```diff
@@ -3209,6 +3210,84 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+        1. Change the layout viewport of the |context| to be |width| pixels wide and |height| pixels high and run the [[cssom-view-1#resizing-viewports]] steps.
+    1. Else:
+        1. Return [=error=] with [=error code=] [=invalid argument=].
+1. Else:
+    1. Reset any previously emulated viewport on the |context|.
```
juliandescottes, line 3286: How does this work with several sessions? Let's say session 1 applied a viewport emulation and session 2 then applied another one. If session 2 "resets" the viewport emulation, I imagine we reset to "no emulation at all", and not just to session 1's emulation?

(and in general I wonder if the "reset" needs to be more explicit, eg which values to we reset to, but I'll let folks more used to writing spec comment on that)

juliandescottes (commented):
thanks for the PR! some fly-by comments.

index.bs
OrKoN: Good point. Since the viewport is essentially a shared global state, I think we should just allow the most recent command to "win", so if a session resets the viewport, it is reset even if some other session has set it previously. I am not sure how to express it in the spec: I have rephrased a bit, PTAL.

index.bs
```diff
@@ -3209,6 +3210,81 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+   <dd>
+    <pre class="cddl remote-cddl">
+      emulation.SetViewport = {
+        method: "emulation.setViewport",
+        params: emulation.SetViewportParameters
+      };
+
+      emulation.SetViewportParameters = {
+        context: browsingContext.BrowsingContext,
+        ? viewport: emulation.Viewport
```
jgraham, line 3248: I don't love making it optional and using the missing value default as "reset", especially given that command parameters are general assumed to be Extensible, and so sending a payload like `{"veiwport: {"width": 800, "height": 600}}` would silently reset the override. I don't know if we have existing precedent in the spec, but using an explicit value like `null` seems preferable.

index.bs
```diff
@@ -3209,6 +3210,81 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+
+1. Let |context id| be the value of the <code>context</code> field of
+   |command parameters|.
+
+1. Let |context| be the result of [=trying=] to [=get a browsing context=] with |context id|.
+
+1. If |context| is not a [=top-level browsing context=], return [=error=] with [=error code=] [=invalid argument=].
+
+1. If |command parameters| [=map/contains=] "<code>viewport</code>", then:
+    1. Let |viewport parameter| be the <code>viewport</code> field of |command parameters|.
```
jgraham, line 3277: FYI I've been switching to writing these things like `|command parameters|["<code>viewport</code>"]`, but it's not a blocker.

index.bs
```diff
@@ -3209,6 +3210,81 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+    1. Let |width| be the <code>width</code> field of |viewport parameter| in CSS pixels.
+    1. Let |height| be the <code>height</code> field of |viewport parameter| in CSS pixels.
+    1. If the |width| or |height| is outside of the implementation-specific or runtime constraints, return an [=error=] with the [=error code=] [=unsupported operation=].
+    1. Change the layout viewport of the |context| to be |width| pixels wide and |height| pixels high and run the [[cssom-view-1#resizing-viewports]] steps.
```
jgraham, line 3281: `layout viewport` should link to something here.

index.bs
```diff
@@ -3209,6 +3210,81 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+1. Else:
+    1. Turn off the viewport emulation on the |context|, reset it to the original value and run the [[cssom-view-1#resizing-viewports]] steps.
```
jgraham, line 3283: "Turn off" and "reset to original value" aren't really defined. I'm not quite sure how to phrase this, but maybe something like "Set the layout viewport of |context| to the implementation-defined default layout viewport for |context|."

index.bs
jgraham: Do we run the resizing steps even if the viewport actually doesn't change?

index.bs
```diff
@@ -3209,6 +3210,81 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+#### The emulation.setViewport Command ####  {#command-emulation-setViewport}
+
+The <dfn export for=commands>emulation.setViewport</dfn> command emulates specific viewport characteristics (e.g. viewport width and viewport height) on the given top-level browsing context.
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl remote-cddl">
+      emulation.SetViewport = {
+        method: "emulation.setViewport",
```
jgraham, line 3242: Are we also going to end up adding DPI settings to the same command? It seems like from a protocol efficiency point of view, having a single command for all the things that people might want to set at the same time might be an improvement.

jgraham (changes requested):

index.bs
jgraham: Yeah, I think the viewport size is inevitably shared global state. I can't really think of a good way to avoid conflicts. Maybe the command could fail if there's already an override set by a different session? Although that's a bit tricky as it involves cross-session state and can leave you in a state that you can't get out of (if the other session closes).

index.bs
OrKoN: I would add device pixel ratio as a separate command (if we ever define it) as it can be used to test things independently from the viewport. E.g., one could test the media query for dpr without changing the viewport. Would the ability to send commands in batch help here?

index.bs
OrKoN: I believe the check for the viewport change is part of the resizing viewport steps:

> If doc’s [viewport](https://www.w3.org/TR/cssom-view-1/#viewport) has had its width or height changed, ...

OrKoN: Added the link.

index.bs

index.bs
OrKoN: There is another edge case around other clients (that BiDi has not way of knowing about) emulating the viewport: e.g., opening DevTools might change the viewport size. I think that having the command apply the specified viewport even if there are other sessions trying to do the same is the least surprising behaviour.

index.bs
```diff
@@ -3212,6 +3213,82 @@ opened</dfn> steps given |window|, |type| and |message|.
 
 </div>
 
+## Emulation ## {#module-emulation}
```
whimboo, line 3216:
```suggestion
## The emulation module ## {#module-emulation}
```

index.bs
```diff
@@ -3212,6 +3213,82 @@ opened</dfn> steps given |window|, |type| and |message|.
...
+        ? viewport: emulation.Viewport
+      }
+
+      emulation.Viewport = {
+        width: js-uint,
+        height: js-uint,
+      }
+    </pre>
+   </dd>
+   <dt>Return Type</dt>
```
whimboo, line 3260: nit: `Result Type` is more common those days and we finally should maybe decide what to use here. There is another issue open for it.

whimboo (commented):
Thanks for working on this spec addition Alex!

Just two comments (see inline) and a question if you would also create related web-platform tests for this requested feature?

OrKoN:
@whimboo yes I will write WPT tests once we agree on the spec. I have a basic test for e2e in chromium-bidi https://github.com/GoogleChromeLabs/chromium-bidi/pull/817/files#diff-2eea1692a9f6db7cd6ce6b8404e31ab647273a0150cfc79fbaca014dacaf290a so I am thinking of adding WPT tests in a similar way.

index.bs
```diff
@@ -3212,6 +3213,82 @@ opened</dfn> steps given |window|, |type| and |message|.
 
 </div>
 
+## The emulation module ## {#module-emulation}
```
jgraham, line 3216: This is probably not worth bikeshedding, but I do find the name "emulation" confusing, since we're not emulating anything. Arguably this stuff could all be in the browsingContext module, since it all applies to a specific browsing context (similarly to the way it's all exposed on `Page` in Puppeteer).

index.bs
OrKoN: Updated

index.bs
OrKoN: Updated.

index.bs
OrKoN: I have updated it to reset the viewport with a null value. So it looks like even if we end up needing the DPI settings in this command we would be able to extend?

index.bs
OrKoN: I chose the emulation name to indicate that the commands in it change some parameters that are usually determined by the browser/page/OS on its own and are not part of regular capabilities of a browsing context (that's why it can be called "emulated" in my opinion). Also, I admit that the name was influenced by CDP :) 

> Arguably this stuff could all be in the browsingContext module

Same can be said about other commands too: e.g., Input is also scoped to the browsing context. 

I am open to moving it to the browsing context domain though if there is a consensus that it is a better place.

whimboo: This is one kind of emulation, and there are many others (see the discussion for the roadmap yesterday). There will be things like geolocation, timezone, locale, user agent, and maybe network throttling later. Maybe a dedicated module for all these different emulation features would make more sense?

OrKoN: Here is the full list of emulations supported by Puppeteer: https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/src/common/EmulationManager.ts#L73 

Some of them might not have been mentioned in the roadmap planning as they are of a lower priority generally but I see how features like these would fit into this module. I guess there are a few questions we could discuss:

1) do features like these deserve an own module or if we should add to existing modules (e.g., potential touch emulation would go into the input module and potential feature for disabling JavaScript would go into script)?
2) if we have a separate module, what should it be called?

jimevans: Is there any mileage in putting them in the `browser` module, which currently only contains the `close` command? I suppose not, since they’re intended to only apply to one context at a time.

OrKoN: Let me move it to the browsing context module then! In the end, fewer modules is probably better and we can revisit once/if new commands are being added.

jgraham (commented):
This basically looks fine to me, just a small comment.

index.bs
```diff
@@ -2730,6 +2729,62 @@ The [=remote end steps=] with |command parameters| are:
...
+1. Let |context id| be the value of the |command parameters|["<code>context</code>"] field.
+
+1. Let |context| be the result of [=trying=] to [=get a browsing context=] with |context id|.
+
+1. If |context| is not a [=top-level browsing context=], return [=error=] with [=error code=] [=invalid argument=].
+
+1. Let |viewport parameter| be the |command parameters|["<code>viewport</code>"] field.
+
+1. If |viewport parameter| is not null:
+    1. Let |width| be the |viewport parameter|'s["<code>width</code>""] field in CSS pixels.
```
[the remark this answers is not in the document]
OrKoN: Updated.

jgraham (approved):
I think this looks good, thanks!

Style wise, the spec hard wraps the text (ideally we'd like some tooling to help with this, but the world is not awash with bikeshed-flavoured-markdown autoformatters :) )

mathiasbynens (approved):

OrKoN:
I will start working on the WPT change in the near future.

juliandescottes (approved):

OrKoN:
PR for tests (1 test currently) https://github.com/web-platform-tests/wpt/pull/40584

whimboo:
Alex, if you could add tests covering invalid scenarios as well, that would be great. It's something that we have done for each and every other command and event as well. Thanks.

Issue #287, opened 2022-09-16
Add support for setting, getting, and deleting cookies


jgraham:

whimboo:
Given that we haven't had an issue for deleting cookies yet and that it actually fits into this issue I've updated the summary of the issue to reflect that as well.

whimboo:
See as well some ongoing discussions on https://github.com/w3c/webdriver-bidi/pull/423.

lolaodelola:
Hi folks, my colleague @jugglinmike  and I have been thinking about ways we could implement Cookies (get, set & delete) and have come up with three possible solutions and would love to hear what folks think.

1. We specify every way that a browser does cookies and include this in the API. So getCookie("foo", container: containerName) would get the named foo from the Firefox container named bar, and getCookie("foo", requestSourceOrigin: [example.com](http://example.com/)) would get the named cookie from the request source origin, etc.

2. We specify a mechanism to include generic properties for your requests. getCookie("foo", source: container | requestSourceOrigin) So you can add whatever you want and implementers can define what those things are and what the browser does in response. This method would mean we don’t have to get into the details of the request, source origin, etc & allow the browsers to interpret as they will.

3. We specify that when getCookie("foo") returns any and everything, for every request source origin, every container. "get" could return an array (which could be empty in some instances), while "set" could create a cookie that applied to all contexts, and "delete" would remove the named cookie in any & all contexts where it was defined.

There are pros and cons to each of these methods, so none is perfect, and I’m not particularly attached to any, so it’d be great to hear what others think.
cc: @mathiasbynens

OrKoN:
@jgraham and I were discussing cookies right now on the webdriver channel. It looks like the option 3 would be preferable and in line with the discussion there but we need to include the cookie key into the matching cookies and probably add a way to compute a cookie key for a given context.

jgraham:
So the problem is forwards compatibility.

Let's say we adopt a design where you have browser-specific partition keys, and any key the client is missing is interpreted as a wildcard.

For getting cookies that's not totally terrible, but there are still sharp edges. For example if initially cookies are only keyed on destination origin (the default), then `getCookies("name", host="example.org")` is going to always just return one cookie. But if it then adds partitioning by source origin, there may well be more than one cookie returned, and that would almost certainly lead to unexpected results in automated tests, even though the site under test would in fact be working correctly.

[...]

PR #429, opened 2023-05-18
Add network request interception

jgraham:
Add a feature for intercepting network requests.

Network request incercepts are created using the network.addIntercept command. This takes a URL pattern describing which resources to match, and a list of phases, which can be "beforeRequestSent", "responseStarted", or "authRequired". It returns a handle that can later be used with network.removeIntercept to unregister the intercept.

When a request is intercepted, the corresponding network events get their `isBLocked` property set to true, and the client has to respond with a command in order to unblock the request.

Blocking only occurs if a client is also subscribed to the relevant network events.

Each request inteception phase has a corresponding event, and one or more commands to continue the request from that point.

At the beforeRequestSent phase, the client may respond with continueRequest, which allows altering the request properties, but continues processing the request via the normal network stack. It may also respond with network.provideResponse, which allows providing a complete response body and prevents all further processing.

At the afterResponse stage, the client may respond with network.continueResponse, which allows altering the response status and headers, or network.provideResponse, which overrides the network response with the client-provided response.

At either phase the client may response with network.failRequest to cause the fetch to end with a network error.

At the authRequired phase, the client must respond with network.continueWithAuth, to handle the request for credentials.

Credentials can also be provided in the responseStarted phase (when the response has an approproate status code and authentication header), in which case those credentials will be used and a network.authRequired event will not be sent.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/429.html" title="Last updated on Aug 17, 2023, 10:14 AM UTC (88e622b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/429/e11ad3a...88e622b.html" title="Last updated on Aug 17, 2023, 10:14 AM UTC (88e622b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -398,9 +398,9 @@ A <dfn>BiDi session</dfn> is a [=/session=] which has the [=BiDi flag=]
 set to true.
 
 <div algorithm>
-The set of <dfn>active BiDi sessions</dfn> is given by:
+The list of <dfn>active BiDi sessions</dfn> is given by:
 
-1. Let |BiDi sessions| be a new [=/set=].
+1. Let |BiDi sessions| be a new [=/list=].
 
 1. For each |session| in [=active sessions=]:
 
@@ -487,25 +487,24 @@ WebDriver BiDi extends the set of [=error codes=] from [[WEBDRIVER|WebDriver]]
 with the following additional codes:
 
 <dl>
-  <dt><dfn>no such script</dfn>
-  <dd>Tried to remove an unknown [=preload script=].
-</dl>
-
-<dl>
   <dt><dfn>no such handle</dfn>
   <dd>Tried to deserialize an unknown <code>RemoteObjectReference</code>.
-</dl>
 
-<dl>
+  <dt><dfn>no such intercept</dfn>
+  <dd>Tried to remove an unknown [=network intercept=].
+
   <dt><dfn>no such node</dfn>
   <dd>Tried to deserialize an unknown <code>SharedReference</code>.
-</dl>
 
-<dl>
+  <dt><dfn>no such request</dfn>
+  <dd>Tried to continue an unknown [=/request=].
+
+  <dt><dfn>no such script</dfn>
+  <dd>Tried to remove an unknown [=preload script=].
+
   <dt><dfn>unable to close browser</dfn>
   <dd>Tried to close the browser, but failed to do so.
 </dl>
-
 
 <pre class="cddl local-cddl">
 ErrorCode = "invalid argument" /
@@ -515,7 +514,9 @@ ErrorCode = "invalid argument" /
             "no such element" /
             "no such frame" /
             "no such handle" /
+            "no such intercept" /
             "no such node" /
+            "no such request" /
             "no such script" /
             "session not created" /
             "unable to capture screen" /
@@ -640,19 +641,6 @@ To <dfn>obtain a set of event names</dfn> given an |name|:
    |events|.
 
 1. Return [=success=] with data |events|.
-
-</div>
-
-<div algorithm>
-To <dfn>emit events</dfn> given |body| and |related browsing contexts|:
-
-1. [=Assert=]: |body| has [=map/size=] 2 and [=contains=] "<code>method</code>"
-   and "<code>params</code>".
-
-1. For each |session| in the [=set of sessions for which an event is enabled=]
-   given |body|["method"] and |related browsing contexts|:
-
-  1. [=Emit an event=] with |session| and |body|.
 
 </div>
 
@@ -3557,6 +3545,13 @@ relating to network requests.
 <pre class="cddl remote-cddl">
 
 NetworkCommand = (
+  network.AddIntercept //
+  network.ContinueRequest //
+  network.ContinueResponse //
+  network.ContinueWithAuth //
+  network.FailRequest //
+  network.ProvideResponse //
+  network.RemoveIntercept
 )
 
 </pre>
@@ -3569,10 +3564,11 @@ NetworkResult = (
 )
 
 NetworkEvent = (
+    network.AuthRequired //
     network.BeforeRequestSent //
     network.FetchError //
-    network.ResponseStarted //
-    network.ResponseCompleted
+    network.ResponseCompleted //
+    network.ResponseStarted
 )
 
 </pre>
@@ -3581,17 +3577,148 @@ A [=remote end=] has a <dfn>before request sent map</dfn> which is initially an
 empty map.  It's used to track the network events for which a
 <code>network.beforeRequestSent</code> event has already been sent.
 
+### Network Intercepts ### {#network-intercepts}
+
+A <dfn>network intercept</dfn> is a mechanism to allow remote ends to intercept
+and modify network requests and responses.
+
+A [=BiDi session=] has an <dfn>intercept map</dfn> which is a [=/map=] between
+intercept id and a [=struct=] with fields <code>url patterns</code> and
+<code>phases</code> that define the properties of active network intercepts. It
+is initially empty.
+
+A [=BiDi session=] has a <dfn>blocked request map</dfn>, used to track the
+requests which are actively being blocked. It is an [=/map=] between request id
+and a [=struct=] with fields <code>request</code>, <code>phase</code>, and
+<code>response</code>. It is initially empty.
+
+<div algorithm>
+
+To <dfn>get the network intercepts</dfn> given |session|, |event|, and |request|:
+
+1. Let |session intercepts| be |session|'s [=intercept map=].
+
+1. Let |intercepts| be an empty list.
+
+1. Run the steps under the first matching condition:
+  <dl>
+    <dt>|event| is "<code>network.beforeRequestSent</code>"
+    <dd>Set |phase| to the <code>beforeRequestSent</code>
+    <dt>|event| is "<code>network.responseStarted</code>"
+    <dd>Set |phase| to "<code>responseStarted</code>".
+    <dt>|event| is "<code>network.authRequired</code>"
+    <dd>Set |phase| to "<code>authRequired</code>"».
+    <dt>|event| is "<code>network.responseCompleted</code>"
+    <dd>Return |intercepts|.
+  </dl>
+
+1. Let |url| be the result of running the [=URL serializer=] with |request|'s
+   [=request/URL=].
+
+1. For each |intercept id| → |intercept| of |session intercepts|:
+
+  1. If |intercept|'s <code>phases</code> [=set/contains=] |phase|:
+
+    1. Let |url patterns| be |intercept|'s <code>url patterns</code>.
+
+    1. If |url patterns| is [=list/empty=]:
+
+      1. [=list/Append=] |intercept id| to |intercepts|.
+
+      1. Continue.
+
+    1. For each |url pattern| in |url patterns|:
+
+      1. If [=match URL pattern=] with |url pattern| and |url|:
+
+        1. [=list/Append=] |intercept id| to |intercepts|.
+
+        1. Continue.
+
+1. Return |intercepts|.
+
+</div>
+
 ### Types ### {#module-network-types}
+
+#### The network.AuthChallenge Type #### {#type-network-AuthChallenge}
+
+<pre class="cddl local-cddl">
+network.AuthChallenge = {
+  scheme: text,
+  realm: text,
+}
+</pre>
+
+<div algorithm>
+
+To <dfn>extract challenges</dfn> given |response|:
+
+Issue: Should we include parameters other than realm?
+
+1. If |response|'s [=response/status=] is 401, let |header name| be
+   `<code>WWW-Authenticate</code>`. Otherwise if |response|'s [=response/status=] is 407, let
+   |header name| be `<code>Proxy-Authenticate</code>`. Otherwise return null.
+
+1. Let |challenges| be a new [=/list=].
+
+1. For each (|name|, |value|) in |response|'s [=response/header list=]:
+
+   Issue: as in Fetch it's unclear if this is the right way to handle multiple
+   headers, parsing issues, etc.
+
+  1. If |name| is a [=byte-case-insensitive=] match for |header name|:
+
+    1. Let |header challenges| be the result of parsing |value| into a list of
+       challenges, each consisting of a scheme and a list of parameters, each of
+       which is a [=tuple=] (name, value), according to the rules of [[!RFC9110]].
+
+    1. For each |header challenge| in |header challenges|:
+
+      1. Let |scheme| be |header challenge|'s scheme.
+
+      1. Let |realm| be the empty string.
+
+      1. For each (|param name|, |param value|) in |header challenge|'s
+         parameters:
+            1. If |param name| equals `<code>realm</code>` let |realm| be
+               [=UTF-8 decode=] |param value|.
+
+      1. Let |challenge| be a new [=/map=] matching the
+         <code>network.AuthChallenge</code> production, with the
+         <code>scheme</code> field set to |scheme| and the <code>realm</code>
+         field set to |realm|.
+
+     1. [=list/Append=] |challenge| to |challenges|.
+
+1. Return |challenges|.
+
+</div>
+
+#### The network.AuthCredentials Type #### {#type-network-AuthCredentials}
+
+<pre class="cddl remote-cddl">
+network.AuthCredentials = {
+  type: "password",
+  username: text,
+  password: text,
+}
+</pre>
+
+The <code>network.AuthCredentials</code> type represents the response to a
+request for authorization credentials.
 
 #### The network.BaseParameters Type #### {#type-network-BaseParameters}
 
 <pre class="cddl local-cddl">
 network.BaseParameters = (
     context: browsingContext.BrowsingContext / null,
+    isBlocked: bool,
     navigation: browsingContext.Navigation / null,
     redirectCount: js-uint,
     request: network.RequestData,
     timestamp: js-uint,
+    ? intercepts: [+network.Intercept]
 )
 </pre>
 
@@ -3602,7 +3729,10 @@ Issue: Consider including the `sharedId` of the document node that initiated the
 request in addition to the context.
 
 <div algorithm>
-To <dfn>get the base network event data</dfn> given |request| and |redirect count|:
+To <dfn>process a network event</dfn> given |session|, |event|, and |request|:
+
+1. Let |intercepts| be the result of [=get the network intercepts=] with
+   |session|, |event|, and |request|.
 
 1. Let |request data| be the result of [=get the request data=] with |request|.
 
@@ -3610,11 +3740,6 @@ To <dfn>get the base network event data</dfn> given |request| and |redirect coun
 1. Let |navigation| be |request|'s [=navigation id=].
 
 1. Let |context id| be null.
-
-1. If |request|'s [=request/client=] is not null, let |related browsing contexts|
-   be the result of [=get related browsing contexts=] with |request|'s
-   [=request/client=]. Otherwise let |related browsing contexts| be an empty
-   set.
 
 1. If |request|'s [=request/window=] is an [=environment settings object=]:
 
@@ -3624,15 +3749,22 @@ To <dfn>get the base network event data</dfn> given |request| and |redirect coun
      settings|' [=environment settings object/global object=], let |context id|
      be the [=browsing context id=] for that context.
 
+1. Let |redirect count| be |request|'s [=redirect count=].
+
 1. Let |timestamp| be a [=time value=] representing the current date and time in UTC.
+
+1. If |intercepts| is not [=list/empty=], let |is blocked| be true, otherwise
+   let |is blocked| be false.
 
 1. Let |params| be [=/map=] matching the <code>network.BaseParameters</code>
    production, with the <code>request</code> field set to |request data|, the
    |navigation| field set to <code>navigation</code>, the <code>context</code>
    field set to |context id|, the <code>timestamp</code> field set to
-   |timestamp|, and the <code>redirectCount</code> field set to |redirect count|.
+   |timestamp|, the <code>redirectCount</code> field set to |redirect count|, the
+   <code>isBlocked</code> field set to |is blocked|, and <code>intercepts</code>
+   field set to |intercepts| if |is blocked| is true, or omitted otherwise.
 
-1. Return (|related browsing contexts|, |params|)
+1. Return |params|
 
 </div>
 
@@ -3658,7 +3790,7 @@ type, any other data is represented in Base64-encoded form as
 <code>network.Base64Value</code>.
 
 <div algorithm>
-To <dfn ignore>deserialize protocol bytes</dfn> given |protocol bytes|:
+To <dfn>deserialize protocol bytes</dfn> given |protocol bytes|:
 
 Note: this takes bytes encoded as a <code>network.BytesValue</code> and returns
 a [=byte sequence=].
@@ -3694,24 +3826,24 @@ Note: this takes a [=byte sequence=] and returns a <code>network.BytesValue</cod
 
 [=Remote end definition=] and [=local end definition=]
 
-<pre class="cddl local-cddl">
+<pre class="cddl local-cddl remote-cddl">
 network.Cookie = {
     name: text,
     value: network.BytesValue,
     domain: text,
     path: text,
-    ? expires: js-uint,
     size: js-uint,
     httpOnly: bool,
     secure: bool,
     sameSite: "strict" / "lax" / "none",
+    ? expires: js-uint,
 };
 </pre>
 
 The <code>network.Cookie</code> type represents a cookie.
 
 <div algorithm>
-To <dfn>get a cookie</dfn> given |stored cookie|:
+To <dfn>serialize cookie</dfn> given |stored cookie|:
 
 Note: The definitions of |stored cookie|'s fields are from [[COOKIES]], except
 samesite-flag, which is from [[SAME-SITE-COOKIES]].
@@ -3748,6 +3880,35 @@ samesite-flag, which is from [[SAME-SITE-COOKIES]].
    set to |size|, the <code>httpOnly</code> field set to |http only|, the
    <code>secure</code> field set to |secure|, and the <code>sameSite</code>
    field set to |same site|.
+
+</div>
+
+#### The network.CookieHeader Type #### {#type-network-CookieHeader}
+
+[=Remote end definition=]
+
+<pre class="cddl remote-cddl">
+network.CookieHeader = {
+    name: text,
+    value: network.BytesValue,
+};
+</pre>
+
+The <code>network.CookieHeader</code> type represents the subset of cookie data
+that's in a <code>Cookie</code> request header.
+
+<div algorithm>
+To <dfn>serialize cookie header</dfn> given |protocol cookie|:
+
+1. Let |name| be [=UTF-8 encode=] |protocol cookie|["<code>name</code>"].
+
+1. Let |value| be [=deserialize protocol bytes=] with |protocol
+   cookie|["<code>value</code>"].
+
+1. Let |header value| be the [=byte sequence=] formed by concatenating |name|,
+   `<code>=</code>`, and |value|
+
+1. Return |header value|.
 
 </div>
 
@@ -3851,7 +4012,7 @@ TODO: Add service worker fields
 
 [=Remote end definition=] and [=local end definition=]
 
-<pre class="cddl local-cddl">
+<pre class="cddl local-cddl remote-cddl">
 network.Header = {
   name: text,
   value: network.BytesValue,
@@ -3872,6 +4033,18 @@ To <dfn>serialize header</dfn> given |name bytes| and |value bytes|:
 1. Return a map matching the <code>network.Header</code> production, with the
    <code>name</code> field set to |name|, and the <code>value</code> field
    set to |value|.
+
+</div>
+
+<div algorithm>
+To <dfn>deserialize header</dfn> given |protocol header|:
+
+1. Let |name| be [=UTF-8 encode=] |protocol header|["<code>name</code>"].
+
+1. Let |value| be [=deserialize protocol bytes=] with |protocol
+   header|["<code>value</code>"].
+
+1. Return a [=/header=] (|name|, |value|).
 
 </div>
 
@@ -3931,6 +4104,16 @@ To <dfn>get the initiator</dfn> given |request|:
    to |request id|.
 
 </div>
+
+#### The network.Intercept Type #### {#type-network-Intercept}
+
+[=Remote end definition=] and [=local end definition=]
+
+<pre class="cddl local-cddl remote-cddl">
+network.Intercept = text
+</pre>
+
+The <code>network.Intercept</code> type represents the id of a [=network intercept=].
 
 #### The network.Request Type #### {#type-network-Request}
 
@@ -4005,7 +4188,7 @@ To <dfn>get the request data</dfn> given |request|:
        the store can be included in a request, but user agents are free to
        impose additional constraints.
 
-       1. Append the result of [=get a cookie=] given |cookie| to |cookies|.
+       1. Append the result of [=serialize cookie=] given |cookie| to |cookies|.
 
 1. Let |timings| be [=get the fetch timings=] with |request|.
 
@@ -4060,7 +4243,8 @@ network.ResponseData = {
     bytesReceived: js-uint,
     headersSize: js-uint / null,
     bodySize: js-uint / null,
-    content: network.ResponseContent
+    content: network.ResponseContent,
+    ?authChallenge: network.AuthChallenge,
 };
 </pre>
 
@@ -4141,6 +4325,8 @@ To <dfn>get the response data</dfn> given |response|:
 
 1. Let |content| be the result of [=get the response content info=] with |response|.
 
+1. Let |auth challenges| be the result of [=extract challenges=] with |response|.
+
 1. Return a [=/map=] matching the <code>network.ResponseData</code> production,
    with the <code>url</code> field set to |url|, the <code>protocol</code> field
    set to |protocol|, the <code>status</code> field set to |status|, the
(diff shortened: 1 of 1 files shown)


jimevans (changes requested):
It looks like this PR removes all of the event definitions. Was that intentional?

**EDIT:** Correction, it looks like the markup is such that the events are not separated into the sidebar into a section under the `Network` module.

index.bs
```diff
@@ -3798,168 +3991,788 @@ To <dfn>get the response data</dfn> given |response|:
...
+
+1. Let |response| be a new [=network error=].
+
+1. TODO: Allow setting the precise kind of error
+
+1. [=Resume=] with "<code>continue request</code>", |request id|, and
+   (|response|, "<code>complete</code>").
+
+1. Return [=success=] with data null.
+
```
jimevans, line 4383: Missing `</div>`?

jimevans (changes requested):

index.bs
```diff
@@ -3798,168 +3991,788 @@ To <dfn>get the response data</dfn> given |response|:
...
+      }
 
-       network.FetchErrorParameters = {
-         network.BaseParameters,
-         errorText: text,
-       }
+      network.ContinueRequestParameters = {
+        request: network.Request,
+        ?body: network.Body,
+        ?headers: [*network.Header],
```
jimevans, line 4072: The `params` property of the `network.beforeRequestSent` event includes the properties of a `network.BaseParameters` object, one of which is the `request` property, the value of which is a `network.RequestData` object. The `network.RequestData` object has a `cookies` property which separates out the cookies of the request into a separate property. We don't allow separate setting of the cookies here. This seems asymmetrical, making it challenging to know how to manipulate add/edit/remove cookies from the request. It is possible to modify the request's cookies by creating a "Cookie" header, but that seems error prone, especially when needing to properly serialize the cookie value in the header.

whimboo:
> When a request is intercepted, the corresponding network events get their `isBLocked` property set to true, and the client has to respond with a command in order to unblock the request.
> 
> Blocking only occurs if a client is also subscribed to the relevant network events.

How would that work when multiple clients are actually connected at the same time and are subscribed with the same interception settings? Do each of them have to respond before the request is unblocked? Nevertheless I can see that different clients might override the behavior as defined by others.

thiagowfx:
Is there any chance this PR could be broken into smaller ones? I find it difficult to review with so many changes at once.

index.bs
```diff
@@ -3798,168 +3991,785 @@ To <dfn>get the response data</dfn> given |response|:
...
+1. TODO: Allow setting the precise kind of error
+
+1. [=Resume=] with "<code>continue request</code>", |request id|, and
+   (|response|, "<code>complete</code>").
+
+1. Return [=success=] with data null.
+
+</div>
+
+#### The network.provideResponse Command ####  {#command-network-provideResponse}
```
thiagowfx, line 4387: FYI only: CDP calls this [fulfillRequest](https://vanilla.aslushnikov.com/?Fetch.fulfillRequest).

No objections if you prefer to keep is as is though.

index.bs
```diff
@@ -3288,27 +3404,67 @@ To <dfn>get the base network event data</dfn> given |request| and |redirect coun
...
+  value: text,
+}
+
+network.Base64Body = {
+  type: "base64",
+  value: text
+}
+</pre>
+
+The <code>network.Body</code> type represents a request or response body,
```
thiagowfx, line 3442: @sadym-chromium @OrKoN @foolip FYI:

On CDP, there seems to be no type for "requests", only for "responses".

The relevant _response_ types are

[Network.getResponseBody](https://vanilla.aslushnikov.com/?Network.getResponseBody) or [Network.getResponseBodyForInterception](https://vanilla.aslushnikov.com/?Network.getResponseBodyForInterception) or [Fetch.getResponseBody](https://vanilla.aslushnikov.com/?Fetch.getResponseBody) or [Fetch.getResponseBodyForInterception](https://vanilla.aslushnikov.com/?Fetch.getResponseBodyForInterception).

What should we do for "requests"?

index.bs
```diff
@@ -3245,17 +3237,143 @@ A [=remote end=] has a <dfn>before request sent map</dfn> which is initially an
 empty map.  It's used to track the network events for which a
 <code>network.beforeRequestSent</code> event has already been sent.
 
+### Network Intercepts ### {#network-intercepts}
+
+A <dfn>network intercept</dfn> is a mechanism to allow remote ends to intercept
+and modify network requests and responses.
+
+A [=BiDi session=] has an <dfn>intercept map</dfn> which is initially an
```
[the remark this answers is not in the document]
thiagowfx: I suppose it's from "Intercept" to "urlPatterns" and "phases".

Can we note this right here? Otherwise the reader needs to go all the way to https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/429.html#command-network-addIntercept in order to figure it out.

index.bs
```diff
@@ -3245,17 +3237,143 @@ A [=remote end=] has a <dfn>before request sent map</dfn> which is initially an
...
+empty [=/map=].  It's used to track the active network intercepts.
+
+A [=BiDi session=] has an <dfn>blocked request map</dfn> which is initially an
```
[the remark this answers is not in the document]
thiagowfx: This seems to be from "RequestId" (string) to ("Request ID" (string),"InterceptPhase", "Response" (js-uint)), but I am not completely sure.

c.f. https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/429.html#command-network-continueRequest

Same note as in the previous map.

index.bs
```diff
@@ -3798,168 +3991,785 @@ To <dfn>get the response data</dfn> given |response|:
...
-        network.ResponseStarted = {
+      <pre class="cddl remote-cddl">
+      network.ContinueResponse = {
+        method: "network.continueResponse",
+        params: network.ContinueResponseParameters
+      }
+
+      network.ContinueResponseParameters = {
+        request: network.Request,
+        ?credentials: network.AuthCredentials,
```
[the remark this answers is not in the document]
OrKoN: https://chromedevtools.github.io/devtools-protocol/tot/Fetch/#method-continueWithAuth ?

index.bs
```diff
@@ -3798,168 +3991,785 @@ To <dfn>get the response data</dfn> given |response|:
...
-       network.BeforeRequestSentParameters = {
-         network.BaseParameters,
-         initiator: network.Initiator,
-       }
+      network.AddInterceptParameters = {
+        phases: [*network.InterceptPhase],
+        ?urlPatterns: [*text],
+      }
+
+      network.InterceptPhase = ("beforeRequestSent" /
```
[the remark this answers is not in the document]
jgraham: It's intentional that it matches the name of the actual events. Although that's longer my guess is that the consistency is more valuable than brevity (and of course clients are free to expose this with a different API).

jgraham:
> Is there any chance this PR could be broken into smaller ones? I find it difficult to review with so many changes at once.

It is a big change, but I don't know what a useful split would look like. Splitting into individual commands wouldn't show the overall structure of the API or lifecycle. I could, I suppose have split out the infra stuff that allows events to block, without actually defining any blocking events, but I don't know that would have made a large difference in this case.

index.bs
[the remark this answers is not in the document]
thiagowfx: OK

index.bs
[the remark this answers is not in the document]
thiagowfx: I see. Then we can just store it for subsequent use later on. A bit of premature optimization, but sounds reasonable.

index.bs
thiagowfx: Resolving. I misunderstood this.

index.bs
```diff
@@ -3798,168 +3991,785 @@ To <dfn>get the response data</dfn> given |response|:
...
+
+  1. Set |response|'s [=response/header list=] to |headers|.
+
+1. If |command parameters| [=map/contains=] "<code>credentials</code>":
+
+   Issue: This doesn't have a way to cancel the auth.
+
+   1. Let |credentials| be |command parameters|["<code>credentials</code>"].
+
+   1. Asssert: |credentials|{"<code>type</code>"] is "<code>password</code>".
```
whimboo, line 4236:
```suggestion
   1. Assert: |credentials|{"<code>type</code>"] is "<code>password</code>".
```

index.bs
```diff
@@ -3798,168 +3991,785 @@ To <dfn>get the response data</dfn> given |response|:
...
+1. If |command parameters| "<code>action</code> is "<code>cancel</code>", set
+   |response|'s authentication credentials <!--TODO: link--> to
+   "<code>cancelled</code>".
+
+1. If |command parameters| "<code>action</code>" is
+   "<code>provideCredentials</code>":
+
+   1. Let |credentials| be |command parameters|["<code>credentials</code>"].
+
+   1. Asssert: |credentials|{"<code>type</code>"] is "<code>password</code>".
```
whimboo, line 4325:
```suggestion
   1. Assert: |credentials|{"<code>type</code>"] is "<code>password</code>".
```

index.bs
```diff
@@ -3798,168 +3991,785 @@ To <dfn>get the response data</dfn> given |response|:
...
+
+1. Let (<var ignore>request</var>, |phase|, <var ignore>response</var>) be
+   |blocked requests|[|request id|].
+
+1. If |phase| is "<code>authRequired</code>", then return [=error=]
+   with [=error code=] [=invalid argument=].
+
+1. Let |response| be a new [=network error=].
+
+1. TODO: Allow setting the precise kind of error
```
whimboo, line 4378: If that's a follow-up maybe you can immediately file a new issue and reference it here?

whimboo (commented):
This is indeed a lot of modifications which make it hard to review. But I agree it's somewhat hard to separate out. Thanks for working on it @jgraham!! I left a couple of comments to clarify and fix. In some cases I might not have the full understanding yet how it would work, so excuse maybe silly questions. :)

index.bs
```diff
@@ -3798,168 +3991,785 @@ To <dfn>get the response data</dfn> given |response|:
...
+
+</div>
+
+#### The network.failRequest Command ####  {#command-network-failRequest}
+
+The <dfn export for=commands>network.failRequest</dfn> command fails a
+fetch that's blocked by a [=network intercept=].
+
+<dl>
+   <dt>Command Type</dt>
```
thiagowfx, line 4343: This entry for `failRequest` is missing "Return Type Empty Result".

index.bs
```diff
@@ -3245,17 +3237,143 @@ A [=remote end=] has a <dfn>before request sent map</dfn> which is initially an
...
+</div>
+
 ### Types ### {#module-network-types}
 
+#### The network.AuthChallenge Type #### {#type-network-AuthChallenge}
+
+<pre class="cddl local-cddl">
+network.AuthChallenge = {
+  scheme: text,
+  realm: text,
```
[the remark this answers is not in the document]
thiagowfx: That's a good point, but there's no other idiomatic term to use here, is there?

index.bs
```diff
@@ -3288,27 +3404,67 @@ To <dfn>get the base network event data</dfn> given |request| and |redirect coun
...
 
 [=Remote end definition=] and [=local end definition=]
 
-<pre class="cddl local-cddl">
+<pre class="cddl local-cddl remote-cddl">
 network.Cookie = {
     name: text,
     ? value: text,
-    ? binaryValue: [ uint ]
+    ? binaryValue: [ *(0..255) ]
```
thiagowfx, line 3467: Should `Cookie` make `value` and `binaryValue` mutually exclusive, like `Header` does?

index.bs
```diff
@@ -3496,11 +3496,15 @@ empty map.  It's used to track the network events for which a
...
-empty [=/map=].  It's used to track the requests which are actively being blocked.
+A [=BiDi session=] has an <dfn>intercept map</dfn> which is a [=/map=] between
+intecept id and a [=tuple=] with fields <code>url patterns</code> and
+<code>phases</code> that define the properties of active network intercepts. It
+is initially empty.
+
+A [=BiDi session=] has an <dfn>blocked request map</dfn>, used to track the
+requests which are actively being blocked. It is an [=map=] between request id
+and a [=tuple=] with fields <code>request</code>, <code>phase</code>, and
+<code>response</code>. It is initially empty.  It's
```
thiagowfx, line 3507: nit: extra `It's`

index.bs
```diff
@@ -4068,13 +4220,579 @@ To <dfn>get the response data</dfn> given |response|:
...
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+      <pre class="cddl remote-cddl">
+      network.ContinueResponse = {
+        method: "network.continueResponse",
+        params: network.ContinueResponseParameters
+      }
+
+      network.ContinueResponseParameters = {
```
jimevans, line 4394: Does this mean it's not possible to modify the response body in this case? Yes, I know `network.provideResponse` exists, but not being able to modify that here seems an oversight.

jimevans: Resolved, via conversation in IRC. Will submit an issue once this PR lands in its initial form.

index.bs
jgraham: I've added an initial cut at handling cookies. Let me know what you think.

index.bs
```diff
@@ -3798,168 +3991,785 @@ To <dfn>get the response data</dfn> given |response|:
...
-   [=redirect count=], then run the [=WebDriver BiDi before request sent=] steps
-   with |request|.
+1. If |blocked requests| does not [=map/contain=] |request id| then return
+   [=error=] with [=error code=] [=invalid argument=].
 
-   Note: This ensures that a <code>network.beforeRequestSent</code> can
-   always be emitted before a <code>network.fetchError</code>, without the
-   caller needing to explicitly invoke the [=WebDriver BiDi before request
-   sent=] steps on every error path.
+   Issue: consider a "<code>no such request</code>" error.
```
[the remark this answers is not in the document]
jgraham: Sure.

index.bs
```diff
@@ -3798,168 +3991,785 @@ To <dfn>get the response data</dfn> given |response|:
...
-1. Assert: |params| matches the <code>network.ResponseCompletedParameters</code>
-   production.
+  1. Let |body| be [=get the body=] with |command
+     parameters|["<code>body</code>"].
 
-1. Let |body| be a map matching the <code>network.ResponseCompleted</code>
-   production, with the <code>params</code> field set to |params|.
+  1. Set |request|'s [=request/body=] to |body|.
+
+1. [=Resume=] with "<code>continue request</code>", |request id|, and (null,
```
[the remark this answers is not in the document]
jgraham: The request object is modified in-place during this algorithm, so we don't need to do anything else with it.

index.bs
```diff
@@ -3798,168 +3991,785 @@ To <dfn>get the response data</dfn> given |response|:
...
+      network.ContinueResponse = {
+        method: "network.continueResponse",
+        params: network.ContinueResponseParameters
+      }
+
+      network.ContinueResponseParameters = {
+        request: network.Request,
+        ?credentials: network.AuthCredentials,
+        ?headers: [*network.Header],
+        ?reasonPhrase: text,
```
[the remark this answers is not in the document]
jgraham: `reasonPhrase` is taken directly from https://httpwg.org/specs/rfc7230.html#status.line but I don't really mind either way.

jimevans (approved):

index.bs
jimevans: I'm pretty happy with this approach. I'd guess that cookie manipulation is going to be the one thing people will want to do when monitoring/modifying network traffic more than just about anything else. This makes that easier.

index.bs
```diff
@@ -4053,164 +4253,973 @@ To <dfn>get the response data</dfn> given |response|:
...
-   </dd>
-</dl>
+<pre class="cddl local-cddl remote-cddl">
+network.SetCookieHeader = {
+    name: text,
+    value: network.BytesValue,
+    ? domain: text,
+    ? httpOnly: bool,
+    ? expires: text,
+    ? maxAge: js-uint,
```
jimevans, line 4273: Minor nitpick: According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie), the Max-Age parameter of the header can be negative, so `js-int`? I know that using zero has the same effect as a negative, but to strictly follow the header allowable values here, we should allow it too. I'm not too worried about it either way, if we want to ignore this corner case.

index.bs
```diff
@@ -4053,164 +4253,973 @@ To <dfn>get the response data</dfn> given |response|:
...
+
+1. Let |request id| be |command parameters|["<code>request</code>"].
+
+1. Let |response| be the result of [=trying=] to [=update the response=] with
+   |session|, "<code>provideResponse</code>", and |command parameters|.
+
+1. If |command parameters| [=map/contains=] "<code>body</code>":
+
+  1. Let |body| be [=deserialize protocol bytes=] with |command
+     parameters["<code>body</code>"].
```
juliandescottes, line 4883:
```suggestion
     parameters|["<code>body</code>"].
```

index.bs
```diff
@@ -4053,164 +4253,973 @@ To <dfn>get the response data</dfn> given |response|:
...
+
+<div algorithm="remote end steps for network.removeIntercept">
+The [=remote end steps=] given |session| and |command parameters| are:
+
+1. Let |intercept| be the value of the "<code>intercept</code>" field in |command
+   parameters|.
+
+1. Let |intercept map| be |session|'s [=intercept map=].
+
+1. If |intercept map| does not <a for=map>contain</a> |intercept|, return
```
juliandescottes, line 4930: Just curious, why not `[=map/contains=]` here?

index.bs
```diff
@@ -4053,164 +4253,973 @@ To <dfn>get the response data</dfn> given |response|:
...
+
+1. Return [=success=] with data null.
+
+</div>
+
+#### The network.provideResponse Command ####  {#command-network-provideResponse}
+
+The <dfn export for=commands>network.provideResponse</dfn> command continues a
+response that's blocked by a [=network intercept=], by providing a complete
+response and bypassing any further network request processing.
```
juliandescottes, line 4843: Can this be called in any phase? If called in beforeRequestSent, do we still get responseStarted/Completed events?

index.bs
```diff
@@ -4053,164 +4253,973 @@ To <dfn>get the response data</dfn> given |response|:
...
 
-       network.BeforeRequestSentParameters = {
-         network.BaseParameters,
-         initiator: network.Initiator,
-       }
-      </pre>
-   </dd>
-</dl>
+<pre class="cddl local-cddl remote-cddl">
+network.SetCookieHeader = {
```
OrKoN, line 4267: What is the reason for not treating cookies as a regular header?

OrKoN: is it only to help the user with the generation of the cookie header string or are there other reasons?

index.bs
```diff
@@ -3635,6 +3801,35 @@ samesite-flag, which is from [[SAME-SITE-COOKIES]].
...
+
+<div algorithm>
+To <dfn>serialize cookie header</dfn> given |protocol cookie|:
+
+1. Let |name| be [=UTF-8 encode=] |protocol cookie|["<code>name</code>"].
+
+1. Let |value| be [=deserialize protocol bytes=] with |protocol
+   cookie|["<code>value</code>"].
+
+1. Let |header value| be the [=byte sequence=] formed by concatenating |name|,
```
mathiasbynens, line 3826: Do we need any special treatment for newline characters here (e.g. disallow, or escape)? Thinking about HTTP header injection.

index.bs
jimevans: @OrKoN It introduces an asymmetry in the API, since cookies are explicitly separated in the request data. Additionally, translating a cookie with the desired data into a `Set-Cookie` header, properly formatted, with the proper fields is likely to be error-prone for client writers.

Either cookies are a special case everywhere, or they’re a special case nowhere. In the existing representation of request data, we’ve already come to a consensus that the former is the right decision.

index.bs
jgraham, line 4928:
```suggestion
1. If |intercept map| does not [=map/contain=] |intercept|, return
```

index.bs
```diff
@@ -4053,164 +4253,973 @@ To <dfn>get the response data</dfn> given |response|:
...
+
+1. Let |intercept| be the value of the "<code>intercept</code>" field in |command
+   parameters|.
+
+1. Let |intercept map| be |session|'s [=intercept map=].
+
+1. If |intercept map| does not <a for=map>contain</a> |intercept|, return
+   [=error=] with [=error code=] [=no such intercept=].
+
+1. <a for=map>Remove</a> |intercept| from |intercept map|.
```
jgraham, line 4931:
```suggestion
1. [=map/Remove=] |intercept| from |intercept map|.
```

index.bs
OrKoN: That means that cookies would be included twice in every request? once in the headers and once in the cookies list, is that right?

jimevans: That depends. Do you mean in requests intercepted by the driver, and reported via events? If so, then yes, you’d see a `Cookie` header, consisting of only name-value pairs, and a list of cookies that contain the additional cookie values (domain, path, expiration, etc.), if you mean in requests and/or responses sent via commands as potentially modified/added/removed/edited by users, then that depends entirely on the parameters passed in by the user in the command to provide the modified request or fulfilled response. Or at least that’s my reading.

Making modifications to cookies in requests and responses is a common enough user action that I’m encouraging of making this part of the API easier to use for users. Forcing the user to treat cookies as “just another header,” especially considering the complexity of formatting the header properly, ans well as the cognitive dissonance with already making cookies “special” in the request interception event data, increases friction for users.

OrKoN: Do you have links to issues where this topic has been discussed before? It feels like this we might need a better integration with future cookie manipulation story (https://github.com/w3c/webdriver-bidi/issues/287) as I am not sure that it is a good thing to unconditionally include serialisation of metadata cookies on each request.

jimevans: I'm not sure I can point to a ton of other issues. I can point to the [comment I made in May](https://github.com/w3c/webdriver-bidi/pull/429#discussion_r1203152220) which led to this addition by @jgraham. The inclusion of cookie metadata in the [predates this PR](https://w3c.github.io/webdriver-bidi/#type-network-RequestData), which has been part of the spec draft for months.

jgraham: I don't know what a "metadata cookie" is, but I don't think this affects the future cookie manipulation story much. 

The key problem we have with general access to the cookie store (or other storage) is that we don't know in general what the keys to the store look like in each browser. For requests and responses that's not relevant because it's just providing the values that are actually in the request/response in a slightly easier to use form. Indeed, that part was already in network request logging, and matches what CDP does.

All that's new here is the ability to also _modify_ cookies using the easier to use form, rather than writing the header values directly. Really the only complexity is that there are now multiple ways to set the same header, and you need to reconcile them. But of course a client can also choose not to use this feature at all.

OrKoN: >  matches what CDP does

I guess that would be `associatedCookies` on the requestWillBeSentExtraInfo event? That's a separate event though and can be ignored for the network interception/logging purposes by clients who don't need it. In WebDriver BiDi the structured cookies will always be sent for all use cases 🤔  

I guess that's okay then (multiple network events in CDP is a big pain so we should avoid having a similar mechanism in BiDi) and some clients will likely need a snapshot of the stored cookies. If in the future there is a need for the ability to opt out from receiving the structured cookies to reduce amount of transferred data between the driver and the client, I guess that would be possible.

> that part was already in network request logging

Yeah, I overlooked that as I was reading the preview version and has not followed the spec closely back then. Thanks for the context!

thiagowfx:
@jgraham We have started the implementation of this feature in the Chromium side.

Do you have any ETA for adding WPT tests for Network Intereception?
(This question is purely for planning purposes, there's no pressure. Also, it would help with coordination e.g. so that we do not both end up writing the same WPT tests.)

juliandescottes:
> @jgraham We have started the implementation of this feature in the Chromium side.
> 
> Do you have any ETA for adding WPT tests for Network Interception? (This question is purely for planning purposes, there's no pressure. Also, it would help with coordination e.g. so that we do not both end up writing the same WPT tests.)

FWIW, I'm going to start the implementation this week but we don't have anything yet. The first bug which should be worked on is [network.addIntercept](https://bugzilla.mozilla.org/show_bug.cgi?id=1826192), so I might be adding basic tests for the API itself. Same for removeIntercept.

However I doubt I'll be able to work on actual network interception tests very soon, because this is a relatively new feature for Firefox in general (for instance current request blocking in FF devtools is not actually blocking requests, only responses...). So that will probably take more time before I can add meaningful tests for say continueWithAuth.

If you already started the implementation, there's a good chance you'll be writing tests before I do anyway? But we can keep in touch about that.

index.bs
```diff
@@ -4109,165 +4301,1226 @@ To <dfn>get the response data</dfn> given |response|:
...
 
-<dl>
-   <dt>Event Type</dt>
+[=Remote end definition=]
+
+<pre class="cddl remote-cddl">
+network.UrlPattern = network.UrlPatternPattern // network.UrlPatternString
+
+network.UrlPatternPattern = {
+    type = "pattern"
```
juliandescottes, line 4446:
```suggestion
    type = "pattern",
```

index.bs
```diff
@@ -4109,165 +4301,1226 @@ To <dfn>get the response data</dfn> given |response|:
...
+    ?protocol: text,
+    ?hostname, text,
+    ?port: text,
+    ?pathname: text,
+    ?search: text,
+}
+
+
+network.UrlPatternString = {
+    type = "string"
```
juliandescottes, line 4456:
```suggestion
    type = "string",
```

index.bs
```diff
@@ -4109,165 +4301,1226 @@ To <dfn>get the response data</dfn> given |response|:
...
+1. Let |intercept| be the string representation of a [[!RFC4122|UUID]].
+
+1. Let |url patterns| be the <code>urlPatterns</code> field of |command
+   parameters| if present, or null otherwise.
+
+1. Let |intercept map| be |session|'s [=intercept map=].
+
+1. Let |parsed patterns| be an empty list.
+
+1. For each |url pattern| in |url patterns|:
```
juliandescottes, line 4730: Is this a problem if `url patterns` is null from step 2? Should we default to empty list instead of null?

juliandescottes (commented):
nits + 1 quick question

index.bs
thiagowfx: Was just implementing this, and yes, empty list would be more convenient.

index.bs
```diff
@@ -4109,165 +4301,1226 @@ To <dfn>get the response data</dfn> given |response|:
...
+
+  1. let |parsed| be the result of [=trying=] to [=parse url pattern=] with |url
+     pattern|.
```
thiagowfx, line 4733:
```suggestion
  1. Let |parsed| be the result of [=trying=] to [=parse url pattern=] with |url
     pattern|.
```

nit

index.bs
```diff
@@ -3556,17 +3552,139 @@ A [=remote end=] has a <dfn>before request sent map</dfn> which is initially an
 empty map.  It's used to track the network events for which a
 <code>network.beforeRequestSent</code> event has already been sent.
 
+### Network Intercepts ### {#network-intercepts}
+
+A <dfn>network intercept</dfn> is a mechanism to allow remote ends to intercept
+and modify network requests and responses.
+
+A [=BiDi session=] has an <dfn>intercept map</dfn> which is a [=/map=] between
+intercept id and a [=tuple=] with fields <code>url patterns</code> and
```
thiagowfx, line 3561:
```suggestion
intercept id and a [=struct=] with fields <code>url patterns</code> and
```

Why struct instead of tuple? Because `addIntercept `step #6 is referring to this as "struct".

(The other way around also works, but in this case the order seems to not matter, so `struct` seems a better fit.)

index.bs
```diff
@@ -3556,17 +3552,139 @@ A [=remote end=] has a <dfn>before request sent map</dfn> which is initially an
...
+<code>phases</code> that define the properties of active network intercepts. It
+is initially empty.
+
+A [=BiDi session=] has a <dfn>blocked request map</dfn>, used to track the
+requests which are actively being blocked. It is an [=/map=] between request id
+and a [=tuple=] with fields <code>request</code>, <code>phase</code>, and
```
thiagowfx, line 3567: s/tuple/struct

Same argument as above.

index.bs
```diff
@@ -3556,17 +3552,139 @@ A [=remote end=] has a <dfn>before request sent map</dfn> which is initially an
...
 
 <pre class="cddl local-cddl">
 network.BaseParameters = (
     context: browsingContext.BrowsingContext / null,
+    isBlocked: bool,
     navigation: browsingContext.Navigation / null,
     redirectCount: js-uint,
     request: network.RequestData,
     timestamp: js-uint,
+    ?intercepts: [*network.Intercept]
```
thiagowfx, line 3687: style: please add a whitespace for all optional arguments, e.g. `? intercepts`.

This is for consistency with the rest of this spec file.

index.bs
```diff
@@ -4109,165 +4301,1226 @@ To <dfn>get the response data</dfn> given |response|:
...
+
+<div algorithm>
+To <dfn>update the response</dfn> given |session|, |command| and |command parameters|:
+
+1. Let |blocked requests| be |session|'s [=blocked request map=].
+
+1. Let |request id| be |command parameters|["<code>request</code>"].
+
+1. If |blocked requests| does not [=map/contain=] |request id| then return
+   [=error=] with [=error code=] [=invalid argument=].
```
thiagowfx, line 4911: Regarding the issue below, `no such request`  is now an existing error.

Should `invalid argument` be replaced with `no such request` then?

thiagowfx (dismissed):
See open comments

index.bs
```diff
@@ -4109,165 +4301,1226 @@ To <dfn>get the response data</dfn> given |response|:
...
+  1. let |parsed| be the result of [=trying=] to [=parse url pattern=] with |url
+     pattern|.
+
+  1. [=list/Append=] |parsed| to |parsed patterns|.
+
+1. Set |intercept map|[|intercept|] to a struct with <code>url patterns</code>
+   |parsed patterns| and <code>phases</code> |command parameters|["<code>phases</code>"].
+
+1. Return a new [=/map=] matching the
+   <code>network.AddNetworkInterceptResult</code> with the
```
thiagowfx, line 4741:
```suggestion
   <code>network.AddInterceptResult</code> with the
```

index.bs
```diff
@@ -3556,17 +3552,139 @@ A [=remote end=] has a <dfn>before request sent map</dfn> which is initially an
...
+1. For each |intercept| → |intercept properties| of |session intercepts|:
+
+  1. If |intercept properties|[<code>phases</code>] [=set/contains=] |phase|:
+
+    1. For each |url pattern| in |intercept properties|[<code>url
+       patterns</code>]:
+
+      1. If [=match URL pattern=] with |url pattern| and |url|:
+
+        1. [list/Append=] |intercept| to |intercepts| and break.
```
thiagowfx, line 3602: Typo. Missing another equals.

index.bs
```diff
@@ -4109,165 +4301,1226 @@ To <dfn>get the response data</dfn> given |response|:
...
+
+      1. Append "<code>:</code>" to |url pattern|.
+
+      1. For each |codepoint| in |port|:
+
+        1. If |codepoint| is not an [=ASCII digit=]:
+
+          1. Return [=error=] with [=error code=] [=invalid argument=].
+
+        1. Append |port| to |url pattern|.
```
juliandescottes, line 4588:
```suggestion
      1. Append |port| to |url pattern|.
```

Indentation seems wrong here, should only be done after checking all codepoints, not for each one of them.

index.bs
```diff
@@ -4109,165 +4301,1226 @@ To <dfn>get the response data</dfn> given |response|:
...
+         |pattern url|.
+
+      1. Set |has hostname| to false.
+
+    1. If |pattern| [=map/contains=] "<code>port</code>":
+
+      1. Let |port| be the result of [=trying=] to [=unescape
+         URL Pattern=] with |pattern|["<code>port</code>"].
+
+      1. Append "<code>:</code>" to |url pattern|.
```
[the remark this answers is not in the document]
juliandescottes: Also we might only want to add `:` after checking that all codepoints are digits? Or maybe we wanted to append each codepoint to `url pattern` and not port... in which case the current order makes sense.

index.bs
```diff
@@ -4109,165 +4301,1226 @@ To <dfn>get the response data</dfn> given |response|:
...
+<div algorithm>
+To <dfn>parse URL pattern</dfn>, given |pattern|:
+
+1. Let |has protocol| be true.
+
+1. Let |has hostname| be true.
+
+1. If |pattern| matches the <code>network.UrlPatternPattern</code> production:
+
+  1. Let |pattern url| be the empty string.
```
juliandescottes, line 4513: The algorithm is mixing `|pattern url|` and `|url pattern|`, which appears at L4580. I imagine `|pattern url|` makes more sense here, since we're building a string which represents the pattern?

index.bs
```diff
@@ -4109,165 +4301,1226 @@ To <dfn>get the response data</dfn> given |response|:
...
+         [=invalid argument=].
+
+      1. Let |hostname| be the result of [=trying=] to [=unescape
+         URL Pattern=] with |pattern|["<code>hostname</code>"].
+
+      1. Let |inside brackets| be false.
+
+      1. For each |codepoint| in |hostname|:
+
+        1. If «U+003A (:), U+002F (/), U+003F (?), U+0023 (#)» [=set/contains=]
```
juliandescottes, line 4553:
```suggestion
        1. If «U+002F (/), U+003F (?), U+0023 (#)» [=set/contains=]
```

I think we need to remove `U+003A` from this check since it's checked at the next step and should be allowed if inside brackets is false.

index.bs
```diff
@@ -4109,165 +4301,1226 @@ To <dfn>get the response data</dfn> given |response|:
...
+          1. Return [=error=] with [=error code=] [=invalid argument=].
+
+        1. If |inside brackets| is false and |codepoint| is U+003A (:):
+
+          1. Return [=error=] with [=error code=] [=invalid argument=].
+
+        1. If |codepoint| is U+005B ([), set |inside brackets| to true.
+
+        1. If |codepoint| is U+005D (]), set |inside brackets| to false.
+
```
juliandescottes, line 4565: Should we try to check that we have matching opening and closing brackets? Is nesting allowed`[[]:]`?

index.bs
```diff
@@ -4109,165 +4301,1226 @@ To <dfn>get the response data</dfn> given |response|:
...
+     <dt>protocol
+     <dd>|url|'s [=url/scheme=] if |has protocol| is true, or null otherwise.
+     <dt>hostname
+     <dd>|url|'s [=url/host=] if |has hostname| is true, or null otherwise.
+     <dt>port
+     <dd>[=Serialize an integer=] with |url|'s [=url/port=], if it's not null,
+         or null otherwise.
+     <dt>pathname
+     <dd>The result of running the [=URL path serializer=] with |url|'s
+         [=url/path=], if it's not [=list/empty=], or null otherwise.
```
juliandescottes, line 4643: [`URL path serializer`](https://url.spec.whatwg.org/#url-path-serializer) takes a url, not a url path

juliandescottes: Apparently `path` can both be a string (opaque path) or a list (path), the current check only mentions list/empty. I think it can also be an empty string in the opaque path state, should we set null in that case as well?

index.bs
jgraham: I think for now I'm going to leave this as-is, but I'll add a note indicating the behaviour.

index.bs
```diff
@@ -4053,164 +4253,973 @@ To <dfn>get the response data</dfn> given |response|:
...
+      1. Set |header|'s value to |cookie header|.
+
+      1. Set |found cookie header| to true.
+
+      1. Break.
+
+  1. If |found cookie header| is false:
+
+    1. Append the [=header=] (`<code>Cookie</code>`, |cookie header|) to
+       |request|'s [=request/headers list=].
```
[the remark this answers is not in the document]
jgraham: Actually I think it's OK? If there's already a cookie header we overwrite it's value. That means if you provide both `cookies` and also `headers` and they both try to set cookies you just get the header value, but it's also unclear to me why you'd do that. I suppose we could make it an error?

index.bs
```diff
@@ -4109,165 +4310,1238 @@ To <dfn>get the response data</dfn> given |response|:
...
+
+1. Let |url| be the result of [=URL parser|parsing=] |url string|.
+
+1. If |url pattern|'s protocol is not null and is not equal to |url|'s
+   [=url/scheme=], return false.
+
+1. If |url pattern|'s hostname is not null and is not equal to |url|'s
+   [=url/host=], return false.
+
+1. If |url pattern|'s port is not null:
```
juliandescottes, line 4682: Wondering what's the best approach here when the pattern port is null.

For `https://example.com`, whether you specified it from a string or a pattern, `url pattern's port` will be null, which means we will match any port. If we don't want to have pattern-like behaviors, then maybe when `url pattern's port` is null, we only match when `url's port` is also null?

To me it seems like "matching any port" should be opt in, but maybe that's not how URLPattern works?

index.bs
```diff
@@ -4109,165 +4310,1238 @@ To <dfn>get the response data</dfn> given |response|:
...
+
+      1. Append |pathname| to |pattern url|.
+
+    1. If |pattern| [=map/contains=] "<code>search</code>":
+
+      1. Let |search| be the result of [=trying=] to [=unescape URL pattern=] with
+         |pattern|["<code>search</code>"].
+
+      1. If |search| does not [=string/starts with|start with=] U+003F (?), then
+         append "<code>?</code>" to |pattern url|.
```
juliandescottes, line 4631: (just thinking out loud...) For the edge case where users want to use query parameters named `?`, they might pass `?=1`, then we won't add the `?`. Meaning the  the user won't have the expected search string in the end...

But it's quite hard to guess what the user really meant in that case, and using other default behaviors (eg always adding `?`, or never adding `?`) also can lead to confusing situations where it will be hard to understand the issue both for us and the users. In summary I think your current approach is the best compromise.

index.bs
juliandescottes: > why port would be more special than any other part of the URL here?

Mostly because http and https have default ports, http with no port usually implies port 80, https with no port implies 443?

juliandescottes: Part of my problem is that while providing no port in the `pattern` type sounds intentional, if you use a string `https://example.com`, it's really hard to tell if you want to match any port or if you expect it to use the default for the protocol. So maybe we need a `has port` flag similar to the other parts?

index.bs
```diff
@@ -4109,165 +4310,1257 @@ To <dfn>get the response data</dfn> given |response|:
...
+
+      1. Set |has protocol| to false.
+
+      1. Append "<code>http</code>" to |pattern url|.
+
+    1. Let |scheme| be [=ASCII lowercase=] with |pattern url|.
+
+    1. Append "<code>:</code>" to |pattern url|.
+
+    1. If |scheme| [=is special=], append "<code>//</code>" to |pattern url|.
```
juliandescottes, line 4551: For non special schemes (let's take `git` for instance), the URL will be parsed differently depending whether we add `//` or not:
- for `git://github.com/whatwg/url.git`, the host will be "github.com"
- for `git:github.com/whatwg/url.git` the host is empty

With the current API users can't really specify the first variant if they use URLPatternPattern. 
But maybe it doesn't matter? I imagine we are not actually interested in handling all possible URLs, but only whatever can be intercepted, so special schemes + a select few (eg `data:` ?). Do we need to surface this somehow in the spec?

index.bs
```diff
@@ -4109,165 +4310,1257 @@ To <dfn>get the response data</dfn> given |response|:
...
+
+    1. If |pattern| [=map/contains=] "<code>hostname</code>":
+
+      1. If |pattern|["<code>hostname</code>"] is the empty string, return [=error=]
+         with [=error code=] [=invalid argument=].
+
+      1. If |scheme| is "<code>file</code>" return [=error=] with [=error code=]
```
juliandescottes, line 4558: It seems that the `file` scheme can also support hostnames, depending on how you use it. 
- `file:/path` -> no hostname
- `file:///path`-> empty hostname
- `file://hostname/path` -> hostname is "hostname" 

With the current spec, `file` patterns defined using URLPatternPattern will all start with `file:///`(or `file://` if users don't specify a path, but the basicURLParser should treat that as if it was `file:///`). Maybe we should lift this restriction?

juliandescottes:
@thiagowfx wanted to do a quick check about wdspec tests. Do you plan to submit tests soon? Otherwise I could provide the basic scaffolding for the network module (which has no command so far) + add intercept validation tests. 

Let me know how you prefer to proceed.

thiagowfx:
@juliandescottes Thanks for asking! I'll be OOO for a couple of days starting this Friday. Therefore, unless I manage to write them until tomorrow EOD, feel free to scaffold them.

juliandescottes:
> @juliandescottes Thanks for asking! I'll be OOO for a couple of days starting this Friday. Therefore, unless I manage to write them until tomorrow EOD, feel free to scaffold them.

Created https://github.com/web-platform-tests/wpt/pull/41291 with my current tests

index.bs
```diff
@@ -4109,165 +4310,1257 @@ To <dfn>get the response data</dfn> given |response|:
...
+
+1. If |url pattern|'s port is not null:
+
+   1. Let |port| be null.
+
+   1. If |url|'s [=url/scheme=] [=is special=] and |url|'s
+      [=url/scheme=]'s [=default port=] is not null, and |url|'s [=url/port=] is
+      null or is equal to [=url/scheme=]'s [=default port=]:
+
+     1. Set |port| to the emtpy string.
```
juliandescottes, line 4703:
```suggestion
     1. Set |port| to the empty string.
```

index.bs
```diff
@@ -4109,165 +4310,1257 @@ To <dfn>get the response data</dfn> given |response|:
...
+   <dl>
+     <dt>protocol
+     <dd>|url|'s [=url/scheme=] if |has protocol| is true, or null otherwise.
+     <dt>hostname
+     <dd>|url|'s [=url/host=] if |has hostname| is true, or null otherwise.
+     <dt>port
+     <dd>
+     1. If |url|'s [=url/scheme=] [=is special=] and |url|'s
+   [=url/scheme=]'s [=default port=] is not null, and |url|'s [=url/port=] is
+   null or is equal to [=url/scheme=]'s [=default port=]:
```
juliandescottes, line 4660: I still feel like if no port is provided, we should match any port (ie have some has port logic). We do that for hostname and protocol, so maybe it would make sense here as well?

On top of that, we can have some odd scenarios when users did not provide a protocol. Since we default to `http` in that case, the algorithm will behave unexpectedly in some cases (eg if provided port is 80). Maybe we should also take `has protocol` into account in this condition?

whimboo: Should we maybe default to `https` those days?

juliandescottes: > Should we maybe default to `https` those days?

Note that In theory, `http` here really only acts as a placeholder so that we can build a URL that we can parse later. You will still have the `has protocol` flag to `false`, which means the pattern will not attempt to match on the protocol. However since this is a special scheme, it does have some side effects: 
- has a default port
- path defaulting to "/"

If we check `has protocol` in the condition as suggested above, then using http or https should be identical. And if not, https just moves the edge case to scenarios where users provide a URLPatterPattern such as `{ port: 443 }` which would start matching `http` or `ws` requests on port 80, `ftp` requests on port 21, which is just as unexpected.

juliandescottes (approved):
While I might still have comments as I go forward with the implementation, IMO this is in a good enough state to land and then iterate with follow ups. Happy to keep providing feedback here, but maybe we should already merge the current version?

thiagowfx:
> While I might still have comments as I go forward with the implementation, IMO this is in a good enough state to land and then iterate with follow ups. Happy to keep providing feedback here, but maybe we should already merge the current version?

+1 - I am happy to approve to unblock/land this if warranted

There will definitely be more issues to be found here; personally I have only deeply reviewed AddIntercept and RemoveIntercept. I am hoping to have more comments as our implementation keeps going.

index.bs
```diff
@@ -4109,165 +4310,1257 @@ To <dfn>get the response data</dfn> given |response|:
...
+
+     1. Set |port| to [=serialize an integer=] with |url|'s [=url/port=].
+
+   1. If |url pattern|'s port is not equal to |port|, return false.
+
+1. If |url pattern|'s pathname is not null and is not equal to the result of
+   running the [=URL path serializer=] with |url|, return false.
+
+1. If |url pattern|'s search is not null and is not equal to |url|'s
+   [=url/query=], return false.
```
juliandescottes, line 4715: Just wanted to check if we are OK with this scenario as I stumbled on it while writing tests.

If I use a URLPatternString based on "http://example.com", the `query` returned by basicURLParser will be null, meaning that url pattern's search here will be null as well. So it will match URLs such as `http://example.com?`, `http://example.com?hello=hi` etc...

Which also means there is no way to say "match only URLs with a null search".  Before starting the implementation I assumed that using a URLPatternString would be close to requesting an exact match, but it's not the case. Is this behavior what we want here?

index.bs
```diff
@@ -4109,165 +4310,1257 @@ To <dfn>get the response data</dfn> given |response|:
...
+     1. Set |port| to the emtpy string.
+
+   1. Otherwise, if |url|'s [=url/port=], is not null:
+
+     1. Set |port| to [=serialize an integer=] with |url|'s [=url/port=].
+
+   1. If |url pattern|'s port is not equal to |port|, return false.
+
+1. If |url pattern|'s pathname is not null and is not equal to the result of
+   running the [=URL path serializer=] with |url|, return false.
```
juliandescottes, line 4712: Another interesting scenario, for a URLPatternPattern such as `{ protocol: "http" }`, since this is a special scheme, the basic URL parser will set the path to a single segment empty string `[""]`, meaning we will have a non-null pathname set to `/`. Such a URLPatternPattern will therefore match any URL using the `http` protocol and with an empty path. Meaning `http://abc.com:1234/?search` will match, but `http://abc.com:1234/path?search` will not. 

I feel like it's unexpected, and we probably want a `has path` logic here?

index.bs
```diff
@@ -3556,17 +3552,148 @@ A [=remote end=] has a <dfn>before request sent map</dfn> which is initially an
...
 
 <pre class="cddl local-cddl">
 network.BaseParameters = (
     context: browsingContext.BrowsingContext / null,
+    isBlocked: bool,
     navigation: browsingContext.Navigation / null,
     redirectCount: js-uint,
     request: network.RequestData,
     timestamp: js-uint,
+    ? intercepts: [*network.Intercept]
```
juliandescottes, line 3696:
```suggestion
    ? intercepts: [+network.Intercept]
```

By definition intercepts cannot be empty? If it's empty, then isBlocked is true and then `intercepts` should be omitted?

index.bs
```diff
@@ -4109,165 +4310,1257 @@ To <dfn>get the response data</dfn> given |response|:
...
+
+1. Let |has protocol| be true.
+
+1. Let |has hostname| be true.
+
+1. If |pattern| matches the <code>network.UrlPatternPattern</code> production:
+
+  1. Let |pattern url| be the empty string.
+
+    1. If |pattern| [=map/contains=] "<code>protocol</code>":
```
whimboo, line 4524: The indentation seem to be totally off here. It would be good to get this fixed before we merge the PR.

index.bs
```diff
@@ -4109,165 +4310,1257 @@ To <dfn>get the response data</dfn> given |response|:
...
+
+      1. Let |protocol| be the result of [=trying=] to [=unescape
+         URL Pattern=] with |pattern|["<code>protocol</code>"].
+
+      1. For each |codepoint| in |protocol|:
+
+      1. If |codepoint| is not [=ASCII alphanumeric=] and «U+002B (+), U+002D
+         (-), U+002E (.)» does not [=set/contains|contain=] |codepoint|:
+
+        1. Return [=error=] with [=error code=] invalid argument.
```
whimboo, line 4537:
```suggestion
      1. For each |codepoint| in |protocol|:

        1. If |codepoint| is not [=ASCII alphanumeric=] and «U+002B (+), U+002D
         (-), U+002E (.)» does not [=set/contains|contain=] |codepoint|:

          1. Return [=error=] with [=error code=] invalid argument.
```

thiagowfx:
One question about the spec (if this is not a good medium to ask, let me know if I should open a new bug):

> A [BiDi session](https://w3c.github.io/webdriver-bidi/#bidi-session) has a blocked request map, used to track the requests which are actively being blocked. It is an [map](https://infra.spec.whatwg.org/#ordered-map) between request id and a [struct](https://infra.spec.whatwg.org/#struct) with fields request, phase, and response. It is initially empty.

Translating to pseudo typescript code, this would roughly look like:

```ts
  /** A map to track the requests which are actively being blocked. */
  readonly #blockedRequestMap = new Map<
    Network.Request,
    {
      request: Network.Request;
      phase?: Network.InterceptPhase;
      response?: Network.ResponseData;
    }
  >();
```

The question is: what is the difference between the map key (the request ID) and the first property of the map value (the request)? Are these fields the same? If so, isn't this redundant? If not, then to what "Request" does the value refer to?

(reference: https://w3c.github.io/webdriver-bidi/#type-network-Request)

jgraham:
This isn't the best place to ask (maybe just IRC/Matrix for this kind of question, or file an new issue), but the request here is https://fetch.spec.whatwg.org/#requests

PR #430, opened 2023-05-20
Replace `null` with `nil` in CDDL spec

christian-bromann:
As far as I can see the CDDL specification doesn't has a native type `null` but there is `nil`. This patch removes occurrences of `null` and replaces it with `nil`.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/430.html" title="Last updated on May 20, 2023, 10:28 PM UTC (027f192)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/430/cc806ae...027f192.html" title="Last updated on May 20, 2023, 10:28 PM UTC (027f192)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -318,7 +318,7 @@ CommandResponse = {
 }
 
 ErrorResponse = {
-  id: js-uint / null,
+  id: js-uint / nil,
   error: ErrorCode,
   message: text,
   ?stacktrace: text,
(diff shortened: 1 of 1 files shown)


christian-bromann:
Nevermind, it has an alias:

```
                  nil = #7.22
                  null = nil
```

ref https://tools.ietf.org/html/draft-ietf-cbor-cddl-08#appendix-D

Issue #383, opened 2023-03-15
"browsingContext.captureScreenshot" should support different image formats


whimboo:
As of now we only support PNG. It would be great if at least JPG would be supported as well. With that addition a `quality` parameter might be helpful as well.

OrKoN:
Puppeteer also needs/supports webp.

thiagowfx:
Reference: https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-captureScreenshot

whimboo:
@jrandolf would you mind adding some wpt tests for this new feature?

PR #436, opened 2023-05-31
`browsingContext.captureScreenshot`: support different image formats

thiagowfx:
Reference: https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-captureScreenshot

Fixes: #383

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/436.html" title="Last updated on May 31, 2023, 7:37 PM UTC (a0b49a3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/436/628c065...a0b49a3.html" title="Last updated on May 31, 2023, 7:37 PM UTC (a0b49a3)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2137,7 +2137,9 @@ Base64-encoded string.
       }
 
       browsingContext.CaptureScreenshotParameters = {
-        context: browsingContext.BrowsingContext
+        context: browsingContext.BrowsingContext,
+        ? format: "jpeg" / "png" / "webp",
+        ? quality: (js-uint .le 100),
       }
       </pre>
    </dd>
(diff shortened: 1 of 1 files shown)


jgraham:
This doesn't work as written because https://w3c.github.io/webdriver/#dfn-encoding-a-canvas-as-base64 hardcodes `image/png` as the type. You need to update that, and also pass in the parameter as an actual MIME type. Arguably the API should just take a full MIME type rather than just using a subtype. Then we wouldn't have to hardcode in the supported formats. Also there should be an early error (either "unsupported operation" or a new one like "unsupported image format") if the requested format isn't supported by the browser (I assume we want that rather than the "fall back to PNG behaviour?).

As a practical matter, do we know what the current support matrix for different formats looks like?

whimboo:
IMHO it would be better for clients to only have to take care / handle the `unsupported operation` error. The message or `data` field could contain more information what's exactly wrong.

thiagowfx:
Closing, and moving comments to https://github.com/w3c/webdriver-bidi/issues/383.

Issue #368, opened 2023-02-22
Use CDDL default value and ranges instead of pros


whimboo:
To better allow automatic validity checks for passed parameter types and values, it would be great to have all the default values and ranges defined in the CDDL sections. A good example of it you can find in the current [PR for `browsingContext.print`](https://github.com/w3c/webdriver-bidi/pull/363):

```
        ?orientation: ("portrait" / "landscape") .default "portrait",
        ?shrinkToFit: bool .default true,
```

@sadym-chromium, @foolip do you have any objections for doing that? Also CC'ing @christian-bromann given that he has interest in automated code generation as well.

jgraham:
I looked at the CDDL spec again and it suggests that `.default` implies a `.ne` condition, so you actually aren't allowed to explicitly provide the default value. If that's true we probably can't use it (I will check how the rust library behaves).

foolip:
If there's a way to specify a default that reduces the amount of prose then that sounds great! If the default value can't be given directly, I guess we'd need "default" dummy and that wouldn't reduce the amount of prose.

whimboo:
@christian-bromann did you already have the time to check the usage of defaults?

christian-bromann:
> did you already have the time to check the usage of defaults?

I started updating the project since I last touched it but have been busy with my SeConf engagements. I will be able to take a look after next week.

christian-bromann:
Update: I improved the [CDDL parser](https://github.com/christian-bromann/cddl) after pulling the latest spec and there are only 2 parsing issues left which I don't know if they are related to my parser or just invalid CDDL.

I also started work on transforming the CDDL interfaces into TypeScript interfaces so that JS clients can consume them.

whimboo:
This is great to hear! Could you tell us which are these 2 remaining issues that are left? We are happy to help in determining if it is related to the CDDL definition or not.

thiagowfx:
jgraham@, via https://github.com/w3c/webdriver-bidi/pull/438:

> So there's two high level questions here:
> * The general one about whether we can use `.default` at all. I hope we can because it makes things easier, but the RFC is [pretty explicit](https://datatracker.ietf.org/doc/html/rfc8610#section-3.8.6) that `.default` implies a `.ne` for the value, i.e. it's forbidden to send the default value explicitly. That's almost never what we want. But the implementations I've tried don't seem to enforce that rule.
[...]

PR #438, opened 2023-05-31
CDDL: use .default wherever possible instead of prose

thiagowfx:
Bug: #368

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/438.html" title="Last updated on May 31, 2023, 10:38 PM UTC (b2441e8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/438/628c065...b2441e8.html" title="Last updated on May 31, 2023, 10:38 PM UTC (b2441e8)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1979,7 +1979,7 @@ To <dfn>get the browsing context info</dfn> given |context|,
 
 <div algorithm>
 To <dfn>await a navigation</dfn> given |context|, |request|, |wait condition|, and optionally
-|history handling| (default: "<code>default</code>") and |ignore cache| (default: false):
+|history handling| (default: "<code>default</code>") and |ignore cache|:
 
 1. Let |navigation id| be the string representation of a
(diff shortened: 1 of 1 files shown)


jgraham:
So there's two high level questions here:
* The general one about whether we can use `.default` at all. I hope we can because it makes things easier, but the RFC is [pretty explicit](https://datatracker.ietf.org/doc/html/rfc8610#section-3.8.6) that `.default` implies a `.ne` for the value, i.e. it's forbidden to send the default value explicitly. That's almost never what we want. But the implementations I've tried don't seem to enforce that rule.
* The use of a default value that's a different type to the specified values e.g. `foo: js-uint .default null`. That seems pretty suspect to me and in particular seems like it might naturally turn into code which erroneously allows `null` to be passed as an explicit value.

thiagowfx:
Anyway, closing for now. Comments will be added to https://github.com/w3c/webdriver-bidi/issues/368.

Issue #171, opened 2022-01-31
Specify focusing another (top-level) browsing context


whimboo:
It should be possible to switch between different top-level browsing contexts. Not sure if setting the focus for individual child browsing contexts would be needed.

thiagowfx:
This would be great to enhance `browsingContext.captureScreenshot`.

Currently, the chromium mapper implementation needs to focus the context in order to take the screenshot (otherwise it blocks / hangs). We're currently doing so with our custom implementation (via CDP). If there were an official BiDi way to do it, then we could just use it instead.

OrKoN:
Completed in https://github.com/w3c/webdriver-bidi/pull/453

PR #440, opened 2023-06-02
context capture screenshot: add optional focus top-level context step

thiagowfx:
CDP's Page.captureScreenshot command[1] is blocked until the active context gets focus.

[1]: https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-captureScreenshot

Follow-up-of: https://github.com/w3c/webdriver-bidi/pull/183
Spec: https://w3c.github.io/webdriver-bidi/#command-browsingContext-captureScreenshot

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/440.html" title="Last updated on Jun 14, 2023, 4:03 PM UTC (063de20)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/440/5d3ee40...063de20.html" title="Last updated on Jun 14, 2023, 4:03 PM UTC (063de20)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2175,6 +2175,9 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
    Issue: This ought to be integrated into the update rendering algorithm
    in some more explicit way.
 
+ 1. An implementation may invoke the [=/focusing steps=] for the
+    [=top-level browsing context=] of |context|.
+
  1. Let |root rect| be |document|’s [=document element=]'s [=rectangle=].
 
  1. Let |canvas| be the result of [=trying=] to [=draw a bounding box from the


thiagowfx:
This PR was created as per @sadym-chromium's suggestion given our [chromium mapper implementation](https://github.com/GoogleChromeLabs/chromium-bidi/blob/097d011241a387e353c95af826fd089370136a77/src/bidiMapper/domains/context/browsingContextImpl.ts#L580).

sadym-chromium (approved):

whimboo:
I'm not that happy about that change given that it might cause side-effects users aren't aware of. For example web pages in background tabs will in most cases get throttled and as such will behave differently as when they are in foreground. 

So I can see especially issues when the focus is not set back to the originally focused tab. I think that this PR needs a discussion.

thiagowfx:
The alternative is to focus on whatever was focused previously, but we do not yet have the ability to determine that, correct? Maybe https://github.com/w3c/webdriver-bidi/issues/171 can help with that.

whimboo:
If some implementation specific steps have to be taken then such code could also revert the focus after the screenshot has been done. That's nothing the client has to be aware of. But again this is causing side-effects that you should be aware of.

jgraham:
Arguably if Chrome can only take screenshots when the context has focus, it should return `unsupported operation` if the context doesn't have focus (this is explicitly allowed by the current spec), and we should have a `browsingContext.focus()` command that clients can call so that we're not unexpectedly changing focus in some browsers but not others.

thiagowfx:
Related: https://github.com/puppeteer/puppeteer/issues/2254

thiagowfx:
Once https://github.com/w3c/webdriver-bidi/pull/453 is landed, do you think it would be feasible to **change** the WPT tests for capture screenshot to first call `browsingContext.bringToFront` before taking a screenshot?

If the answer is no: Then what if we **added** more tests instead? We could keep the current ones + add extra ones that perform the focus before taking screenshot.

whimboo:
I think it would be fine to have screenshot tests that test foreground and background. Maybe the background tests only test the basic features then and all the available variants are done in a foreground tab.

thiagowfx:
We will instead leverage https://github.com/w3c/webdriver-bidi/pull/453 to explicitly focus. Otherwise the implementation will throw an error.

WPT tests should be updated accordingly once the command is added to the spec (as whimboo@ seems to agree above).

thiagowfx:
> I think it would be fine to have screenshot tests that test foreground and background.

For completeness: https://bugs.chromium.org/p/chromium/issues/detail?id=32667 to support background screenshots

thiagowfx:
This issue was closed because of: (i) https://github.com/web-platform-tests/wpt/pull/40965 and (ii) WG consensus that this is probably not a good idea.

From @jgraham via IRC:

>  "optional" anything is bad, and that given changing focus has clear observable side effects, it's not an exception to that.

Issue #186, opened 2022-03-17
Enforce zero build warnings in CI


jgraham:
We don't enforce zero build warnings in CI and as a result there have been a few instances where changes have landed with warnings that turned out to be bugs. Even in cases where we had warnigns that aren't bugs it makes it easier for everyone if you know there should always be zero warnings. Therefore  propose that we fail the CI if there's a warning.

whimboo:
As mentioned to @jgraham on element maybe we could even have a pre-commit hook setup so that the bikeshed tests get run locally before being able to even push a PR which triggers certain amount of CI tasks?

thiagowfx:
If desired, I am happy to add pre-commit.com integration that does the above.

PR #442, opened 2023-06-05
add pre-commit.com integration

thiagowfx:
Bug: https://github.com/w3c/webdriver-bidi/issues/186

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/442.html" title="Last updated on Jun 6, 2023, 3:38 PM UTC (61a50b3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/442/5d3ee40...61a50b3.html" title="Last updated on Jun 6, 2023, 3:38 PM UTC (61a50b3)">Diff</a>


.pre-commit-config.yaml
```diff
@@ -0,0 +1,47 @@
...
+#   $ pre-commit run --all-files
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.4.0
+    hooks:
+      - id: check-executables-have-shebangs
+      - id: check-yaml
+      - id: end-of-file-fixer
+      - id: trailing-whitespace
+  - repo: https://github.com/pre-commit/mirrors-eslint
```
thiagowfx, line 28: Could potentially remove ESLint and prettier in order to remove diff noise.

thiagowfx: Removing prettier for now. Could be added later.

index.bs
```diff
@@ -1811,6 +1825,7 @@ navigation status</dfn> struct, which has the following items:
 
 <pre class="cddl remote-cddl">
 BrowsingContextCommand = (
+; keep-sorted start
```
whimboo, line 1828:
```suggestion
  ; keep-sorted start
```

Interesting idea. Also could we actually have a default set so that we do not have to repeat it in all the cases?

thiagowfx: If using the keep-sorted package, unfortunately this is the only way (repeating it in all cases).

A more robust solution would be to have a bikeshed formatter that automatically does so (a la python-isort), but I'm not aware of any.

thiagowfx:
Any concerns with moving forward with this? Is it overkill to add it to the agenda of the monthly meeting? We should probably not waste much time discussing infra there (it would be better to bikeshed about the spec itself only :P), so it would be nicer if this were resolved offline.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,40 @@
...
+# Install pre-commit itself:
+#   $ pip install pre-commit
+#     OR
+#   $ brew install pre-commit
+#
+# Update all pre-commit hooks to latest version:
+#   $ pre-commit autoupdate
+#
+# Install all pre-commit hooks locally:
+#   $ pre-commit install
```
whimboo, line 13: Shall we actually perform that automatically with the test script so that the pre-commit hook is triggered locally as well? It would be great to see the failures / errors before committing changes and not have to wait for the results in CI.

jgraham (changes requested):
I don't love this change.

I don't mind the idea of lints, but forcing them to run before you're allowed to commit changes doesn't seem ideal. Of course you can opt not to install the hooks, but then this amounts to using some relatively obscure external tool for linting.

The actual spec changes basically fall into two categories: typo fixes, which are fine, and a requirement to add boilerplate around every CDDL group, which is not.

It will take more time, but if we want to apply lints to the CDDL I suggest we invest the effort to do that directly and without requiring the spec editor to add explicit opt-ins all over the place.

thiagowfx:
> I don't mind the idea of lints, but forcing them to run before you're allowed to commit changes

Despite the name ("pre-commit"), no changes are needed for clients. In the end this will just become a presubmit check (surfaced as a github action) that will need to be addressed before the PR is merged. It does not block commits, and it does not block pushes. It would only block PR merges - and even so, only if it is configured in the repository, otherwise it would be non-blocking.

If your concern is regarding the editor experience, note that absolutely nothing changes in the current workflow. Contributors are still allowed to make changes locally and commit and push freely as they see fit. The only difference that this change makes is to add an additional check in the _PR only_ which, again, doesn't even have to be enforced.

>  Of course you can opt not to install the hooks, but then this amounts to using some relatively obscure external tool for linting.

In general, the CI logs should display what changes are needed. In principle it's not necessary to install any external tool (it's optional). This would be the case if we were installing e.g. a formatter (prettier, yapf, or similar). But in this case I am being careful not to require any external tool - if there are lint errors, all changes can be reasonably done manually, without the need of installing another tool - at least with the current set of rules I included.

> and a requirement to add boilerplate around every CDDL group, which is not.

I am fine with removing the `keep-sorted` rule (and thus all the boilerplates around CDDL groups). Your concern herein is reasonable. Other than this, and given what I wrote above, would you still be concerned with this addition to the repo?

.pre-commit-config.yaml
thiagowfx: Yes, this is feasible.

However I don't know if it's worth the effort, it depends on jgraham@, I guess :-)

whimboo: Yes, looks like he doesn't want it by default, but maybe if we keep at least some of the checks we could have an argument for the scripts to opt-in for those checks.

thiagowfx:
Closing, given lack of support / interest. Leftovers are in https://github.com/w3c/webdriver-bidi/pull/463.

PR #445, opened 2023-06-06
Fix "get shared id for a node" to use node's browsing context

whimboo:
When a script evaluation retrieves an element from a different browsing context the `get shared id for a node` method cannot use the browsing context of the realm the command is executing in, but has to use the browsing context of the node. Otherwise the reference will be added to the wrong navigable entry within seenNodes and as such can cause failures in WebDriver classic.

I'll add tests as part of https://bugzilla.mozilla.org/show_bug.cgi?id=1830884.

@jgraham can you please review?

***
<a href="https://pr-preview.s3.amazonaws.com/whimboo/webdriver-bidi/pull/445.html" title="Last updated on Jun 6, 2023, 2:00 PM UTC (bfb7bcd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/445/5d3ee40...whimboo:bfb7bcd.html" title="Last updated on Jun 6, 2023, 2:00 PM UTC (bfb7bcd)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5234,13 +5234,13 @@ To get the <dfn>handle for an object</dfn> given |realm|, |ownership type| and
 </div>
 
 <div algorithm>
-To <dfn>get shared id for a node</dfn> given |node|, |realm| and |session|:
+To <dfn>get shared id for a node</dfn> given |node|, and |session|:
 
 1. Let |node| be [=unwrapped=] |node|.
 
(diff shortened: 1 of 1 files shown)


jgraham (changes requested):
I also think we can stop passing `realm` into this algorithm.

whimboo:
With the recent push all the CI checks pass now.

PR #452, opened 2023-06-22
Serialize a context property for WindowProxy

jgraham:
This means that given e.g. an iframe element it's possible to get the context it contains by serializing its `contentWindow` attribute.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/452.html" title="Last updated on Jun 23, 2023, 1:13 PM UTC (2d98cde)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/452/ce67907...2d98cde.html" title="Last updated on Jun 23, 2023, 1:13 PM UTC (2d98cde)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5221,8 +5221,13 @@ script.NodeProperties = {
 
 script.WindowProxyRemoteValue = {
   type: "window",
+  value: script.WindowProxyProperties,
   ? handle: script.Handle,
-  ? internalId: script.InternalId,
+  ? internalId: script.InternalId
+}
+
+script.WindowProxyProperties = {
+  context: browsingContext.BrowsingContext
 }
 </pre>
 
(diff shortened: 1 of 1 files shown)


jimevans (approved):
This feels like the right approach, and if I'm reading it right, imposes no additional issues on implementers, since they already have the `browsingContext` for each thing corresponding to a window.

sadym-chromium (approved):

whimboo (approved):
Looks good to me as well.

We need tests for that as well. Problem is that we cannot easily support it in Firefox right now. So @OrKoN could someone from Google provide the relevant tests? Thanks!

thiagowfx:
> We need tests for that as well. Problem is that we cannot easily support it in Firefox right now. So @OrKoN could someone from Google provide the relevant tests? Thanks!

To be clear, should these be serialization tests, similarly to https://github.com/web-platform-tests/wpt/pull/40814?

If so, then I could look into adding them. Let me know.

whimboo:
> To be clear, should these be serialization tests, similarly to [web-platform-tests/wpt#40814](https://github.com/web-platform-tests/wpt/pull/40814)?

Yes, when we serialize a `WindowProxy` the `context` has to be present.

PR #453, opened 2023-06-23
Add `browsingContext.activate`

OrKoN:
https://github.com/w3c/webdriver-bidi/issues/171

***
<a href="https://pr-preview.s3.amazonaws.com/OrKoN/webdriver-bidi/pull/453.html" title="Last updated on Jul 7, 2023, 9:38 AM UTC (28ab483)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/453/bc1806d...OrKoN:28ab483.html" title="Last updated on Jul 7, 2023, 9:38 AM UTC (28ab483)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -183,6 +183,11 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
     text: window open steps; url: window-object.html#window-open-steps
     text: worker event loop; url: webappapis.html#worker-event-loop-2
     text: worklet global scopes; url:worklets.html#concept-document-worklet-global-scopes
+    text: traversable navigable; url:document-sequences.html#traversable-navigable
+    text: system visibility state; url: document-sequences.html#system-visibility-state
+    text: visible; url: document-sequences.html#system-visibility-state
+    text: hidden; url: document-sequences.html#system-visibility-state
+    text: navigables; url: document-sequences.html#navigables
 spec: RESOURCE-TIMING; urlPrefix: https://w3c.github.io/resource-timing/
   type: dfn
     text: convert fetch timestamp; url: dfn-convert-fetch-timestamp
@@ -2173,7 +2178,11 @@ The [=remote end steps=] with |command parameters| are:
 
 1. If |context| is not a [=top-level browsing context=], return [=error=] with [=error code=] [=invalid argument=].
 
-1. Run implementation-specific steps to bring the |context| to the top of the implementation-specific view hierarchy.
+1. Run implementation-specific steps so that |context|'s [=traversable navigable=]'s [=system visibility state=] becomes [=visible=]. If this is not possible return [=error=] with error code [=unsupported operation=].
+
+    Note: This can have the side effect of making currently [=visible=] [=navigables=] [=hidden=].
+
+    Issue: It's not clear if we should have explicit requirements for cases like a minimized window; should that cause the window to become unminimized?
 
 1. Run implementation-specific steps to set the [=top-level traversable/system focus=] on the |context| if it is not focused.
 
--- a/index.bs
+++ b/index.bs
@@ -2182,7 +2182,7 @@ The [=remote end steps=] with |command parameters| are:
 
     Note: This can have the side effect of making currently [=visible=] [=navigables=] [=hidden=].
 
-    Issue: It's not clear if we should have explicit requirements for cases like a minimized window; should that cause the window to become unminimized?
+    Note: This can change the underlying OS state by causing the window to become unminimized or by other side effects related to changing the [=system visibility state=].
 
 1. Run implementation-specific steps to set the [=top-level traversable/system focus=] on the |context| if it is not focused.
 
--- a/index.bs
+++ b/index.bs
@@ -2187,7 +2187,7 @@ The [=remote end steps=] with |command parameters| are:
 
 1. Run implementation-specific steps to set the [=top-level traversable/system focus=] on the |context| if it is not focused.
 
-    Note: This does not explicitly change the [=focused area of the document=] unless it is defined by other specifications.
+    Note: This does not change the [=focused area of the document=] except as required to by other specifications.
 
 1. Return [=success=] with data null.
 
--- a/index.bs
+++ b/index.bs
@@ -185,6 +185,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
     text: worklet global scopes; url:worklets.html#concept-document-worklet-global-scopes
     text: traversable navigable; url:document-sequences.html#traversable-navigable
     text: system visibility state; url: document-sequences.html#system-visibility-state
+    text: focused area of the document; url: document-sequences.html#focused-area-of-the-document
     text: visible; url: document-sequences.html#system-visibility-state
     text: hidden; url: document-sequences.html#system-visibility-state
     text: navigables; url: document-sequences.html#navigables
@@ -2185,6 +2186,8 @@ The [=remote end steps=] with |command parameters| are:
     Note: This can change the underlying OS state by causing the window to become unminimized or by other side effects related to changing the [=system visibility state=].
 
 1. Run implementation-specific steps to set the [=top-level traversable/system focus=] on the |context| if it is not focused.
+
+    Note: This does not explicitly change the [=focused area of the document=] unless it is defined by other specifications.
 
 1. Return [=success=] with data null.
 
--- a/index.bs
+++ b/index.bs
@@ -2142,20 +2142,20 @@ document loading at which a navigation command will return.
 
 ### Commands ### {#module-browsingContext-commands}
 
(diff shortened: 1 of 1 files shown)


jgraham (changes requested):

index.bs
```diff
@@ -2124,6 +2124,50 @@ document loading at which a navigation command will return.
 
 ### Commands ### {#module-browsingContext-commands}
 
+#### The browsingContext.bringToFront Command ####  {#command-browsingContext-bringToFront}
```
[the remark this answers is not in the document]
OrKoN: I think focusing on an element can be done via `script` evaluation so I am not sure if we would need to add it. At the same time, I feel like the name `focus` would not reflect that it also does activate the tab making it visible. Happy to rename to `focus` or `activate` though if you think this is a better name.

index.bs
```diff
@@ -2124,6 +2124,50 @@ document loading at which a navigation command will return.
...
+
+<div algorithm="remote end steps for browsingContext.bringToFront">
+
+The [=remote end steps=] with |command parameters| are:
+
+1. Let |context id| be the value of the |command parameters|["<code>context</code>"] field.
+
+1. Let |context| be the result of [=trying=] to [=get a browsing context=] with |context id|.
+
+1. If |context| is not a [=top-level browsing context=], return [=error=] with [=error code=] [=invalid argument=].
```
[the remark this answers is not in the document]
OrKoN: I think there is no fundamental limitation but the main motivation for this command is changing the system focus (bring the tab to the front) and frames can be focused via script evaluation AFAIK.

index.bs
```diff
@@ -2124,6 +2124,50 @@ document loading at which a navigation command will return.
...
+
+1. Run implementation-specific steps to bring the |context| to the top of the implementation-specific view hierarchy.
```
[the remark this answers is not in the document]
OrKoN: Thanks! I get an error for `[=system visibility state=]`:

```
 LINE ~2165: No 'dfn' refs found for 'system visibility state' that are marked for export.
[=system visibility state=]
```

does it mean it cannot be referenced?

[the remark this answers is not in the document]
jgraham: It seems to not be marked for export. In theory we could ask for it to be marked for export, but in practice you can just add it to the HTML spec part of the `<pre class=anchors>` section at the top (hopefully it's obvious from the existing examples how to do that).

thiagowfx (approved):

index.bs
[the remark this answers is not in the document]

index.bs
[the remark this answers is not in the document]
whimboo: I for myself would vote for `activate` given that this better reflects the activation of the tab.

Do we need special steps to ensure that the window itself is in the foreground (top-most) in case remote ends cannot correctly handle the focus when a window is hidden behind another one? In Firefox this is not a problem, but I wonder about other implementations.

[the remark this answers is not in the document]
thiagowfx: I have no objections with `activate()`.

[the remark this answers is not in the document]
jrandolf: +1 for `activate`.

thiagowfx (approved):

whimboo:
@jgraham do you still have any objections? Please let us know. Otherwise I would say we could merge the PR.

@OrKoN are you going to create wdspec tests for this new command?

index.bs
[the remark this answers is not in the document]
mathiasbynens: +1 to `activate()`

mathiasbynens (approved):

index.bs
```diff
@@ -2137,6 +2142,54 @@ document loading at which a navigation command will return.
...
+
+1. Let |context| be the result of [=trying=] to [=get a browsing context=] with |context id|.
+
+1. If |context| is not a [=top-level browsing context=], return [=error=] with [=error code=] [=invalid argument=].
+
+1. Run implementation-specific steps so that |context|'s [=traversable navigable=]'s [=system visibility state=] becomes [=visible=]. If this is not possible return [=error=] with error code [=unsupported operation=].
+
+    Note: This can have the side effect of making currently [=visible=] [=navigables=] [=hidden=].
+
+    Issue: It's not clear if we should have explicit requirements for cases like a minimized window; should that cause the window to become unminimized?
```
mathiasbynens, line 2185: https://html.spec.whatwg.org/multipage/interaction.html#page-visibility:system-visibility-state says:

> A [traversable navigable](https://html.spec.whatwg.org/multipage/document-sequences.html#traversable-navigable)'s [system visibility state](https://html.spec.whatwg.org/multipage/document-sequences.html#system-visibility-state), including its initial value upon creation, is determined by the user agent. It represents, for example, whether the browser window is minimized, a browser tab is currently in the background, or a system element such as a task switcher obscures the page.

To me, this reads as if we might already have these requirements implicitly with the current wording, and there’s no further action needed.

mathiasbynens (approved):

OrKoN:
@whimboo yes, I will start on WPT

index.bs
jgraham: I think if we're assuming that this has to adjust the underlying OS state in some cases we should call that out at least as a note. It also means we can't test this command correctly without using classic for window manipulation, or adding the window manipulation commands to BiDi.

index.bs
```diff
@@ -2137,6 +2142,54 @@ document loading at which a navigation command will return.
...
+
+1. Run implementation-specific steps to set the [=top-level traversable/system focus=] on the |context| if it is not focused.
```
jgraham, line 2187: So, to be clear, this doesn't change the [focused area of the document](https://html.spec.whatwg.org/#focused-area-of-the-document) in the browsing context? That might also be worth calling out.

jgraham (changes requested):

index.bs
OrKoN, line 2185:
```suggestion
    Note: This can change the underlying OS state by causing the window to become unminimized or by other side effects related to changing the [=system visibility state=].
```

OrKoN: @jgraham should we phrase it like this then? (given the comment from Mathias and your response above)

jgraham: That looks fine.

index.bs
OrKoN: I have been reading spec parts about the focused area of the document but the term is still not completely clear to me. My current understanding is that any DOM element can be the focused area of the document and so I would expect the focused area to change if the system focus + making the  change causes it to: for example, if you insert an element with an autofocus attribute into a background tab (the focused element does not change), as soon as you activate the context, the autofocus applies (it looks like the system focus makes the document fully active and that will apply various focus algorithms based on that). Any suggestions on the best way to describe this here?

OrKoN: Added a note that the focused area of the document should not be explicitly changed.

jgraham: OK, I think that's enough. I was thinking of the case where there's already an element with focus, to make it clear that we shouldn't end up with focus on the document (or somewhere in the browser UI).

index.bs
```diff
@@ -2137,6 +2143,56 @@ document loading at which a navigation command will return.
...
+
+1. Run implementation-specific steps so that |context|'s [=traversable navigable=]'s [=system visibility state=] becomes [=visible=]. If this is not possible return [=error=] with error code [=unsupported operation=].
+
+    Note: This can have the side effect of making currently [=visible=] [=navigables=] [=hidden=].
+
+    Note: This can change the underlying OS state by causing the window to become unminimized or by other side effects related to changing the [=system visibility state=].
+
+1. Run implementation-specific steps to set the [=top-level traversable/system focus=] on the |context| if it is not focused.
+
+    Note: This does not explicitly change the [=focused area of the document=] unless it is defined by other specifications.
```
jgraham, line 2190:
```suggestion
    Note: This does not change the [=focused area of the document=] except as required to by other specifications.
```

jgraham (approved):

index.bs
OrKoN: I think this is generally the behaviour in our implementation (no explicit change of the focused area) but we found one edge case is when a new document with an element with the autofocus attribute is inserted while the page is in the background. In that situation, the element gets focused on activation but I believe this is correct according to the HTML spec.

jrandolf: If someone inserts an element into a background tab with autofocus, then, although activation will trigger the focus, the initial cause was the person adding the autofocus element into the background tab. They presumably did this knowing this will cause activation to trigger focus, so in truth, activation by itself didn't change the focused area of the document, but a combination of actions. So it still applies that activation (by itself) doesn't trigger a change in focus area.

OrKoN:
I will start working on WPT tests this week.

jgraham (approved):

OrKoN:
Draft tests https://github.com/web-platform-tests/wpt/pull/40964

PR #454, opened 2023-06-23
Add support for clipping screenshots

jgraham:
Support clipping screenshots, either directly, by providing coordinates relative to the viewport, or by providing an element, and using its bounding rect as the coordinates to clip.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/454.html" title="Last updated on Jun 30, 2023, 1:33 PM UTC (6836090)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/454/88f40d0...6836090.html" title="Last updated on Jun 30, 2023, 1:33 PM UTC (6836090)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2170,10 +2170,10 @@ Base64-encoded string.
 
       browsingContext.BoxClipRectangle = {
          type: "viewport",
-         x: js-uint,
-         y: js-uint,
-         width: js-uint,
-         height: js-uint
+         x: float,
+         y: float,
+         width: float,
+         height: float
       }
       </pre>
    </dd>
@@ -2272,9 +2272,13 @@ To <dfn>render viewport to a canvas</dfn> given |viewport| and |rect|:
 
 1. Let |scale| be [=determine the device pixel ratio=] for the output device of |viewport|.
 
-1. Let |paint width| be |rect|'s [=width dimension=] multiplied by |scale|.
+1. Let |paint width| be |rect|'s [=width dimension=] multiplied by |scale|,
+   rounded to the nearest integer, so it matches the width of |rect| in device
+   pixels.
 
-1. Let |paint height| be |rect|'s [=height dimension=] multiplied by |scale|.
+1. Let |paint height| be |rect|'s [=height dimension=] multiplied by |scale|,
+   rounded to the nearest integer, so it matches the height of |rect| in device
+   pixels.
 
 1. Let |canvas| be a new {{HTMLCanvasElement}} with {{HTMLCanvasElement/width}} |paint
    width| and {{HTMLCanvasElement/height}} |paint height|.
--- a/index.bs
+++ b/index.bs
@@ -2203,10 +2203,10 @@ Note: This ensures that the resulting rect has positive [=width dimension=] and
 
 1. Let |height| be |rect|'s [=height dimension=].
 
-1. If |width| is less than 0, set |x| to |x| - |width| and then set |width| to
+1. If |width| is less than 0, set |x| to |x| + |width| and then set |width| to
    -|width|.
 
-1. If |height| is less than 0, set |y| to |y| - |height| and then set |height|
+1. If |height| is less than 0, set |y| to |y| + |height| and then set |height|
    to -|height|.
 
 1. Return a new {{DOMRectReadOnly}} with [=x coordinate=] |x|, [=y coordinate=]
--- a/index.bs
+++ b/index.bs
@@ -74,6 +74,7 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
     text: remote end steps; url: dfn-remote-end-steps
     text: remote end; url: dfn-remote-ends
     text: reset the input state; url: dfn-reset-the-input-state
+    text: scroll into view; url: dfn-scrolls-into-view
     text: session ID; url: dfn-session-id
     text: session not created; url: dfn-session-not-created
     text: session; url: dfn-sessions
@@ -81,6 +82,7 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
     text: success; url: dfn-success
     text: try; url: dfn-try
     text: trying; url: dfn-try
+    text: unable to capture screen; url: dfn-unable-to-capture-screen
     text: unknown command; url: dfn-unknown-command
     text: unknown error; url: dfn-unknown-error
     text: unsupported operation; url: dfn-unsupported-operation
@@ -143,8 +145,14 @@ spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
 spec: GEOMETRY; urlPrefix: https://drafts.fxtf.org/geometry/
   type: dfn
     text: rectangle; url: rectangle
+    text: x coordinate; url: rectangle-x-coordinate
+    text: y coordinate; url: rectangle-y-coordinate
+    text: width dimension; url: rectangle-width-dimension
+    text: height dimension; url: rectangle-height-dimension
 spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
   type: dfn
+    text: 2D; url: canvas.html#concept-canvas-2d
+    text: 2D context creation algorithm; url: canvas.html/#2d-context-creation-algorithm
     text: a browsing context is discarded; url: window-object.html#a-browsing-context-is-discarded
     text: active browsing context; url: document-sequences.html#nav-bc
     text: active document; for: browsing context; url: document-sequences.html#active-document
@@ -154,6 +162,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
     text: create a classic script; url: webappapis.html#creating-a-classic-script
     text: create a new browsing context; url: browsers.html#creating-a-new-browsing-context
     text: confirm; url: timers-and-user-prompts.html#dom-confirm
+    text: context mode; url: /canvas.html#offscreencanvas-context-mode
     text: default classic script fetch options; url: webappapis.html#default-classic-script-fetch-options
     text: environment settings object's Realm; url: webappapis.html#environment-settings-object's-realm
     text: handled; url: webappapis.html#concept-error-handled
@@ -183,6 +192,9 @@ spec: HR-TIME; urlPrefix: https://w3c.github.io/hr-time/
 spec: RFC4648; urlPrefix: https://tools.ietf.org/html/rfc4648
   type: dfn
     text: Base64 Encode; url: section-4
+spec: CSSOM-VIEW; urlPrefix: https://drafts.csswg.org/cssom-view/
+  type: dfn
+    text: visual viewport; url: #visual-viewport
 </pre>
 
 <pre class="biblio">
@@ -488,6 +500,7 @@ ErrorCode = ("invalid argument" /
              "no such node" /
              "no such script" /
              "session not created" /
+             "unable to capture screen" /
              "unable to close browser" /
              "unknown command" /
              "unknown error" /
@@ -2140,7 +2153,25 @@ Base64-encoded string.
       }
 
       browsingContext.CaptureScreenshotParameters = {
-        context: browsingContext.BrowsingContext
+        context: browsingContext.BrowsingContext,
+        ?clip: browsingContext.ClipRectangle
+      }
+
+      browsingContext.ClipRectangle = (browsingContext.ElementClipRectangle //
+                                       browsingContext.BoxClipRectangle)
+
+      browsingContext.ElementClipRectangle = {
+        type: "element",
+        element: script.SharedReference,
+        ?scrollIntoView: bool
+      }
+
+      browsingContext.BoxClipRectangle = {
+         type: "viewport",
+         x: js-uint,
+         y: js-uint,
+         width: js-uint,
+         height: js-uint
       }
       </pre>
    </dd>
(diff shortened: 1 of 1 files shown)


OrKoN:
@sadym-chromium PTAL

index.bs
```diff
@@ -2140,7 +2153,25 @@ Base64-encoded string.
       }
 
       browsingContext.CaptureScreenshotParameters = {
-        context: browsingContext.BrowsingContext
+        context: browsingContext.BrowsingContext,
+        ?clip: browsingContext.ClipRectangle
```
whimboo, line 2157:
```suggestion
        ? clip: browsingContext.ClipRectangle
```

index.bs
```diff
@@ -2140,7 +2153,25 @@ Base64-encoded string.
...
+      }
+
+      browsingContext.ClipRectangle = (browsingContext.ElementClipRectangle //
+                                       browsingContext.BoxClipRectangle)
```
whimboo, line 2161:
```suggestion
      browsingContext.ClipRectangle = (
        browsingContext.BoxClipRectangle //
        browsingContext.ElementClipRectangle
      )
```

index.bs
```diff
@@ -2140,7 +2153,25 @@ Base64-encoded string.
...
+
+      browsingContext.ElementClipRectangle = {
+        type: "element",
+        element: script.SharedReference,
+        ?scrollIntoView: bool
```
whimboo, line 2166:
```suggestion
        ? scrollIntoView: bool
```

Question: Do we actually capture the full element when parts of it are outside the visible viewport? Or would be the `full` screenshot support necessary for that?

index.bs
```diff
@@ -2156,10 +2187,117 @@ Base64-encoded string.
...
+Note: This ensures that the resulting rect has positive [=width dimension=] and
+[=height dimension=].
+
+1. Let |x| be |rect|'s [=x coordinate=].
+
+1. Let |width| be |rect|'s [=width dimension=].
+
+1. Let |y| be |rect|'s [=y coordinate=].
+
+1. Let |height| be |rect|'s [=height dimension=].
```
whimboo, line 2202:
```suggestion
1. Let |y| be |rect|'s [=y coordinate=].

1. Let |width| be |rect|'s [=width dimension=].

1. Let |height| be |rect|'s [=height dimension=].
```

index.bs
```diff
@@ -2156,10 +2187,117 @@ Base64-encoded string.
...
+
+1. If |width| is less than 0, set |x| to |x| - |width| and then set |width| to
```
whimboo, line 2204: The width and height cannot be smaller than `0` given that both are of type `js-uint`.

index.bs
```diff
@@ -2169,16 +2307,65 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+         |clip|["<code>scrollIntoView</code>"]:
+
+        1. [=Scroll into view=] with |element|.
+
+      1. Let |clip rect| be [=get the bounding box=] for |element|.
+
+    Otherwise, |clip| matches the
+    <code>browsingContext.BoxClipRectange</code> production:
+
+     1. Let |clip rect| be a {{DOMRectReadOnly}} with x coordinate |clip|["<code>x</code>"],
```
whimboo, line 2355:
```suggestion
      1. Let |clip rect| be a {{DOMRectReadOnly}} with x coordinate |clip|["<code>x</code>"],
```

index.bs
```diff
@@ -2156,10 +2187,117 @@ Base64-encoded string.
...
+
+1. Return a new {{DOMRectReadOnly}} with [=x coordinate=] |x_0|, [=y coordinate=]
+   |y_0|, [=width dimension=] |width| and [=height dimension=] |height|.
+
+</div>
+
+<div algorithm>
+To <dfn>render viewport to a canvas</dfn> given |viewport| and |rect|:
+
+1. Assert: |rect| is normlaised and represents a non-empty sub-region of |viewport|,
```
whimboo, line 2260:
```suggestion
1. Assert: |rect| is normalized and represents a non-empty sub-region of |viewport|,
```

whimboo (changes requested):
Generally it looks fine. Please see the various inline comments.

index.bs
jgraham: No, we always clip to the viewport. I don't know if that's the behaviour of classic, but if you want the full visible part of the element in the future we'll add full-document screenshot support and combining _that_ with an element clip will get you the full element (or at least any part of it that's visible).

index.bs
jgraham: Note to self: those types are wrong. And we probably need to round to pixels after scaling when we actually render.

index.bs
whimboo: Ok, that sounds fine with me.

index.bs
whimboo: I assume that you are still planning to do an update for that? I'll defer the next review until you closed this issue.

jgraham: Update now.

index.bs
```diff
@@ -2169,16 +2313,65 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+1. Let |viewport| be |document|'s [=visual viewport=].
+
 1. Immediately after the next invocation of the [=run the animation frame
    callbacks=] algorithm for |document|:
 
    Issue: This ought to be integrated into the update rendering algorithm
    in some more explicit way.
 
- 1. Let |root rect| be |document|’s [=document element=]'s [=rectangle=].
+1. Let |viewport rect| be a {{DOMRectReadOnly}} with [=x coordinate=] 0, [=y
```
whimboo, line 2324: This is only used in the otherwise section below in line 2367 so we potentiall can get this moved there.

index.bs
```diff
@@ -2169,16 +2313,65 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+   coordinate=] 0, [=width dimension=] |viewport| width, and [=height
+   dimension=] |viewport| height.
+
+  1. If |command parameters| contains "<code>clip</code>":
```
whimboo, line 2328: The numeration is off here. It should not be in the 2nd level. This applies to other lines below as well.

index.bs
```diff
@@ -2140,7 +2153,27 @@ Base64-encoded string.
...
+        element: script.SharedReference,
+        ? scrollIntoView: bool
+      }
+
+      browsingContext.BoxClipRectangle = {
+         type: "viewport",
+         x: float,
+         y: float,
+         width: float,
+         height: float
```
whimboo, line 2176: Out of interest, which scenarios do you think that we can handle with negative width and height?

jgraham: The DOMRect allows the coordinates to be arbitrary floats, so this is matching that. We end up normalizing to positive width/height in the algorithm anyway (basically just subtracting the width from x if it's negative and then replacing it with its absolute value). But also we have to allow fractional values for high dpi screens, and there isn't an easy way to specify non-negative floats, so I think this makes sense.

whimboo: Thanks. That's fair.

index.bs
```diff
@@ -2169,16 +2313,65 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+
+      1. If |clip| contains "<code>scrollIntoView</code>" and
+         |clip|["<code>scrollIntoView</code>"]:
+
+        1. [=Scroll into view=] with |element|.
+
+      1. Let |clip rect| be [=get the bounding box=] for |element|.
+
+    Otherwise, |clip| matches the
+    <code>browsingContext.BoxClipRectange</code> production:
```
thiagowfx, line 2359: typo

thiagowfx:
LGTM at a high-level.

index.bs
```diff
@@ -154,6 +162,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
     text: create a classic script; url: webappapis.html#creating-a-classic-script
     text: create a new browsing context; url: browsers.html#creating-a-new-browsing-context
     text: confirm; url: timers-and-user-prompts.html#dom-confirm
+    text: context mode; url: /canvas.html#offscreencanvas-context-mode
```
juliandescottes, line 165:
```suggestion
    text: context mode; url: canvas.html#offscreencanvas-context-mode
```

juliandescottes (approved):
lgtm

index.bs
jgraham: I generally rearranged this section to make it easier to follow.

jgraham:
One question: are people happy with using "`viewport`" as the name for the explicit offset coordinate system? The only reason not to would be fullscreen screenshots, where the top left of the viewport would presumably be the top-left of the document (although I suppose we could also say it's actually the top-left of the original viewport, although I don't know quite what the use case for that is).

whimboo:
> One question: are people happy with using "`viewport`" as the name for the explicit offset coordinate system? The only reason not to would be fullscreen screenshots, where the top left of the viewport would presumably be the top-left of the document (although I suppose we could also say it's actually the top-left of the original viewport, although I don't know quite what the use case for that is).

Well, what is the original viewport? Think about that a custom viewport could have been set via `browsingContext.setViewport`. Then this one should be the reference for calculating the rectangle for clipping.

jgraham:
By "original viewport" I mean whatever's being used for layout of the page. That would indeed include any webdriver specific overrides.

The way I'd expect full page screenshots to work is to define a new viewport that has the dimensions of the scroll width and scroll height of the document, and then render to that viewport. This is extremely sketchy because we generally don't have more than one rendering of a given document at a time, and presumably we want this rendering to happen without firing any events (e.g. paint events, intersection observer events, etc.). Maybe print preview is the closest existing analogy.

If we didn't do that, and tried to say that you reuse the existing viewport but render even those parts which are outside the visible area then you'd end up with different behaviour for elements that are positioned relative to the viewport e.g. a `position:fixed` heading with `top: ; left: 0` on a document that's scrolled would render with its top left at  `(scrollX, scrollY)` (in document coordinates) rather than at `(0, 0)` as you might expect.

The API-visible effect of this is that when doing a full-page screenshot "viewport" coordinates would be relative to this second viewport, rather than the actual viewport being used to display the document on the screen. That makes sense if you understand the model, but you do need to understand the model.

whimboo (commented):
Looks fine to me. If something will not work that well we could always have a follow-up.

We can take care of the wpt tests when we start the implementation in Firefox via https://bugzilla.mozilla.org/show_bug.cgi?id=1840998.

whimboo (approved):

thiagowfx (approved):

thiagowfx:
Spec looks fine, we didn't have a chance to prototype it in Chromium yet though. @OrKoN do you also want to take a look before submission?

OrKoN:
Are there WPT tests in progress for this feature?

whimboo:
> Are there WPT tests in progress for this feature?

See my review comment above. We could get started working on it soon. So let us know if we can already get these PR merged.

index.bs
```diff
@@ -2156,10 +2189,121 @@ Base64-encoded string.
...
+
+1. Let |x| be |rect|'s [=x coordinate=].
+
+1. Let |y| be |rect|'s [=y coordinate=].
+
+1. Let |width| be |rect|'s [=width dimension=].
+
+1. Let |height| be |rect|'s [=height dimension=].
+
+1. If |width| is less than 0, set |x| to |x| - |width| and then set |width| to
```
OrKoN, line 2206: should it be `|x| + |width|` given that the width is negative? IIUC, the idea is to flip the origin of the rect? for example, `x = 10; width = -10;` should become `x = 0; width = 10;`? unless `|width|` means taking an absolute number?

ollie-iterators:
This seems like an important PR. I have been trying to find a way to get screenshots of just an element before

index.bs
thiagowfx: Good catch! The `|` definitely make it seem like a module (absolute value), but in this context it is not. I believe your comment is correct.

Likewise, the same applies to the `height`.

jgraham: Indeed, good catch.

index.bs
```diff
@@ -2156,10 +2189,121 @@ Base64-encoded string.
...
+1. If |height| is less than 0, set |y| to |y| + |height| and then set |height|
+   to -|height|.
+
+1. Return a new {{DOMRectReadOnly}} with [=x coordinate=] |x|, [=y coordinate=]
+   |y|, [=width dimension=] |width| and [=height dimension=] |height|.
+
+</div>
+
+<div algorithm>
+To <dfn>rectangle intersection</dfn> given |rect1| and |rect2|
```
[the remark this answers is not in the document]
jgraham: Generally defining things in terms of JS is a bad idea; it's very heavyweight to call JS code from a spec context, and if you're not careful it leaves a bunch of questions unanswered (e.g. in this case, which realm are you running in? what happens if `IntersectionObserver` is redefined in that realm? What happens if the code throws an exception? Are you correctly waiting for promises to resolve? How do you read the result back into spec types?). 

For all these reasons, it's strongly preferred to call spec algorithms rather than trying to define things in terms of js. In this case the underlying spec algorithm seems to be https://w3c.github.io/IntersectionObserver/#compute-the-intersection but I think that doesn't quite fit our requirements, which really are just about figuring out the overlap of two rectangles.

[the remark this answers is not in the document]
OrKoN: Makes sense!

OrKoN (approved):
LGTM

PR #460, opened 2023-07-03
Use parenthesis for base types

jrandolf:
CDDL doesn't allow composition with non-groups.

Basically, you have the following ABNF:

```
   grpent = [occur S] [memberkey S] type
            / [occur S] groupname [genericarg]  ; preempted by above
            / [occur S] "(" S group S ")"
```

When you you don't use parenthesis, it matches the first branch, but this is bad because for maps (types declared with { ... }), this is not valid ("If the memberkey is not given, the entry can only be used for matching arrays, not for maps." (Ctrl + F this)).

Thus we need to declare the base types with parenthesis.

See https://www.rfc-editor.org/rfc/rfc8610.html#appendix-C for the relevant spec.

***
<a href="https://pr-preview.s3.amazonaws.com/jrandolf/webdriver-bidi/pull/460.html" title="Last updated on Jul 3, 2023, 9:23 AM UTC (1431ad5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/460/e8bb82d...jrandolf:1431ad5.html" title="Last updated on Jul 3, 2023, 9:23 AM UTC (1431ad5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3502,13 +3502,13 @@ empty map.  It's used to track the network events for which a
 #### The network.BaseParameters Type #### {#type-network-BaseParameters}
 
 <pre class="cddl local-cddl">
-network.BaseParameters = {
+network.BaseParameters = (
     context: browsingContext.BrowsingContext / null,
     navigation: browsingContext.Navigation / null,
     redirectCount: js-uint,
     request: network.RequestData,
     timestamp: js-uint,
-}
+)
 </pre>
 
 The <code>network.BaseParameters</code> type is an abstract type representing
@@ -4948,10 +4948,10 @@ script.RealmInfo = (
   script.WorkletRealmInfo
 )
 
-script.BaseRealmInfo = {
+script.BaseRealmInfo = (
   realm: script.Realm,
   origin: text
-}
+)
 
 script.WindowRealmInfo = {
   script.BaseRealmInfo,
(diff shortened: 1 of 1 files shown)


OrKoN:
@whimboo @jgraham could you please take a look?

thiagowfx (approved):

jgraham (approved):
OK. I don't really understand if this is intended to be syntactically invalid (obviously it matches the ABNF, but maybe violates the additional prose rules), or if it matches something but not what we want (in which case what?).

In either case it's annoying to have constructs that the tooling accepts but which apparently don't make sense :/

jrandolf:
The difference is not readily apparent. You see the difference when you consider arrays.

For example, consider the following:

```
apple = ( uint )
banana = ( apple )
mango = [
 * banana
]
banana-2 = { apple }
mango-2 = [
 * banana-2
]
```

`banana` is a group, so what this means is `mango = apple[]` in TS because groups are spread. `banana-2` is a type, so what this means is `mango-2 = banana-2[]` because types are not spread.

Fun fact:

You can have the CDDL `test = ( test )` which is valid, but literally means nothing for tooling because you can't determine if `test` is a group name or a type name. If you had `test = ( test .default 5 )`, the `test` would be a type because `.default` can only be used on types.

PR #461, opened 2023-07-03
Use `.and` instead of composition

jrandolf:
CDDL doesn't allow composition of maps.

See https://github.com/w3c/webdriver-bidi/pull/460

***
<a href="https://pr-preview.s3.amazonaws.com/jrandolf/webdriver-bidi/pull/461.html" title="Last updated on Jul 3, 2023, 1:53 PM UTC (9442d71)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/461/e8bb82d...jrandolf:9442d71.html" title="Last updated on Jul 3, 2023, 1:53 PM UTC (9442d71)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5299,16 +5299,14 @@ script.FunctionRemoteValue = {
 }
 
 script.RegExpRemoteValue = {
-  script.RegExpLocalValue,
   ? handle: script.Handle,
   ? internalId: script.InternalId,
-}
+} .and script.RegExpLocalValue
 
 script.DateRemoteValue = {
-  script.DateLocalValue,
   ? handle: script.Handle,
   ? internalId: script.InternalId,
-}
+} .and script.DateLocalValue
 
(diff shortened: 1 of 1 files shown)


thiagowfx (approved):

jgraham (changes requested):
I don't think this is right? 

```

   A ".and" control on a type indicates that the data item matches both
   the left-hand-side type and the type given as the right-hand side.
   (Formally, the resulting type is the intersection of the two types
   given.)
```

We don't want an intersection, but a union.

jrandolf:
I'm pretty sure we want an intersection type. We want `script.RegExpRemoteValue` to contain the keys for `script.RegExpLocalValue` and the keys presented in the map.

jgraham (approved):
OK, based on discussions, it seems like I didn't understand what CDDL intended here.

Issue #444, opened 2023-06-06
Move CDDL job from Node 12 to Node 16 due to deprecated features


whimboo:
See the annotations for the CDDL job:
https://github.com/w3c/webdriver-bidi/actions/runs/5186558692?pr=442

> Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions-rs/toolchain@v1. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

and

> The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

thiagowfx:
One idea: https://github.com/actions-rs/toolchain/issues/216#issuecomment-1291613319

dtolnay:
> This is likely blocked on [actions-rs/toolchain#219](https://github.com/actions-rs/toolchain/issues/219) and [actions-rs/toolchain#221](https://github.com/actions-rs/toolchain/issues/221)

The whole actions-rs org is abandoned for years, so I would recommend not blocking yourself on it being revived.

whimboo:
Thank you David! It's great to see that `[dtolnay/rust-toolchain](https://github.com/dtolnay/rust-toolchain)` exists and works fine for our purpose.

PR #470, opened 2023-07-05
Switch to new rust-toolchain because actions-rs/toolchain is unmaintained

whimboo:
Lets see if the [`dtolnay/rust-toolchain`](https://github.com/dtolnay/rust-toolchain) works for us so that we can get rid of the deprecation warnings.

Closes: #444


whimboo:
This actually works just fine and will fix #444.

thiagowfx (approved):

OrKoN (approved):

PR #471, opened 2023-07-05
Add method enum for script domain

jrandolf:
This PR adds some helper types for understanding relevant method names for events and commands.

If this is merged, I'll add this for other domains as well.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/471.html" title="Last updated on Jul 5, 2023, 8:34 AM UTC (6ea7d48)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/471/4f6ffa4...6ea7d48.html" title="Last updated on Jul 5, 2023, 8:34 AM UTC (6ea7d48)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4275,6 +4275,15 @@ ScriptCommand = (
   script.GetRealms //
   script.RemovePreloadScriptCommand
 )
+
+ScriptCommandMethod = (
+  "script.addPreloadScript" /
+  "script.callFunction" /
+  "script.disown" /
+  "script.evaluate" /
+  "script.getRealms" /
+  "script.removePreloadScript"
+)
 </pre>
 
 [=local end definition=]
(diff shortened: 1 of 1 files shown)


jgraham:
How important is this to you?

From my point of view this makes the spec harder to write and somewhat harder to read (the `.within` isn't really adding any useful information for humans, so is basically just noise). I can see why you'd want it for a generic CDDL-to-interface conversion, but is there an alternate possibility of writing the logic to generate command name / event name enums as a special case?

OrKoN (commented):
I am not convinced that we want to add duplication to the spec for the sake of simplifying generators. The tools could probably collect all method literals if they need to be aggregated in a type (for validation or other purposes). As far as I can see, validation of the literal types without an union type is possible too.

thiagowfx:
I am in favor of improving the robustness of the generator but indeed this change adds redundancy (and maintenance difficulty) to the spec. If you happen to find another (simpler) way, feel free to send another PR.

christian-bromann:
> The CDDL generator I created

Hey @jrandolf 👋 I am interested in your generator. I build one myself called [`cddl2ts`](https://www.npmjs.com/package/cddl2ts). Can you share anything about it?

jrandolf:
Sure! I think you are also on the Matrix group right? We can talk there!

PR #476, opened 2023-07-06
Add regular expression for page ranges

jrandolf:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/476.html" title="Last updated on Jul 6, 2023, 4:50 PM UTC (69580e7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/476/ddabede...69580e7.html" title="Last updated on Jul 6, 2023, 4:50 PM UTC (69580e7)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2737,13 +2737,15 @@ PDF document represented as a Base64-encoded string.
         params: browsingContext.PrintParameters
       }
 
+      browsingContext.PageRange = text .regexp "^(?:[0-9]+)?(?:-(?:[0-9]+)?)?$"
+
       browsingContext.PrintParameters = {
         context: browsingContext.BrowsingContext,
         ? background: bool .default false,
         ? margin: browsingContext.PrintMarginParameters,
         ? orientation: ("portrait" / "landscape") .default "portrait",
         ? page: browsingContext.PrintPageParameters,
-        ? pageRanges: [*(js-uint / text)],
+        ? pageRanges: [*(js-uint / browsingContext.PageRange )],
         ? scale: (0.1..2.0) .default 1.0,
         ? shrinkToFit: bool .default true,
       }


index.bs
```diff
@@ -2743,7 +2743,7 @@ PDF document represented as a Base64-encoded string.
         ? margin: browsingContext.PrintMarginParameters,
         ? orientation: ("portrait" / "landscape") .default "portrait",
         ? page: browsingContext.PrintPageParameters,
-        ? pageRanges: [*(js-uint / text)],
+        ? pageRanges: [*(js-uint / text .regexp "^[0-9]+(?:-[0-9]+)?$" )],
```
thiagowfx, line 2746: The algorithm (https://w3c.github.io/webdriver/#dfn-parse-a-page-range) is responsible to validate this.

Adding a regexp here is "premature optimization".

I don't think it hurts, and it happens to play well with our zod generator, but see what the Mozilla folks think about it.

PS. This matches "1-3", but it does not match "1-" nor "-3", however AFAIK it should (please double check).

index.bs
```diff
@@ -2737,13 +2737,15 @@ PDF document represented as a Base64-encoded string.
         params: browsingContext.PrintParameters
       }
 
+      browsingContext.PageRange = text .regexp "^(?:[0-9]+)?(?:-(?:[0-9]+)?)?$"
```
jimevans, line 2740: Is the intent here just to limit what constitutes a valid value for the text the user sends, if they choose to send text as the argument for `pageRanges`?

index.bs
thiagowfx: Now it currently matches "-". Is it allowed?

index.bs
thiagowfx: Yes. For example, "5-3" would be matched however it's not valid (AFAIUI). I think @jrandolf is only interested in the overall text format, not necessarily in the full validation.

Still - see comment above - "`'-'" is matched but it should not be allowed.

jrandolf: Indeed.

jimevans: Fair enough. Nothing to do for local ends/clients, then. This is all on the remote-end implementation.

index.bs
jrandolf: Yes, it is allowed.

thiagowfx (approved):

index.bs
jgraham, line 2740: So, I'm nervous about introducing regex into the spec, because I suspect in practice there are enough differences between dialects that they are likely to lead to interop problems.

In particular the regexp above is wrong per spec (though probably works OK in a js implementation) because CDDL specifies that it follows https://www.w3.org/TR/2004/REC-xmlschema-2-20041028/#regexs. That implicitly anchors expressions, so `^` and `$` match literals rather than start/end of the string. I also don't think it has capture groups, so I'm not sure what `?:` would actually do (I think the `:` would match itself, but the `?` might be a syntax error? Not sure; I've literally never looked at this spec before).

Given that, I think we'd either have to intentionally violate the CDDL spec and pick a different syntax (e.g. ECMAScript, although that itself might not work for code generation in languages with different regex implementations e.g. Go or Rust), or use the XSD format and rely on everyone implementing a conversion into an actual implemented regex syntax. Or just not use regex and avoid the entire problem, at the cost of having to write a bit of extra code whenever there's string parsing. Which is hopefully "almost never", since we can generally choose to use structured data instead.

So I think my preference is to drop this PR.

jgraham (changes requested):

thiagowfx (changes requested):
Removing approval given @jgraham's comment.

index.bs
jrandolf: I'd be in favor of the universal minimal syntax. Based on the XSD, it looks like that format is the minimal.

That said, how about we change page ranges to use `(* [js-uint,js-uint] | js-uint)`? I think we can avoid this all together if we just allowed this.

thiagowfx: The string possibility comes from webdriver classic: https://w3c.github.io/webdriver/#print

I believe we need to be permissive in order to keep compatibility with it, otherwise it would be a regression/downgrade.

jgraham: `(* [js-uint,js-uint] | js-uint)` would also be wrong because one can omit the ends of the range. So we'd have to allow `null` in either the first or second position, but not both. I don't think this would be incompatible with the goal of having classic be implementable on top of BiDi, but it would add some complexity. The main argument against using the fully array-based syntax is that it's trickier to write by hand (and less like what you write in the print dialog).

jrandolf:
Closing in favor of https://github.com/w3c/webdriver-bidi/pull/481

PR #477, opened 2023-07-06
Use `script.LocalValue` for arguments and `this`

jrandolf:
According to the steps, `script.LocalValue` must include `script.RemoteReference` and `script.ChannelValue`. A priori, there is no purpose in `script.ArgumentValue`

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/477.html" title="Last updated on Jul 10, 2023, 5:14 PM UTC (829f709)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/477/cbc17e9...829f709.html" title="Last updated on Jul 10, 2023, 5:14 PM UTC (829f709)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4440,7 +4440,7 @@ To <dfn export>run WebDriver BiDi preload scripts</dfn> given |environment setti
 script.Channel = text;
 </pre>
 
-The <code>script.Channel</code> type represents the id of a specific channel
+The <code><dfn>script.Channel</dfn></code> type represents the id of a specific channel
 used to send custom messages from the [=remote end=] to the [=local end=].
 
 #### The script.ChannelValue Type #### {#type-script-ChannelValue}
@@ -4580,11 +4580,13 @@ strong [=/map=] from handle ids to their corresponding objects.
 
 #### The script.LocalValue Type #### {#type-script-LocalValue}
 
-[=Remote end definition=] and [=local end definition=]
+[=Remote end definition=]
 
-<pre class="cddl remote-cddl local-cddl">
+<pre class="cddl remote-cddl">
 script.LocalValue = (
+  script.RemoteReference //
   script.PrimitiveProtocolValue //
+  script.ChannelValue //
   script.ArrayLocalValue //
   script.DateLocalValue //
   script.MapLocalValue //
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -4528,7 +4528,9 @@ strong [=/map=] from handle ids to their corresponding objects.
 
 <pre class="cddl remote-cddl local-cddl">
 script.LocalValue = (
```
thiagowfx, line 4530: It makes me wonder whether `LocalValue` should be renamed to `ArgumentValue` now (as it is effectively used as an argument).

thiagowfx (approved):
Approving, but please wait for review from Mozilla just to double check it's correct.

index.bs
jrandolf: LocalValue derives its name from the spec instructions.

index.bs
```diff
@@ -4584,7 +4584,9 @@ strong [=/map=] from handle ids to their corresponding objects.
 
 <pre class="cddl remote-cddl local-cddl">
```
jgraham, line 4585: I don't think this should be `local-cddl`, right? These values are never parsed by the client.

index.bs
```diff
@@ -6007,7 +6009,7 @@ ownership will be treated.
 
 [=Remote end definition=]
 
-<pre class="cddl remote-cddl">
+<pre class="cddl remote-cddl local-cddl">
```
jgraham, line 6012: Also this.

jgraham (changes requested):
I think this is OK. The main change seems to be that we allow clients to send maps with channels or remote values as arguments.

The description of `script.LocalValue` needs to be updated to match the change.

I think adding all these types to `local-cddl` is wrong; I don't think clients ever have to parse these values.

index.bs
```diff
@@ -4447,7 +4447,7 @@ used to send custom messages from the [=remote end=] to the [=local end=].
...
```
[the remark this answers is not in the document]
jrandolf: It can't be since `script.LocalValue` exists in `local-cddl`.

index.bs
```diff
@@ -5188,7 +5190,7 @@ The <code>script.RealmType</code> type represents the different types of Realm.
...
```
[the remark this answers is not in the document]
jrandolf: It can't be since `script.LocalValue` exists in `local-cddl`.

index.bs
jrandolf: It can't be since `script.LocalValue` exists in `local-cddl`.

index.bs
jrandolf: They aren't parsed, but they are passed by the client.

jrandolf:
> I think adding all these types to local-cddl is wrong; I don't think clients ever have to parse these values.

Sure, but CDDL complains about unknown types if you don't include it because LocalValue references it and LocalValue exists in `local-cddl`.

jrandolf:
As discussed offline, the solution is to just make `script.LocalValue` only defined in `remote.cddl`.

jgraham (approved):

PR #479, opened 2023-07-07
`ErrorResponse` should always have id

Lightning00Blade:
https://w3c.github.io/webdriver-bidi/#send-an-error-response

***
<a href="https://pr-preview.s3.amazonaws.com/Lightning00Blade/webdriver-bidi/pull/479.html" title="Last updated on Jul 7, 2023, 8:47 AM UTC (fe5a28a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/479/ddabede...Lightning00Blade:fe5a28a.html" title="Last updated on Jul 7, 2023, 8:47 AM UTC (fe5a28a)">Diff</a>


OrKoN (changes requested):
I have double checked the spec and it does not look like we are able to change it. In case, if a command cannot be parsed or does not match any commands, the ID can be null.

thiagowfx:
> NOTE: command id can be null, in which case the id field will also be set to null, not omitted from response.

Issue #480, opened 2023-07-07
Extensible and Type Discrimination


jrandolf:
While working on Puppeteer, we've noticed it's very difficult to discriminate response types because response types have extensibility. This causes a reduction problem because, e.g., an EmptyResponse can in theory contain the keys of `ErrorResponse`, but the values are something completely different.

I propose instead of spreading extensibility, we go along with one of the following standard options:

 - Define a private key, say `? extensions: (*text => any)`, that allows implementation-specific extensions, or
 - Create a registry like how RFCs typically do.

CC: @jgraham @whimboo

jgraham:
Isn't this a fundamental problem if you just look at the CDDL? Per the conversation about intersection types, I think something that matches `ErrorResponse` would also match `EmptyResponse`, and so if the assumption is that implementations will in fact add arbitrary extra data to the response guided only by the CDDL requirements, then we can't spec this away.

On the other hand, nothing in the spec currently allows browsers to actually add any extra data to responses. The point of Extensible was "clients have to accept payloads with extra fields in these places". But if that's always true, then it isn't adding anything.

I think what's actually needed is some constraints on how implementations can represent unilateral extensions of the protocol (which are hopefully rare, but for which there is precedent in e.g. capabilities). In that case we require that any extra fields have a `:` in the name, and commit to never specifying a field with such a name.

jgraham:
For the specific case of `Message` types, putting an explicit `type` field on seems like the obvious thing that we should have done, i.e. `type: "success"`, `type: "error"`, and `type: "event"`.

jrandolf:
> For the specific case of `Message` types, putting an explicit `type` field on seems like the obvious thing that we should have done (and still can do), i.e. `type: "success"`, `type: "error"`, and `type: "event"`.

Perhaps which you try this then.

Issue #274, opened 2022-09-08
Clarify extensibility rules


jgraham:
For compatibility reasons we've had an understanding that we want to allow commands and responses to be extensible. In the CDDL this is handled with boilerplate in the definitions like:
```
*text => any
```
Which amounts to ignoring unknown fields.

However we have not been consistent about adding this in all the definitions.

This recently came up in the context of the `Target` type, which is defined as

```
RealmTarget = {realm: Realm};

ContextTarget = {
  context: BrowsingContext,
  ?sandbox: text
}

Target = (
  RealmTarget //
  ContextTarget
);
```

The question was: should it be allowed to have something that matches _both_ the `RealmTarget` production and the `ContextTarget` production? In that case the _behaviour_ is well-defined (because the prose says to first check if it matches `ContextTarget`), but it seems likely that the user made an error in that case and we should report that rather than silently resolving the ambiguity.

So, how should we handle this in general? Extensibility is useful and required in some cases. But it's not clear if we need it for every type.

Technically, the problem here is that we're hoping to rely on structural typing i.e. we distinguish the target types by the fields they have rather than by name. In contrast for commands and responses where we clearly do want extensibility each type is explicitly named, so there's no possibility of confusion.

So that suggests a general rule: if we want to allow a type with multiple variants to be extended by clients we should always provide an explicit name for each varaint. If there's no explicit name we should not allow additional fields in that type.

Does that sound like a reasonable approach that meets our use cases? And if it does, are we happy with `Target` being non-extensible, or should we add an explicit type field?

sadym-chromium:
I'm inclined not to throw exception in such a case, or in case of extended values like:
```
NullValue = {
  type: "null",
  value: "SOME_VALUE"
}
```
My motivation is the following: do we want to restrict any additional fields to `ScriptTarget`, or just having `realm` and `context` or `sandbox` together? And the first seems an overkill, and the I'm not sure how the second can be specified.

foolip:
Related: https://github.com/w3c/webdriver-bidi/issues/79

foolip:
This problem also reminds me strongly of [distinguishable](https://webidl.spec.whatwg.org/#dfn-distinguishable) dictionaries in Web IDL. There, unions of dictionaries aren't allowed, because type testing would require checking the object properties.

In BiDi, one fix would be to always use different parameters for these target types. But that's not very ergonomic I suspect.

[...]

Issue #79, opened 2020-12-08
Forward compatibility and the use of {*text => any} in CDDL


foolip:
In https://github.com/w3c/webdriver-bidi/pull/77#discussion_r537671899 I learned what the `{*text => any}` in our CDDL is for, it's to allow arbitrary extra keys in an object. I don't know what part of the CDDL spec to read to understand this, but https://tools.ietf.org/html/rfc8610#appendix-H has it in an example.

In order to allow future extensions of the (unversioned) protocol, we will want to accept extra keys everywhere, which means we'll need ensure `{*text => any}` is used wherever extra keys would otherwise be rejected. It looks like `SubscribeParameters` is a case where we don't have it, so that `session.subscribe` with params `{ events: 'browsingContext.*', futureExtraStuff: true }` would fail.

Can tooling help us ensure that we don't miss `{*text => any}` anywhere? Or is there a way of using CDDL where the default behavior is inverted?

bwalderman:
I did another scan through the CDDL spec. This pattern is first described in [section 3.5.1](https://tools.ietf.org/html/rfc8610#section-3.5.1) :

> If the addition of more entries not specified by the current
>   specification is desired, one can add this possibility explicitly:
>
>                          PersonalData = {
>                            ? displayName: tstr,
>                            NameComponents,
>                            ? age: uint,
>                            * tstr => any
>                          } 

I didn't see any kind of shorthand to implicitly accept extra keys everywhere. We could improve the situation somewhat using a new group definition that includes the `*text => any` snippet instead of copy/pasting it:

```cddl
Extensible = (*text => any)

SubscribeParameters = {
  events: [*text],
  ?contexts: [*BrowsingContext],
  Extensible
}
``` 

This still has the problem of potentially forgetting to include `Extensible` everywhere though. Tooling might be a good idea.

foolip:
I like the `Extensible` idea, it makes it more clear to people not very familiar with CDDL (like me) what's going on.

PR #482, opened 2023-07-10
Remove `Extensible` for `Metadata`

jrandolf:
The `Extensible` mechanism makes type discrimination impossible for WebDriver Bidi. In particular, it's impossible to reduce between various productions that have common keys since `Extensible` is a catch-all.

To fix this, this PR introduces the `Metadata` type that replaces `Extensible`. The purpose of `Metadata` is well-defined and carries out the task of `Extensible` without polluting productions.

Fixed: https://github.com/w3c/webdriver-bidi/issues/480

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/482.html" title="Last updated on Jul 10, 2023, 10:26 AM UTC (262fa3f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/482/6736ff3...262fa3f.html" title="Last updated on Jul 10, 2023, 10:26 AM UTC (262fa3f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -302,8 +302,8 @@ the remainder of the specification.
 <pre class="cddl remote-cddl">
 Command = {
   id: js-uint,
+  ? metadata: Metadata,
   CommandData,
-  Extensible,
 }
 
 CommandData = (
@@ -316,7 +316,7 @@ CommandData = (
 )
 
 EmptyParams = {
-   Extensible
+  ? metadata: Metadata,
 }
 </pre>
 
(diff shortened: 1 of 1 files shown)


jrandolf:
> I am not sure it helps with the original issue that Extensible is likely solving (these issues look related #79 #274): as I understand it is to allow the compatibility between protocol versions, which I don't see how metadata would solve. cc @jgraham

I see. Then I think this PR is irrelevant.

PR #493, opened 2023-07-14
Extend roadmap: window management

OrKoN:


roadmap.md
```diff
@@ -90,3 +90,10 @@ This scenario loads a web page and prints it as a PDF. In spec terms, this invol
 
 - [x] Some items from the previous milestones
 - [x] [Printing to PDF as Base64-encoded string](https://w3c.github.io/webdriver-bidi/#command-browsingContext-print)
+
+### Window management
+
+This scenario changes the browser window dimensions and position, maximizes, minimizes the window and toggles the full screen mode.
```
mathiasbynens, line 96:
```suggestion
This scenario changes the browser window dimensions and position, maximizes/minimizes the window, and toggles full-screen mode. 
```

roadmap.md
```diff
@@ -90,3 +90,10 @@ This scenario loads a web page and prints it as a PDF. In spec terms, this invol
 
 - [x] Some items from the previous milestones
 - [x] [Printing to PDF as Base64-encoded string](https://w3c.github.io/webdriver-bidi/#command-browsingContext-print)
+
+### Window management
+
+This scenario changes the browser window dimensions and position, maximizes/minimizes the window, and toggles full-screen mode.
```
mathiasbynens, line 96: One thing we tried to do for the initial scenarios is explain why a user might want this. Maybe something like:

> This scenario loads a web page with a responsive UI, …

E.g. how would a web page respond to the window being maximized/minimized?

OrKoN: I am not sure it will be feasible to list all possible side effects but I assume the general use case is 1) for testing web observable side effects of this 2) for testing the browser alongside other desktop apps

OrKoN: We don't need it for Puppeteer but it seems high on the list of other tools.

roadmap.md
```diff
@@ -90,3 +90,10 @@ This scenario loads a web page and prints it as a PDF. In spec terms, this invol
 
 - [x] Some items from the previous milestones
 - [x] [Printing to PDF as Base64-encoded string](https://w3c.github.io/webdriver-bidi/#command-browsingContext-print)
+
+### Window management
+
+This scenario changes the browser window dimensions and position, maximizes/minimizes the window, and toggles full-screen mode. 
+
+- [x] Some items from the previous milestones
+- [] [Add commands for resizing and positioning browser windowsg](https://github.com/w3c/webdriver-bidi/issues/398)
```
OrKoN, line 99:
```suggestion
- [ ] [Add commands for resizing and positioning browser windowsg](https://github.com/w3c/webdriver-bidi/issues/398)
```

roadmap.md
jgraham: I guess the user scenarios are:
* Ensuring the page responds as expected when the viewport dimensions change
* Verifying that the page works as expected when the document is hidden (minimize window is the easiest way to ensure this in platforms that support it)
* Ensuring that the page works as expected in fullscreen mode (especially given that there's an API for that, and pages can have specific fullscreen behaviour).

OrKoN, line 96:
```suggestion
This scenario changes the browser window dimensions and position, maximizes/minimizes the window, and toggles full-screen mode while testing the side effects of these actions on the web page. 
```

OrKoN: > Ensuring the page responds as expected when the viewport dimensions change

browsingContext.set_viewport can be used for this.

> Verifying that the page works as expected when the document is hidden (minimize window is the easiest way to ensure this in platforms that support it)

browsingContext.activate would be the cross platform way.

At the same time implementing only the fullscreen toggle seems very limiting.

jgraham: > browsingContext.activate would be the cross platform way.

Not really; if you have two browsing contexts they might both be visible at the same time (e.g. because they're in different windows rather than different tabs), so there's no guarantee that activating one context makes the other hidden. But, on platforms that support it, minimize does pretty much guarantee that any contexts in that (OS) window are hidden.

OrKoN: > Not really; if you have two browsing contexts they might both be visible at the same time (e.g. because they're in different windows rather than different tabs), so there's no guarantee that activating one context makes the other hidden. But, on platforms that support it, minimize does pretty much guarantee that any contexts in that (OS) window are hidden.

is there a real-world use case for testing that two contexts visible at the same time in different windows? my comment was for testing the visibility change in isolation.

ollie-iterators:
"windowsg" should be "windows"

jgraham (approved):

whimboo:
@mathiasbynens do you want to review again after your last comments?

@OrKoN looks like there is a conflict that you need to solve.

mathiasbynens:
I’ve resolved the conflict and as a drive-by changed the header text to match the existing style.

PR #499, opened 2023-07-19
Spec issue: This assumes for-in works on iterators

lolaodelola:
This is a proposed solution for by converting the iterator into a list so that it can be iterated on, however similar to other parts of WebDriver BiDi as it is currently, this solution mishandles completion records.

***
<a href="https://pr-preview.s3.amazonaws.com/bocoup/webdriver-bidi/pull/499.html" title="Last updated on Jul 24, 2023, 1:38 PM UTC (3566ecc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/499/63ae894...bocoup:3566ecc.html" title="Last updated on Jul 24, 2023, 1:38 PM UTC (3566ecc)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -118,11 +118,13 @@ spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
     text: Date.toISOString; url: sec-date.prototype.toisostring
     text: EnumerableOwnPropertyNames; url: sec-enumerableownpropertynames
     text: Get; url: sec-get
+    text: GetIterator; url: sec-getiterator
     text: HasProperty; url: sec-hasproperty
     text: IsArray; url: sec-isarray
     text: IsCallable; url: sec-iscallable
     text: IsPromise; url: sec-ispromise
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -6001,7 +6001,7 @@ To <dfn>serialize as a mapping</dfn> given |iterable|, |serialization options|,
 
 1. Let |serialized| be a new list.
 
-1. For |item| in |iterable|:
+1. For |item| in [=IteratorToList=]([=GetIterator=](|iterable|)):
```
jugglinmike, line 6004: [The latest version of ECMA262's GetIterator abstract operation](https://tc39.es/ecma262/#sec-getiterator) requires two arguments--we'll want to specify "sync" in this case.

jugglinmike (commented):
It looks like BiDi's "serialize as a list" algorithm also uses "for in" with an iterable, so we should probably address that, as well.

PR #500, opened 2023-07-19
Spec issue: Differentiate between `List` and `Array`

lolaodelola:
This solution returns a list instead of an array so that it can be used later.

***
<a href="https://pr-preview.s3.amazonaws.com/bocoup/webdriver-bidi/pull/500.html" title="Last updated on Jul 27, 2023, 5:12 PM UTC (bc29541)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/500/63ae894...bocoup:bc29541.html" title="Last updated on Jul 27, 2023, 5:12 PM UTC (bc29541)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4711,8 +4711,6 @@ and |session|:
 
 <div algorithm>
 
-Issue: TODO distinguish between `List` and `Array`.
-
 To <dfn>deserialize value list</dfn> given |serialized value list|, |realm| and
 |session|:
 
--- a/index.bs
+++ b/index.bs
@@ -4709,6 +4709,8 @@ and |session|:
 
 </div>
 
+Issue:  TODO distinguish between `List` and `Array`. This has been partially completed as the "value" parameter
+of "deserialize value list" could be an ECMAScript Array or an ECMAScript Set.
 <div algorithm>
 
 To <dfn>deserialize value list</dfn> given |serialized value list|, |realm| and
--- a/index.bs
+++ b/index.bs
@@ -4708,9 +4708,6 @@ and |session|:
 1. Return [=success=] with data |deserialized key-value list|.
 
 </div>
-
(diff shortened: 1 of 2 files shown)


jugglinmike (commented):
For other reviewers, when the commit message explains that the list can be used later, it's referring to "deserialize key-value list"'s two call sites. Both call sites treat this algorithm's return value as a list.

lolaodelola:
The deploy is failing with
```
  # Failed
  node -e 'require("@actions/core").setFailed("Failed. See details above.")'
  shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
  env:
    PNPM_HOME: /home/runner/setup-pnpm/node_modules/.bin
    PYTHONUSERBASE: /home/runner/work/_actions/w3c/spec-prod/v2/python_modules
```

I'm unsure where this error is coming from or what's causing it. When I build the HTML locally, I'm error free.

index.bs
```diff
@@ -4705,14 +4705,14 @@ and |session|:
    1. Append [=CreateArrayFromList=](«|deserialized key|, |deserialized value|») to
       |deserialized key-value list|.
 
-1. Return [=success=] with data [=CreateArrayFromList=](|deserialized key-value list|).
+1. Return [=success=] with data |deserialized key-value list|.
 
 </div>
 
+Issue:  TODO distinguish between `List` and `Array`. This has been partially completed as the "value" parameter
```
jgraham, line 4712: The `|value|` parameter is always an infra List, not an ECMAScript type at all, so I don't understand the updated comment.

jgraham (changes requested):

index.bs
lolaodelola: @jgraham Ah ok, I can remove the comment then.

jgraham (approved):

PR #501, opened 2023-07-19
Cookies

lolaodelola:
WIP

Issue https://github.com/w3c/webdriver-bidi/issues/287

***
<a href="https://pr-preview.s3.amazonaws.com/bocoup/webdriver-bidi/pull/501.html" title="Last updated on Jan 4, 2024, 2:45 AM UTC (5830f5b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/501/9ddaeb4...bocoup:5830f5b.html" title="Last updated on Jan 4, 2024, 2:45 AM UTC (5830f5b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -9188,7 +9188,7 @@ The <dfn export for=commands>storage.getCookies</dfn> command retrieves zero or 
         }
 
         storage.GetCookiesParameters = {
-          cookie: storage.CookieFilter,
+          filter: storage.CookieFilter,
           ? partition: storage.PartitionKey / browsingContext.BrowsingContext,
         }
     </pre>
@@ -9207,7 +9207,7 @@ The <dfn export for=commands>storage.getCookies</dfn> command retrieves zero or 
 <div algorithm="remote end steps for script.getCookies">
 The [=remote end steps=] with <var ignore>session</var> and |command parameters| are:
 
-1. Let |filter| be the value of the <code>cookie</code> field of |command parameters|.
+1. Let |filter| be the value of the <code>filter</code> field of |command parameters|.
 
 1. Let |cookie domain| be the value of the <code>domain</code> field of |filter| if it is present or null if it isn't.
 
@@ -9340,7 +9340,7 @@ The <dfn export for=commands>storage.deleteCookies</dfn> command removes zero or
         }
 
         storage.DeleteCookiesParameters = {
-          cookie: storage.CookieFilter,
+          filter: storage.CookieFilter,
           ? partition: storage.PartitionKey / browsingContext.BrowsingContext,
         }
     </pre>
@@ -9358,7 +9358,7 @@ The <dfn export for=commands>storage.deleteCookies</dfn> command removes zero or
 <div algorithm="remote end steps for script.deleteCookies">
 The [=remote end steps=] with <var ignore>session</var> and |command parameters| are:
 
-1. Let |filter| be the value of the <code>cookie</code> field of |command parameters|.
+1. Let |filter| be the value of the <code>filter</code> field of |command parameters|.
 
 1. Let |cookie domain| be the value of the <code>domain</code> field of |filter| if it is present or null if it isn't.
 
--- a/index.bs
+++ b/index.bs
@@ -9096,9 +9096,9 @@ StorageCommand = (
 
 <pre class="cddl local-cddl">
 StorageResult = (
+  storage.DeleteCookiesResult /
   storage.GetCookiesResult /
-  storage.SetCookieResult /
-  storage.DeleteCookiesResult
+  storage.SetCookieResult
 )
 </pre>
 
--- a/index.bs
+++ b/index.bs
@@ -3625,11 +3625,6 @@ opened</dfn> steps given |window|, |type| and |message|.
 
 </div>
 
-#### The browsingContext.CookieCreated Event #### {#event-browsingContext-cookieCreated}
-
-#### The browsingContext.CookieDestroyed Event #### {#event-browsingContext-cookieDestroyed}
-
-
 ## The network Module ## {#module-network}
 
 The <dfn export for=modules>network</dfn> module contains commands and events
--- a/index.bs
+++ b/index.bs
@@ -9161,9 +9161,9 @@ of the [=table of standard storage partition key attributes=] where the
 [=storage partition key=] corresponds to a standard storage partition, or an
 [=extension storage partition key attribute=] where it does not. The precise
 entries are implementation-defined and are determined by the storage
-partitioning adopted by the implementation. This list should not include
-attributes which appear as keys in the [=default values for storage partition
-key attributes=].
+partitioning adopted by the implementation. This list includes only partition keys
+for which no default is available. As such the list must not share any entries with the keys of 
+ [=default values for storage partition key attributes=].
 
 <div algorithm>
 To <dfn>expand a storage partition spec</dfn> given |partition spec|:
--- a/index.bs
+++ b/index.bs
@@ -9110,9 +9110,9 @@ StorageResult = (
 
 <pre class="cddl local-cddl remote-cddl">
 storage.PartitionKey = {
-    ? userContext: text,
-    ? sourceOrigin: text,
-    Extensible,
+  ? userContext: text,
+  ? sourceOrigin: text,
+  Extensible,
 }
 </pre>
 
--- a/index.bs
+++ b/index.bs
@@ -8517,12 +8517,6 @@ events related to storage.
 
 ### Definition ### {#module-storage-definition}
 
-[=Local end definition=]
-
-### Types ### {#module-storage-types}
-
-#### The storage.Cookie Type #### {#type-storage-Cookie}
-
 [=Remote end definition=]
 
 <pre class="cddl remote-cddl">
@@ -8541,6 +8535,10 @@ StorageResult = (
   storage.SetCookieResult
 )
 </pre>
+
+### Types ### {#module-storage-types}
+
+#### The storage.Cookie Type #### {#type-storage-Cookie}
 
 [=Remote end definition=] and [=Local end definition=]
 
--- a/index.bs
+++ b/index.bs
@@ -9054,12 +9054,12 @@ StorageResult = (
 
 #### The storage.PartitionKey Type #### {#type-storage-PartitionKey}
 
-[=Remote end definition=] and [=Local end definition=]
+[=Local end definition=]
 
 <pre class="cddl local-cddl remote-cddl">
 storage.PartitionKey = {
-    ? userContext: text,
-    ? sourceOrigin: text,
+    userContext: text,
+    sourceOrigin: text,
     Extensible,
 }
 </pre>
@@ -9073,7 +9073,11 @@ A [=remote end=] has a [=/map=] of <dfn>default values for storage partition key
 <div algorithm>
 To <dfn>expand a storage partition spec</dfn> given |partition spec| and |cookie domain|:
 
-1. If |partition spec| is a [=browsing context id=]:
+1. If |partition spec| is null:
+
+  1. Let |partition key| be an empty [=/map=].
+
+1. Otherwise, if |partition spec|["<code>type</code>"] is "<code>context</code>":
 
   1. Let |browsing context| be the result of [=trying=] to [=get a browsing context=] given |partition spec|.
 
@@ -9081,13 +9085,9 @@ To <dfn>expand a storage partition spec</dfn> given |partition spec| and |cookie
 
   1. Return [=success=] with data |partition key|.
 
-1. Otherwise, if |partition spec| is null:
-
-  1. Let |partition key| be an empty [=/map=].
-
 1. Otherwise:
 
-  1. Assert: |partition spec| is a [=/map=] with zero or more entries.
+  1. Assert: |partition spec|["<code>type</code>"] is "<code>storageKey</code>".
 
   1. Let |partition key| be |partition spec|.
 
@@ -9189,9 +9189,26 @@ The <dfn export for=commands>storage.getCookies</dfn> command retrieves zero or 
           Extensible,
         }
 
+        storage.BrowsingContextPartitionDescriptor = {
+          type: "context",
+          context: browsingContext.BrowsingContext
+        }
+
+        storage.StorageKeyPartitionDescriptor = {
+          type: "storageKey",
+          ? userContext: text,
+          ? sourceOrigin: text,
+          Extensible,
+        }
+
+        storage.PartitionDescriptor = (
+          storage.BrowsingContextPartitionDescriptor /
+          storage.StorageKeyPartitionDescriptor
+        )
+
         storage.GetCookiesParameters = {
           ? filter: storage.CookieFilter,
-          ? partition: storage.PartitionKey / browsingContext.BrowsingContext,
+          ? partition: storage.PartitionDescriptor,
         }
     </pre>
    </dd>
@@ -9263,7 +9280,7 @@ The <dfn export for=commands>storage.setCookie</dfn> command creates a new [=coo
 
         storage.SetCookieParameters = {
           cookie: storage.PartialCookie,
-          ? partition: storage.PartitionKey / browsingContext.BrowsingContext,
+          ? partition: storage.PartitionDescriptor,
         }
     </pre>
    </dd>
@@ -9344,7 +9361,7 @@ The <dfn export for=commands>storage.deleteCookies</dfn> command removes zero or
 
         storage.DeleteCookiesParameters = {
           ? filter: storage.CookieFilter,
-          ? partition: storage.PartitionKey / browsingContext.BrowsingContext,
+          ? partition: storage.PartitionDescriptor,
         }
     </pre>
    </dd>
--- a/index.bs
+++ b/index.bs
@@ -7113,7 +7113,9 @@ storage.Cookie = {
     name: text;
     value: text;
     host: text;
-    partitionKey: text|undefined;
+    ? partitionKey: {
+        Extensible
+    };
 }
 </pre>
 
--- a/index.bs
+++ b/index.bs
@@ -9188,7 +9188,7 @@ The <dfn export for=commands>storage.getCookies</dfn> command retrieves zero or 
         }
 
         storage.GetCookiesParameters = {
-          filter: storage.CookieFilter,
+          ? filter: storage.CookieFilter,
           ? partition: storage.PartitionKey / browsingContext.BrowsingContext,
         }
     </pre>
@@ -9207,7 +9207,8 @@ The <dfn export for=commands>storage.getCookies</dfn> command retrieves zero or 
 <div algorithm="remote end steps for script.getCookies">
 The [=remote end steps=] with <var ignore>session</var> and |command parameters| are:
 
-1. Let |filter| be the value of the <code>filter</code> field of |command parameters|.
+1. Let |filter| be the value of the <code>filter</code> field of |command parameters|
+  if it is present or an empty [=/map=] if it isn't.
 
 1. Let |cookie domain| be the value of the <code>domain</code> field of |filter| if it is present or null if it isn't.
 
@@ -9340,7 +9341,7 @@ The <dfn export for=commands>storage.deleteCookies</dfn> command removes zero or
         }
 
         storage.DeleteCookiesParameters = {
-          filter: storage.CookieFilter,
+          ? filter: storage.CookieFilter,
           ? partition: storage.PartitionKey / browsingContext.BrowsingContext,
         }
     </pre>
@@ -9358,7 +9359,8 @@ The <dfn export for=commands>storage.deleteCookies</dfn> command removes zero or
 <div algorithm="remote end steps for script.deleteCookies">
 The [=remote end steps=] with <var ignore>session</var> and |command parameters| are:
 
-1. Let |filter| be the value of the <code>filter</code> field of |command parameters|.
+1. Let |filter| be the value of the <code>filter</code> field of |command parameters|
+  if it is present or an empty [=/map=] if it isn't.
 
 1. Let |cookie domain| be the value of the <code>domain</code> field of |filter| if it is present or null if it isn't.
 
--- a/index.bs
+++ b/index.bs
@@ -9086,9 +9086,9 @@ A <dfn>storage partition key</dfn> is a [=/map=] which uniquely identifies a [=s
 
 <pre class="cddl remote-cddl">
 StorageCommand = (
+  storage.DeleteCookies //
   storage.GetCookies //
-  storage.SetCookie //
-  storage.DeleteCookies
+  storage.SetCookie
 )
 </pre>
 
--- a/index.bs
+++ b/index.bs
@@ -9094,9 +9094,9 @@ To <dfn>expand a storage partition key spec</dfn> given |partition key spec| and
 <div algorithm>
 To <dfn>get the cookie store</dfn> given |storage partition key|:
 
-1. If a [=cookie store=] exists for |storage partition key|:
+1. If a [=cookie store=] |match| exists for |storage partition key|:
 
-  1. Return the matching [=cookie store=].
+  1. Return [=success=] with data |match|.
 
 1. Return [=error=] with [=error code=] [=no such storage partition=].
 
--- a/index.bs
+++ b/index.bs
@@ -3945,6 +3945,8 @@ Note: this takes a [=byte sequence=] and returns a <code>network.BytesValue</cod
 
 <pre class="cddl local-cddl remote-cddl">
 
+network.SameSite = "strict" / "lax" / "none"
+
 <!--
 Modifications to this definition should be reflected in
 `network.SetCookieHeader`, `storage.CookieFilter`, and `storage.PartialCookie`.
@@ -3957,7 +3959,7 @@ network.Cookie = {
     size: js-uint,
     httpOnly: bool,
     secure: bool,
-    sameSite: "strict" / "lax" / "none",
+    sameSite: network.SameSite,
     ? expires: js-uint,
 };
 </pre>
@@ -4480,7 +4482,7 @@ network.SetCookieHeader = {
     ? expires: text,
     ? maxAge: js-int,
     ? path: text,
-    ? sameSite: "strict" / "lax" / "none",
+    ? sameSite: network.SameSite,
     ? secure: bool,
 }
 </pre>
@@ -8757,7 +8759,7 @@ The <dfn export for=commands>storage.getCookies</dfn> command retrieves zero or 
           ? size: js-uint,
           ? httpOnly: bool,
           ? secure: bool,
-          ? sameSite: "strict" / "lax" / "none",
+          ? sameSite: network.SameSite,
           ? expires: js-uint,
         }
 
@@ -8824,7 +8826,7 @@ The <dfn export for=commands>storage.setCookie</dfn> command creates a new [=coo
           ? size: js-uint,
           ? httpOnly: bool,
           ? secure: bool,
-          ? sameSite: "strict" / "lax" / "none",
+          ? sameSite: network.SameSite,
           ? expires: js-uint,
         }
 
--- a/index.bs
+++ b/index.bs
@@ -9154,8 +9154,6 @@ doesn't provide an explicit value. The precise
 entries are implementation-defined and are determined by the storage
 partitioning adopted by the implementation.
 
-Note: Default values are used by [=storage.getCookies=], [=storage.setCookie=],
-and [=storage.deleteCookies=] when the [=local end=] does not specify a value.
 
 A [=remote end=] has a [=/list=] of <dfn>required partition key
 attributes</dfn> which contains zero or more entries. Each key must be a member
--- a/index.bs
+++ b/index.bs
@@ -316,7 +316,8 @@ CommandData = (
   InputCommand //
   NetworkCommand //
   ScriptCommand //
-  SessionCommand
+  SessionCommand //
+  StorageCommand
 )
 
 EmptyParams = {
@@ -354,7 +355,8 @@ ResultData = (
   EmptyResult //
   NetworkResult //
   ScriptResult //
-  SessionResult
+  SessionResult //
+  StorageResult
 )
 
 EmptyResult = {
@@ -1848,10 +1850,7 @@ BrowsingContextCommand = (
   browsingContext.Navigate //
   browsingContext.Print //
   browsingContext.Reload //
-  browsingContext.SetViewport //
-  browsingContext.GetCookie //
-  browsingContext.SetCookie //
-  browsingContext.DeleteCookie
+  browsingContext.SetViewport
 )
 </pre>
 
@@ -1863,9 +1862,7 @@ BrowsingContextResult = (
   browsingContext.CreateResult //
   browsingContext.GetTreeResult //
   browsingContext.NavigateResult //
-  browsingContext.PrintResult //
-  browsingContext.GetCookieResult //
-  browsingContext.SetCookieResult
+  browsingContext.PrintResult
 )
 
 BrowsingContextEvent = (
@@ -2156,25 +2153,6 @@ browsingContext.ReadinessState = "none" / "interactive" / "complete"
 
 The <code>browsingContext.ReadinessState</code> type represents the stage of
 document loading at which a navigation command will return.
-
-#### The browsingContext.Cookie Type #### {#type-browsingContext-Cookie}
-
-[=remote end definition=] and [=local end definition=]
-
-<pre class="cddl local-cddl remote-cddl">
-browsingContext.Cookie = {
-    name: text;
-    value: text;
-    host: text;
-    partitionKey: text|undefined;
-}
-
-</pre>
-
-The <code>browsingContext.Cookie</code> type provides details of a given cookie.
-
-Issue Confirm if `cookie` type should be consolidated in one place `browsingContext.Cookie` or `Network.Cookie`
-
 
 ### Commands ### {#module-browsingContext-commands}
 
@@ -3083,65 +3061,6 @@ The [=remote end steps=] with |command parameters| are:
 
 </div>
 
-#### The browsingContext.getCookie Command ####  {#command-browsingContext-getCookie}
-
-The <dfn export for=commands>browsingContext.getCookie</dfn> command  specific cookie characteristics (e.g. ) on the given top-level browsing context.
-
-<dl>
-   <dt>Command Type</dt>
-   <dd>
-    <pre class="cddl remote-cddl">
-        browsingContext.GetCookie = {
-
-        }
-    </pre>
-   </dd>
-   <dt>Result Type</dt>
-   <dd>
-    <pre class="cddl local-cddl">
-      browsingContext.GetCookieResult = {
-        data: text
-      }
-    </pre>
-   </dd>
-</dl>
-
-#### The browsingContext.setCookie Command ####  {#command-browsingContext-setCookie}
-
-The <dfn export for=commands>browsingContext.setCookie</dfn> command  specific cookie characteristics (e.g. ) on the
-given top-level browsing context.
-
-<dl>
-   <dt>Command Type</dt>
-   <dd>
-    <pre class="cddl remote-cddl">
-        browsingContext.SetCookie = {
-
-        }
-    </pre>
-   </dd>
-   <dt>Result Type</dt>
-   <dd>
-    <pre class="cddl local-cddl">
-      browsingContext.SetCookieResult = {
-        data: text
-      }
-    </pre>
-   </dd>
-</dl>
-
-#### The browsingContext.deleteCookie Command ####  {#command-browsingContext-deleteCookie}
-
-<dl>
-   <dt>Command Type</dt>
-   <dd>
-    <pre class="cddl remote-cddl">
-        browsingContext.DeleteCookie = {
-
-        }
-    </pre>
-   </dd>
-</dl>
 
 ### Events ### {#module-contexts-events}
 
@@ -7117,6 +7036,116 @@ worker=] algorithm:
    production, with the <code>params</code> field set to |params|.
 
 </div>
+
+## The storage Module ## {#module-storage}
+
+The <dfn export for=modules>storage</dfn> module contains functionality and
+events related to storage.
+
+### Definition ### {#module-storage-definition}
+
+[=Local end definition=]
+
+### Types ### {#module-storage-types}
+
+#### The storage.Cookie Type #### {#type-storage-Cookie}
+
+[=Remote end definition=]
+
+<pre class="cddl remote-cddl">
+StorageCommand = (
+  storage.GetCookie //
+  storage.SetCookie //
+  storage.DeleteCookie
+)
+</pre>
+
+[=Local end definition=]
+
+<pre class="cddl local-cddl">
+StorageResult = (
+  storage.GetCookieResult //
+  storage.SetCookieResult
+)
+</pre>
+
+[=Remote end definition=] and [=Local end definition=]
+
+<pre class="cddl local-cddl remote-cddl">
+storage.Cookie = {
+    name: text;
+    value: text;
+    host: text;
+    partitionKey: text|undefined;
+}
+</pre>
+
+The <code>storage.Cookie</code> type provides details of a given cookie.
+
+Issue: Confirm if <code>cookie</code> type should be consolidated in one place <code>storage.Cookie</code> or <code>Network.Cookie</code>.
+
+### Commands ### {#module-storage-commands}
+
+#### The storage.getCookie Command ####  {#command-storage-getCookie}
+
+The <dfn export for=commands>storage.getCookie</dfn> command  specific cookie characteristics (e.g. ) on the given top-level browsing context.
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl remote-cddl">
+        storage.GetCookie = {
+
+        }
+    </pre>
+   </dd>
+   <dt>Result Type</dt>
+   <dd>
+    <pre class="cddl local-cddl">
+      storage.GetCookieResult = {
+        data: text
+      }
+    </pre>
+   </dd>
+</dl>
+
+#### The storage.setCookie Command ####  {#command-storage-setCookie}
+
+The <dfn export for=commands>storage.setCookie</dfn> command  specific cookie characteristics (e.g. ) on the
+given top-level browsing context.
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl remote-cddl">
+        storage.SetCookie = {
+
+        }
+    </pre>
+   </dd>
+   <dt>Result Type</dt>
+   <dd>
+    <pre class="cddl local-cddl">
+      storage.SetCookieResult = {
+        data: text
+      }
+    </pre>
+   </dd>
+</dl>
+
+#### The storage.deleteCookie Command ####  {#command-storage-deleteCookie}
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl remote-cddl">
+        storage.DeleteCookie = {
+
+        }
+    </pre>
+   </dd>
+</dl>
+
 
 ## The log Module ## {#module-log}
 
--- a/index.bs
+++ b/index.bs
@@ -1879,9 +1879,7 @@ BrowsingContextEvent = (
   browsingContext.NavigationAborted //
   browsingContext.NavigationFailed //
   browsingContext.UserPromptClosed //
-  browsingContext.UserPromptOpened //
-  browsingContext.CookieCreated //
-  browsingContext.CookieDestroyed
+  browsingContext.UserPromptOpened
 )
 </pre>
 
@@ -2158,6 +2156,25 @@ browsingContext.ReadinessState = "none" / "interactive" / "complete"
 
 The <code>browsingContext.ReadinessState</code> type represents the stage of
 document loading at which a navigation command will return.
+
+#### The browsingContext.Cookie Type #### {#type-browsingContext-Cookie}
+
+[=remote end definition=] and [=local end definition=]
+
+<pre class="cddl local-cddl remote-cddl">
+browsingContext.Cookie = {
+    name: text;
+    value: text;
+    host: text;
+    partitionKey: text|undefined;
+}
+
+</pre>
+
+The <code>browsingContext.Cookie</code> type provides details of a given cookie.
+
+Issue Confirm if `cookie` type should be consolidated in one place `browsingContext.Cookie` or `Network.Cookie`
+
 
 ### Commands ### {#module-browsingContext-commands}
 
@@ -3068,9 +3085,63 @@ The [=remote end steps=] with |command parameters| are:
 
 #### The browsingContext.getCookie Command ####  {#command-browsingContext-getCookie}
 
+The <dfn export for=commands>browsingContext.getCookie</dfn> command  specific cookie characteristics (e.g. ) on the given top-level browsing context.
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl remote-cddl">
+        browsingContext.GetCookie = {
+
+        }
+    </pre>
+   </dd>
+   <dt>Result Type</dt>
+   <dd>
+    <pre class="cddl local-cddl">
+      browsingContext.GetCookieResult = {
+        data: text
+      }
+    </pre>
+   </dd>
+</dl>
+
 #### The browsingContext.setCookie Command ####  {#command-browsingContext-setCookie}
 
+The <dfn export for=commands>browsingContext.setCookie</dfn> command  specific cookie characteristics (e.g. ) on the
+given top-level browsing context.
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl remote-cddl">
+        browsingContext.SetCookie = {
+
+        }
+    </pre>
+   </dd>
+   <dt>Result Type</dt>
+   <dd>
+    <pre class="cddl local-cddl">
+      browsingContext.SetCookieResult = {
+        data: text
+      }
+    </pre>
+   </dd>
+</dl>
+
 #### The browsingContext.deleteCookie Command ####  {#command-browsingContext-deleteCookie}
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl remote-cddl">
+        browsingContext.DeleteCookie = {
+
+        }
+    </pre>
+   </dd>
+</dl>
 
 ### Events ### {#module-contexts-events}
 
--- a/index.bs
+++ b/index.bs
@@ -578,8 +578,8 @@ with the following additional codes:
   <dt><dfn>unable to set cookie</dfn>
   <dd>Tried to create a cookie, but the user agent rejected it.
 
-  <dt><dfn>unknown source origin</dfn>
-  <dd>Tried to interact with data in storage, but the source origin could not be inferred.
+  <dt><dfn>underspecified storage partition</dfn>
+  <dd>Tried to interact with data in a storage partition which was not adequately specified.
 </dl>
 
 <pre class="cddl local-cddl">
@@ -600,10 +600,10 @@ ErrorCode = "invalid argument" /
             "unable to capture screen" /
             "unable to close browser" /
             "unable to set cookie" /
+            "underspecified storage partition" /
             "unknown command" /
             "unknown error" /
-            "unsupported operation" /
-            "unknown source origin"
+            "unsupported operation"
 </pre>
 
 ## Events ## {#events}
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2152,6 +2157,25 @@ browsingContext.ReadinessState = "none" / "interactive" / "complete"
 The <code>browsingContext.ReadinessState</code> type represents the stage of
 document loading at which a navigation command will return.
 
+#### The browsingContext.Cookie Type #### {#type-browsingContext-Cookie}
```
OrKoN, line 2160: we probably want cookies outside of the browsing context module (except for the command to get a partition key or other commands that require the context as input)? I am thinking a new domain called `storage` would be good?  cc @jgraham

lolaodelola: That sounds fine to me. Currently I haven't included a command to get a cookie's partition key, are you thinking something like `broswsingContext.getCookiePartitionKey` for that?

jgraham: Yes to a `storage` domain.

For the partition key, I think one question is whether it's sufficient to be able to get the key only for already loaded contexts. That is roughly equivalent to not exposing the key at all, but only being able to set cookies for a currently loaded page, because you always need to load a page, get the key, and then use that to set cookies.

I slightly wondered if we should define some partition key fields and have something in the capabilities response that returns the list of keys recognised by the current browser. So we might define `sourceOrigin` as a key. Then a browser with only SOP would just return an empty key set like:
```js
storagePartitionKeys: {}
```
A browser that required keying on `sourceOrigin` and couldn't provide any defaults would return something like:
```js
storagePartitionKeys: {sourceOrigin: {required: true}}
```
And then something like current gecko with containers could provide a custom key like `moz:userContext` like so:
```js
storagePartitionKeys: {sourceOrigin: {required: true}, "moz:userContext": {default: null}}
```
So then clients/users could know that to set storage properties you'd _have_ to provide a value for the `sourceOrigin` key and that `moz:userContext` is another part of the key, but if you don't provide it you'll get some default value that hopefully does the right thing for most cases. This doesn't totally solve the forwards compatibility problem, but at least makes it possible for clients to implement a good UX (by telling you exactly what you need to change if there's a change, and by allowing users to provide a set of properties that's a superset of what the client expects).

lolaodelola: I just want to make a clarification here, when we're talking about partition key for cookies, are we talking about partitioned cookies as defined in [CHIPS](https://developer.mozilla.org/en-US/docs/Web/Privacy/Partitioned_cookies) or are we talking about cookies defined within Mozilla's containers for different user contexts?

jugglinmike:
> I am thinking a new domain called `storage` would be good?

I believe "domain" is a term from Chrome DevTools Protocol which roughly corresponds to the BiDi draft's "module." Please let me know if I'm mistaken!

It's my understanding that there's no interest in extending the original WebDriver specification with implementation-defined cookie partitioning. That's why I've wrapped WebDriver's "process capabilities" with a new algorithm which adds partition information. To reduce the risk of confusion, I've referenced the original algorithm with the word "classic" and used "BiDi" in the name of the new algorithm (though either one of those alone would be sufficient to disambiguate).

@jgraham's [recent comment](https://github.com/w3c/webdriver-bidi/pull/501#discussion_r1272444304) describes a scenario involving multiple partition keys, so I've changed the type in `storage.Cookie` to an optional extensible object (rather than [`string|undefined`](https://github.com/w3c/webdriver-bidi/issues/287#issuecomment-1635432871)).

[The draft currently defines a type named `network.Cookie`.](https://w3c.github.io/webdriver-bidi/#type-network-Cookie) Is there a need for that type to differ from `storage.Cookie`? I believe the organization of types is not observable, so would it be possible to extend `network.Cookie` and use it from the new `storage` module?

jgraham:
Yes, reusing `network.Cookie` is the right thing to do here.

On the more general design here, I think we agreed at TPAC to drop the capabilities part for now.

Instead, in the meantime, we can have something like
```
storage.PartitionKey = {
  ?container: text,
  ?sourceOrigin: text,  
}
```

Then when you set a cookie you have something like:

```
storage.SetCookieParameters = {
  cookies: [ +network.Cookie ],
  ? partitionKey: storage.PartitionKey,
}
```

and `storage.SetCookieResult` which will look like
```
storage.SetCookieResult = {
   partitionKey: partitionKey
}
```

In the return value all partition keys that were used must be set to the value that was used, and any that were not used must be unset. For `sourceOrigin` the default must be the same as the Cookie's destination origin (there might be some detail here because cookies are not precisely keyed on origin). For `container` the default must be the default container (which we might call null or empty string, I'm not sure).

jugglinmike:
Hi folks, I've completed a first draft; I'll share a few specific thoughts as in-line comments.

index.bs
```diff
@@ -8508,6 +8527,357 @@ worker=] algorithm:
...
+        }
+
+        storage.CookieFilter = {
+          ? ~ network.Cookie
+        }
+
+        storage.GetCookiesParameters = {
+          cookie: storage.CookieFilter,
+          ? partitionKey: storage.PartitionKey,
+        }
```
jugglinmike, line 8727: These commands can't use [BiDi's `network.Cookie` type](https://w3c.github.io/webdriver-bidi/#type-network-Cookie) directly because in these contexts, many of the entries are optional. In the interest of limiting duplication, I've attempted to specify the `storage.CookieFilter` and `storage.PartialCookie` types in terms of `network.Cookie`, but I'm not certain the CDDL syntax I've applied is appropriate. (You can find more detail in [the message I sent to the CBOR mailing list](https://mailarchive.ietf.org/arch/msg/cbor/41_rLuyFj8upo2wJFksagTOa-HY/).)

index.bs
```diff
@@ -8508,6 +8527,357 @@ worker=] algorithm:
...
+</div>
+
+<div algorithm>
+To <dfn>get matching cookies</dfn> given |cookie store| and |filter|:
+
+1. Let |cookies| be a new list.
+1. For each |stored cookie| in |cookie store|:
+  1. If [=match cookie=] with |stored cookie| and |filter| is true:
+    1. Append the result of [=serialize cookie=] given |stored cookie| to |cookies|.
+1. Return |cookies|.
```
jugglinmike, line 8701: This could be expressed without an algorithm like "match cookie", but doing so would involve labeling a loop in order to continue from within a nested loop. Factoring the operation into two simpler algorithms seems preferable.

index.bs
```diff
@@ -8508,6 +8527,357 @@ worker=] algorithm:
...
+1. Let |cookies| be the result of [=get matching cookies=] with |store| and |filter|.
+1. Let |body| be a [=/map=] matching the <code>storage.GetCookiesResult</code> production,
+  with the <code>cookies</code> field set to |cookies|.
+1. Return [=success=] with data |body|.
+
+</div>
+
+#### The storage.setCookie Command ####  {#command-storage-setCookie}
+
+The <dfn export for=commands>storage.setCookie</dfn> command creates a new [=cookie=] in a cookie store, replacing any cookie in that store which matches according to [[COOKIES]].
```
jugglinmike, line 8758: A command like `storage.setCookies` would involve a variable number of fallible operations. That might mean aborting on the first failure, or it might mean attempting all operations regardless of individual failures. I can see the value in both, so it seemed prudent to instead specify `storage.setCookie` and allow consumers to build their preference on top of that.

index.bs
```diff
@@ -8508,6 +8527,357 @@ worker=] algorithm:
...
+1. If a [=cookie store=] exists for |storage partition key|:
+  1. Return the matching [=cookie store=].
+1. Return [=error=] with [=error code=] [=no such storage partition=].
+
+</div>
+
+<p>The following <dfn>table for cookie conversion</dfn>
+ defines the cookie concepts relevant to WebDriver BiDi,
+ how these are referred to in [[COOKIES]],
+ as well as what keys they map to in a [=serialized cookie=].
```
jugglinmike, line 8602: If we want to re-use [the corresponding table in WebDriver Classic](https://w3c.github.io/webdriver/#dfn-table-for-cookie-conversion), we'll have to special-case the "Cookie expiry time" concept because WebDriver classic uses `"expiry"` as the JSON key, but [BiDi's `network.Cookie`](https://w3c.github.io/webdriver-bidi/#type-network-Cookie) uses `"expires"`.

index.bs
```diff
@@ -8508,6 +8527,357 @@ worker=] algorithm:
...
+      }
+    </pre>
+   </dd>
+</dl>
+
+<div algorithm="remote end steps for script.setCookie">
+The [=remote end steps=] with <var ignore>session</var> and |command parameters| are:
+
+1. Let |cookie spec| be the value of the <code>cookie</code> field of |command parameters|.
+1. If |cookie spec|["<code>size</code>"] exists:
```
jgraham, line 8796: We should definitely change the CDDL instead of having this, but also in general we _don't_ raise errors for additional fields.

jgraham (changes requested):
At a high level this is looking pretty good to me.

One formatting comment: in the rest of the spec we leave a blank line between algorithm steps in order to make it easier to read.

index.bs
jimevans: You're right, @jgraham, I'm not fond of it, but as it duplicates existing Classic behavior, I'm not going to kick about it. Round-trips are perf killers in high-latency scenarios.

index.bs
sadym-chromium: I believe it's not too late for changing the [BiDi's `network.Cookie`](https://w3c.github.io/webdriver-bidi/#type-network-Cookie) "expires" to "expiry" to align with WebDriver Classic.

index.bs
```diff
@@ -8508,6 +8527,357 @@ worker=] algorithm:
...
+
+### Types ### {#module-storage-types}
+
+#### The storage.PartitionKey Type #### {#type-storage-PartitionKey}
+
+[=Remote end definition=] and [=Local end definition=]
+
+<pre class="cddl local-cddl remote-cddl">
+storage.PartitionKey {
+    ? container: text,
```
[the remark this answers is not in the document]
jugglinmike: I've renamed the key, but I'll hold off on restructuring until there's a stronger signal about whether it's mandatory.

index.bs
```diff
@@ -8508,6 +8527,357 @@ worker=] algorithm:
...
+          cookie: storage.CookieFilter,
+          ? partitionKey: storage.PartitionKey,
+        }
+    </pre>
+   </dd>
+   <dt>Result Type</dt>
+   <dd>
+    <pre class="cddl local-cddl">
+      storage.GetCookiesResult = {
+        cookies: [*network.Cookie]
```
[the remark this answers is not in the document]
jugglinmike: Sounds good.

index.bs
jugglinmike: Okay, I've made `Storage.CookieFilter` and `Storage.PartialCookie` extensible.

index.bs
```diff
@@ -8508,6 +8527,357 @@ worker=] algorithm:
...
+        storage.DeleteCookiesParameters = {
+          cookie: storage.CookieFilter,
+          ? partitionKey: storage.PartitionKey,
+        }
+    </pre>
+   </dd>
+   <dt>Result Type</dt>
+   <dd>
+     <pre class="cddl">
+        EmptyResult
```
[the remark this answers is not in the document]
jugglinmike: Got it. This makes `Storage.SetCookieResult` and `Storage.DeleteCookiesResult` identical, but I've defined them as distinct types, anyway.

index.bs
```diff
@@ -8508,6 +8527,357 @@ worker=] algorithm:
...
+    ? sourceOrigin: text,
+    Extensible,
+}
+</pre>
+
+The <code>storage.PartitionKey</code> type describes namespaces within which the user agent may organize persistent data such as [=cookies=] and local storage.
+
+A [=remote end=] has a [=/set=] of <dfn>storage partition key attributes</dfn> which contains zero or more implementation-defined string values.
+
+A [=remote end=] has a [=/map=] of <dfn>default values for storage partition key attributes</dfn> which contains zero or more implementation-defined entries. Each key must be a member of the [=storage partition key attributes=] which the [=remote end=] does not require as input to [=storage.getCookies=], [=storage.setCookie=], and [=storage.deleteCookies=]; the corresponding value is a the value which the [=remote end=] uses when one is not provided.
```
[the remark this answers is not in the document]
jugglinmike: Does this mean that we need to require the `domain` key in `storage.getCookies` and `storage.deleteCookies` (just as we already do for `storage.setCookie`)? Or maybe require at least one, e.g. either:

    getCookies({domain: 'zombo.com'})

or

    getCookies({}, {partitionKey: {sourceOrigin: 'zombo.com'}})

?

Or could a wildcard value actually work for "get" and "delete" operations? Unlike "set" (where implementations may not be able to create cookies for all domains *a piori*), it seems like "get" and "delete" can describe meaningful operations in the absence of any particular domain, e.g. "retrieve matching cookies from all source origins."

index.bs
```diff
@@ -8508,6 +8527,357 @@ worker=] algorithm:
...
+        storage.PartialCookie = {
+          name: text,
+          value: network.BytesValue,
+          domain: text,
+          ? ~ network.Cookie
+        }
+
+        storage.SetCookieParameters = {
+          cookie: storage.PartialCookie,
+          ? partitionKey: storage.PartitionKey,
```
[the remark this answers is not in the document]
jugglinmike: Could you clarify what you mean by "context" in this case? I believe [the existing "browsing context"](https://w3c.github.io/webdriver-bidi/#type-browsingContext-Browsingcontext) and [the proposed "user context"](https://github.com/w3c/webdriver-bidi/pull/570) can both have associated domains, so it seems like either could fit in your example.

jugglinmike:
> One formatting comment: in the rest of the spec we leave a blank line between algorithm steps in order to make it easier to read.

I don't know how I missed that. Fixed.

> I'm quite in favour of just duplicating the definition with explicit markers for optionality, rather than trying to use clever CDDL which readers are unlikely to figure out the meaning of without being CDDL experts :)

You got it. I've included an HTML comment which reminds contributors to keep these representations in sync (the CDDL build script ignores HTML comments inside the spec's definition elements). I've also introduced a named type for the `sameSite` properties to reduce the noise a little.

index.bs
jugglinmike: Oh, that's great! [Here's a patch for that.](https://github.com/w3c/webdriver-bidi/pull/596)

jugglinmike:
[We just today updated BiDi to use "expiry" instead of "expires"](https://github.com/w3c/webdriver-bidi/pull/596) to better align with WebDriver Classic and allow reuse (thanks again to @sadym-chromium for [calling out the feasibility of that change](https://github.com/w3c/webdriver-bidi/pull/501#discussion_r1383254805)). I've merged that change into this branch so that the new types also use "expiry" and so that we can rely on [WebDriver Classic's "table for cookie conversion"](https://w3c.github.io/webdriver/#dfn-table-for-cookie-conversion) instead of defining a new one.

index.bs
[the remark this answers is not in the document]
OrKoN: I believe both could be applicable here. But I guess we could start with the browsing context since the user contexts are not fully defined yet.

index.bs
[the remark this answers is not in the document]
jgraham: I think it's theoretically possible to get all cookies in a partition (at least in gecko, c.f. https://searchfox.org/mozilla-central/source/netwerk/cookie/nsICookieManager.idl#232-245; note that "origin attributes" here is basically the same as "partition").

But I don't think it's possible to get all cookies from all partitions at once. 

So I think you either need to exactly specify which partition you want _or_ you need to specify the domain so that we can compute a default partition (or specify a navigable / context which is implicitly associated with a specific partition).

index.bs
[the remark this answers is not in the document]
jgraham: Yes, browsing contexts (navigables).

(I don't think a user context has an associated domain? It's "just" a storage partition key. One can imagine having different domains load in different user contexts by default or something, and certainly each user context might at any one time have multiple navigables, all with documents that have origins covering multiple domains, but I don't see how there would be a useful 1:1 mapping).

jugglinmike:
Thanks for the input, @OrKoN and @jgraham! I've responded here on the main thread to promote visibility (my latest changes have replaced the changes upon which the original feedback was based).

> I think it's theoretically possible to get all cookies in a partition (at least in gecko, c.f. https://searchfox.org/mozilla-central/source/netwerk/cookie/nsICookieManager.idl#232-245; note that "origin attributes" here is basically the same as "partition").
>
> But I don't think it's possible to get all cookies from all partitions at once.
>
> So I think you either need to exactly specify which partition you want _or_ you need to specify the domain so that we can compute a default partition (or specify a navigable / context which is implicitly associated with a specific partition).

Got it. I've made the text require one of those three pieces of information. I've also persisted the map named "default values for storage partition key attributes" and its associated semantics because they seem necessary to support implementation-defined extensions to the partition key.

> Yes, browsing contexts (navigables).

Okay. The latest iteration replaces the concept of a "partition key spec" (which was a nullable map describing zero or more entries of a partition key) with a more generic "partition spec" (which is a nullable value that's either a map as before, or a browsing context id). At the protocol level, I've suggested referring to this value simply as "partition".

Although @jgraham [mentioned](https://github.com/w3c/webdriver-bidi/pull/501#discussion_r1397128933) that browsing contexts are implicitly associated with a specific partition, I believe this patch needs to make that association explicit. In order to do that, I've also added formal definitions for "storage partition" and "storage partition key."

(Separately, I've also corrected the return type for the "get the cookie store" algorithm. In order to be used with [WebDriver's "try"](https://w3c.github.io/webdriver/#dfn-try), it has to return either a "success" value or an "error" value.)

index.bs
```diff
@@ -8972,6 +9018,371 @@ worker=] algorithm:
...
+
+1. If |partition spec| is a [=browsing context id=]:
+
+  1. Let |browsing context| be the result of [=trying=] to [=get a browsing context=] given |partition spec|.
+
+  1. Let |partition key| be the [=storage partition key|key=] of |browsing context|'s [=associated storage partition=].
+
+  1. Return [=success=] with data |partition key|.
+
+1. Otherwise, if |partition spec| is null:
```
jugglinmike, line 9084: Due to the early return, this step could be expressed with an independent "If" (that is, by removing the word "otherwise"). I'm calling this out because I know opinions differ on which style is preferable.

index.bs
```diff
@@ -8972,6 +9018,371 @@ worker=] algorithm:
...
+          ? httpOnly: bool,
+          ? secure: bool,
+          ? sameSite: network.SameSite,
+          ? expiry: js-uint,
+          Extensible,
+        }
+
+        storage.GetCookiesParameters = {
+          cookie: storage.CookieFilter,
+          ? partition: storage.PartitionKey / browsingContext.BrowsingContext,
```
jgraham, line 9192: So when we have this kind of choice we're preferring named types over structural typing. So in this design the options should be something like:
```
storage.GetCookiesParameters = {
[…]
? partition: storage.BrowsingContextPartition / storage.StorageKeyPartition
}

storage.BrowsingContextPartition = {
  type: "context",
  context: browsingContext.BrowsingContext
 }
 
 storage.StorageKeyPartition = {
   type: "storage-key",
   ? userContext: text,
   ? sourceOrigin: text,
 }
 ```

**Comment by lolaodelola** [comment]
Hi @gsnedders,

We're putting finishing touches on this piece of work & would like a rep from Apple to review and make sure it's up to scratch for you folks too as Mozilla & Google have given their input.

File: index.bs [hunk 14]
```diff
@@ -8972,6 +9018,371 @@ worker=] algorithm:
...
+
+    1. [=map/Set=] |partition key|[|name|] to |default value|.
+
+1. If |partition key|["<code>sourceOrigin</code>"] does not [=map/exist=]:
+
+  1. If |cookie domain| is null, then:
+
+    1. Return [=error=] with [=error code=] [=unknown source origin=].
+
+  1. [=map/Set=] |partition key|["<code>sourceOrigin</code>"] to |cookie domain|.
```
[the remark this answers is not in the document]
jugglinmike: That's a good point. It seems to me like we have two options:

- in the absence of an explicitly-provided `sourceOrigin` value, use the cookie "domain" attribute as the fallback only if it includes a scheme and port (and return and error if it does not)
- require `sourceOrigin` to be specified

Is that right?

index.bs
jugglinmike: That seems more amenable to future extensions, though I'm curious if there are any other motivations.

I think referring to these as types of "Partitions" could be misleading because it implies that they're mutually exclusive. As far as I understand, one could use either type to describe the same storage partition. This would be more clear to me if we qualified these types with a suffix like "Descriptor." The resultant verbosity could be reduced with a new group:

```diff
 storage.GetCookiesParameters = {
   […]
-  ? partition: storage.BrowsingContextPartition / storage.StorageKeyPartition
+  ? partition: storage.PartitionDescriptor
 }
+
+storage.PartitionDescriptor = (
+  storage.BrowsingContextPartitionDescriptor /
+  storage.StorageKeyPartitionDescriptor
+)
 
-storage.BrowsingContextPartition = {
+storage.BrowsingContextPartitionDescriptor = {
   type: "context",
   context: browsingContext.BrowsingContext
 }
 
-storage.StorageKeyPartition = {
+storage.StorageKeyPartitionDescriptor = {
   type: "storageKey",
   ? userContext: text,
   ? sourceOrigin: text,
+  Extensible,
 }
```

(Also, whether we call it `StorageKeyPartition` or `StorageKeyPartitionDescriptor`, I think [our previous discussion about allowing additional fields](https://github.com/w3c/webdriver-bidi/pull/501#discussion_r1378734361) still applies.)

This leaves room for a type name like `storage.PartitionKey` as the return value for "set cookie" and "delete cookie". I think there's value in a distinct type (rather than using `storage.StorageKeyPartition`/`storage.StorageKeyPartitionDescriptor`) because the `type` field is superfluous in that context and because the `userContext` and `sourceOrigin` field would not be optional.

```diff
 storage.PartitionKey {
-    ? userContext: text,
+    userContext: text,
-    ? sourceOrigin: text,
+    sourceOrigin: text,
     Extensible,
 }
```

That design would also involve a delineation between remote end definitions ( `storage.PartitionDescriptor`, `storage.BrowsingContextPartitionDescriptor`, and `storage.StorageKeyPartitionDescriptor`) and local end definitions (`storage.PartitionKey`).

What do you think?

index.bs
```diff
@@ -4386,7 +4428,7 @@ network.Cookie = {
     size: js-uint,
     httpOnly: bool,
     secure: bool,
-    sameSite: "strict" / "lax" / "none",
+    sameSite: network.SameSite,
```
sadym-chromium, line 4431: Probably, along with `domain` and `path, user wants to know `port` as well.

sadym-chromium: Of cause, this can be done out of scope of this PR

index.bs
```diff
@@ -8972,6 +9018,371 @@ worker=] algorithm:
...
+          ? size: js-uint,
+          ? httpOnly: bool,
+          ? secure: bool,
+          ? sameSite: network.SameSite,
+          ? expiry: js-uint,
+          Extensible,
+        }
+
+        storage.GetCookiesParameters = {
+          cookie: storage.CookieFilter,
```
sadym-chromium, line 9191: 1. Maybe `filter` to avoid confusion?
2. Does it make sense to make it optional?

index.bs
mathiasbynens: I believe `domain`, `path`, and the other properties listed here are cookie attributes:

image

`port` is not a cookie attribute.

index.bs
```diff
@@ -8972,6 +9018,371 @@ worker=] algorithm:
...
+1. For each |name| → |value| in |filter|:
+
+  1. If |name| is "<code>value</code>":
+
+    1. Set |value| to [=deserialize protocol bytes=] with |value|.
+
+  1. Let |field name| be the field name corresponding to the JSON key |name| in
+    the [=table for cookie conversion=].
+
+  1. If |stored cookie|[|field name|] does not equal |value|:
```
sadym-chromium, line 9135: I'm not sure how **`equal`** defined for dictionary, and specifically for `network.BytesValue`, which has `value` and `type` fields.

sadym-chromium:
Note: please be aware that running `./scripts/test.sh` is failing:
```
error: parser errors
    ┌─ input:854:3
    │
854 │   storage.DeleteCookiesResult
    │   ^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing definition for rule storage.DeleteCookiesResult
    ·
857 │ storage.PartitionKey {
    │ ^^^^^^^^^^^^^^^^^^^^ expected assignment token '=', '/=' or '//=' after rule identifier
    ·
865 │   partitionKey: storage.partitionKey,
    │                 ^^^^^^^^^^^^^^^^^^^^ missing definition for rule storage.partitionKey
    ·
869 │   partitionKey: storage.partitionKey
    │                 ^^^^^^^^^^^^^^^^^^^^ missing definition for rule storage.partitionKey

Error: "incremental parsing error"
```

index.bs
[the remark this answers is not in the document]
jgraham: So I asked a bit about this, and although I didn't (yet) get a very clear answer it looks like gecko is using (scheme, host, port) as the key (with some extra considerations around not setting cookies for public suffix parts of the domain). It also appears that it has support for the port being unset, but I'm not yet totally sure if that means something other than the default port.

So yes, I think there are a few options:
1. Use the `domain` value where possible. However that (afaict?) doesn't ever have a host and port. So we'd have to pick a default scheme at least. Or maybe we could set cookies for both `http` and `https`, with the default port, although the spec currently doesn't anticipate a 1:many relationship between the request and the actual cookies set. Also many testing setups might use non-default ports for local servers.
2. Error if `sourceOrigin` isn't provided. This has the considerable disadvantage that it makes it easy to write tests that don't work cross-browser.
3. Allow the `domain` value to be a full origin, but only set the actual domain part on the cookie. Although in that case one might wonder whether `domain: "https://example.com:8000"` should also set the `secure` flag. In general making the fields not correspond directly to the actual cookie fields seems likely to cause confusion.
4. Allow an implementation-defined default behaviour (e.g. it could pick a default scheme and port based on sites that have been visited in the current session). But that seems incredibly hard to reason about or make work cross browser.

So unfortunately I think that the best thing for now might be to error out if the source origin isn't supplied and then allow clients to provide a more user friendly API (or maybe during implementation it would become obvious how we could relax the constraint). But I'd be interested to hear what client authors think about this approach.

index.bs
```diff
@@ -8972,6 +9018,371 @@ worker=] algorithm:
...
+          cookie: storage.PartialCookie,
+          ? partition: storage.PartitionKey / browsingContext.BrowsingContext,
+        }
+    </pre>
+   </dd>
+   <dt>Result Type</dt>
+   <dd>
+    <pre class="cddl local-cddl">
+      storage.SetCookieResult = {
+        partitionKey: storage.partitionKey
```
OrKoN, line 9271:
```suggestion
        partitionKey: storage.PartitionKey
```

index.bs
jgraham: > That seems more amenable to future extensions, though I'm curious if there are any other motivations.

Mostly that. I don't know that everyone agrees with the design decision, but in my opinion it helps make parsing into program types unambiguous, and as you say makes future extensions easier.

> What do you think?

I don't have a strong feeling about internal type names, so your suggestions seem fine. I agree that separating out the type that we return from the input type makes sense given that they have different validity constraints.

jugglinmike:
@sadym-chromium @OrKoN sorry to take your time with such obvious mistakes as syntax validation errors. I've been working without the benefit of running this project's test script locally because I couldn't properly install its dependencies. I initially assumed this was an inconsequential problem with my system and decided to rely on the project's continuous integration environment. [Now, though, I'm starting to wonder if the problem exists there, as well.](https://github.com/w3c/webdriver-bidi/issues/614)

index.bs
jugglinmike: Which operand of "equal" are you referring to? I believe the left-hand side is limited to strings, integers, and booleans (since they come from the stored cookie as described by [WebDriver Classic's table](https://w3c.github.io/webdriver/#dfn-table-for-cookie-conversion)). The corresponding types in the "filter" map a defined similarly, with the exception of the property named "value". I think this is what you had in mind with your reference to `network.BytesValue`, but I believe I handled that case with the earlier step's invocation of the "deserialize protocol bytes" algorithm.

index.bs
jugglinmike: Those both sound like good ideas to me! I've pushed up a commit to enact that change. (I assumed your suggestion also applies to the "storage.DeleteCookies" commands because the API and semantics are so similar. Please let me know if I'm mistaken about that.)

sadym-chromium:
https://github.com/bocoup/webdriver-bidi/pull/2

index.bs
```diff
@@ -8972,6 +9018,373 @@ worker=] algorithm:
...
+1. Let |partition key| be the result of [=trying=] to [=expand a storage partition spec|expand=] |partition spec| using |cookie domain|.
+
+1. Let |store| be the result of [=trying=] to [=get the cookie store=] with
+  |partition key|.
+
+1. Let |cookies| be the result of [=get matching cookies=] with |store| and |filter|.
+
+1. Let |body| be a [=/map=] matching the <code>storage.GetCookiesResult</code> production,
+  with the <code>cookies</code> field set to |cookies| and the <code>partitionKey</code>
+  field set to |partition key|.
```
sadym-chromium, line 9227: The `partition key` field here means "this cookies were found for the given partition key". Let's say a browser supports partition keys `sourceOrigin` and `other_supported`, but not partition key `unsupported`. 
The user sends `storage.getCookies` with **partition**: `{sourceOrigin: 'foo', unsupported: 'baz'}`. The response would contain **partitionKey**: `{sourceOrigin: 'foo', other_supported: 'default_value', unsupported: 'baz'}`. 

I wonder if we should provide a way for users to understand that the `unsupported` partition key is not supported by the given browser.

index.bs
```diff
@@ -8972,6 +9018,373 @@ worker=] algorithm:
...
+  1. [=map/Set=] |partition key|["<code>sourceOrigin</code>"] to |cookie domain|.
+
+1. Return [=success=] with data |partition key|.
+
+</div>
+
+<div algorithm>
+To <dfn>get the cookie store</dfn> given |storage partition key|:
+
+1. If a [=cookie store=] |match| exists for |storage partition key|:
```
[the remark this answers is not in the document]
jugglinmike: Here, I intended "match" as a noun (not a verb), but I didn't recognize the confusion that could result from using a homonym here. This operation is right on the edge of what we're normatively specifying, so I think this is best resolved with more detail rather than a simple word swap. [I've made an attempt along those lines](https://github.com/w3c/webdriver-bidi/pull/501/commits/d7e61f03357e1852259b4521d2b8e56f0d1ffba9), please let me know what you think!

index.bs
sadym-chromium: you are right, it is addressed. Thanks!

index.bs
sadym-chromium: Basically, the question is if we want to provide the actual partition key for each returned cookie individually?

index.bs
```diff
@@ -8972,6 +9018,392 @@ worker=] algorithm:
...
+)
+</pre>
+
+[=Local end definition=]
+
+<pre class="cddl local-cddl">
+StorageResult = (
+  storage.GetCookiesResult //
+  storage.SetCookieResult //
+  storage.DeleteCookiesResult
```
sadym-chromium, line 9049: According to [cddl spec](https://www.tech-invite.com/y85/tinv-ietf-rfc-8610.html#e-2-2-2), this should be a single `/`.
```suggestion
  storage.GetCookiesResult /
  storage.SetCookieResult /
  storage.DeleteCookiesResult
```

index.bs
```diff
@@ -8972,6 +9018,392 @@ worker=] algorithm:
...
+
+        storage.DeleteCookiesParameters = {
+          ? filter: storage.CookieFilter,
+          ? partition: storage.PartitionDescriptor,
+        }
+    </pre>
+   </dd>
+   <dt>Result Type</dt>
+   <dd>
+     <pre class="cddl">
```
sadym-chromium, line 9370:
```suggestion
     <pre class="cddl local-cddl">
```

index.bs
```diff
@@ -8972,6 +9018,392 @@ worker=] algorithm:
...
+The <dfn export for=commands>storage.getCookies</dfn> command retrieves zero or more [=cookies=] which [=match cookie|match=] a set of provided parameters.
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl remote-cddl">
+        storage.GetCookies = {
+          method: "storage.getCookies",
+          params: storage.GetCookiesParameters
+        }
```
sadym-chromium, line 9173: `storage.GetCookies` is a group <- `StorageResult` group <- `ResultData`, which is used as a type in `CommandResponse` map contains `ResultData` as a type, so there map around it will be created later on. Meaning the there should be parentheses used:
```suggestion
        storage.GetCookies = (
          method: "storage.getCookies",
          params: storage.GetCookiesParameters
        )
```

index.bs
```diff
@@ -8972,6 +9018,392 @@ worker=] algorithm:
...
+The <dfn export for=commands>storage.setCookie</dfn> command creates a new [=cookie=] in a cookie store, replacing any cookie in that store which matches according to [[COOKIES]].
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl remote-cddl">
+        storage.SetCookie = {
+          method: "storage.setCookie",
+          params: storage.SetCookieParameters,
+        }
```
sadym-chromium, line 9263: Same as `storage.GetCookies`:
```suggestion
        storage.SetCookie = (
          method: "storage.setCookie",
          params: storage.SetCookieParameters,
        )
```

index.bs
```diff
@@ -8972,6 +9018,392 @@ worker=] algorithm:
...
+The <dfn export for=commands>storage.deleteCookies</dfn> command removes zero or more [=cookies=] which [=match cookie|match=] a set of provided parameters.
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl remote-cddl">
+        storage.DeleteCookies = {
+          method: "storage.deleteCookies",
+          params: storage.DeleteCookiesParameters,
+        }
```
sadym-chromium, line 9360: Same as `storage.GetCookies`:
```suggestion
        storage.DeleteCookies = (
          method: "storage.deleteCookies",
          params: storage.DeleteCookiesParameters,
        )
```

index.bs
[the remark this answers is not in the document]
jugglinmike: Thanks, @jgraham; I believe [this commit](https://github.com/w3c/webdriver-bidi/pull/501/commits/2315ab086700606905b3e10be466e1bfe1f7a51e) implements your feedback. In this revision, the two partition key attributes we've been discussing (`userContext` and `sourceOrigin`) no longer appear to have any normative requirements at all. Is that because I've misunderstood the intent?

index.bs
jugglinmike: @sadym-chromium I think I agree: it seems misleading for the user agent to reflect partition key properties that it does not recognize. This should be fairly straightforward to specify in the "expand a storage partition spec" algorithm (especially [if my latest change is appropriate](https://github.com/w3c/webdriver-bidi/pull/501#discussion_r1424814640)), though I could use some clarification regarding your follow-up question:

> Basically, the question is if we want to provide the actual partition key for each returned cookie individually?

As I understand it, the `storage.getCookies` command is only capable of returning cookies stored in one specific storage partition. From that perspective, it seems redundant for the remote end to return the partition key for each cookie individually.

jgraham: One per command rather than one per cookie seems good.

index.bs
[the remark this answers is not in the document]
jgraham: I think under the definition of `PartitionKey` you can't just say "keys are implementation defined" but need to actually specify for those keys in the spec what property of the storage partition they actually correspond to, noting that implementations are not required to partition on all keys. 

I also think we should probably require that additional keys are vendor prefixed in until standardised; that would avoid a situation where two browsers implemented a partition with slightly different semantics and it was impossible to specify a key that worked in both.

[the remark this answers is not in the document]
jugglinmike: @jgraham I believe that the "sourceOrigin" corresponds to the first item in [the Storage standard's "storage key" tuple](https://storage.spec.whatwg.org/#storage-key), but I'm confused about what makes BiDi's "storage partition key" conceptually distinct from that standard's "storage key." While the "storage key" doesn't [yet](https://privacycg.github.io/storage-partitioning/) include a notion of "user context" (nor is is extensible), I'm suddenly wondering why we shouldn't amend it based on the needs we've identified here. Could you help me understand the interaction between these two concepts?

index.bs
```diff
@@ -9024,6 +9070,386 @@ worker=] algorithm:
 
 </div>
 
+## The storage Module ## {#module-storage}
```
whimboo, line 9073: Would you mind to move this whole section down so that we can keep an alphabetical order of modules? Thanks!

index.bs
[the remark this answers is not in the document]
jgraham: Conceptually they're pretty similar, I think. Although Cookies are a bit special because they're intrinsically keyed on destination domain (and historically _not_ keyed on source origin), whereas other storage types are only keyed on source (i.e. document) origin and don't have the concept of a destination. 

I think it might be easier just to define the concepts we need for BiDi rather than trying to build on other DOM specs; hopefully if we're defining the same underlying reality we converge on the same model and we can eventually share a defintion. But at least for BiDi we're not really expecting anyone to _change_ their storage model, so I think a small amount of duplication is mostly harmless and means we're not so exposed to changes in other standards as they try to unify a DOM-exposed version of how this works.

index.bs
jugglinmike: @whimboo I'm happy to move the section, but since the list is not sorted alphabetically, I need more specific instruction about where it the storage module should go.

```
$$('[id^=module-]').map(el => el.innerText).join('\n')
"module name
6.1. The session Module
6.1.1. Definition
6.1.2. Types
6.1.3. Commands
6.2. The browser Module
6.2.1. Definition
6.2.2. Commands
6.3. The browsingContext Module
6.3.1. Definition
6.3.2. Types
6.3.3. Commands
6.3.4. Events
6.4. The network Module
6.4.1. Definition
6.4.3. Types
6.4.4. Commands
6.4.5. Events
6.5. The script Module
6.5.1. Definition
6.5.3. Types
6.5.4. Commands
6.5.5. Events
6.6. The storage Module
6.6.1. Definition
6.6.2. Types
6.6.3. Commands
6.7. The log Module
6.7.1. Definition
6.7.2. Types
6.7.3. Events
6.8. The input Module
6.8.1. Definition
6.8.2. Types
6.8.3. Commands" 
```

whimboo: Oh, I thought that only `session` is not sorted. But given that also `log` and `input` are affected I would suggest we keep it after `script` and once this PR is merged we can re-organize the list of modules.

jugglinmike: Maybe we could rename session to "000_session" while we're at it.

index.bs
```diff
@@ -9024,6 +9070,386 @@ worker=] algorithm:
...
+
+### Types ### {#module-storage-types}
+
+#### The storage.PartitionKey Type #### {#type-storage-PartitionKey}
+
+[=Local end definition=]
+
+<pre class="cddl local-cddl remote-cddl">
+storage.PartitionKey = {
+    userContext: text,
```
sadym-chromium, line 9113: I believe this parameter is either optional or nullable, to point out the default user context.

index.bs
```diff
@@ -9024,6 +9070,386 @@ worker=] algorithm:
...
+1. Let |partition key| be the result of [=trying=] to [=expand a storage partition spec|expand=] |partition spec|.
+
+1. Let |store| be the result of [=trying=] to [=get the cookie store=] with
+  |partition key|.
+
+1. Let |cookies| be the result of [=get matching cookies=] with |store| and |filter|.
+
+1. Let |body| be a [=/map=] matching the <code>storage.GetCookiesResult</code> production,
+  with the <code>cookies</code> field set to |cookies| and the <code>partitionKey</code>
+  field set to |partition key|.
```
sadym-chromium, line 9296: At this point the partition key can be an empty map (if `partition spec` is was set, and endpoint does not require `userContext` and `sourceOrigin`), which conflicts with the result type `partitionKey: storage.PartitionKey`.

index.bs
```diff
@@ -9024,6 +9070,386 @@ worker=] algorithm:
...
+        )
+
+        <!--
+        Modifications to this definition should be reflected in
+        `network.Cookie`, `network.SetCookieHeader`, and `storage.CookieFilter`.
+        -->
+        storage.PartialCookie = {
+          name: text,
+          value: network.BytesValue,
+          domain: text,
```
sadym-chromium, line 9322: Do we want to duplicate domain, if the `browsingContext` is provided?

jgraham: This is the domain the cookie will be sent to, not the domain of the page the cookie will be sent from, so I'm not sure what the duplication is?

sadym-chromium: Oh, my bad. Thanks!

sadym-chromium: Does it make sense to set it optional if the expected domain to be sent to is the same as the one in partition descriptor?

index.bs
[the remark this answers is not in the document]
jugglinmike: @jgraham If I understand you correctly, this is to be an informative explanation (since normatively defining something like Firefox's Total Cookie Protection seems out of scope for BiDi). That's how I've attempted to integrate your feedback in [the latest commit](https://github.com/w3c/webdriver-bidi/pull/501/commits/4d74327953991bb1866ec9629af4b8061db3d24e) (that also introduces the restrictions on implementation-defined attributes that you recommended). User contexts seems a little more straightforward thanks to gh-570; I've just inserted an issue reference so that this patch can land before that one. Does this look good to you?

I'm also thinking about renaming the term "default values for storage partition key attributes" to "optional storage partition key attributes." The latter would highlight the relationship with the "required [storage] partition key attributes," but on the other hand, it would obscure their difference in type. Any thoughts on that change?

index.bs
jgraham: That feels like too much magic to me.

index.bs
```diff
@@ -9118,11 +9119,44 @@ storage.PartitionKey = {
...
+  <tr>
+    <td>"<code>userContext</code>"
+    <td>
+      A user context id
+
+      Issue(w3c/webdriver-bidi#570): User contexts are in the process of being
+      specified.
+  <tr>
+    <td>"<code>sourceOrigin"</code>
+    <td>The [=origin=] from which the storage partition is to be accessed
```
jgraham: I don't love "is to be accessed". Maybe`The [=origin=] of resources that can access the storage partition` or something.

index.bs
```diff
@@ -9118,11 +9119,44 @@ storage.PartitionKey = {
...
+
+[=Remote ends=] may support any number of <dfn>extension storage partition key
+attributes</dfn>. In order to avoid conflicts with other implementations, these
+attributes must begin with a unique identifier for the vendor and user-agent
+followed by U+003A (:).
+
+A [=remote end=] has a [=/map=] of <dfn>default values for storage partition
+key attributes</dfn> which contains zero or more implementation-defined
+entries. Each key must be a member of the [=table of standard storage partition
+key attributes=] or an [=extension storage partition key attribute=].
```
jgraham, line 9150:
```
A [=remote end=] has a [=/map=] of <dfn>default values for storage partition
key attributes</dfn> which contains zero or more entries. Each key must be a 
member of the [=table of standard storage partition key attributes=] where the
storage partition key corresponds to a standard partition, or an 
[=extension storage partition key attribute=] where it does not. The precise entries 
are implementation-defined, and are determined by the storage partitioning adopted
by the implementation.
```

index.bs
```diff
@@ -9118,11 +9119,44 @@ storage.PartitionKey = {
...
+
+Note: Default values are used by [=storage.getCookies=], [=storage.setCookie=],
+and [=storage.deleteCookies=] when the [=local end=] does not specify a value.
+
+A [=remote end=] has a [=/list=] of <dfn>required partition key
```
jgraham, line 9155: Maybe also rephrase this more like the one above.

index.bs
jgraham: Yes, everything here should be optional. If the implementation doesn't support the key it shouldn't return it.

index.bs
```diff
@@ -9024,6 +9070,419 @@ worker=] algorithm:
...
+
+    1. [=map/Set=] |partition key|[|name|] to |default value|.
+
+1. For each |name| in the remote end's [=required partition key attributes=]:
+
+  1. If |partition key|[|name|] does not [=map/exist=]:
+
+    1. Return [=error=] with [=error code=] [=underspecified storage partition=].
+
+1. Return [=success=] with data |partition key|.
```
jgraham, line 9193: This will still have a `type` attribute. It will also still include anything that was supplied but isn't used. We really want to end up with just the things that actually form part of the used keys.

jgraham (changes requested):

index.bs
jugglinmike: Addressed via commit 35d7fad.

index.bs
jugglinmike: Addressed via commit 35d7fad.

index.bs
jugglinmike: Addressed via commit 35d7fad.

index.bs
jugglinmike: Addressed via commit 9eebfc1 (note that I switched to using "the serialization of the origin" rather than "the origin")

index.bs
jugglinmike: Addressed via commit 9eebfc1.

index.bs
jugglinmike: Addressed via commit 9eebfc1.

index.bs
```diff
@@ -9145,48 +9145,54 @@ attributes must begin with a unique identifier for the vendor and user-agent
...
-  1. If |partition key|[|name|] does not [=map/exist=]:
+  1. Let |value| be |partition spec|[|name|] if it [=map/exists=] or |default value| otherwise.
 
-    1. [=map/Set=] |partition key|[|name|] to |default value|.
+  1. [=map/Set=] |partition key|[|name|] to |value|.
 
 1. For each |name| in the remote end's [=required partition key attributes=]:
 
-  1. If |partition key|[|name|] does not [=map/exist=]:
+  1. If |partition spec|[|name|] [=map/exists=]:
```
jgraham, line 9191: So this means that you're treating "required" and "default" as mutually exclusive? i.e. the full key will be all the required keys plus all the keys with a default value? I was imagining that "required" would be everything and "default" would be the subset for which a known default exists. Either way works, but I found this quite surprising, so we at least need to be clearer in the definitions what's expected.

jgraham (commented):
This feels really close to landing; I think we should just clarify how "required" and "default" partition keys relate to each other, but at that point we're good to go.

jugglinmike:
Thanks, @jgraham! I've personally been considering this "requirement" from the perspective of the local end, where the attributes with default values are implicitly not required. I believe your expectation comes from the remote end's perspective, where the requirement stands regardless of whether each property was explicitly specified.

It seems like your recommendation to have guidance applies in either case--that is, whether the list SHOULD or SHOULD NOT include property values from the map. I've added some text which supports the "mutually exclusive"/"local end" framing since that seems like the lowest-friction solution at this late stage of the design process.

This has me wondering, though: do we really need two data structures here? If we only allowed implementers to specify which attributes are optional (and treat all other attributes are required), then would that limit them in any meaningful way?

If not, then we could avoid confusion (and simplify the spec generally) by eliminating this avenue of implementer choice. If, instead of "required partition key attributes", we had a definition like:

>     A remote end has a set of <dfn>recognized storage partition key attributes</dfn>
>     which is the union of the attributes in the [=table of standard storage
>     partition key attributes=] and the [=remote end=]'s [=extension storage
>     partition key attributes=].
>
>     Note: The [=local end=] must specify any partition key attribute that a
>     [=remote end=] [=recognized storage partition key attributes|recognizes=] but
>     does not define a [default values for storage partition key
>     attributes|default value=].

I believe that would be sufficient to express the desired semantics of the "expand" algorithm.

index.bs
```diff
@@ -9024,6 +9070,427 @@ worker=] algorithm:
...
+[=Remote ends=] may support any number of <dfn>extension storage partition key
+attributes</dfn>. In order to avoid conflicts with other implementations, these
+attributes must begin with a unique identifier for the vendor and user-agent
+followed by U+003A (:).
+
+A [=remote end=] has a [=/map=] of <dfn>default values for storage partition
+key attributes</dfn> which contains zero or more entries. Each key must be a
+member of the [=table of standard storage partition key attributes=] where the
+[=storage partition key=] corresponds to a standard storage partition, or an
+[=extension storage partition key attribute=] where it does not. The precise
```
jgraham, line 9151:
```suggestion
[=extension storage partition key attribute=] where it does not, and the values
represent the default value of that partition key that will be used when the user 
doesn't provide an explicit value. The precise
```

index.bs
```diff
@@ -9024,6 +9070,427 @@ worker=] algorithm:
...
+entries are implementation-defined and are determined by the storage
+partitioning adopted by the implementation.
+
+Note: Default values are used by [=storage.getCookies=], [=storage.setCookie=],
+and [=storage.deleteCookies=] when the [=local end=] does not specify a value.
```
jgraham, line 9156:
```suggestion
```

index.bs
```diff
@@ -9024,6 +9070,427 @@ worker=] algorithm:
...
+
+A [=remote end=] has a [=/list=] of <dfn>required partition key
+attributes</dfn> which contains zero or more entries. Each key must be a member
+of the [=table of standard storage partition key attributes=] where the
+[=storage partition key=] corresponds to a standard storage partition, or an
+[=extension storage partition key attribute=] where it does not. The precise
+entries are implementation-defined and are determined by the storage
+partitioning adopted by the implementation. This list should not include
+attributes which appear as keys in the [=default values for storage partition
+key attributes=].
```
jgraham, line 9166:
```suggestion
partitioning adopted by the implementation. This list includes only partition keys
for which no default is available. As such the list must not share any entries with the keys of 
 [=default values for storage partition key attributes=].
```

jgraham:
Writing the spec from the perspective of the local end isn't going to work out well; in practice the spec is specifying what happens in remote ends, and local ends are essentially free to do whatever they like to achieve their desired result. Of course by constraining the behaviour of remote ends the spec also de-facto constrains the things that local ends can do to achieve their goals. But note that, for example, we don't have a local end testsuite, only a remote end one. That reflects the fact that the normative criteria in the spec are all on the remote end.

I've made some suggestions for small improvements that I think clarify things for now. With those I'm basically happy for this to land, although as with all other commands I expect we'll need some later cleanup as we implement and as the interaction between this feature and other features in the spec is worked out.

jgraham (approved):

jugglinmike:
Thanks, @jgraham. I'm still interested in learning if the simplification I proposed is feasible, but like you, I'm satisfied with where we've landed for now.

whimboo (dismissed):
Thanks for all the updates @jugglinmike! I did a overall check only for now given that there are some issues to fix.

index.bs
```diff
@@ -9024,6 +9070,427 @@ worker=] algorithm:
...
+<table>
+  <tr>
+    <th>Attribute
+    <th>Definition
+  <tr>
+    <td>"<code>userContext</code>"
+    <td>
+      A user context id
+
+      Issue(w3c/webdriver-bidi#570): User contexts are in the process of being
```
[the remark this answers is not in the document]
jugglinmike: You got it

index.bs
```diff
@@ -9024,6 +9070,427 @@ worker=] algorithm:
...
+        storage.CookieFilter = {
+          ? name: text,
+          ? value: network.BytesValue,
+          ? domain: text,
+          ? path: text,
+          ? size: js-uint,
+          ? httpOnly: bool,
+          ? secure: bool,
+          ? sameSite: network.SameSite,
+          ? expiry: js-uint,
```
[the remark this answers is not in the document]
jugglinmike: This document doesn't follow a convention concerning if/how optionality impacts sort order (or concerning sorting in general). Would you like to sort with or without regard for optionality? Or would you prefer to introduce the new convention atomically?

index.bs
```diff
@@ -9024,6 +9070,427 @@ worker=] algorithm:
...
+<div algorithm="remote end steps for storage.getCookies">
+The [=remote end steps=] with <var ignore>session</var> and |command parameters| are:
+
+1. Let |filter| be the value of the <code>filter</code> field of |command parameters|
+  if it is present or an empty [=/map=] if it isn't.
+
+1. Let |partition spec| be the value of the <code>partition</code> field of
+  |command parameters| if it is present or null if it isn't.
+
+1. Let |partition key| be the result of [=trying=] to [=expand a storage partition spec|expand=] |partition spec|.
```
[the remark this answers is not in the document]
jugglinmike: Sure (That was intentional; I thought the phrase "expand partition key" is more readable than "expand a partition key with partition key" and further, that context sufficiently disambiguates the more concise form. I'm taking your interpretation of an editorial error as evidence to the contrary.)

index.bs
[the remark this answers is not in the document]
whimboo: Very cool. Thanks a lot!

index.bs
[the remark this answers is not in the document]
whimboo: Generally we apply this sorting in our CDDL definitions. Usually we have the mandatory group first, and then the optional one, whereby within each group we sort alphabetically. I checked quite a few definitions and that is true, only some do not follow. With introducing complete new commands and types we should follow that.

sadym-chromium (approved):

index.bs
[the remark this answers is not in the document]
whimboo: We can do the sorting in a follow-up PR. It's not blocking.

whimboo:
We have full review from Mozilla and Google. So lets get it merged! 🥇

lutien:
I've picked up the work on `storage.getCookies` command in the scope of this bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1854580.
@sadym-chromium, let me know if you started with the wpt tests already. If not, I'm going to add them with the implementation.

sadym-chromium:
> I've picked up the work on `storage.getCookies` command in the scope of this bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1854580. @sadym-chromium, let me know if you started with the wpt tests already. If not, I'm going to add them with the implementation.

I started working on implementation, but not on WPT. We can split the work on WPT

lutien:
> I started working on implementation, but not on WPT. We can split the work on WPT

Sounds good to split :) 
What do you think what would be the best way to do it? I could only think about that I take tests for `getCookies` and you for `setCookies`, but I'm open to suggestions

PR #512, opened 2023-08-21
Fix CDDL

jrandolf:
We need to use `/` instead of `//` as `//` are meant to union groups, not types.

For example, `ResultData` is used as the type for `CommandResponse.result`, but `ResultData` is a group (since it uses `//`). Only types may be used as values.

This will be validated once `cddl` 1.0.0 comes out.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/512.html" title="Last updated on Aug 21, 2023, 12:09 PM UTC (19ecb28)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/512/45ceaac...19ecb28.html" title="Last updated on Aug 21, 2023, 12:09 PM UTC (19ecb28)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -331,8 +331,8 @@ EmptyParams = {
 
 <pre class="cddl local-cddl">
 Message = (
-  CommandResponse //
-  ErrorResponse //
+  CommandResponse /
+  ErrorResponse /
   Event
 )
 
@@ -353,10 +353,10 @@ ErrorResponse = {
 }
 
 ResultData = (
-  BrowsingContextResult //
-  EmptyResult //
-  NetworkResult //
-  ScriptResult //
+  BrowsingContextResult /
+  EmptyResult /
+  NetworkResult /
+  ScriptResult /
   SessionResult
 )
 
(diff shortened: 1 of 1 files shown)


OrKoN (approved):

OrKoN:
@jgraham @whimboo PTAL

thiagowfx (approved):

index.bs
```diff
@@ -1432,10 +1432,10 @@ This is a [=static command=].
    <dt>Command Type</dt>
    <dd>
       <pre class="cddl remote-cddl">
-      session.Status = {
+      session.Status = (
```
juliandescottes, line 1435: Can you explain why those were changed from `{` to `(`

jrandolf: Because `{` are used for types, not groups.

jrandolf: More specifically, `SessionCommand` is a set of group choices (not type choices because it uses `//`), so its constituents need to be groups, i.e. rules defined with `(`.

The reason `SessionCommand` **must** be a set of group choices is because it later gets spread into `Command` via `CommandData`. Only a set of group choices are allowed to be spread, not a type (like `{`).

juliandescottes (approved):
Makes sense, thanks for the explanation

Issue #507, opened 2023-08-17
Fix URLPattern matching


jgraham:
@juliandescottes left a bunch of comments on https://github.com/w3c/webdriver-bidi/pull/429 which weren't resolved before the PR landed, in the interests of making some progress.

The general theory here is that we want the behaviour to be a subset of https://wicg.github.io/urlpattern/ with only support for literals (no wildcards, regexps, etc.)

juliandescottes:
Some of the follow up items I mentioned:

- If a "string" pattern has no `search`, it should not match URLs with a search. For instance, `{type: "string", pattern: "http://example.com"}` should not match "http://example.com/?test". The implementation of URLPattern in Chrome matches this. In theory it should also not match "http://example.com/?", but Chrome does not differentiate between empty search and null search.
- If a "pattern" pattern has no `pathname` property set, it should match URLs with any pathname. For instance, `{type: "pattern", protocol: "http"}` should match "http://abc.com/somepath". Chrome's URLPattern matches this. Same for port and search

juliandescottes:
While the second bullet point is rather straightforward, for the first one it's a bit more complicated.

Our BiDi spec uses the URL parser / basic URL parser to process string URLs. Based on that spec, when there is no `?` in the URL, the Basic URL Parser should return null for the URL's query. And then, if we process this according to the URL Pattern spec, this means we should consider this as a wildcard and match everything. See https://wicg.github.io/urlpattern/#compile-a-component.
 
However, the URLPattern spec is not using the URL Parser to process string input constructor. It's following the steps at https://wicg.github.io/urlpattern/#constructor-string-parsing, which if I'm reading them correctly, is setting an empty string by default for the `search` (step 2.2.1.4). And that would match what Chrome's behavior I believe.

We should either stop using the URL Parser for processing URLPatternString patterns, or at least modify the defaults enough in order to match the logic of the URLPattern parser.

juliandescottes:
It's also worth noting that the match algorithm from URLPattern will force the empty string as the default value for several components, including search, see https://wicg.github.io/urlpattern/#match 

> Set hostname to url’s [host](https://url.spec.whatwg.org/#concept-url-host) or the empty string if the value is null.
Set port to url’s [port](https://url.spec.whatwg.org/#concept-url-port) or the empty string if the value is null.
Set search to url’s [query](https://url.spec.whatwg.org/#concept-url-query) or the empty string if the value is null.

[...]

PR #513, opened 2023-08-22
Fix missing components in pattern patterns and default search value for string pattern

juliandescottes:
Attempts to fix a few issues:
- when no value is provided for protocol, pathname or search in a URLPatternPattern, we should not attempt to match against it
- with a URLPatternString, the search component should default to empty string, not null


index.bs
```diff
@@ -4684,26 +4702,47 @@ To <dfn>parse URL pattern</dfn>, given |pattern|:
...
+          [=default port=] is not null, and |url|'s [=url/port=] is null or is
+          equal to [=url/scheme=]'s [=default port=]:
 
-     1. Otherwise:
+         1. The empty string.
 
-       1. null.
+       1. Otherwise, if |url|'s [=url/port=] is not null:
+
+         1. [=Serialize an integer=] with |url|'s [=url/port=].
```
thiagowfx, line 4719: Do you also need to mention here the integer range for a valid port number?

juliandescottes: At the moment we don't have to: this will already raise an error when trying to parse the URL, near the end of the parsing step. 

Now if we stop using the basic URL parser here (and it might make sense based on https://github.com/w3c/webdriver-bidi/issues/507#issuecomment-1688462008), adding this validation would make sense.

juliandescottes:
cc @jgraham for review

juliandescottes:
Small review ping for @jgraham

index.bs
```diff
@@ -4684,26 +4702,47 @@ To <dfn>parse URL pattern</dfn>, given |pattern|:
...
      <dt>search
-     <dd>|url|'s [=url/query=].
+     <dd>
+     1. If |has search| is false:
+
+       1. null.
+
+     1. Otherwise:
+
+       1. The empty string if |url|'s [=url/path=] is null, or |url|'s [=url/query=] otherwise.
```
jgraham, line 4745: Why do we set the empty string if the path is null?

juliandescottes: The idea is to match the [URLPattern's spec match algorithm](https://wicg.github.io/urlpattern/#match), which defaults to an empty string if the url's search is null. 

Basically this algorithm will never differentiate between empty string search or no search, so I always turn null into empty string, both here, and when reading the components of the URL to match.

juliandescottes: I missed that this was `path` instead of `query` should be fixed now!

jgraham (approved):

juliandescottes:
@thiagowfx maybe you want to take another look at this for a final review?

index.bs
```diff
@@ -4748,8 +4787,13 @@ To <dfn>match URL pattern</dfn> given |url pattern| and |url string|:
 1. If |url pattern|'s pathname is not null and is not equal to the result of
    running the [=URL path serializer=] with |url|, return false.
 
-1. If |url pattern|'s search is not null and is not equal to |url|'s
-   [=url/query=], return false.
+1. If |url pattern|'s search is not null:
+
+   1. Let |url query| be |url|'s [=url/query=]
```
thiagowfx, line 4792:
```suggestion
   1. Let |url query| be |url|'s [=url/query=].
```

thiagowfx (approved):
Are you planning to add WPT tests for these edge cases?

juliandescottes:
Thanks for the review! 

> Are you planning to add WPT tests for these edge cases?

I tried to cover it already url_patterns_tentative.py (that's why we have the "tentative" suffix actually). Once this PR lands I will remove the suffix. That will happen downstream at the same time as I update our implementation to match the new spec: https://bugzilla.mozilla.org/show_bug.cgi?id=1849743

Issue #494, opened 2023-07-14
Set input type=file content


OrKoN:
It is mentioned in the roadmap spreadsheet, but not tracked as an issue here.

CDP equivalent: DOM.setFileInputFiles

OrKoN:
assigning to @jrandolf as discussed before.

whimboo:
@OrKoN would this still be an event you are interested in? The PR from @jrandolf-2 got closed a while ago.

OrKoN:
The PR is still open but it needs to be updated to leverage user prompt handler configuration to decide how to handle file dialogs. cc @sadym-chromium

sadym-chromium:
Addressed

PR #514, opened 2023-08-22
Add `input.SetFiles`

jrandolf:
Issue #494

File dialog PR: https://github.com/w3c/webdriver-bidi/pull/568

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/514.html" title="Last updated on Jan 12, 2024, 1:38 PM UTC (337fe1f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/514/6ebacb6...337fe1f.html" title="Last updated on Jan 12, 2024, 1:38 PM UTC (337fe1f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -606,7 +606,7 @@ with the following additional codes:
 
   <dt><dfn>underspecified storage partition</dfn>
   <dd>Tried to interact with data in a storage partition which was not adequately specified.
-  
+
   <dt><dfn>unable to set file input</dfn>
   <dd>Tried to set a file input, but failed to do so.
 </dl>
@@ -10134,8 +10134,8 @@ The [=remote end steps=] given |session|, and |command parameters| are:
 
 #### The input.setFiles Command ####  {#command-input-setFiles}
 
-The <dfn export for=commands>input.setFiles</dfn> command sets a given file
-input with a given set of file paths.
+The <dfn export for=commands>input.setFiles</dfn> command sets the <code>files</code> property of a given <code>input<code> element with type <code>file</code>
+to a set of file paths.
 
 <dl>
    <dt>Command Type</dt>
@@ -10183,23 +10183,28 @@ The [=remote end steps=] given |session| and |command parameters| are:
 1. Let |element| be the result of [=trying=] to [=deserialize remote reference=]
    with |command parameters|["<code>element</code>"], |realm|, and |session|.
 
+1. If |element| doesn't implement {{Element}}, return [=error=] with [=error code=]
+   [=no such element=].
+
+1. If |element| doesn't implement {{HTMLInputElement}},
+   |element|["<code>type</code>"] is not "file", or
+   |element|["<code>disabled</code>"] is true, return [=error=] with [=error
+   code=]
+   [=unable to set file input=].
+
 1. Let |files| be the value of the |command parameters|["<code>files</code>"]
    field.
 
 1. Let |selected files| be |element|'s [=selected files=].
-   
-1. If the [=set/size=] of the [=set/intersection=] of |files| and |selected files| is
-   equal to the [=set/size=] of |selected files|:
-   
-   1. [=Queue an element task=] on the [=user interaction task source=] given
-      |element| to fire an event named <code>cancel</code> at |element|, with
-      the <code>bubbles</code> attribute initialized to true.
 
-      Note: Although there may not be any dialog, cancellation in a browser is
-      typically determined by changes in file selection. In other words, if
-      there is no change, a "cancel" event should be sent.
-   
-   1. Return [=success=] with data null.
+1. If the [=set/size=] of the [=set/intersection=] of |files| and |selected
+   files| is equal to the [=set/size=] of |selected files|, [=queue an element
+   task=] on the [=user interaction task source=] given |element| to fire an
+   event named <code>cancel</code> at |element|, with the <code>bubbles</code>
+   attribute initialized to true.
+
+   Note: Cancellation in a browser is typically determined by changes in file
+   selection. In other words, if there is no change, a "cancel" event is sent.
 
 1. Otherwise, [=update the file selection=] for |element| with |files| as the
    user's selection.
--- a/index.bs
+++ b/index.bs
@@ -10134,7 +10134,7 @@ The [=remote end steps=] given |session|, and |command parameters| are:
 
 #### The input.setFiles Command ####  {#command-input-setFiles}
 
-The <dfn export for=commands>input.setFiles</dfn> command sets the <code>files</code> property of a given <code>input<code> element with type <code>file</code>
+The <dfn export for=commands>input.setFiles</dfn> command sets the <code>files</code> property of a given <code>input</code> element with type <code>file</code>
 to a set of file paths.
 
 <dl>
--- a/index.bs
+++ b/index.bs
@@ -199,6 +199,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
     text: hidden; url: document-sequences.html#system-visibility-state
     text: history handling behavior; url: browsing-the-web.html#history-handling-behavior
     text: innerText getter steps; url:dom.html#dom-innertext
+    text: input type; url: input.html#dom-input-type
     text: navigables; url: document-sequences.html#navigables
     text: navigation id; url: browsing-the-web.html#navigation-id
     text: origin-clean; url: canvas.html#concept-canvas-origin-clean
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -8923,6 +8930,79 @@ The [=remote end steps=] given |session|, and |command parameters| are:
...
+   |command parameters|.
+
+1. Let |context| be the result of [=trying=] to [=get a browsing context=]
+   with |context id|.
+
+1. Let |top-level context| be |context|'s [=top-level browsing context=].
+
+<!-- converting to realm here is kind of silly since we immediately convert back
+      in [=deserialize remote reference=] -->
+1. Let |document| be |top-level context|'s [=active document=].
```
OrKoN, line 8976: why do we need to access the document of the top-level browsing context?

OrKoN: it sounds like environment settings can be retried from the `|context|'s [=active document=].`?

index.bs
```diff
@@ -8923,6 +8930,79 @@ The [=remote end steps=] given |session|, and |command parameters| are:
...
+
+1. If |element| doesn't implement {{HTMLInputElement}} return [=error=] with [=error code=]
+   [=unable to set file input=].
+
+1. If the [=input type=] of |element| doesn't match <code>"file"</code> return [=error=] with [=error code=]
+   [=unable to set file input=].
+
+1. Let |files| be the value of the <code>files</code> field of |command parameters|.
+
+1. [=Update the file selection=] for |element| with |files| as if a user
```
OrKoN, line 8999: `as if a user selected |files| through the [=show the picker|file picker=] prompt` seems redundant? as it's not clear if it's actionable and in what way?

jrandolf: I think the problem here is this is implementation-specific, so I'm not sure if we can be more specific. @whimboo Is this sufficient? From Chromium's standpoint, this has a precise meaning because we are able to set an input with an arbitrary set of paths provided by either CDP or the file picker prompt.

jrandolf: Note: `Update the file selection` points to a spec-defined action.

OrKoN:
@jgraham @whimboo @juliandescottes PTAL Perhaps you have suggestions for the open discussions and some other suggestions.

index.bs
whimboo: Please also see https://github.com/w3c/webdriver/issues/1742 for an open issue with classic. We should make sure to implement it correctly in BiDi.

index.bs
```diff
@@ -9036,6 +9044,98 @@ The [=remote end steps=] given |session|, and |command parameters| are:
...
+      [=error code=] [=unable to set file input=].
+
+   1. If the [=input type=] of |element| doesn't match <code>"file"</code>
+      return [=error=] with [=error code=] [=unable to set file input=].
+
+   1. [=trying|Try=] [=show the picker|showing the picker=] with |element|.
+
+1. Let |file dialogs| be |context|'s [=active file dialogs=].
+
+1. Assert: |file dialogs| [=contains=] |element|.
```
OrKoN, line 9113: Let's make it a conditional and move it to before trying to show a picker?

index.bs
```diff
@@ -9036,6 +9044,98 @@ The [=remote end steps=] given |session|, and |command parameters| are:
...
+    <pre class="cddl remote-cddl">
+      input.SetFiles = (
+        method: "input.setFiles",
+        params: input.SetFilesParameters
+      )
+
+      input.SetFilesParameters = {
+        context: browsingContext.BrowsingContext,
+        files: [*text]
+        ? element: script.SharedReference,
```
OrKoN, line 9064:
```suggestion
        element: script.SharedReference,
```

index.bs
```diff
@@ -9036,6 +9044,98 @@ The [=remote end steps=] given |session|, and |command parameters| are:
...
+   [=relevant global object=]'s <a>associated <code>Document</code></a> is
+   |document|.
+
+1. Let |realm| be |environment settings|'s [=realm execution context=]'s
+   Realm component.
+
+1. Let |reference| be the value of the <code>element</code> field of
+   |command parameters|.
+
+1. If |reference| is not null,
```
OrKoN, line 9098: it looks like the rest of the algorithm requires an element right?

index.bs
```diff
@@ -9036,6 +9044,98 @@ The [=remote end steps=] given |session|, and |command parameters| are:
...
+   1. Let |element| be the result of [=trying=] to [=deserialize remote
+      reference=] with |reference|, |realm|, and |session|.
+
+   1. If |element| doesn't implement {{HTMLInputElement}} return [=error=] with
+      [=error code=] [=unable to set file input=].
+
+   1. If the [=input type=] of |element| doesn't match <code>"file"</code>
+      return [=error=] with [=error code=] [=unable to set file input=].
+
+   1. [=trying|Try=] [=show the picker|showing the picker=] with |element|.
```
OrKoN, line 9109: I am not sure we really need to trigger the picker when we set the files. Perhaps, we can setFiles on an input unconditionally no matter if the dialog was shown or not.

jgraham: Yes, agreed. We should close the picker if it's already showing when this is called, but if it's not showing we should simply set the input value directly.

index.bs
```diff
@@ -9036,6 +9044,98 @@ The [=remote end steps=] given |session|, and |command parameters| are:
...
+1. Assert: |file dialogs| [=contains=] |element|.
+
+1. [=set/Remove=] |element| from |file dialogs|.
+
+1. Let |files| be the value of the <code>files</code> field of |command
+   parameters|.
+
+1. Let |selected files| be |element|'s [=selected files=].
+   
+1. If the [=set/size=] of the [=set/intersection=] of |files| and |selected files| is
```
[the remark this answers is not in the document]
jrandolf: If the selected files and files is the same, then we should send a cancel event. This follows the current behavior of Chromium for file selection. Does this follow for Firefox?

OrKoN (approved):
LGTM

index.bs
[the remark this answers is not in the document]
jrandolf: The previous logic was incorrect. Perhaps it looks right now.

index.bs
```diff
@@ -9547,6 +9554,83 @@ The [=remote end steps=] given |session|, and |command parameters| are:
...
+1. Let |realm| be |environment settings|'s [=realm execution context=]'s
+   Realm component.
+
+1. Let |reference| be the value of the |command
+   parameters|["<code>element</code>"] field.
+
+1. Let |files| be the value of the |command parameters|["<code>files</code>"]
+   field.
+
+1. Let |selected files| be |element|'s [=selected files=].
```
OrKoN, line 9611: Looks like the |element| definition got lost

whimboo:
Looks like this PR stalled a bit. @jrandolf mind giving a quick update what's the status?

jrandolf:
@whimboo Waiting on File dialog PR: https://github.com/w3c/webdriver-bidi/pull/568

index.bs
```diff
@@ -9636,6 +9643,82 @@ The [=remote end steps=] given |session|, and |command parameters| are:
...
+1. Let |environment settings| be the [=environment settings object=] whose
+   [=relevant global object=]'s <a>associated <code>Document</code></a> is
+   |document|.
+
+1. Let |realm| be |environment settings|'s [=realm execution context=]'s
+   Realm component.
+
+1. Let |element| be the result of [=trying=] to [=deserialize remote reference=]
+   with |command parameters|["<code>element</code>"], |realm|, and |session|.
+
```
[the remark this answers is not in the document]
whimboo: @jrandolf are you going to address this comment as well? Maybe we should even check that it is an `input` element of type `file`?

index.bs
```diff
@@ -10125,6 +10132,86 @@ The [=remote end steps=] given |session|, and |command parameters| are:
...
+1. If |element| doesn't implement {{HTMLInputElement}}, return [=error=] with [=error code=]
```
whimboo, line 10186: I actually wonder if we should only fail if the element doesn't implement `Element` here, and do the input check together with type `file` in the next step. I feel that this would make more sense given that we actually might have found an element but not with the right type.

jrandolf: Makes sense. Done.

whimboo (approved):
Thanks for the updates. For me it looks fine. @jgraham can you please take a final look as well?

jgraham (approved):

index.bs
```diff
@@ -10125,6 +10132,87 @@ The [=remote end steps=] given |session|, and |command parameters| are:
 
 </div>
 
+#### The input.setFiles Command ####  {#command-input-setFiles}
+
+The <dfn export for=commands>input.setFiles</dfn> command sets the <code>files</code> property of a given <code>input<code> element with type <code>file</code>
```
jgraham, line 10137:
```suggestion
The <dfn export for=commands>input.setFiles</dfn> command sets the <code>files</code> property of a given <code>input</code> element with type <code>file</code>
```

whimboo:
@jrandolf are you planning to write necessary wpt tests for this new feature?

PR #518, opened 2023-08-23
Extend `browsingContext.close` params with `promptUnload`

### LINKED ISSUE #462: Add flag to `browsingContext.close` to skip beforeunload
Original issue https://github.com/w3c/webdriver-bidi/issues/170, Original discussion comment - https://github.com/w3c/webdriver-bidi/issues/170#issuecomment-1034033101
The issue was no discussed in depth as that focused more on https://github.com/w3c/webdriver-bidi/issues/187
Currently when we close a BrowsingContext we need to run [close-a-top-level-traversable](https://html.spec.whatwg.org/multipage/document-sequences.html#close-a-top-level-traversable) in there [check if unloading is canceled](https://html.spec.whatwg.org/multipage/browsing-the-web.html#checking-if-unloading-is-canceled). 
From an user perspective I may want to close the BrowsingContext to free resources regardless if there is unload handlers or not.

Puppeteer CDP already supports this at [Page.close](https://pptr.dev/api/puppeteer.page.close), 
where default is to skip `beforeunload`.

Lightning00Blade:
Closes #462

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2470,7 +2470,8 @@ The <dfn export for=commands>browsingContext.close</dfn> command closes a
       }
 
       browsingContext.CloseParameters = {
-        context: browsingContext.BrowsingContext
+        context: browsingContext.BrowsingContext,
+        ? promptUnload: bool .default false
       }
       </pre>
    </dd>
@@ -2489,6 +2490,9 @@ The [=remote end steps=] with |command parameters| are:
   1. Let |context id| be the value of the <code>context</code> field of
      |command parameters|.
 
+  1. Let |prompt unload| be the value of the <code>promptUnload</code> field of
+     |command parameters|.
+
   1. Let |context| be the result of [=trying=] to [=get a browsing context=]
      with |context id|.
 
@@ -2497,7 +2501,13 @@ The [=remote end steps=] with |command parameters| are:
   1. If |context| is not a [=top-level browsing context=], return [=error=] with
      [=error code=] [=invalid argument=].
 
-  1. [=Close=] |context|.
+  1. If |prompt unload| is true:
+     
+     1. [=Close=] |context|.
+
+  1. Else:
+   
+     1. [=Close=] |context| without [=prompting to unload=].
 
   1. Return [=success=] with data null.
 


OrKoN (approved):
LGTM (wait for other reviewers)

index.bs
```diff
@@ -2489,6 +2490,9 @@ The [=remote end steps=] with |command parameters| are:
   1. Let |context id| be the value of the <code>context</code> field of
      |command parameters|.
 
+  1. Let |prompt unload| be the value of the <code>promptUnload</code> field of
+     |command parameters|.
```
juliandescottes, line 2494: Since the parameter is optional I think we need to check if |command parameters| contains "promptUnload" first?

juliandescottes: My mistake, since there is a default value this is probably not needed.

OrKoN:
@juliandescottes @jgraham @whimboo would it be possible for you to review this one? it seems uncontroversial

juliandescottes (approved):
That looks fine to me, thanks

index.bs
```diff
@@ -2497,7 +2501,13 @@ The [=remote end steps=] with |command parameters| are:
   1. If |context| is not a [=top-level browsing context=], return [=error=] with
      [=error code=] [=invalid argument=].
 
-  1. [=Close=] |context|.
+  1. If |prompt unload| is true:
+     
+     1. [=Close=] |context|.
+
+  1. Else:
```
OrKoN, line 2508:
```suggestion
  1. Otherwise:
```

OrKoN:
@jgraham should we land this change or do you have concerns about exposing this?

thiagowfx:
Is there a bug or PR for the corresponding WPT test?

jgraham:
I think this is OK in principle.

I also think we should fix this to not work by "magic" but actually modify https://html.spec.whatwg.org/#steps-to-fire-beforeunload to skip the prompt in some cases. It seems like we could reuse the existing `WebDriver BiDi user prompt opened` steps by allowing them to return a value that specifies whether to show the prompt or not, and keeping a set of browsing contexts for which the unload prompt should be skipped.

That doesn't block landing this one, but we should at least file an issue.

Lightning00Blade:
Issue on HTML spec - https://github.com/whatwg/html/issues/9890 , please add anything I may have missed.
Merging this per the last comment.

whimboo:
@Lightning00Blade Thanks for your work! Are you going to also create a test for that?

Lightning00Blade:
Yes I will add WPT test for it, I already updated Chromium-BiDi https://github.com/GoogleChromeLabs/chromium-bidi/pull/1508 to support it, so it should be easier to create them now.

whimboo:
> Yes I will add WPT test for it, I already updated Chromium-BiDi [GoogleChromeLabs/chromium-bidi#1508](https://github.com/GoogleChromeLabs/chromium-bidi/pull/1508) to support it, so it should be easier to create them now.

Hi @Lightning00Blade. Quick question if you still have it in your queue. Thanks.

whimboo:
FYI the tests will be added via https://github.com/web-platform-tests/wpt/pull/44004.

Issue #473, opened 2023-07-05
`browsingContext.print`: should emit `InvalidArgument` when printing an empty content area


thiagowfx:
- WPT repository issue: https://github.com/web-platform-tests/wpt/issues/40870
- WPT repository PR: https://github.com/web-platform-tests/wpt/pull/40872

The `test_margin_same_as_page_dimension` test in `webdriver/tests/bidi/browsing_context/print /margin.py` in WPT currently emits a PDF and checks that it has empty content:

```python
@pytest.mark.parametrize(
    "margin",
    [
        {"top": 27.94},
        {"left": 21.59},
        {"right": 21.59},
        {"bottom": 27.94},
        {"top": 27.94, "left": 21.59, "right": 21.59, "bottom": 27.94},
    ],
    ids=[
        "top",
        "left",
        "right",
        "bottom",
        "all",
    ],
)
async def test_margin_same_as_page_dimension(
    bidi_session,
    top_context,
    inline,
    assert_pdf_content,
    margin,
):
    page = inline("Text")
    await bidi_session.browsing_context.navigate(
        context=top_context["context"], url=page, wait="complete"
    )
    print_value = await bidi_session.browsing_context.print(
        context=top_context["context"],
        shrink_to_fit=False,
        margin=margin,
    )

    # Check that content is out of page dimensions.
    await assert_pdf_content(print_value, [{"type": "string", "value": ""}])
```

For context, this happens to works on Firefox, but it does not work on Chromium, but this is not the point (the spec should be browser agnostic). On Chromium, it throws `webdriver.bidi.error.UnknownErrorException: unknown error (invalid print parameters: content area is empty)`, which is technically correct. Chromium cannot handle printing an empty content area.

Putting vendors aside, the purpose of this issue is to question the validity of printing a page in either of these scenarios, which all produce an empty content area:

- `width` is zero
- `height` is zero
- left margin plus right margin is greater than or equal to `width` (=effectively zero width)
- top margin plus bottom margin is greater than or equal to `height` (=effectively zero height)

My understanding is that currently the BiDi spec implicitly allows this condition (@whimboo seems to [agree](https://github.com/web-platform-tests/wpt/issues/40870#issuecomment-1620638128)).

That said, with further investigation...

@whimboo ([source](https://github.com/web-platform-tests/wpt/pull/40872#discussion_r1252764498))

[...]

PR #522, opened 2023-08-31
browsing context: print: enforce minimum page size dimensions

thiagowfx:
The minimum possible page size is 1x1 point.

1 point is equal to (2.54 / 72) cm.

Docs: https://www.w3.org/TR/css3-values/#absolute-lengths
Docs: https://stackoverflow.com/questions/139655/convert-pixels-to-points
WPT: https://github.com/web-platform-tests/wpt/pull/40872
Fixed: #473

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/522.html" title="Last updated on Aug 31, 2023, 2:56 PM UTC (14dacc4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/522/ec67d1d...14dacc4.html" title="Last updated on Aug 31, 2023, 2:56 PM UTC (14dacc4)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -205,6 +205,9 @@ spec: HR-TIME; urlPrefix: https://w3c.github.io/hr-time/
 spec: RFC4648; urlPrefix: https://tools.ietf.org/html/rfc4648
   type: dfn
     text: Base64 Encode; url: section-4
+spec: CSS-VALUES-3; urlPrefix: https://drafts.csswg.org/css-values-3/
+  type: dfn
+    text: absolute lengths; url: #absolute-lengths
 spec: CSSOM-VIEW; urlPrefix: https://drafts.csswg.org/cssom-view/
   type: dfn
     text: visual viewport; url: #visual-viewport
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2825,8 +2825,8 @@ PDF document represented as a Base64-encoded string.
       }
 
       browsingContext.PrintPageParameters = {
-        ? height: (float .ge 0.0) .default 27.94,
-        ? width: (float .ge 0.0) .default 21.59,
+        ? height: (float .ge (2.54 / 72)) .default 27.94,
+        ? width: (float .ge (2.54 / 72)) .default 21.59,
```
thiagowfx, line 2829: Unfortunately this is a repeating decimal, so I opted for the fraction representation instead. We could also round up (but not down), if preferred.

index.bs
```diff
@@ -2862,6 +2862,8 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
    <code>browsingContext.PrintPageParameters</code> with the fields set to
    their default values.
 
+Note: The minimum page size is 1 point, which is (2.54 / 72) cm.
```
thiagowfx, line 2865: This is just for documentation purposes. It's not strictly needed. It could be nice to link to https://www.w3.org/TR/css3-values/#absolute-lengths.

thiagowfx: Added the aforementioned link inline.

Lightning00Blade (approved):
LGTM

thiagowfx:
It would be nice to have another vendor than Google review it as well: cc @whimboo @jgraham

thiagowfx:
Ping

whimboo (approved):
Sorry for the delay. These changes are in sync with the outcome of the discussion during our W3C meeting. And yes points are the right unit to use here given that this is generally used for printing.

Issue #506, opened 2023-08-17
Meta: WebDriver BiDi extensions


mathiasbynens:
WebDriver Classic allowed other specs & vendors to define [extension commands](https://w3c.github.io/webdriver/#extensions-0) and [capabilities](https://w3c.github.io/webdriver/#dfn-additional-webdriver-capability).

Similarly, WebDriver BiDi allows [extension modules](https://w3c.github.io/webdriver-bidi/#extension-modules). Let’s use this issue to track such modules in other specs and repos:

- [x] https://github.com/WebBluetoothCG/web-bluetooth/issues/613

OrKoN:
cc @mathiasbynens @whimboo should we track extensions via this issue or have it on the space page?

whimboo:
What is the `space page`?

OrKoN:
@whimboo sorry, a typo, I meant to be "spec" (i.e., https://w3c.github.io/webdriver-bidi/)

PR #523, opened 2023-08-31
Extend roadmap: permissions automation

outofambit:
My team is working on automated bluetooth testing in WPT, which has no way to automate managing device permissions. Instead of implementing something chrome specific, we’d like to explore device access/permission automation in BiDi. We’ll follow up with a design approach in an issue but wanted to start alignment conversations now.

- BiDi extensions: https://github.com/w3c/webdriver-bidi/issues/506
- Manage Bluetooth device permissions: https://github.com/WebBluetoothCG/web-bluetooth/issues/613

### DIFF
--- a/roadmap.md
+++ b/roadmap.md
@@ -107,8 +107,8 @@ This scenario changes the browser window dimensions and position, maximizes/mini
 - [x] Some items from the previous milestones
 - [ ] [Add commands for resizing and positioning browser windows](https://github.com/w3c/webdriver-bidi/issues/398)
 
-### Managing permissions
+### Extensions
 
-This scenario automates user granting permission for a browser to access different private information.
+[Define how extension modules could be created.](https://github.com/w3c/webdriver-bidi/issues/506)
 
-- [ ] Bluetooth device chooser
+- [ ] Document a general device access pattern (for use with Web Bluetooth and other APIs)
--- a/roadmap.md
+++ b/roadmap.md
@@ -111,4 +111,4 @@ This scenario changes the browser window dimensions and position, maximizes/mini
 
 This scenario automates user granting permission for a browser to access different private information.
 
-- [ ] Device access: bluetooth
+- [ ] Bluetooth device chooser
--- a/roadmap.md
+++ b/roadmap.md
@@ -109,6 +109,6 @@ This scenario changes the browser window dimensions and position, maximizes/mini
 
 ### Extensions
 
-[Define how extension modules could be created.](https://github.com/w3c/webdriver-bidi/issues/506)
+[As a spec author I want to extend WebDriver BiDi functionality, therefore I define a WebDriver BiDi extension in the spec following WebDriver BiDi extension guidelines.](https://github.com/w3c/webdriver-bidi/issues/506)
 
-- [ ] Document a general device access pattern (for use with Web Bluetooth and other APIs)
+- [ ] [Manage Bluetooth device permissions](https://github.com/WebBluetoothCG/web-bluetooth/issues/613)
--- a/roadmap.md
+++ b/roadmap.md
@@ -106,3 +106,9 @@ This scenario changes the browser window dimensions and position, maximizes/mini
 
 - [x] Some items from the previous milestones
 - [ ] [Add commands for resizing and positioning browser windows](https://github.com/w3c/webdriver-bidi/issues/398)
+
+### Managing permissions
+
+This scenario automates user granting permission for a browser to access different private information.
+
+- [ ] Device access: bluetooth


outofambit:
cc @lolaodelola @mzgoddard @boazsender

christian-bromann:
Hey @outofambit I am not quite sure but would fall this under this specification: https://w3c.github.io/permissions ?

OrKoN:
I think permissions were discussed in https://github.com/w3c/webdriver-bidi/issues/403 and looks like the agreement want that we want to do something in WebDriver BiDi. I think we should discuss it again.

mathiasbynens:
This could become the first ever WebDriver BiDi (rather than WebDriver Classic) extension, as in, another spec defines its own commands/events. Exciting! cc @thiagowfx

jugglinmike:
Folks looking to take this on might be interested in [the 2017 collaboration which informed the design of the WebDriver Classic Extension Command](https://github.com/w3c/permissions/pull/151). While some details may be worth replicating, I imagine that thread will be just as useful in determining what to discard (due to the bi-directional paradigm shift, BiDi's richer abstractions for browsing contexts, and six years of refinements to the Permissions specification).

The editorial history of that section may also be instructive--[it was removed in 2021 for lack of implementation interest](https://github.com/w3c/permissions/issues/258) and [reinstated a few months later](https://github.com/w3c/permissions/pull/346).

> This could become the first ever WebDriver BiDi (rather than WebDriver Classic) extension, as in, another spec defines its own commands/events. Exciting! cc @thiagowfx

If my experience with WebDriver is any indication, we can also expect it to surface latent issues in the specification's extension mechanism.

outofambit:
hi @OrKoN! I made a small update to the phrasing in the roadmap item to emphasize the "chooser" aspect of the bluetooth permission automation since that is somewhat unique compared to other permissions APIs. Based on the conversation today, do you think we can merge this PR in?

thiagowfx (approved):
LGTM but please wait for review from another project member as well

OrKoN:
My understanding based on the discussion was that we want the specification for the Bluetooth specific part to be part of the Bluetooth spec and perhaps only define a general pattern for such interactions in the WebDriver BiDi spec. Perhaps we should add a more general item to the roadmap about supporting/clarifying how extensions could be defined and documenting the more general device access pattern (similar to the [DeviceAccess](https://chromedevtools.github.io/devtools-protocol/tot/DeviceAccess/) in CDP)? For example, the existing permissions extension in WebDriver https://w3c.github.io/permissions/#automation

Also, should we file an issue in https://github.com/WebBluetoothCG/web-bluetooth/issues/ for specifying the Bluetooth specific extension and an issue in this repo for providing guidelines for extensions? We could link those issues in the roadmap. WDYT?

thiagowfx:
I'm going to do the implementation in the chromium mapper side of whatever comes out of this spec, so please keep me in the loop as well, thank you Evelyn :-)

roadmap.md
```diff
@@ -106,3 +106,9 @@ This scenario changes the browser window dimensions and position, maximizes/mini
 
 - [x] Some items from the previous milestones
 - [ ] [Add commands for resizing and positioning browser windows](https://github.com/w3c/webdriver-bidi/issues/398)
+
+### Extensions
+
+[Define how extension modules could be created.](https://github.com/w3c/webdriver-bidi/issues/506)
```
OrKoN, line 112: this line should describe the scenario, for example, something along the lines of "as a spec author I want to extend WebDriver BiDi functionality, therefore I define a WebDriver BiDi extension in the spec following WebDriver BiDi extension guidelines". The current text should become an action item in the list below.

@jgraham @whimboo do you think it's a good item for the roadmap? with the proposed changes it looks good to me. We can probably exclude the device access pattern as I am not sure it is a cross-browser pattern.

outofambit: @OrKoN Made this change, thank you for clarifying the right framing here. I also modified the bullet point to link to the issue in https://github.com/WebBluetoothCG/web-bluetooth/, does that seem right to you?

OrKoN (commented):
This looks fine to me, please wait for the feedback from other vendors.

thiagowfx:
cc @jgraham @whimboo PTAL

thiagowfx:
Friendly ping^ (https://github.com/w3c/webdriver-bidi/pull/523#pullrequestreview-1650111021)

roadmap.md
```diff
@@ -106,3 +106,9 @@ This scenario changes the browser window dimensions and position, maximizes/mini
 
 - [x] Some items from the previous milestones
 - [ ] [Add commands for resizing and positioning browser windows](https://github.com/w3c/webdriver-bidi/issues/398)
+
+### Extensions
+
+[As a spec author I want to extend WebDriver BiDi functionality, therefore I define a WebDriver BiDi extension in the spec following WebDriver BiDi extension guidelines.](https://github.com/w3c/webdriver-bidi/issues/506)
+
+- [ ] [Manage Bluetooth device permissions](https://github.com/WebBluetoothCG/web-bluetooth/issues/613)
```
whimboo, line 114: Are you planning to have a long list of all the different permissions APIs listed here in the future, or should we have just a generic item for specifically the `Permissions`. I would find that more helpful given that there will be much more other extensions in the future like WebAuthn (which we potentially should support as well as for classic).

OrKoN:
I think we will not extend the roadmap with permissions and web bluetooth since those are defined in external specifications and automation algorithms can be defined there. This already happened for Permissions and WebBluetooth:

- https://www.w3.org/TR/permissions/#automation-webdriver-bidi
- https://webbluetoothcg.github.io/web-bluetooth/#automated-testing

Issue #527, opened 2023-09-05
The `BrowsingContext.reload` command needs to return a NavigateResult


whimboo:
Currently the [`browsingContext.reload`](https://w3c.github.io/webdriver-bidi/#command-browsingContext-reload) command returns an `EmptyResult` and not a `NavigationResult`. This is problematic because clients will require the `navigation id` to check for the navigation completed (in combination with the `wait` argument for `none` and `interactive`) or even for specific network events.

jgraham:
Note that the spec here is just broken: https://w3c.github.io/webdriver-bidi/#command-browsingContext-reload has unconditional `return` statements in both steps 9 and 10, and the one in step 9 will return a `browsingContext.NavigateResult`, contrary to the CDDL which claims `EmptyResult`.

It appears that #382 broke this as it didn't realise we were already returning a `browsingContext.NavigateResult` and so assumed the return type was empty.

jgraham:
https://github.com/w3c/webdriver-bidi/pull/528

whimboo:
The PR has been merged.

thiagowfx:
The `None` case cannot guarantee an updated `navigation id`, as there's no await.

Should we document that?

thiagowfx:
Chromium mapper PR, with tests: https://github.com/GoogleChromeLabs/chromium-bidi/pull/1255

`None` cannot have such tests, as the new ID is typically the same as the old ID (as we do not await for page lifecycle events).

thiagowfx:
cc @OrKoN @Lightning00Blade

jgraham:
I think per spec the reload will cancel the existing navigation and should get a new ID. That seems pretty much like what a user would expect, so I'm not sure what you want us to document, or why this issue was reopened.

thiagowfx:
This is specifically about `wait=None`. Because we return NavigationResult very quickly, before the navigation has finished, the `navigation id` will be the same as the previous instance.

My question is whether we should return `EmptyResult` for `wait=None`, or document that the return ID may not necessarily be the new ID, as there's no "await" (unlike "interactive" and "complete").

whimboo:
> This is specifically about `wait=None`. Because we return NavigationResult very quickly, before the navigation (reload) has finished, the `navigation id` will be the same as the previous instance.
>
> My question is whether we should return `EmptyResult` for `wait=None`, or document that the return ID may not necessarily be the new ID, as there's no "await" (unlike "interactive" and "complete").

[...]

PR #528, opened 2023-09-05
Set the correct result type for browsingContext.Reload

jgraham:
The prose here has two different return statements, the first that returns a `browsingContext.NavigateResult` and the second that returns `null`.

Based on the history it seems like the intent was always to return `browsingContext.NavigateResult`, and the `return null` was erronously added during cleanup when it was noticed that the return type CDDL was missing, and so was assuemd to be empty.

From a user perspective, having some way to track progress of a reload, just like any other navigation, seems important, so this is the right behaviour.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/528.html" title="Last updated on Sep 5, 2023, 11:50 AM UTC (fb2d59c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/528/835858f...fb2d59c.html" title="Last updated on Sep 5, 2023, 11:50 AM UTC (fb2d59c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2958,7 +2958,7 @@ browsing context.
    <dt>Return Type</dt>
    <dd>
       <pre class="cddl">
-         EmptyResult
+      browsingContext.NavigateResult
       </pre>
    </dd>
 </dl>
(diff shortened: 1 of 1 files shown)


whimboo:
Fixes #527.

whimboo (approved):
Looks good to me. @OrKoN @thiagowfx can you please review as well?

I can update the tests together with https://bugzilla.mozilla.org/show_bug.cgi?id=1851583 once this PR got merged.

thiagowfx (approved):

OrKoN (approved):

PR #533, opened 2023-09-08
Add serialization steps for iterators

lutien:
***
<a href="https://pr-preview.s3.amazonaws.com/lutien/webdriver-bidi/pull/533.html" title="Last updated on Sep 19, 2023, 1:51 PM UTC (004860f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/533/d81292d...lutien:004860f.html" title="Last updated on Sep 19, 2023, 1:51 PM UTC (004860f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -148,6 +148,8 @@ spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
     text: running execution context; url: running-execution-context
     text: time value; url: sec-time-values-and-time-range
     text: throw completion; url: sec-completion-record-specification-type
+    text: %AsyncIteratorPrototype%; url: sec-asynciteratorprototype
+    text: %IteratorPrototype%; url: sec-%25iteratorprototype%25-object
 spec: GEOMETRY; urlPrefix: https://drafts.fxtf.org/geometry/
   type: dfn
     text: rectangle; url: rectangle
@@ -6845,8 +6847,6 @@ Issue: Add WASM types?
 
 Issue: Should WindowProxy get attributes in a similar style to Node?
 
-Issue: Describe `script.IteratorRemoteValue` serialization.
-
 Issue: Describe `script.GeneratorRemoteValue` serialization.
 
 Issue: Describe `script.ProxyRemoteValue` serialization.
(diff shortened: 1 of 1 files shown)


whimboo (commented):
Thanks @lutien. Would you mind to also [remove the listed issues](https://w3c.github.io/webdriver-bidi/#issue-ace5d7fc)?

lutien:
> Thanks @lutien. Would you mind to also [remove the listed issues](https://w3c.github.io/webdriver-bidi/#issue-ace5d7fc)?

Sure, removed

index.bs
```diff
@@ -7055,11 +7049,26 @@ an |ownership type|, a |serialization internal map|, a |realm| and a |session|:
         production in the [=local end definition=], with the <code>handle</code>
         property set to |handle id| if it's not null, or omitted otherwise.
 
+    <dt>|value| has a \[[SyncIteratorRecord]] [=internal slot=]
```
whimboo, line 7052: This is for a sync iterators only but we have to support async ones as well. Not sure if both could be merged here or if that would require separate types? @jgraham might be better to comment on that.

lutien: As far as I understood from here: https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-createasyncfromsynciterator, async iterators should also have `SyncIteratorRecord` internal slot.

jgraham: I think this is wrong either way; the specification can create async iterators from sync iterators by creating objects with `[[SyncIteratorRecord]]` internal slots, but a [general iterator](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-iterator-interface) only has to have a `next` method. But maybe we want to check for `%IteratorPrototype%` or `%AsyncIteratorPrototype%` on the prototype chain?

jgraham (changes requested):
@mathiasbynens might also want to look at this one since I think he's been more recently involved with ECMAScript internals.

index.bs
lutien: Maksym in [this issue](https://github.com/w3c/webdriver-bidi/issues/537) also suggested checking if `next` method is callable for iterators. @sadym-chromium, could you maybe elaborate on your suggestion?

lutien:
To unblock some of our work, I've moved serialization steps for generator and proxy into [a separate PR](https://github.com/w3c/webdriver-bidi/pull/552). This PR is now only about serialization of iterators.

index.bs
lutien: Alright, I've tried to update the check for iterators in the terms of inheritance from `%IteratorPrototype%` or `%AsyncIteratorPrototype%`. Let me know what you think.

sadym-chromium: I'm quite sceptical about calling an `iterator` any object with callable property `next`.
I'd question if we have real-world scenarios, where iterators serialization is required, and if not - just consider it's an object.

sadym-chromium:
Moving the comment to the top level conversation:

> I'm quite sceptical about calling an `iterator` any object with callable property `next`.
I'd question if we have real-world scenarios, where iterators serialization is required, and if not - just consider it's an object.

PR #535, opened 2023-09-11
Add file dialog events and methods

jrandolf:
#494

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/535.html" title="Last updated on Sep 29, 2023, 12:42 PM UTC (92cdb9e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/535/22e6df6...92cdb9e.html" title="Last updated on Sep 29, 2023, 12:42 PM UTC (92cdb9e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -493,6 +493,9 @@ WebDriver BiDi extends the set of [=error codes=] from [[WEBDRIVER|WebDriver]]
 with the following additional codes:
 
 <dl>
+  <dt><dfn>no such file dialog</dfn>
+  <dd>Tried to interact with an unknown [=current file dialog|file dialog=]</code>.
+
   <dt><dfn>no such handle</dfn>
   <dd>Tried to deserialize an unknown <code>RemoteObjectReference</code>.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -8991,41 +9149,85 @@ input with a given set of file paths.
...
+      </pre>
+   </dd>
+</dl>
+
+<div algorithm>
+
+The [=remote end event trigger=] is the <dfn export>intercept WebDriver BiDi
+file dialog</dfn> steps given |context| and |element|.
+
+1. Assert: |context| is a [=top-level browsing context=].
```
OrKoN, line 9206: File dialogs can be not only from the top-level contexts, right?

OrKoN:
We are going to review internally first.

index.bs
jrandolf: Indeed. I'll resolve this once I've made the change.

index.bs
```diff
@@ -8960,24 +8963,179 @@ The [=remote end steps=] given |session|, and |command parameters| are:
...
+         method: "input.setFiles",
+         params: input.SetFilesParameters
       )
 
-      input.SetFilesParameters = {
-        context: browsingContext.BrowsingContext,
-        element: script.SharedReference,
-        files: [*text]
+      input.SetFilesParameters = input.FileDialogInfo .and {
+         files: [*text]
```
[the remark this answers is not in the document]
jrandolf: Is there a use-case for setting files when the user was never prompted for a dialog?

index.bs
```diff
@@ -8960,24 +8963,179 @@ The [=remote end steps=] given |session|, and |command parameters| are:
...
+   |command parameters| if it is present or null if it isn't.
+
+1. Let |enabled| be the value of the <code>enabled</code> field of |command
+   parameters|.
+
+1. If |enabled| is null, set |enabled| to true.
+
+1. For each |context| in |list of contexts|:
+
+   1. If |context| is not a [=top-level browsing context=], return [=error=] with
```
OrKoN, line 9016: In this method too, I think it does not need to be restricted to top-level contexts

index.bs
```diff
@@ -8960,24 +8963,179 @@ The [=remote end steps=] given |session|, and |command parameters| are:
 
 </div>
 
+#### The input.interceptFileDialogs Command ####  {#command-input-interceptFileDialogs}
+
+The <dfn export for=commands>input.interceptFileDialogs</dfn> command
```
OrKoN, line 8968: should we perhaps intercept by the fact of the event subscription instead of the dedicated method? it probably makes sense to have interception always enabled otherwise the dialog will block the execution?

index.bs
```diff
@@ -8960,24 +8963,179 @@ The [=remote end steps=] given |session|, and |command parameters| are:
...
+   1. If |context| is not a [=top-level browsing context=], return [=error=] with
+      [=error code=] [=invalid argument=].
+
+   1. Set the [=file dialog interception flag=] for |context| to |enabled|.
+
+1. Return [=success=] with data null.
+
+</div>
+
+#### The input.displayFileDialog Command ####  {#command-input-displayFileDialog}
```
OrKoN, line 9025: what's the use case for displaying the dialog to the user?

index.bs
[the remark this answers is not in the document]
OrKoN: I think the use case would be to fill out the forms without testing/verifying if they really result in a dialog (that is an existing use case supported by Puppeteer).

index.bs
jrandolf: It allows introspection of file dialog prompt if a manual workflow. I believe this makes this feature complete, but we can omit this for the initial proposal.

OrKoN (changes requested):
Let's update the spec with the following:

1) mention that the dialog is not blocking  and it is automatically dismissed.
2) add a shared reference to the element that is the target of the dialog.
3)  keep the shared reference as the parameter on setFiles

jgraham:
> mention that the dialog is not blocking and it is automatically dismissed.

I'm actually not sure we agreed to that (maybe by default if you're subscribed to the events?) But we need some kind of per-session configuration that enables either auto-dissmiss or not depending on the use case.

jrandolf:
> > mention that the dialog is not blocking and it is automatically dismissed.
> 
> I'm actually not sure we agreed to that (maybe by default if you're subscribed to the events?) But we need some kind of per-session configuration that enables either auto-dismiss or not depending on the use case.

I think @OrKoN mean that it is automatically dismissed when a session is subscribed to the event.

> But we need some kind of per-session configuration that enables either auto-dismiss or not depending on the use case.

So if one session is subscribed to the interception event, other sessions _cannot_ be not auto-dismiss in the same multiple users of the same browser cannot use the browser if any of the users is picking a file.

whimboo:
@jrandolf could you maybe explain why this PR got closed?

jrandolf:
> @jrandolf could you maybe explain why this PR got closed?

I'll be separating this PR into smaller, more independent features.

PR #534, opened 2023-09-11
Enforce minimum page size dimensions

jrandolf:
This original PR (https://github.com/w3c/webdriver-bidi/pull/522) used invalid syntax for the minimum size.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/534.html" title="Last updated on Sep 11, 2023, 10:33 AM UTC (be7c553)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/534/3535c86...be7c553.html" title="Last updated on Sep 11, 2023, 10:33 AM UTC (be7c553)">Diff</a>

PR #536, opened 2023-09-11
`browsingContext.print`: minimum page size: round down instead of up

thiagowfx:
This is a small correction of #534.

The minimum page size is (2.54 / 72) = 0.03527...

We need to round it down to 0.0352 instead of up to 0.0353 in order to make 0.03527 (1x1 point) an acceptable value.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/536.html" title="Last updated on Sep 11, 2023, 6:43 PM UTC (5ccce6e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/536/bcf5601...5ccce6e.html" title="Last updated on Sep 11, 2023, 6:43 PM UTC (5ccce6e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2827,11 +2827,11 @@ PDF document represented as a Base64-encoded string.
         ? top: (float .ge 0.0) .default 1.0,
       }
 
-      ; Minimum size is 1pt x 1pt. Conversion follows from 
+      ; Minimum size is 1pt x 1pt. Conversion follows from
       ; https://www.w3.org/TR/css3-values/#absolute-lengths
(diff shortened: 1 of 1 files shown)


whimboo (approved):
Hm, good point. It's probably more likely to hit that case as running into the still existent invalid small gap between 2.54/72 and 0.0352. Nevertheless having to use fractional numbers is always not great, especially when it comes to checking the equality.

Lightning00Blade (approved):

PR #538, opened 2023-09-12
Editorial: formalize screenshot waiting algorithm

jugglinmike:
Alternatively, we might define an algorithm for HTML to explicitly invoke during [Update the rendering](https://html.spec.whatwg.org/multipage/webappapis.html#update-the-rendering). There's certainly precedence for this, as HTML already signals three other specifications near the moment we're interested in here in BiDi.

By integrating with `requestAnimationFrame`, this approach precludes coordination with HTML, but it's not without its downsides, e.g.

- calling out to ECMAScript might be distracting for some readers
- the scheduling is observable to authors' code in that it increments the target's [animation frame callback identifier](https://html.spec.whatwg.org/multipage/imagebitmap-and-animations.html#animation-frame-callback-identifier)
- authors' rAF callbacks may execute after this algorithm resumes

...and with that, I've all but talked myself out of this patch :P

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -179,7 +179,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
     text: prompt to unload; url: browsing-the-web.html#prompt-to-unload-a-document
     text: report an error; url: webappapis.html#report-the-error
     text: remove a browsing context; url: browsers.html#bcg-remove
-    text: run the animation frame callbacks; url: imagebitmap-and-animations.html#run-the-animation-frame-callbacks
+    text: requestAnimationFrame; url: imagebitmap-and-animations.html#dom-animationframeprovider-requestanimationframe
     text: same origin domain; url: browsers.html#same-origin-domain
     text: session history; url: history.html#session-history
     text: set up a window environment settings object; url: window-object.html#set-up-a-window-environment-settings-object
@@ -2369,6 +2369,24 @@ To <dfn>render viewport to a canvas</dfn> given |viewport| and |rect|:
 
 </div>
 
+<div algorithm>
+To <dfn>await the next animation frame</dfn> given |window|:
+
+1. Let |render id| be the string representation of a
+   [[!RFC4122|UUID]] based on truly random, or pseudo-random numbers.
+
+1. Let |steps| be the following steps:
+
+  1. [=Resume=] with "<code>animation frame</code>", |render id|, and ().
+
+1. Let |callback| be [=CreateBuiltinFunction=](|steps|, 0, "", « »).
+
+1. [=Call=]([=requestAnimationFrame=], |window|, «|callback|»).
+
+1. [=Await=] with «"<code>animation frame</code>"», and |render id|.
+
+</div>
+
 The [=remote end steps=] with <var ignore>session</var> and |command parameters| are:
 
 1. Let |context id| be |command parameters|["<code>context</code>"].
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2369,6 +2369,24 @@ To <dfn>render viewport to a canvas</dfn> given |viewport| and |rect|:
 
 </div>
 
+<div algorithm>
+To <dfn>await the next animation frame</dfn> given |window|:
+
+1. Let |render id| be the string representation of a
```
jrandolf, line 2375: Let's do `opaque implementation-specific string`. That way implementations aren't bound by UUID.

jrandolf (commented):
I think this sounds reasonable. Would like some confirmation from others in Firefox before we give our LGTM.

index.bs
jgraham: I note in passing that render id isn't observable either way, so making it a UUID in the spec or not doesn't actually constrain implementations.

jgraham (changes requested):
I believe the correct way to do this would be with a specific step in https://html.spec.whatwg.org/#update-the-rendering (c.f. mark paint timing) rather than invoking DOM APIs.

index.bs
jugglinmike: As long as it doesn't matter either way, I'm happy to take @jrandolf's suggestion since that seems less distracting (though I used HTML's wording: "new unique opaque string").

jugglinmike:
> I believe the correct way to do this would be with a specific step in https://html.spec.whatwg.org/#update-the-rendering (c.f. mark paint timing) rather than invoking DOM APIs.

Great minds think alike--that's exactly what I was getting at in the description of this pull request! I've added a commit to use that approach, instead. I'm expecting folks will want to workshop the design some more, so I'm going to hold off on submitting anything to HTML. But to help explain this patch, here's how it could be integrated in its current form:

```diff
      <li><p>For each <span>fully active</span> <code>Document</code> <var>doc</var> in
      <var>docs</var> run <span>process top layer removals</span> given <var>doc</var>.</p></li>

+     <li><p>For each <span>fully active</span> <code>Document</code> <var>doc</var> in
+     <var>docs</var> run <span>resume suspended renderings</span> given <var>doc</var>'s
+     <span>browsing context</span>. <ref>WEBDRIVERBIDI</ref></p></li>
     </ol>
    </li>
```

jugglinmike:
I just realized that BiDi has a formal mechanism for tracking patches to external specs. I've pushed up a commit to include the change I described in [my previous comment](https://github.com/w3c/webdriver-bidi/pull/538#issuecomment-1736374118), and I've resolved merge conflicts from recent changes in the main branch.

What do you think, @jgraham?

index.bs
```diff
@@ -1971,6 +1971,11 @@ implicitly set when the context is created. For browsing contexts with an
 associated WebDriver [=window handle=] the [=/browsing context id=] must be the
 same as the [=window handle=].
 
+Each [=/browsing context=] has an associated <dfn>list of suspended rendering
```
jgraham, line 1974: Let's add properties directly on the remote end (or session) rather than on the browsing context. In this case a map between browsing context id and animation frame callback ids would work.

jgraham (commented):
This basically seems oK< but I really would prefer the patch in HTML if possible. We don't have great luck with HTML considering other specifications when refactoring, and having local patches makes it worse.

PR #539, opened 2023-09-12
Editorial: use method name as defined by ECMA262

jugglinmike:
~~The third argument to the Call abstract operation, if present, is expected to be a List.~~

~~https://tc39.es/ecma262/#sec-call~~

Refer to `Date.prototype.toISOString` by that name in order to more clearly communicate its intended use as an instance method.

***
<a href="https://pr-preview.s3.amazonaws.com/bocoup/webdriver-bidi/pull/539.html" title="Last updated on Sep 13, 2023, 4:40 PM UTC (afcefce)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/539/43abc3c...bocoup:afcefce.html" title="Last updated on Sep 13, 2023, 4:40 PM UTC (afcefce)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6996,7 +6996,7 @@ an |ownership type|, a |serialization internal map|, a |realm| and a |session|:
 
     <dt>|value| has a \[[DateValue]] [=internal slot=].
     <dd>
-    1. Set |serialized| to [=Call=]([=Date.toISOString=], |value|).
+    1. Set |serialized| to [=Call=]([=Date.toISOString=], «|value|»).
 
     1. Assert: |serialized| is not a [=throw completion=].
 
--- a/index.bs
+++ b/index.bs
@@ -115,7 +115,7 @@ spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
     text: CreateSetIterator; url: sec-createsetiterator
     text: Date; url: sec-date-constructor
     text: Date Time String Format; url: sec-date-time-string-format
-    text: Date.toISOString; url: sec-date.prototype.toisostring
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -6996,7 +6996,7 @@ an |ownership type|, a |serialization internal map|, a |realm| and a |session|:
 
     <dt>|value| has a \[[DateValue]] [=internal slot=].
     <dd>
-    1. Set |serialized| to [=Call=]([=Date.toISOString=], |value|).
+    1. Set |serialized| to [=Call=]([=Date.toISOString=], «|value|»).
```
whimboo, line 6999: Good catch, but doesn't it also require the `V` (`this`) argument?

```suggestion
    1. Set |serialized| to [=Call=]([=Date.toISOString=], null, «|value|»).
```

jugglinmike: Oh, yes, of course. In fact, that's the intended purpose of `|value|` as written without this patch, as `toISOString` method is an instance method for `Date` objects rather than a static method on the constructor. In other words, the spec text is correct.

That said, it looks like we were both confused by this proposal's use of `Date.toISOString` to refer to `Date.prototype.ISOString`. Rather than close this pull request outright, I've repurposed it to suggest renaming the reference to match how it is defined in ECMA262. If that seems desirable, then I'll be happy to rebase the branch in order to correct the commit message.

whimboo: Ah good point! And indeed fixing the reference to correctly point to the prototype chain makes that work.

whimboo (approved):

Issue #447, opened 2023-06-08
Consider a method to allow sending initial commands as part of session creation


jgraham:
A common part of the workflow for starting a WebDriver-BiDi session is performing initial configuration. This can involve various things e.g.:
* Configuring global browser settings (e.g. proxy configuration, prefs, window size)
* Subscribing to certain events
* Installing preload scripts
* Adding network request intercepts
* Configuring BiDi settings (there isn't much of this so far, but in the future maybe e.g. setting log level filters)

Currently some of this is done in capabilities, and some of it is done in individual commands that all need to be run once the session is configured.

Ideally we wouldn't use capabilities for anything that happens after the browser is launched. They have this complex matching behaviour that doesn't really make sense for most configuration use cases. In particular we could aim to end up in a place where by the time you're connecting to an actual browser the capabilities have all been processed and can be ignored entirely (obviously a classic-style Driver is performing the role of a browser launcher, and so would consider configuration required at browser launch time).

(We also have "capabilities" which only exist in the return value; that's a rather different concern, although I note that we've failed to update the return value with new "optional" features).

Although removing existing capabilities might not be possible,  I at least don't think we should add new post-launch configuration to capabilities.

So assuming we don't add all this configuration to capabilities directly it's currently up to the client to send all the configuration commands just after the session has started. That's OK, but it does add a bunch of roundtrip latency.

Early on there was some discussion about sending batches of commands. This seems like a clear case where that might be useful e.g. we give the `session.new` payload a field like `commands` which takes a list of commands. The `session.new` response is delayed until after all those commands have been dispatched and the response gets an extra field containing the return value of each command (which might be an error).

Alternatively we could have a more generic mechanism that doesn't special-case new session e.g. a `session.sendCommands` command whose payload would be a list of commands to run in-order and whose return value would be a list of responses to those commands.

OrKoN:
Wanted to add a comment that to minimize the roundtrip latency somewhat the client can schedule all requests at once and then wait for all responses. It works if the next command does not depend on the output of the previous command.

jimevans:
After the discussion at TPAC, having a command batch facility, with the ability to share response data between commands in the batch would unlock a huge number of scenarios that would accomplish a lot in the way of respecting high-latency connections and reducing round-trips between the local and remote ends. I would most assuredly like to see such a thing created.

jgraham:
OK, I'm going to update this issue to be about the more generic option.

I think it would be useful to document known use cases.

[...]

PR #547, opened 2023-09-14
Implement browsingContext.locateNodes

### LINKED ISSUE #150: Discussion: need for `browsingContext.findElement` command

jimevans:
Fixes #150.

This PR implements the `browsingContext.locateNodes` command, which allows the user to locate nodes in the DOM using means other than returning them via JavaScript. This also allows for future expansion of location strategies should they become available.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/547.html" title="Last updated on Nov 2, 2023, 1:37 PM UTC (ae9938d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/547/a3a09be...ae9938d.html" title="Last updated on Nov 2, 2023, 1:37 PM UTC (ae9938d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -177,6 +177,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
     text: handled; url: webappapis.html#concept-error-handled
     text: hidden; url: document-sequences.html#system-visibility-state
     text: history handling behavior; url: browsing-the-web.html#history-handling-behavior
+    text: innerText getter steps; url:dom.html#dom-innertext
     text: navigables; url: document-sequences.html#navigables
     text: navigation id; url: browsing-the-web.html#navigation-id
     text: parent; for:navigable; url: document-sequences.html#nav-parent
@@ -2844,14 +2845,14 @@ To <dfn>locate nodes using XPath</dfn> with given |context target|,
 
    1. Let |evaluateResult| be the result of calling [=evaluate=] on |context target|'s
       [=active document=], with arguments |selector|, |context node|, null,
-      [=ORDERED_NODE_SNAPSHOT_TYPE=], and null.
+      [=ORDERED_NODE_SNAPSHOT_TYPE=], and null. If this throws an [=XPathException=]
+      return [=error=] with [=error code=] [=invalid selector=], otherwise if this
+      throws any other exception return [=error=] with [=error code=] [=unknown error=].
 
    1. Let |index| be 0.
 
    1. Let |length| be the result of getting the <code>snapshotLength</code> property
-      from |evaluateResult|. If this throws an [=XPathException=] return [=error=]
-      with [=error code=] [=invalid selector=], otherwise if this throws any other
-      exception return [=error=] with [=error code=] [=unknown error=].
+      from |evaluateResult|.
 
    1. Repeat, while |index| is less than |length|:
 
@@ -2864,7 +2865,7 @@ To <dfn>locate nodes using XPath</dfn> with given |context target|,
          |returned nodes| is greater than or equal to |maximum returned nodes|,
          return |returned nodes|.
 
-      1. Increment |index| by 1. 
+      1. Set |index| to |index| + 1. 
 
 1. Return |returned nodes|
 
@@ -2881,7 +2882,8 @@ To <dfn>locate nodes using regexp</dfn> with given |context target|, |context no
 
 1. For each |element node| in |element nodes|:
 
-  1. Let |node text| be the <code>innerText</code> property of |element node|.
+  1. Let |node text| be the result of calling the [=innerText getter steps=] with
+     |element node| as the [=this=] value.
 
   1. Let |is text match| be the result of calling [=test=] with |regexp| as
      this and |node text| as the argument.
@@ -2920,8 +2922,6 @@ The [=remote end steps=] with |session| and |command parameters| are:
 
 1. Let |locator| be the value of the <code>locator</code> field
    of |command parameters|.
-
-1. Assert: |locator| is not null.
 
 1. Let |start nodes parameter| be the value of the <code>startNodes</code> field
    of |command parameters|.
@@ -2976,7 +2976,7 @@ The [=remote end steps=] with |session| and |command parameters| are:
             given |current context target|, |context nodes|, |regexp|, |maximum returned
             nodes| and |session|.
 
-1. If |maximum returned nodes| is greater than 0, truncate |result nodes| to the
+1. Assert  |maximum returned nodes| is greater than 0, truncate |result nodes| to the
    first |maximum returned nodes| values.
 
 1. Let |serialization options| be the <code>serializationOptions</code> field of
--- a/index.bs
+++ b/index.bs
@@ -2952,7 +2952,7 @@ and |session|:
           |returned nodes| is greater than or equal to |maximum returned nodes|,
           return |returned nodes|.
          
-         Otherwise, perform the following steps:
+       Otherwise, perform the following steps:
 
        1. Let |child nodes| be an empty list and, for each node |child| in the
           <a spec=dom>children</a> of |context node|:
--- a/index.bs
+++ b/index.bs
@@ -2875,7 +2875,8 @@ To <dfn>locate nodes using text</dfn> with given |context target|, |context node
 
   1. Let |node text| be the <code>innerText</code> property of |element node|.
 
-  1. Let |is text match| be the result of calling [=test=] on |regexp| with |node text|.
+  1. Let |is text match| be the result of calling [=test=] with |regexp| as
+     this and |node text| as the argument.
 
   1. If |is text match| is true, append |element node| to |returned nodes|.
 
--- a/index.bs
+++ b/index.bs
@@ -2879,25 +2879,27 @@ To <dfn>locate nodes using text</dfn> with given |context target|, |context node
   1. In the following list of conditions and associated steps, run the first set 
      of steps for which the associated condition is true, if any:
 
-     <dl>|match content| is "partial" and |match case| is true
-     <dd>
-       1. If |node text| contains the substring |selector|, including case,
-          append |element node| to |returned nodes|.
+     <dl>
 
-     <dl>|match content| is "partial" and |match case| is false
-     <dd>
-       1. If |node text| contains the substring |selector|, ignoring case,
-          append |element node| to |returned nodes|.
+       <dt>|match content| is "partial" and |match case| is true
+       <dd>
+         1. If |node text| contains the substring |selector|, including case,
+            append |element node| to |returned nodes|.
 
-     <dl>|match content| is "full" and |match case| is true
-     <dd>
-       1. If |node text| is equal to |selector|, including case, append
-          |element node| to |returned nodes|.
+       <dt>|match content| is "partial" and |match case| is false
+       <dd>
+         1. If |node text| contains the substring |selector|, ignoring case,
+            append |element node| to |returned nodes|.
 
-     <dl>|match content| is "full" and |match case| is false
-     <dd>
-       1. If |node text| is equal to |selector|, ignoring case, append
-          |element node| to |returned nodes|.
+       <dt>|match content| is "full" and |match case| is true
+       <dd>
+         1. If |node text| is equal to |selector|, including case, append
+            |element node| to |returned nodes|.
+
+       <dt>|match content| is "full" and |match case| is false
+       <dd>
+         1. If |node text| is equal to |selector|, ignoring case, append
+            |element node| to |returned nodes|.
 
 1. Return |returned nodes|.
 
--- a/index.bs
+++ b/index.bs
@@ -3011,13 +3011,15 @@ The [=remote end steps=] with |session| and |command parameters| are:
 
 1. Assert [=list/size=] of |result nodes| is less than or equal to |maximum returned nodes|.
 
-1. Let |serialization options| be the <code>serializationOptions</code> field of
-   |command parameters| if present, otherwise a [=/map=] matching the 
+1. If |command parameters| [=map/contains=] "<code>serializationOptions</code>",
+   let |serialization options| be |command parameters|["<code>serializationOptions</code>"|.
+   Otherwise let |serialization options| be a [=/map=] matching the 
    <code>script.SerializationOptions</code> production with the fields
    set to their default values.
 
-1. Let |result ownership| be the value of the <code>ownership</code> field of
-   |command parameters|, if it exits, otherwise "none".
+1. If |command parameters| [=map/contains=] "<code>ownership</code>",
+   let |result ownership| be |command parameters|["<code>ownership</code>"|.
+   Otherwise let |result ownership| be "none".
 
 1. Let |serialized nodes| be the result of [=serialize as a remote value=] with
    |result nodes|, |serialization options|, |result ownership|, a new map as
--- a/index.bs
+++ b/index.bs
@@ -2844,7 +2844,7 @@ list of all nodes matching the specified locator.
 To <dfn>locate nodes using CSS</dfn> with given |context target|, |context nodes|,
 |selector|, |maximum returned node count|, and <var ignore>session</var>:
 
-1. Let |returned nodes| be an empty list.
+1. Let |returned nodes| be an empty [=/list=].
 
 1. Let |parse result| be the result of [=parse a selector=] given |selector|.
 
@@ -2858,7 +2858,7 @@ To <dfn>locate nodes using CSS</dfn> with given |context target|, |context nodes
 
   1. For each |element| in |elements|:
 
-    1. [=list/append=] |element| to |returned nodes|.
+    1. [=list/Append=] |element| to |returned nodes|.
 
     1. If |maximum returned node count| is greater than 0 and [=list/size=] of
        |returned nodes| is greater than or equal to |maximum returned node count|,
@@ -2878,7 +2878,7 @@ is phrased as if making calls to the XPath DOM APIs. However this is to be under
 as equivalent to spec-internal calls directly accessing the underlying algorithms,
 without going via the ECMAScript runtime.
 
-1. Let |returned nodes| be an empty list.
+1. Let |returned nodes| be an empty [=/list=].
 
 1. For each |context node| of |context nodes|:
 
@@ -2899,7 +2899,7 @@ without going via the ECMAScript runtime.
       1. Let |node| be the result of calling [=snapshotItem=] with |evaluate result|
          as this and |index| as the argument.
 
-      1. [=list/append=] |node| to |returned nodes|.
+      1. [=list/Append=] |node| to |returned nodes|.
       
       1. If |maximum returned node count| is greater than 0 and [=list/size=] of
          |returned nodes| is greater than or equal to |maximum returned node count|,
@@ -2916,7 +2916,7 @@ To <dfn>locate nodes using text</dfn> with given |context target|, |context node
 |selector|, |max depth|, |match type|, |ignore case|, |maximum returned node count|,
 and |session|:
 
-1. Let |returned nodes| be an empty list.
+1. Let |returned nodes| be an empty [=/list=].
 
 1. For each |context node| in |context nodes|:
 
@@ -2930,25 +2930,25 @@ and |session|:
 
        <dt>|match type| is "full" and |ignore case| is false
        <dd>
-         1. If |node text| [=strings/string-is] |selector|, [=list/append=]
+         1. If |node text| [=is=] |selector|, [=list/append=]
             |context node| to |returned nodes|.
 
        <dt>|match type| is "full" and |ignore case| is true
        <dd>
-         1. If |node text| [=strings/string-is] |selector|, ignoring case,
+         1. If |node text| [=is=] |selector|, ignoring case,
             [=list/append=] |context node| to |returned nodes|.
 
        <dt>Otherwise
        <dd>
-         1. If |ignore case| is false and |selector| is a [=strings/code-point-substring=]
+         1. If |ignore case| is false and |selector| is a [=code point substring=]
             of |node text| set |contains text| to true.
 
-         1. If |ignore case| is false and |selector| is a [=strings/code-point-substring=]
+         1. If |ignore case| is false and |selector| is a [=code point substring=]
             of |node text|, ignoring case, set |contains text| to true.
 
          1. If |contains text| is true, perform the following steps:
 
-           1. Let |child nodes| be an empty list and, for each node |child| in the
+           1. Let |child nodes| be an empty [=/list=] and, for each node |child| in the
               <a spec=dom>children</a> of |context node|:
 
              1. If |child| implements {{Element}}, [=list/append=] |child| to
@@ -3010,22 +3010,20 @@ The [=remote end steps=] with |session| and |command parameters| are:
    |start nodes parameter| be |command parameters|["<code>startNodes</code>"].
    Otherwise let |start nodes parameter| be null.
 
-1. Let |context nodes| be an empty list.
+1. Let |context nodes| be an empty [=/list=].
 
-1. If |start nodes parameter| is null, [=list/append=] [=document element=] to
-   |context nodes|. Otherwise, for each |serialized start node| in
-   |start nodes parameter|:
+1. If |start nodes parameter| is null, [=list/append=] the [=document element=]
+   to of |context|'s [=active document=] |context nodes|. Otherwise, for each
+   |serialized start node| in |start nodes parameter|:
 
    1. Let |start node| be the result of [=trying=] to [=deserialize shared reference=] given
       |serialized start node|, |realm| and |session|.
 
-   1. [=list/append=] |start node| to |context nodes|.
+   1. [=list/Append=] |start node| to |context nodes|.
 
-1. If the <code>length</code> property of |context nodes| is 0, append the node 
-   corresponding to the <code>documentElement</code> property of |context|'s
-   [=active document=] to |context nodes|.
+1. Assert [=list/size=] of |context nodes is greater than 0.
 
-1. Let |result nodes| be an empty list.
+1. Let |result nodes| be an empty [=/list=].
 
 1. Let |type| be the value of the <code>type</code> property of |locator|.
 
--- a/index.bs
+++ b/index.bs
@@ -2120,7 +2120,7 @@ To <dfn>await a navigation</dfn> given |context|, |request|, |wait condition|, a
 <pre class="cddl remote-cddl local-cddl">
 browsingContext.Locator = (
    browsingContext.CssLocator /
-   browsingContext.TextLocator /
+   browsingContext.InnerTextLocator /
    browsingContext.XPathLocator
 )
 
@@ -2129,8 +2129,8 @@ browsingContext.CssLocator = {
    value: text
 }
 
-browsingContext.TextLocator {
-   type: "text",
+browsingContext.InnerTextLocator {
+   type: "innerText",
    value: text,
    ? ignoreCase: bool
    ? matchType: "full" / "partial",
@@ -2914,8 +2914,8 @@ without going via the ECMAScript runtime.
 
 </div>
 
-<div algorithm="locate nodes using text">
-To <dfn>locate nodes using text</dfn> with given |context target|, |context nodes|,
+<div algorithm="locate nodes using inner text">
+To <dfn>locate nodes using inner text</dfn> with given |context target|, |context nodes|,
 |selector|, |max depth|, |match type|, |ignore case|, |maximum returned node count|,
 and |session|:
 
@@ -2932,7 +2932,7 @@ and |session|:
 
   1. If |ignore case| is false, let |node text| be |node inner text|. Otherwise,
      let |node text| be the result of executing [=toUppercase=] with |node inner text|
-     according to the Unicode Default Case Conversion algorithm. 
+     according to the Unicode Default Case Conversion algorithm.
 
   1. If |match type| is "full" and |node text| [=is=] |search text| [=list/append=]
      |context node| to |returned nodes|. Otherwise, perform the following steps:
@@ -3053,7 +3053,7 @@ The [=remote end steps=] with |session| and |command parameters| are:
          1. If |locator| [=map/contains=] <code>matchType</code>, let |match type| be 
             |locator|["<code>matchType</code>"]. Otherwise, let |match type| be "full".
 
-         1. Let |result nodes| be a result of [=trying=] to [=locate nodes using text=]
+         1. Let |result nodes| be a result of [=trying=] to [=locate nodes using inner text=]
             given |current context target|, |context nodes|, |selector|, |max depth|,
             |match type|, |ignore case|, |maximum returned node count| and |session|.
 
--- a/index.bs
+++ b/index.bs
@@ -2790,6 +2790,8 @@ list of all nodes matching the specified locators.
          locator: browsingContext.Locator,
          startNodes: [ * script.SharedReference ],
          ? maxNodes: js-uint,
+         ? ownership: script.ResultOwnership,
+         ? serializationOptions: script.SerializationOptions
       }
       </pre>
    </dd>
@@ -2798,7 +2800,7 @@ list of all nodes matching the specified locators.
     <pre class="cddl local-cddl">
         browsingContext.LocateNodesResult = {
             context: browsingContext.BrowsingContext,
-            nodeSharedIds: [ * script.SharedId ]
+            nodes: [ * script.RemoteNodeValue ]
         }
     </pre>
    </dd>
@@ -2852,9 +2854,6 @@ To <dfn>locate nodes using XPath</dfn> with given |context target|,
       1. Let |node| be the result of calling [=snapshotItem=] with |evaluateResult|
          as this and |index| as the argument.
 
-      1. If |node| is not an element return [=error=] with [=error code=]
-         [=invalid selector=].
-      
       1. Append |node| to |returned nodes|.
       
       1. Increment |index| by 1. 
@@ -2964,26 +2963,21 @@ The [=remote end steps=] with |session| and |command parameters| are:
 1. If |maximum returned nodes| is greater than 0, truncate |result nodes| to the
    first |maximum returned nodes| values.
 
-1. Let |serialization options| be a [=/map=] matching the 
+1. Let |serialization options| be the <code>serializationOptions</code> field of
+   |command parameters| if present, otherwise a [=/map=] matching the 
    <code>script.SerializationOptions</code> production with the fields
    set to their default values.
 
-1. Let |result ownership| be "none".
+1. Let |result ownership| be the value of the <code>ownership</code> field of
+   |command parameters|, if it exits, otherwise "none".
 
-1. Let |serialized result| be the result of [=serialize as a remote value=] with
+1. Let |serialized nodes| be the result of [=serialize as a remote value=] with
    |result nodes|, |serialization options|, |result ownership|, a new map as
    <var ignore>serialization internal map</var>, |realm| and |session|.
 
-1. Let |serialized shared ids| be a new array.
-
-1. For each |serialized result node| in |serialized result|:
-
-   1. Append the value of the <code>sharedId</code> field of |serialized result node|
-      to |serialized shared ids|.
-
 1. Let |result| be a [=/map=] matching the <code>browsingContext.LocateNodesResult</code>
    production, with the <code>context</code> field set to the |context id|
-   and the <code>nodeSharedIds</code> field set |serialized shared ids|.
+   and the <code>nodes</code> field set |serialized nodes|.
 
 1. Return [=success=] with data |result|.
 
--- a/index.bs
+++ b/index.bs
@@ -2823,7 +2823,9 @@ To <dfn>locate nodes using CSS</dfn> with given |context target|, |context nodes
      |parse result| and |context target|’s [=active document=] [=root=] using
      [=scoping root=] |context node|.
 
-  1. Append |elements| to |returned nodes|
+  1. For each |element| in |elements|:
+
+    1. Append |element| to |returned nodes|
 
 1. Return |returned nodes|.
 
--- a/index.bs
+++ b/index.bs
@@ -2981,7 +2981,7 @@ The [=remote end steps=] with |session| and |command parameters| are:
          1. Let |match content| be the value of the <code>matchContent</code> field of
             |locator| if present, otherwise "full".
 
-         1. Let |match case| be the value of the <code>match case</code> field of
+         1. Let |match case| be the value of the <code>matchCase</code> field of
             |locator| if present, otherwise true.
 
          1. Let |context nodes| be a result of [=trying=] to [=locate nodes using text=]
--- a/index.bs
+++ b/index.bs
@@ -2135,7 +2135,7 @@ browsingContext.CssLocator = {
    value: text
 }
 
-browsingContext.InnerTextLocator {
+browsingContext.InnerTextLocator = {
    type: "innerText",
    value: text,
    ? ignoreCase: bool
@@ -2143,7 +2143,7 @@ browsingContext.InnerTextLocator {
    ? maxDepth: js-uint,
 }
 
-browsingContext.XPathLocator {
+browsingContext.XPathLocator = {
    type: "xpath",
    value: text
 }
--- a/index.bs
+++ b/index.bs
@@ -2847,7 +2847,7 @@ To <dfn>locate nodes using XPath</dfn> with given |context target|,
    1. Let |length| be the result of getting the <code>snapshotLength</code> property
       from |evaluateResult|. If this throws an [=XPathException=] return [=error=]
       with [=error code=] [=invalid selector=], otherwise if this throws any other
-      exception return return [=error=] with [=error code=] [=unknown error=].
+      exception return [=error=] with [=error code=] [=unknown error=].
 
    1. Repeat, while |index| is less than |length|:
 
--- a/index.bs
+++ b/index.bs
@@ -223,6 +223,7 @@ spec: XPATH; urlPrefix: https://www.w3.org/TR/DOM-Level-3-XPath/xpath.html
   type: dfn
     text: evaluate; url: #XPathEvaluator-evaluate
     text: ORDERED_NODE_SNAPSHOT_TYPE; url: #XPathResult-ORDERED-NODE-SNAPSHOT-TYPE
+    text: snapshotItem; url: #XPathResult-snapshotItem
     text: XPathException; url: #XPathException
 </pre>
 
@@ -2849,7 +2850,7 @@ To <dfn>locate nodes using XPath</dfn> with given |context target|,
 
    1. Repeat, while |index| is less than |length|:
 
-      1. Let |node| be the result of calling snapshotItem with evaluateResult
+      1. Let |node| be the result of calling [=snapshotItem=] with |evaluateResult|
          as this and |index| as the argument.
 
       1. If |node| is not an element return [=error=] with [=error code=]
--- a/index.bs
+++ b/index.bs
@@ -2960,35 +2960,33 @@ and |session|:
      let |node text| be the result of [=toUppercase=] with |node inner text|
      according to the [=Unicode Default Case Conversion algorithm=].
 
-  1. If |match type| is "full" and |node text| [=is=] |search text|, [=list/append=]
-     |context node| to |returned nodes|.
+  1. If |search text| is a [=code point substring=] of |node text|, perform the
+     following steps:
      
-  1. Otherwise, perform the following steps:
+    1. Let |child nodes| be an empty [=/list=] and, for each node |child| in the
+       <a spec=dom>children</a> of |context node|:
 
-    1. If |search text| is a [=code point substring=] of |node text|, perform the
-       following steps:
+      1. If |child| implements {{Element}}, [=list/append=] |child| to |child nodes|.
 
-      1. Let |child nodes| be an empty [=/list=] and, for each node |child| in the
-         <a spec=dom>children</a> of |context node|:
+    1. If [=list/size=] of |child nodes| is equal to 0 or |max depth| is equal to 0,
+       perform the following steps:
 
-        1. If |child| implements {{Element}}, [=list/append=] |child| to
-           |child nodes|.
+      1. If |match type| is <code>"full"</code> and |node text| [=is=] |search text|,
+         [=list/append=] |context node| to |returned nodes|.
 
-      1. If [=list/size=] of |child nodes| is equal to 0 or |max depth|
-         is equal to 0, [=list/append=] |context node| to |returned nodes|.
-         
-      1. Otherwise, perform the following steps:
+      1. Otherwise, [=list/append=] |context node| to |returned nodes|.
 
-        1. Set |max depth| to |max depth| - 1. 
+    1. Otherwise, perform the following steps:
 
-        1. Let |child node matches| be the result of [=locate nodes using inner text=]
-           with |context target|, |child nodes|, |selector|, |max depth|,
-           |match type|, |ignore case|, |maximum returned node count|, and |session|.
+      1. Set |max depth| to |max depth| - 1. 
 
-        1. If [=list/size=] of |child node matches| is equal to 0 and
-           |match type| is "partial", [=list/append=] |context node| to
-           |returned nodes|. Otherwise, [=list/extend=] |returned nodes|
-           with |child node matches|.
+      1. Let |child node matches| be the result of [=locate nodes using inner text=]
+         with |context target|, |child nodes|, |selector|, |max depth|,
+         |match type|, |ignore case|, |maximum returned node count|, and |session|.
+
+      1. If [=list/size=] of |child nodes matches| is equal to 0 and |match type| is
+         <code>"partial"</code>, append |context node| to |returned nodes|. Otherwise,
+         [=list/extend=] |returned nodes| with |child node matches|.
 
 1. If |maximum returned node count| is not null, [=list/remove=] all entries
    in |returned nodes| with an index greater than or equal to |maximum returned
--- a/index.bs
+++ b/index.bs
@@ -2822,7 +2822,7 @@ list of all nodes matching the specified locator.
       browsingContext.LocateNodesParameters = {
          context: browsingContext.BrowsingContext,
          locator: browsingContext.Locator,
-         ? maxNodes: (js-uint .ge 1),
+         ? maxReturnedNodeCount: (js-uint .ge 1),
          ? ownership: script.ResultOwnership,
          ? serializationOptions: script.SerializationOptions
          ? startNodes: [ + script.SharedReference ],
@@ -2842,7 +2842,7 @@ list of all nodes matching the specified locator.
 
 <div algorithm="locate nodes using CSS">
 To <dfn>locate nodes using CSS</dfn> with given |context target|, |context nodes|,
-|selector|, |maximum returned nodes|, and <var ignore>session</var>:
+|selector|, |maximum returned node count|, and <var ignore>session</var>:
 
 1. Let |returned nodes| be an empty list.
 
@@ -2860,8 +2860,8 @@ To <dfn>locate nodes using CSS</dfn> with given |context target|, |context nodes
 
     1. [=list/append=] |element| to |returned nodes|.
 
-    1. If |maximum returned nodes| is greater than 0 and [=list/size=] of
-       |returned nodes| is greater than or equal to |maximum returned nodes|,
+    1. If |maximum returned node count| is greater than 0 and [=list/size=] of
+       |returned nodes| is greater than or equal to |maximum returned node count|,
        return |returned nodes|.
 
 1. Return |returned nodes|.
@@ -2871,7 +2871,7 @@ To <dfn>locate nodes using CSS</dfn> with given |context target|, |context nodes
 <div algorithm="locate nodes using XPath">
 
 To <dfn>locate nodes using XPath</dfn> with given |context target|,
-|context nodes|, |selector|, |maximum returned nodes|, and <var ignore>session</var>:
+|context nodes|, |selector|, |maximum returned node count|, and <var ignore>session</var>:
 
 Note: Owing to the unmaintained state of the XPath specification, this algorithm
 is phrased as if making calls to the XPath DOM APIs. However this is to be understood
@@ -2901,8 +2901,8 @@ without going via the ECMAScript runtime.
 
       1. [=list/append=] |node| to |returned nodes|.
       
-      1. If |maximum returned nodes| is greater than 0 and [=list/size=] of
-         |returned nodes| is greater than or equal to |maximum returned nodes|,
+      1. If |maximum returned node count| is greater than 0 and [=list/size=] of
+         |returned nodes| is greater than or equal to |maximum returned node count|,
          return |returned nodes|.
 
       1. Set |index| to |index| + 1. 
@@ -2913,7 +2913,7 @@ without going via the ECMAScript runtime.
 
 <div algorithm="locate nodes using text">
 To <dfn>locate nodes using text</dfn> with given |context target|, |context nodes|,
-|selector|, |max depth|, |match type|, |ignore case|, |maximum returned nodes|,
+|selector|, |max depth|, |match type|, |ignore case|, |maximum returned node count|,
 and |session|:
 
 1. Let |returned nodes| be an empty list.
@@ -2957,8 +2957,8 @@ and |session|:
         
        1. [=list/append=] |context node| to |returned nodes|.
 
-       1. If |maximum returned nodes| is not null and [=list/size=] of
-          |returned nodes| is greater than or equal to |maximum returned nodes|,
+       1. If |maximum returned node count| is not null and [=list/size=] of
+          |returned nodes| is greater than or equal to |maximum returned node count|,
           return |returned nodes|.
          
        Otherwise, perform the following steps:
@@ -2972,8 +2972,8 @@ and |session|:
        
          1. [=list/append=] |context node| to |returned nodes|.
 
-         1. If |maximum returned nodes| is not null and [=list/size=] of
-            |returned nodes| is greater than or equal to |maximum returned nodes|,
+         1. If |maximum returned node count| is not null and [=list/size=] of
+            |returned nodes| is greater than or equal to |maximum returned node count|,
             return |returned nodes|.
          
          Otherwise, perform the following steps:
@@ -2982,7 +2982,7 @@ and |session|:
 
          1. [=list/extend=] |returned nodes| with the result of [=locate nodes using text=]
             with |context target|, |child nodes|, |selector|, |max depth|,
-            |match type|, |ignore case|, |maximum returned nodes|, and |session|.
+            |match type|, |ignore case|, |maximum returned node count|, and |session|.
 
 1. Return |returned nodes|.
 
@@ -2999,11 +2999,11 @@ The [=remote end steps=] with |session| and |command parameters| are:
 
 1. Assert: |context| is not null.
 
-1. If |command parameters| [=map/contains=] "<code>maxNodes</code>", let
-   |maximum returned nodes| be |command parameters|["<code>maxNodes</code>"].
-   Otherwise let |maximum returned nodes| be null.
+1. If |command parameters| [=map/contains=] "<code>maxReturnedNodeCount</code>", let
+   |maximum returned node count| be |command parameters|["<code>maxReturnedNodeCount</code>"].
+   Otherwise let |maximum returned node count| be null.
 
-1. If |maximum returned nodes| is not null, and |maximum returned nodes|
+1. If |maximum returned node count| is not null, and |maximum returned node count|
    is less than 1, return return [=error=] with [=error code=] [=unsupported operation=]. 
 
 1. Let |current context target| be the a [=/map=] matching of the
@@ -3076,9 +3076,9 @@ The [=remote end steps=] with |session| and |command parameters| are:
 
          1. Let |result nodes| be a result of [=trying=] to [=locate nodes using text=]
             given |current context target|, |context nodes|, |selector|, |max depth|,
-            |match type|, |ignore case|, |maximum returned nodes| and |session|.
+            |match type|, |ignore case|, |maximum returned node count| and |session|.
 
-1. Assert [=list/size=] of |result nodes| is less than or equal to |maximum returned nodes|.
+1. Assert [=list/size=] of |result nodes| is less than or equal to |maximum returned node count|.
 
 1. If |command parameters| [=map/contains=] "<code>serializationOptions</code>",
    let |serialization options| be |command parameters|["<code>serializationOptions</code>"].
--- a/index.bs
+++ b/index.bs
@@ -148,6 +148,7 @@ spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
     text: running execution context; url: running-execution-context
     text: time value; url: sec-time-values-and-time-range
     text: throw completion; url: sec-completion-record-specification-type
+    text: test; url: #sec-regexp.prototype.test
 spec: GEOMETRY; urlPrefix: https://drafts.fxtf.org/geometry/
   type: dfn
     text: rectangle; url: rectangle
@@ -2123,9 +2124,7 @@ browsingContext.CssLocator = {
 
 browsingContext.TextLocator {
    type: "text",
-   value: text,
-   ? matchCase: boolean,
-   ? matchContent: "full" / "partial"
+   value: script.RegExpLocalValue
 }
 
 browsingContext.XPathLocator {
@@ -2868,7 +2867,7 @@ To <dfn>locate nodes using XPath</dfn> with given |context target|,
 
 <div algorithm="locate nodes using text">
 To <dfn>locate nodes using text</dfn> with given |context target|, |context nodes|,
-|selector|, |match content|, |match case| and <var ignore>session</var>:
+|regexp| and <var ignore>session</var>:
 
 1. Let |returned nodes| be an empty array.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2089,6 +2091,48 @@ To <dfn>await a navigation</dfn> given |context|, |request|, |wait condition|, a
...
+   browsingContext.XPathLocator
+)
+
+browsingContext.CssLocator = {
+   type: "css",
+   value: text
+}
+
+browsingContext.FrameLocator = {
+   type: "frame"
```
shs96c, line 2113: Is the `context nodes` parameter missing here?

index.bs
```diff
@@ -2089,6 +2091,48 @@ To <dfn>await a navigation</dfn> given |context|, |request|, |wait condition|, a
...
+   ? match: "full" / "partial"
+}
+
+browsingContext.XPathLocator {
+   type: "xpath",
+   value: text
+}
+</pre>
+
+The <code>browsingContext.Locator</code> type is a provides details on the strategy
```
shs96c, line 2133: `s/ is a //`

index.bs
```diff
@@ -2725,6 +2769,284 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+      browsingContext.LocateNodes = (
+        method: "browsingContext.locateNodes",
+        params: browsingContext.LocateNodesParameters
+      )
+
+      browsingContext.LocateNodesParameters = {
+         context: browsingContext.BrowsingContext,
+         locators: [ + browsingContext.Locator ],
+         ? startNode: script.SharedReference,
+         ? waitTimeout: js-uint
```
shs96c, line 2791: Perhaps an optional `maxCount` to limit the number of returned nodes? Would be useful for selenium's `findElement` method, where returning a large number of nodes is wasteful.

index.bs
```diff
@@ -2725,6 +2769,284 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+
+<div algorithm="locate nodes using XPath">
+To <dfn>locate nodes using XPath</dfn> with given |context|, |context nodes|
+|selector| and <var ignore>session</var>:
+
+1. Let |returned nodes| be an empty array.
+
+1. For each |context node| of |context nodes|:
+
+   1. Let |evaluateResult| be the result of calling evaluate on |context|, with arguments |selector|,
```
shs96c, line 2832: Nit: do we need to link `evaluate` to something?

index.bs
```diff
@@ -2725,6 +2769,284 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+
+   1. Let |length| be the result of getting the property "snapshotLength" from
+      |evaluateResult|. If this throws an XPathException return [=error=] with
+      [=error code=] [=invalid selector=], otherwise if this throws any other
+      exception return return [=error=] with [=error code=] [=unknown error=].
+
+   1. Repeat, while |index| is less than |length|:
+
+      1. Let |node| be the result of calling snapshotItem with evaluateResult
+         as this and index as the argument.
```
shs96c, line 2845: `|index|`?

index.bs
```diff
@@ -2725,6 +2769,284 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+   [=deserialize local value=] given |start node parameter|, and |session|.
+
+1. Let |wait timeout| be the value of the <code>waitTimeout</code> field of |command
+   parameters| if present, or 0 otherwise.
+
+1. Let |expiration time| be the current time plus |wait timeout| milliseconds.
+
+1. Let |result nodes| be an empty array.
+
+1. Perform the following steps until the current time is greater than |expiration time|
```
shs96c, line 2967: There's a potential race here: the `expiration time` may have already been reached at this step. We should clarify that these steps must be executed at least once.

index.bs
```diff
@@ -2725,6 +2769,284 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+   or |result nodes| is not empty.
```
shs96c, line 2968: Do we want to specify a retry period? Or leave that TBD?

index.bs
```diff
@@ -2725,6 +2769,284 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+
+  1. Let |context nodes| be an empty array.
+
+  1. Append |start node| to |context nodes|.
+
+  1. For each |locator| in |locators|:
+
+     1. Let |type| be the value of the type property of |locator|.
+   
+     1. In the following list of conditions and associated steps, run the first set
```
shs96c, line 2978: Do we want to use the same idea of the element location strategy table from Classic so that implementors can add additional location strategies? Or should we have a "find any platform-specific implementation that matches the input properties, if any" at the end?

index.bs
```diff
@@ -2725,6 +2769,284 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+           <dt>|type| is the string "<code>text</code>"
+           <dd>
+              1. Let |selector| be the value of the value property of |locator|.
+
+              1. Let |match| be the value of the match property of |locator| if
+                 present, otherwise "full".
+
+              1. Let |context nodes| be a result of [=trying=] to [=locate nodes using text=]
+                 given |context|, |context nodes|, |selector|, |match| and |session|
+
```
shs96c, line 3026: What happens if there is no match?

shs96c (commented):
Very excited to see this in the spec :)

index.bs
OrKoN, line 2791: do we want timeouts to be part of the protocol definitions? this kind of timeouts looks like it should be a client side timeout since the locator is not abortable.

index.bs
```diff
@@ -2725,6 +2769,284 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+     return [=error=] with [=error code=] [=invalid selector=]. 
+
+  1. Append |elements| to |returned nodes|
+
+1. Return |returned nodes|.
+
+</div>
+
+<div algorithm="locate nodes using XPath">
+To <dfn>locate nodes using XPath</dfn> with given |context|, |context nodes|
```
OrKoN, line 2825: should accept a document or realm instead of the context? 1) to avoid execution on a wrong document if it changes for the context 2) to allow locators to run a sandbox, so that the script locator can use the objects existing in the sandbox.

index.bs
```diff
@@ -2725,6 +2769,284 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+      
+      1. Append |node| to |returned nodes|.
+      
+      1. Increment |index| by 1. 
+
+1. Return |returned nodes|
+
+</div>
+
+<div algorithm="locate nodes using text">
```
OrKoN, line 2858: does this locator exist in WebDriver Classic? Is it the one called Link text?

index.bs
```diff
@@ -2725,6 +2769,284 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+      <pre class="cddl remote-cddl">
+      browsingContext.LocateNodes = (
+        method: "browsingContext.locateNodes",
+        params: browsingContext.LocateNodesParameters
+      )
+
+      browsingContext.LocateNodesParameters = {
```
OrKoN, line 2787: do we need to add the ownership parameters to retain results?

index.bs
```diff
@@ -2725,6 +2769,284 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
 
 1. Return [=success=] with data null.
 
+</div>
+#### The browsingContext.locateNodes Command ####  {#command-browsingContext-locateNodes}
```
OrKoN, line 2773: feel-free-to-ignore suggestion: it might make it easier to review if we start with a single locator and then extend the list in PR

index.bs
jimevans: It does not, but is included to be able to construct "link text" and "partial link text." The intended payload for that would be something like:

```json
{
  "context": "some context",
  "locators": [
    {
      "type": "css",
      "value": "a"
    },
    {
      "type": "text"
      "value": "desired link text",
      "match": "full"
    }
  ]
}
```
Where `match` would be `full` for find by link text, and `partial` for find by partial link text.

index.bs
jimevans: It is not. This locator is only chainable; it can't be used on its own. The intent is to combine it with whatever locator returns the frame/iframe element. All this does is switch the "active" context. As a side effect, it also enables the classic "switch to frame" command.

index.bs
jimevans: Probably, but the decision was to first enable the locators defined by Classic, so that's what I was attempting to do. If it makes it easier, I can remove the others, and we can go from there, I suppose, but I have no idea if that really makes sense. The text would simply have to be replaced once one locator was agreed on.

index.bs
jimevans: We've never done that for locating elements before, but we certainly can if we want to. I'm entirely agnostic on that.

jimevans: @OrKoN I'm not entirely sure what you mean by "not abortable;" the algorithm for element finding does abort upon timing out. Having a timeout for this command solves a number of issues, including being able to wait for an element's presence without a duplicating `waitForElement` command.

I mean, there's another design entirely. where we treat finding elements like network interception, firing off a find command, and having an event tell us when the element is found, if we _really_ want to be edgy and async, but that seems really challenging to get right, and not ergonomic to what developers (especially non-JavaScript developers) expect.

index.bs
jimevans: We do. That's the kind of thing I was explicitly talking with @jgraham about help me with as I was signing off the Zoom meeting today.

index.bs
jimevans: I'm agnostic on that. The retry interval can be implementation-based as far as I'm concerned.

index.bs
jimevans: I don't know. Do we?

index.bs
jimevans: Probably true. Please feel free to provide suggested spec prose for how to do that when all the user knows about is the contextId, since this is the `browserContext` module.

index.bs
jimevans: Good catch. Fixed in subsequent commit.

index.bs
jimevans: Absolutely correct. Fixed in subsequent commit.

index.bs
jimevans: Fixed.

index.bs
jimevans: Should be that you get an empty list returned.

index.bs
jimevans: I don't know. I haven't thought that part through yet.

index.bs
OrKoN: I meant that if you, for example, start running a JS locator (applicable to many other locators but maybe others are less likely to timeout), there is no way to abort it even if timeout happens. So what would implementations do is to just throw an ignoring the locator results. The same can be achieved on the client but stopping to await the response message from the protocol. For the "waitForElement" use case it would be indeed useful, but we also need to define a bunch of other parameters such as how many elements to wait for including, potentially, waiting for elements to disappear (i.e., having zero elements matching the selector).

index.bs
OrKoN: I think we would need that if we want the user to make use of the returned handles. Currently, it looks like we just return a snapshot of the elements (I see resultOwnerhip set to none). We also probably want to think about realms and sandboxes (maybe not immediately in this proposal).

index.bs
shs96c: I think keeping the PR "as is" is fine, and will help us avoid churn

index.bs
shs96c: Let's get this PR landed. We can tweak later.

shs96c: I think we probably need to re-introduce timeouts from the Classic spec, and then allow those to be used. A per-request timeout, per-command is going to get pretty ugly, pretty quickly. I'd opt to leave `waitTimeout` out, and then implement it later using the global timeouts.

index.bs
shs96c: Again, we can land and iterate. I think for the locators we have, this is just fine.

index.bs
shs96c: I think it may be better to indicate to the user that an `invalid selector` was used. I can imagine trying to debug a typo in a local end implementation....

index.bs
```diff
@@ -2089,6 +2091,48 @@ To <dfn>await a navigation</dfn> given |context|, |request|, |wait condition|, a
 
 </div>
 
+#### The browsingContext.Locator Type #### {#type-browsingContext-Locator}
+
+[=remote end definition=] and [=local end definition=]
+
+<pre class="cddl remote-cddl local-cddl">
+browsingContext.Locator = (
```
AutomatedTester, line 2099: Should we have `JavaScriptLocator` in here?

index.bs
jimevans: I've changed the location algorithms to use `script.ContextTarget` instead of `browsingContext.BrowsingContext`. I think this might address this concern, along with the optional `sandbox` argument.

index.bs
jimevans: @OrKoN I've added an optional `ownership` property to the command parameters. Does that alleviate this concern?

index.bs
jimevans: @shs96c What about the perfectly legitimate case where you're attempting to find an element by its text, and no element with that text actually exists in the document? Since we're doing the equivalent of Classic's `Find Elements` for this command, an empty list (not an exception) is the right return type.

index.bs
jimevans: @AutomatedTester I would like to have it in the spec, but for this go-round, I can remove it. Or someone else can from the branch, I'm not that wedded to it.

jgraham (changes requested):
So in general, it's unclear to me what level this API is supposed to work at. It seems like the original intent was to push something like Playwright's API directly into the WebDriver protocol. But I think I'd prefer not to try to transcribe the current state of the art into a JSON API, but instead provide the necessary primitives without baking in lots of internal looping.

Also from a spec point of view, I think this really should assume that's it's not running in a js context, and that implementations might work by directly calling internal browser methods on the DOM, rather than going via a scripting context.

index.bs
```diff
@@ -2089,6 +2091,48 @@ To <dfn>await a navigation</dfn> given |context|, |request|, |wait condition|, a
...
+browsingContext.CssLocator = {
+   type: "css",
+   value: text
+}
+
+browsingContext.FrameLocator = {
+   type: "frame"
+}
+
+browsingContext.JavaScriptLocator {
```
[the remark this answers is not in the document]
jimevans: After several request to do exactly this, I've removed this locator from this iteration.

index.bs
jimevans: Removed in subsequent commit.

index.bs
OrKoN: Just a note that if this API needs to be compatible with Puppeteer (Playwright? and maybe others), we need a way to retain ownership because it is what Puppeteer element handling model assumes (and we use script.evaluate for that, that supports that). Also, if this API ends up being defined in terms of DOM APIs, having an option to specify realms seems useful since the results may differ across realms (e.g., the main realm might have DOM APIs overriden). It probably does not make sense to define this API in terms of DOM APIs though: I guess in such a  case one could use DOM APIs directly via scripts. Still we would likely need a way to control the result ownership.

index.bs
jimevans: @jgraham The concrete use case for the `TextLocator` is "find some element with tag other than `<a>` by its visible text. Once we accept that's a legitimate use case (and I'm willing to believe that it is, so strongly that I'm willing to abandon this PR and let someone else author it), then replicating "find element by [partial] link text" makes sense to define as a combination of these primitives.

index.bs
OrKoN: > if you wanted to find a node that was subsequently deleted

the node might get detached from DOM so having the protocol retain the object is useful, if, for example, you are testing for specific content of the element in subsequent calls and you don't care if it is connected to DOM at a later point. If the object is not retained, in highly dynamic apps it might require constant re-querying to do anything in multiple operations. Also, it might work or not depending on gc.

index.bs
```diff
@@ -2725,6 +2769,306 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+   of |command parameters|.
+
+1. Assert: |locators| is not null.
+
+1. Assert: |locators| length is > 0.
+
+1. Let |start node parameter| be the value of the <code>startNode</code> field
+   of |command parameters|, if present, or null otherwise.
+
+1. Let |start node| be the result of [=trying=] to [=deserialize local value=] given
```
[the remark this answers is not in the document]
jimevans: I'm sure you're right. That's my incompetence showing.

index.bs
OrKoN: @jgraham yeah, we have some examples in Chrome DevTools frontend e2e tests where UI updates (usually driven by the browser) are done by removing the content and inserting new DOM elements. References to the old elements are not kept. The test goes like this: 1) get an element 2) check its attributes or content (sometimes more complex stuff) via a second call. In this case, you either need to retry if the element the test managed to grab got GCed, or know really well that the element will not be removed by the next update (e.g., by preventing updates). Apart from that, our overall API in Puppeteer is that the element is retained unless you dispose it (alternatively, it is disposed by navigation).

We currently don't plan to use this API for Puppeteer anyway so the ownership capabilities can be also added later (if there is a need).

index.bs
```diff
@@ -2725,6 +2769,306 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+        params: browsingContext.LocateNodesParameters
+      )
+
+      browsingContext.LocateNodesParameters = {
+         context: browsingContext.BrowsingContext,
+         locators: [ + browsingContext.Locator ],
+         ? maxNodes: js-uint,
+         ? ownership: script.ResultOwnership,
+         ? sandbox: text,
+         ? startNode: script.SharedReference,
```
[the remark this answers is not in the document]
jimevans: Suggestion taken.

index.bs
```diff
@@ -2725,6 +2769,306 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+              1. Let |selector| be the value of the value property of |locator|.
+
+              1. Let |match| be the value of the match property of |locator| if
+                 present, otherwise "full".
+
+              1. Let |context nodes| be a result of [=trying=] to [=locate nodes using text=]
+                 given |current context target|, |context nodes|, |selector|, |match| and
+                 |session|.
+
+   1. Append |context nodes| to |result nodes|.
```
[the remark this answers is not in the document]
jimevans: The algorithm has changed to return only the shared IDs, so this should no longer apply.

index.bs
jimevans: At others' request, I've removed this loop from the algorithm. This question is no longer applicable.

index.bs
jimevans: Fixed.

index.bs
jimevans: I've removed the JavaScript-based locator for now. Implementation of the general case of #447 would be a more appropriate way to handle this.

index.bs
jimevans: @jgraham I've removed the locator chaining from this PR.

index.bs
```diff
@@ -2089,6 +2091,48 @@ To <dfn>await a navigation</dfn> given |context|, |request|, |wait condition|, a
...
+   browsingContext.TextLocator /
+   browsingContext.XPathLocator
+)
+
+browsingContext.CssLocator = {
+   type: "css",
+   value: text
+}
+
+browsingContext.FrameLocator = {
```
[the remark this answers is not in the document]
jimevans: I've removed this locator.

index.bs
```diff
@@ -2725,6 +2769,306 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+
+<div algorithm="locate nodes using CSS">
+To <dfn>locate nodes using CSS</dfn> with given |context target|, |context nodes|,
+|selector| and <var ignore>session</var>:
+
+1. Let |returned nodes| be an empty array.
+
+1. For each |context node| of |context nodes|:
+
+  1. Let |elements| be the result of calling querySelectorAll() on |context target|
```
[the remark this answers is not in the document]
jimevans: @jgraham I think I've updated the spec to handle this correctly now, with the proper references to external specs. Let me know if I've done something incorrectly.

index.bs
```diff
@@ -2725,6 +2769,306 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+
+<div algorithm="locate nodes using XPath">
+To <dfn>locate nodes using XPath</dfn> with given |context target|,
+|context nodes|, |selector| and <var ignore>session</var>:
+
+1. Let |returned nodes| be an empty array.
+
+1. For each |context node| of |context nodes|:
+
+   1. Let |evaluateResult| be the result of calling evaluate on |context target|,
```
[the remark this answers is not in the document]
jimevans: I think I've properly defined this in terms of DOM APIs now.

index.bs
```diff
@@ -2725,6 +2769,306 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+        method: "browsingContext.locateNodes",
+        params: browsingContext.LocateNodesParameters
+      )
+
+      browsingContext.LocateNodesParameters = {
+         context: browsingContext.BrowsingContext,
+         locators: [ + browsingContext.Locator ],
+         ? maxNodes: js-uint,
+         ? ownership: script.ResultOwnership,
+         ? sandbox: text,
```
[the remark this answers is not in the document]
jimevans: Removed in subsequent commit.

index.bs
jimevans: @OrKoN These algorithms have been entirely rewritten. A re-review is required.

index.bs
```diff
@@ -2725,6 +2769,306 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+   parameters| if present, or 0 otherwise.
+
+1. Let |expiration time| be the current time plus |wait timeout| milliseconds.
+
+1. Let |result nodes| be an empty array.
+
+1. Perform the following steps at least once, and repeat until the current time is
+   greater than |expiration time| or |result nodes| is not empty.
+
+  1. Let |context nodes| be an empty array.
```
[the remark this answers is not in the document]
jimevans: Since I've removed the iteration, I'm guessing this comment no longer applies.

index.bs
whimboo: @jgraham for `sharedId` we keep references to elements in weak sets but do not hold a strong reference to it. So if an element is removed from the DOM and got GC'ed the `sharedId` in comparison to the `handle` is no longer valid. That's reminds me that I should write a wdspec test for that!

jimevans: Folks, I'm going to resolve this conversation, as I've redefined the algorithm to only return a list of `sharedId`s. If anyone would rather see different behavior, I'm open to iterating on the proposal, but I warn I won't have a ton of time for revising this PR going forward. We're at the point where someone probably needs to take over authorship from me.

whimboo: Please note that my comment might not block getting the first iteration of this command landed. Having handling of ownership included could be a next step.

index.bs
```diff
@@ -2725,6 +2773,249 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+         ? maxNodes: js-uint,
+      }
+      </pre>
+   </dd>
+   <dt>Result Type</dt>
+   <dd>
+    <pre class="cddl local-cddl">
+        browsingContext.LocateNodesResult = {
+            context: browsingContext.BrowsingContext,
+            nodeSharedIds: [ * script.SharedId ]
```
OrKoN, line 2804: Still a comment related to the ownership discussion before: would it be possible to return script.NodeListRemoteValue here so that the command can be extended without breaking changes in the future?

OrKoN: cc @jgraham @whimboo

OrKoN: ah sorry, it does not have a shared id

whimboo: Well, I think what we need here is not a list of shared ids but a list of `NodeRemoteValue`. That will make it way easier later to get ownership support added because a `handle` is already present.

whimboo: Oh, looks like I missed this comment from @jgraham. Nevertheless I can see the benefits by only returning the `sharedId`'s to reduce the amount of data in the payload of the command's response. But it would require the client to construct a `NodeRemoteObject` on its own, also it's not known which special type those nodes have. Also this will be especially hard to use for batched commands where such a returned list cannot easily be used by the next command as actual `NodeRemoteValue`'s.

I've added the `needs-discussion` label so that we can at latest discuss that in the next group meeting if no other consensus has been found before.

jgraham: I think we should return a list of `script.RemoteReference`. Or maybe even `script.RemoteValue` since you can get XPath to return things that aren't actually nodes.

index.bs
```diff
@@ -2725,6 +2773,249 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+      from |evaluateResult|. If this throws an [=XPathException=] return [=error=]
+      with [=error code=] [=invalid selector=], otherwise if this throws any other
+      exception return return [=error=] with [=error code=] [=unknown error=].
+
+   1. Repeat, while |index| is less than |length|:
+
+      1. Let |node| be the result of calling [=snapshotItem=] with |evaluateResult|
+         as this and |index| as the argument.
+
+      1. If |node| is not an element return [=error=] with [=error code=]
```
jgraham, line 2858: It definitely seems like we could support returning non-Element nodes here. Maybe even non-nodes if we can select them.

index.bs
whimboo: Indeed. That's a good point. I agree on that.

index.bs
```diff
@@ -2089,6 +2105,38 @@ To <dfn>await a navigation</dfn> given |context|, |request|, |wait condition|, a
...
+   browsingContext.TextLocator /
+   browsingContext.XPathLocator
+)
+
+browsingContext.CssLocator = {
+   type: "css",
+   value: text
+}
+
+browsingContext.TextLocator {
```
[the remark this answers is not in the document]
OrKoN: I think there are also useful functions to normalise spaces and do other transformations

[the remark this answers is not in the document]
jimevans: XPath does nothing to take into account visibility of the contained text; innerText does, if memory serves. And the `text()` function _only_ refers to the text node of the node itself, not descendants.

index.bs
OrKoN: What about a list of script.NodeRemoteValue then? to indicate that primitive values and non-node objects cannot be returned?

jimevans: @OrKoN @jgraham XPath _can_ return non-nodes from `evaluate`, but we are restricting the return type to be a set of nodes by using `ORDERED_NODE_SNAPSHOT_TYPE` as the result type argument when we call `evaluate`. This still makes sense to me, as the name of the method is, after all, `locateNodes`. I'll remove the restriction around the returned nodes needing to be elements, and I'll return to making the command  return `RemoteNodeValue`s. I'll also return the optional `ownership` argument so that ownership of the resulting nodes can be properly handled by clients.

index.bs
jimevans: @jgraham We can support returning non-element nodes, and I'll update the PR to reflect that. However, I would like to retain the restriction to only return nodes from XPath, since the command is `locateNodes`.

index.bs
```diff
@@ -2723,6 +2770,218 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+         ? serializationOptions: script.SerializationOptions
+      }
+      </pre>
+   </dd>
+   <dt>Result Type</dt>
+   <dd>
+    <pre class="cddl local-cddl">
+        browsingContext.LocateNodesResult = {
+            context: browsingContext.BrowsingContext,
+            nodes: [ * script.RemoteNodeValue ]
```
sadym-chromium, line 2803:
```suggestion
            nodes: [ * script. NodeRemoteValue ]
```

index.bs
```diff
@@ -2723,6 +2770,218 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+   1. Let |evaluateResult| be the result of calling [=evaluate=] on |context target|'s
+      [=active document=], with arguments |selector|, |context node|, null,
+      [=ORDERED_NODE_SNAPSHOT_TYPE=], and null.
+
+   1. Let |index| be 0.
+
+   1. Let |length| be the result of getting the <code>snapshotLength</code> property
+      from |evaluateResult|. If this throws an [=XPathException=] return [=error=]
+      with [=error code=] [=invalid selector=], otherwise if this throws any other
+      exception return return [=error=] with [=error code=] [=unknown error=].
```
sadym-chromium, line 2850:
```suggestion
      exception return [=error=] with [=error code=] [=unknown error=].
```

index.bs
```diff
@@ -2723,6 +2770,218 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+      1. Append |node| to |returned nodes|.
+      
+      1. Increment |index| by 1. 
+
+1. Return |returned nodes|
+
+</div>
+
+<div algorithm="locate nodes using text">
+To <dfn>locate nodes using text</dfn> with given |context target|, |context nodes|,
```
sadym-chromium, line 2866: Maybe rename algo to `using regexp`?
```suggestion
To <dfn>locate nodes using regexp</dfn> with given |context target|, |context nodes|,
```

index.bs
```diff
@@ -2723,6 +2770,218 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+
+   1. Append |start node| to |context nodes|.
+
+1. If the <code>length</code> property of |context nodes| is 0, append the node 
+   corresponding to the <code>documentElement</code> property of |context|'s
+   [=active document=] to |context nodes|.
+
+1. Let |result nodes| be an empty array.
+
+1. Let |type| be the value of the type property of |locator|.
```
sadym-chromium, line 2934: nit:
```suggestion
1. Let |type| be the value of the <code>type</code> property of |locator|.
```

index.bs
```diff
@@ -2723,6 +2770,218 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+         ? serializationOptions: script.SerializationOptions
+      }
+      </pre>
+   </dd>
+   <dt>Result Type</dt>
+   <dd>
+    <pre class="cddl local-cddl">
+        browsingContext.LocateNodesResult = {
+            context: browsingContext.BrowsingContext,
+            nodes: [ * script. NodeRemoteValue ]
```
OrKoN, line 2803:
```suggestion
            nodes: [ * script.NodeRemoteValue ]
```

index.bs
jimevans: @sadym-chromium Done.

index.bs
```diff
@@ -2723,6 +2770,218 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
 
 1. Return [=success=] with data null.
 
+</div>
+#### The browsingContext.locateNodes Command ####  {#command-browsingContext-locateNodes}
+
+The <dfn export for=commands>browsingContext.locateNodes</dfn> command returns a
+list of all nodes matching the specified locators.
```
whimboo, line 2777: You are using the plural for locator here. I assume that this is a mistake, or shall we consider having the locator to be a list? Probably not given that it would make it hard to distinguish the results without having to put those into their own buckets as well.

index.bs
```diff
@@ -2723,6 +2770,218 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+To <dfn>locate nodes using CSS</dfn> with given |context target|, |context nodes|,
+|selector| and <var ignore>session</var>:
+
+1. Let |returned nodes| be an empty array.
+
+1. For each |context node| of |context nodes|:
+
+  1. Let |parse result| be the result of [=parse a selector=] given |selector|.
+
+  1. If |parse result| is failure, return [=error=] with [=error code=] [=invalid selector=].
```
whimboo, line 2819: Why do we have to do parse the selector for each and every context node? It stays the same and as such we can move this out to just before the loop.

The same applies to the other methods as well.

whimboo (commented):
Thanks Jim! I had first pass through your PR and I noticed some things you might want to have a look at. It's mostly performance related.

index.bs
jimevans: This is a good catch. In the original implementation, there was a notion of chaining locators and the `locator` property, was, in fact, a list, to allow round-trip minimization. I've been talked out of that proposal, in favor of a [more general solution for batching commands](https://github.com/w3c/webdriver-bidi/issues/447). I'll update the PR to reflect this.

index.bs
jimevans: Good point. I don't think the other algorithms suffer from the same issue. For XPath, we have to call `evaluate` using the selector on each of the context nodes, so there is no "parse the XPath to make sure it's valid" step before calling `evaluate`. For find by text, the regexp is parsed before entering the loop for the context nodes.

index.bs
```diff
@@ -2723,6 +2770,218 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+1. If the <code>length</code> property of |context nodes| is 0, append the node 
+   corresponding to the <code>documentElement</code> property of |context|'s
+   [=active document=] to |context nodes|.
+
+1. Let |result nodes| be an empty array.
+
+1. Let |type| be the value of the <code>type</code> property of |locator|.
+
+1. In the following list of conditions and associated steps, run the first set 
+   of steps for which the associated condition is true, if any:
```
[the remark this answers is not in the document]
jimevans: I've made this change in a subsequent commit.

index.bs
```diff
@@ -2723,6 +2770,233 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+   of |command parameters|.
+
+1. Assert: |locator| is not null.
+
+1. Let |start nodes parameter| be the value of the <code>startNodes</code> field
+   of |command parameters|.
+
+1. Assert: |start nodes parameter| is not null.
+
+1. Let |context nodes| be an empty array.
```
jgraham, line 2931: Again, use lists not arrays.

index.bs
```diff
@@ -2723,6 +2770,233 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+
+1. For each |serialized start node| in |start nodes parameter|:
+
+   1. Let |start node| be the result of [=trying=] to [=deserialize shared reference=] given
+      |serialized start node|, |realm| and |session|.
+
+   1. Append |start node| to |context nodes|.
+
+1. If the <code>length</code> property of |context nodes| is 0, append the node
```
jgraham, line 2940: Don't get properties, just refer directly to the spec, in this case https://dom.spec.whatwg.org/#document-element

Also this phrasing means that passing in `startNodes: []` ends up not being an error but implicitly selecting the document element. That seems pretty non-obvious; I'd make `startNodes` optional, but require it to contain at least one element if provided.

index.bs
jgraham: I agree that no match should return an empty array. Invalid selector seems strictly wrong since there are many cases (e.g. polling for an element) where a valid selector will result in zero matches.

index.bs
```diff
@@ -2723,6 +2770,233 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+      <dd>
+         1. Let |selector| be the value of the <code>value</code> property of |locator|.
+
+         1. Let |regexp| be the value of the <code>data</code> property of |selector|.
+
+         1. Let |result nodes| be a result of [=trying=] to [=locate nodes using regexp=]
+            given |current context target|, |context nodes|, |regexp|, |maximum returned
+            nodes| and |session|.
+
+1. If |maximum returned nodes| is greater than 0, truncate |result nodes| to the
```
jgraham, line 2979: Since each algorithm is already truncating, I don't think that's possible? It could be an assert.

jgraham (changes requested):

index.bs
```diff
@@ -2723,6 +2770,233 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+
+1. For each |context node| of |context nodes|:
+
+   1. Let |evaluateResult| be the result of calling [=evaluate=] on |context target|'s
+      [=active document=], with arguments |selector|, |context node|, null,
+      [=ORDERED_NODE_SNAPSHOT_TYPE=], and null.
+
+   1. Let |index| be 0.
+
+   1. Let |length| be the result of getting the <code>snapshotLength</code> property
```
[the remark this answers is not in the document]
jimevans: You're right. The exception should be detected when `evaluate` is called. Will fix.

index.bs
```diff
@@ -2723,6 +2770,233 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+      1. Let |node| be the result of calling [=snapshotItem=] with |evaluateResult|
+         as this and |index| as the argument.
+
+      1. Append |node| to |returned nodes|.
+      
+      1. If |maximum returned nodes| is greater than 0 and <code>length</code> of
+         |returned nodes| is greater than or equal to |maximum returned nodes|,
+         return |returned nodes|.
+
+      1. Increment |index| by 1.
```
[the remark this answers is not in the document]
jimevans: Will so update.

index.bs
```diff
@@ -2723,6 +2770,233 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+   <code>context</code> field set to the [=browsing context id=]
+   of |context|.
+
+1. Let |realm| be the result of [=trying=] to [=get a realm from a target=]
+   given |current context target|.
+
+1. Let |locator| be the value of the <code>locator</code> field
+   of |command parameters|.
+
+1. Assert: |locator| is not null.
```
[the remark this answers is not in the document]
jimevans: Fair. Will remove.

index.bs
jimevans: You're right. That was a recent change made yesterday, and I forgot to update this step. Will fix.

index.bs
```diff
@@ -2723,6 +2770,233 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+|regexp|, |maximum returned nodes|, and <var ignore>session</var>:
+
+1. Let |returned nodes| be an empty array.
+
+1. Let |element nodes| be the result of [=locate nodes using CSS=] given |context target|,
+   |context nodes|, "*" and <var ignore>session</var>.
+
+1. For each |element node| in |element nodes|:
+
+  1. Let |node text| be the <code>innerText</code> property of |element node|.
```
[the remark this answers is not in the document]
jimevans: Thanks for that pointer. Will so update.

index.bs
```diff
@@ -2723,6 +2770,233 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+        }
+    </pre>
+   </dd>
+</dl>
+
+<div algorithm="locate nodes using CSS">
+To <dfn>locate nodes using CSS</dfn> with given |context target|, |context nodes|,
+|selector|, |maximum returned nodes|, and <var ignore>session</var>:
+
+1. Let |returned nodes| be an empty array.
```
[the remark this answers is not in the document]
jimevans: @jgraham I think I've caught all of these instances. Please feel free to rereview.

index.bs
```diff
@@ -2723,6 +2770,233 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+
+  1. Let |elements| be the result of [=match a selector against a tree=] with 
+     |parse result| and |context target|’s [=active document=] [=root=] using
+     [=scoping root=] |context node|.
+
+  1. For each |element| in |elements|:
+
+    1. Append |element| to |returned nodes|.
+
+    1. If |maximum returned nodes| is greater than 0 and <code>length</code> of
```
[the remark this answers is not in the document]
jimevans: @jgraham I think I've caught all of these instances. Please feel free to rereview.

index.bs
```diff
@@ -2723,6 +2770,233 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+   |context nodes|, "*" and <var ignore>session</var>.
+
+1. For each |element node| in |element nodes|:
+
+  1. Let |node text| be the <code>innerText</code> property of |element node|.
+
+  1. Let |is text match| be the result of calling [=test=] with |regexp| as
+     this and |node text| as the argument.
+
+  1. If |is text match| is true, append |element node| to |returned nodes|.
```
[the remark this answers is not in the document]
jimevans: @jgraham I think I've caught all of these instances. Please feel free to rereview.

index.bs
```diff
@@ -2723,6 +2770,233 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+
+1. Let |context id| be the value of the <code>context</code> field of
+   |command parameters|.
+
+1. Let |context| be the result of [=trying=] to [=get a browsing context=]
+   with |context id|.
+
+1. Assert: |context| is not null.
+
+1. Let |maximum returned nodes| be the value of the <code>maxNodes</code>
```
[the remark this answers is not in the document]
jimevans: I've updated this method of accessing parameters. I've also updated `maxNodes` to use its absence to indicate "no maximum".

index.bs
jimevans: I think I've covered all of these.

index.bs
jimevans: @jgraham Double-check my change here on how the algorithm works now. I don't know if the prose is correct. I've made `startNodes` optional, and have it's absence indicate "use the document element as the sole context node."

index.bs
```diff
@@ -2723,6 +2770,233 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+         return |returned nodes|.
+
+      1. Increment |index| by 1. 
+
+1. Return |returned nodes|
+
+</div>
+
+<div algorithm="locate nodes using regexp">
+To <dfn>locate nodes using regexp</dfn> with given |context target|, |context nodes|,
```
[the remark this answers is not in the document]
jimevans: @jgraham I'm concerned about both of these issues too. Regarding the algorithm, I don't see how to better specify it, though. I'm certainly open to suggestions.

And no, that's not how I _want_ the algorithm to work. What I'd _actually_ want is the "deepest" element returned (the `<span>` in your example), but if you're already worried about perf, I don't see how to check `node.contains` and replace "shallower" nodes with the correct "deeper" node. Again I'm open to suggestions.

index.bs
```diff
@@ -2723,6 +2770,233 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+1. Let |returned nodes| be an empty array.
+
+1. Let |element nodes| be the result of [=locate nodes using CSS=] given |context target|,
+   |context nodes|, "*" and <var ignore>session</var>.
+
+1. For each |element node| in |element nodes|:
+
+  1. Let |node text| be the <code>innerText</code> property of |element node|.
+
+  1. Let |is text match| be the result of calling [=test=] with |regexp| as
```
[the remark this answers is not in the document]
jimevans: I've removed the use of regexp from the draft, so this comment no longer applies.

index.bs
```diff
@@ -2723,6 +2770,233 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+       |returned nodes| is greater than or equal to |maximum returned nodes|,
+       return |returned nodes|.
+
+1. Return |returned nodes|.
+
+</div>
+
+<div algorithm="locate nodes using XPath">
+To <dfn>locate nodes using XPath</dfn> with given |context target|,
+|context nodes|, |selector|, |maximum returned nodes|, and <var ignore>session</var>:
```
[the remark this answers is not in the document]
jimevans: I _think_ I've done this correctly. Please advise if not.

index.bs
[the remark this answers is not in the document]
jimevans: I've refined the algorithm to no longer use a RegExp, and it now allows the user to control the recursion level using a `maxDepth` property. It still has the potential to be N² performant, but it's now much more controllable.

index.bs
```diff
@@ -2089,6 +2109,39 @@ To <dfn>await a navigation</dfn> given |context|, |request|, |wait condition|, a
...
+   browsingContext.TextLocator /
+   browsingContext.XPathLocator
+)
+
+browsingContext.CssLocator = {
+   type: "css",
+   value: text
+}
+
+browsingContext.TextLocator {
```
jrandolf, line 2128: I'm slightly against the idea of having a text selector with WebDriver BiDi. The reasons are:

 - There is no support of ShadowDOM,
 - Text could also include `input` values and other properties,
 - It's almost guaranteed to be less capable than a well-written text searcher built client-side.

To stay compatible with WebDriver, I'd be in favor of just adding the `link text` and `partial link text` algorithm verbatim here. On the other hand, we could also update the algorithm to be actually useful (i.e. fixing the above problems), however this is itself a large problem.

All in all, perhaps we can split this PR into two PRs, one for text locators and one for the rest.

whimboo: Splitting up sounds great. This would allow us to start sooner than later on the implementation.

jgraham: > There is no support of ShadowDOM,

That's currently true, in the sense that one doesn't get the text from within shadow DOM. But it's also true that it's impossible for the js-based solutions to work with closed shadow DOM, whereas a native implementation could do. 

In general, better support for shadow DOM seems like a good reason to build in element selection rather than depending only on js, rather than the other way around.

> Text could also include input values and other properties,

I can't really tell what you mean here. `innerText` doesn't contain the input value. 

> It's almost guaranteed to be less capable than a well-written text searcher built client-side.

I don't think this is true. A text-searcher written client side can't pierce closed shadow DOM without making multiple roundtrips, and getting the same layout information as `innerText` uses is challenging. On the other hand depending specifically on `innerText` does sacrifice quite a lot of flexibility and some performance (assuming one doesn't have the ability to cache partial innerText values).

So it seems like engineering tradeoffs.

index.bs
```diff
@@ -2089,6 +2109,39 @@ To <dfn>await a navigation</dfn> given |context|, |request|, |wait condition|, a
...
+   type: "text",
+   value: text,
+   ? ignoreCase: bool
+   ? matchType: "full" / "partial",
+   ? maxDepth: js-uint,
```
thiagowfx, line 2133: What is the difference between `maxDepth` here and `maxNodes` below?

index.bs
```diff
@@ -2750,6 +2803,302 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+
+      1. [=list/append=] |node| to |returned nodes|.
+      
+      1. If |maximum returned nodes| is greater than 0 and [=list/size=] of
+         |returned nodes| is greater than or equal to |maximum returned nodes|,
+         return |returned nodes|.
+
+      1. Set |index| to |index| + 1. 
+
+1. Return |returned nodes|
```
thiagowfx, line 2909:
```suggestion
1. Return |returned nodes|.
```

index.bs
thiagowfx: Never mind: This is explanatory from reading the algorithm below.

`maxNodes` caps the amount of returned nodes. Initially I was thinking it referred to the depth. I would argue the name could be improved (e.g. `maxNumberOfNodes` would certainly be unambiguous), feel free to bikeshed if warranted, but no strong opinions here.

index.bs
jrandolf: > I don't think this is true.
 
I made an assumption that a theoretical text searcher in WebDriver BiDi would not be descriptive enough to allow handling this case. Given that querying closed shadow roots are not part of any spec, I assume there are many obstacles with writing a text searcher that can fix all details.

> I can't really tell what you mean here.

In Puppeteer and Playwright, `input.value` is considered to be the "innerText" of an input element (since input elements don't have innerText).

AutomatedTester: > It's almost guaranteed to be less capable than a well-written text searcher built client-side.

Who is going to write it? I don't think it will get written when it is really useful for the community to have from the start.

Users will Hightlight text and then paste that into their test and say "find this text". This is how a lot of users have used Find By Link Text in webdriver classic. 

> Text could also include input values and other properties,
Users already know that to get the value of the input they need to get it via a property or attributes. I can't see them looking for elements based on text that way. 

The use of innerText here would would be better. Browser vendors can spend time improving the algorithm to make it less expensive. The Selenium code around visibility doesn't do any form of caching and it's never been too bad expect on text heavy websites like the HTML spec.

index.bs
jimevans: @thiagowfx I've renamed the parameter to `maxReturnedNodeCount`. Does that satisfy your concern?

thiagowfx: Yes, thanks! "Count" makes it unambiguous to me.

index.bs
jrandolf: > Who is going to write it? I don't think it will get written when it is really useful for the community to have from the start.

There was an implication that it's already been done; e.g. Puppeteer and Playwright.

jimevans: @jrandolf It's been repeatedly made clear that Puppeteer has no intention of using this API (unless something has changed to make that statement obsolete), and Playwright has expressed an explicit disinterest in supporting use cases that this would enable (e.g., non-Web or mobile use cases). I've already culled a great deal from the initial proposal that would make things far more useful (explicit ShadowRoot and Frame locators, batching of location steps to minimize round trips, and so on). This is a proverbial hill I'm willing to make a stand on.

I do not want to artificially cripple a "find by visible text" locator (e.g., by making it only applicable to `<a>` elements), so as to allow frameworks based on WebDriver BiDi a chance to compete with frameworks based on other, non-open, proprietary technologies. The fact that those other frameworks may consider the `value` property of elements that do not produce [CSS boxes when rendered](https://html.spec.whatwg.org/multipage/dom.html#rendered-text-collection-steps) as part of the "innerText" of the element is immaterial to me. This is especially so since (1) that's not relying on standardized behavior, and (2) those projects are unlikely to use this mechanism in the first place, by their own admission.

jgraham (changes requested):
FWIW I wrote a quick, not well tested, implementation of the text match: https://jsfiddle.net/dmgzsoL2/

index.bs
```diff
@@ -2751,6 +2804,302 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+     <dl>
+
+       <dt>|match type| is "partial" and |ignore case| is false
+       <dd>
+         1. If |node text| contains the substring |selector|, including case,
+            set |is match| to true.
+
+       <dt>|match type| is "partial" and |ignore case| is true
+       <dd>
+         1. If |node text| contains the substring |selector|, ignoring case,
```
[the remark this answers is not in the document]
jimevans: @jgraham I knew even as I was writing this that it was grossly unspecified. When we were using RegExp, at least that had the quality of relying on another well-understood set of semantics, even if it were less performant. I don’t know the web spec ecosystem well enough to know how to properly specify string and substring matches, with or without case sensitivity, without resorting to ECMAScript. Please provide the correct language, or at least more detailed information about how to properly adapt the spec prose to which you linked if you care to see this portion of the PR improved.

index.bs
```diff
@@ -2751,6 +2804,302 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+       <dd>
+         1. If |node text| is equal to |selector|, including case, 
+            set |is match| to true.
+
+       <dt>|match type| is "full" and |ignore case| is true
+       <dd>
+         1. If |node text| is equal to |selector|, ignoring case,
+            set |is match| to true.
+
+   1. If |is match| is true, perform the following steps:
```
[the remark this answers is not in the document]
jimevans: That analysis sounds right to me. Back to the proverbial drawing board.

index.bs
```diff
@@ -2751,6 +2804,302 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+
+     1. If |max depth| is not null and |max depth| is equal to 0, perform the
+        following steps:
+        
+       1. [=list/append=] |context node| to |returned nodes|.
```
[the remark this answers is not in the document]
jimevans: Coming at this from a (relatively poor quality, apparently) programming background and not a spec writer’s background is hindering my progress here. I’m having difficulty figuring out whether items passed into an algorithm are treated as by value or by reference or how a `break` or `continue` structure should be handled.

index.bs
```diff
@@ -2751,6 +2804,302 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+
+         1. If |maximum returned node count| is not null and [=list/size=] of
+            |returned nodes| is greater than or equal to |maximum returned node count|,
+            return |returned nodes|.
+         
+         Otherwise, perform the following steps:
+
+         1. Set |max depth| to |max depth| - 1 if |max depth| is not null.
+
+         1. [=list/extend=] |returned nodes| with the result of [=locate nodes using text=]
```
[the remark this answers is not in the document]
jimevans: Probably correct. I’ll reevaluate once I’ve had a coffee. It’s early here yet.

index.bs
AutomatedTester: There were 2 main value propositions here as I see it.

Firstly, remove the need to inject JS over the wire to do simple things. Time and again we've had members going "we'll just inject javaScript" which is over time is going impact the performance of people using cloud services. If we inject things on session create, or in the mobile case, we going negatively impact the way webdriver works. WebDriver Classic works really well on mobile devices (except for native mobile dialogs). The non-web case also benefits from this as they can then figure out what to do find elements without having to break down the JS and guess where the user wants the element from.

Secondly, it was a way to give a meaningful way to find the elements with minimal round trips. Again this benefits cloud providers immensely, including Playwright as they branch out into the cloud space, hopefully helping people run their tests faster. It normalises what tools like NightwatchJS, Playwright, Puppeteer are doing with a lot of round trips and tries to minimise that. Other JS frameworks like Testing Library would benefit from this. The outlier is going to be closed shadow roots (and that should have been solved by the web components group ages ago but that's a different argument)

index.bs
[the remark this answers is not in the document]
jimevans: For what it's worth, I think I've corrected the algorithm to do what it's expected to do now.

index.bs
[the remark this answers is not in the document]
jimevans: Have corrected the algorithm for this case.

index.bs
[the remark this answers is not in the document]
jimevans: I've refactored the algorithm to use a single return now.

index.bs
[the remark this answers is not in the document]
jimevans: I've now linked to the Unicode specification. If that's not sufficient, feel free to suggest a different solution.

index.bs
jrandolf: I think what would satisfy me is just renaming this to `inner-text` rather than `text`.

jimevans: @jrandolf So renamed in latest commit.

index.bs
```diff
@@ -2778,6 +2831,276 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+            |locator|["<code>ignoreCase</code>"]. Otherwise, let |ignore case| be false.
+
+         1. If |locator| [=map/contains=] <code>matchType</code>, let |match type| be 
+            |locator|["<code>matchType</code>"]. Otherwise, let |match type| be "full".
+
+         1. Let |result nodes| be a result of [=trying=] to [=locate nodes using inner text=]
+            given |current context target|, |context nodes|, |selector|, |max depth|,
+            |match type|, |ignore case|, |maximum returned node count| and |session|.
+
+1. Assert [=list/size=] of |result nodes| is less than or equal to |maximum returned node count|.
```
jgraham, line 3082: This doesn't work if it's null.

index.bs
```diff
@@ -2778,6 +2831,276 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+        method: "browsingContext.locateNodes",
+        params: browsingContext.LocateNodesParameters
+      )
+
+      browsingContext.LocateNodesParameters = {
+         context: browsingContext.BrowsingContext,
+         locator: browsingContext.Locator,
+         ? maxReturnedNodeCount: (js-uint .ge 1),
+         ? ownership: script.ResultOwnership,
+         ? serializationOptions: script.SerializationOptions
```
jgraham, line 2854: I think we also need to allow passing in a sandbox name, so that you can create the handles in the right sandbox.

jgraham (changes requested):

index.bs
```diff
@@ -2778,6 +2831,276 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+      <pre class="cddl remote-cddl">
+      browsingContext.LocateNodes = (
+        method: "browsingContext.locateNodes",
+        params: browsingContext.LocateNodesParameters
+      )
+
+      browsingContext.LocateNodesParameters = {
+         context: browsingContext.BrowsingContext,
+         locator: browsingContext.Locator,
+         ? maxReturnedNodeCount: (js-uint .ge 1),
```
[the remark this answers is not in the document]
jimevans: `maxNodes` caused confusion with at least one WG member. I'm happy to change it back. Will update to `maxNodeCount`.

index.bs
```diff
@@ -2778,6 +2831,276 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+      from |evaluate result|.
+
+   1. Repeat, while |index| is less than |length|:
+
+      1. Let |node| be the result of calling [=snapshotItem=] with |evaluate result|
+         as this and |index| as the argument.
+
+      1. [=list/Append=] |node| to |returned nodes|.
+      
+      1. If |maximum returned node count| is greater than 0 and [=list/size=] of
```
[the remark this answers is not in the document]
jimevans: Ibid.

index.bs
```diff
@@ -2778,6 +2831,276 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+<div algorithm="locate nodes using inner text">
+To <dfn>locate nodes using inner text</dfn> with given |context target|, |context nodes|,
+|selector|, |max depth|, |match type|, |ignore case|, |maximum returned node count|,
+and |session|:
+
+1. Let |returned nodes| be an empty [=/list=].
+
+1. If |ignore case| is false, let |search text| be |selector|. Otherwise, let
+   |search text| be the result of executing [=toUppercase=] with |selector|
+   according to the Unicode Default Case Conversion algorithm.
```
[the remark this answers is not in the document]
jimevans: `toUppercase` already does link. I'll add a link to the "Unicode Default Case Conversion algorithm."

index.bs
```diff
@@ -2778,6 +2831,276 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+
+<div algorithm="locate nodes using inner text">
+To <dfn>locate nodes using inner text</dfn> with given |context target|, |context nodes|,
+|selector|, |max depth|, |match type|, |ignore case|, |maximum returned node count|,
+and |session|:
+
+1. Let |returned nodes| be an empty [=/list=].
+
+1. If |ignore case| is false, let |search text| be |selector|. Otherwise, let
+   |search text| be the result of executing [=toUppercase=] with |selector|
```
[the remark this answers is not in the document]
jimevans: Fixed in next commit

index.bs
```diff
@@ -2778,6 +2831,276 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+   according to the Unicode Default Case Conversion algorithm.
+
+1. For each |context node| in |context nodes|:
+
+  1. Let |node inner text| be the result of calling the [=innerText getter steps=] with
+     |context node| as the [=this=] value.
+
+  1. If |ignore case| is false, let |node text| be |node inner text|. Otherwise,
+     let |node text| be the result of executing [=toUppercase=] with |node inner text|
```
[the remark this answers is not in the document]
jimevans: Fixed in next commit

index.bs
```diff
@@ -2778,6 +2831,276 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+     according to the Unicode Default Case Conversion algorithm.
+
+  1. If |match type| is "full" and |node text| [=is=] |search text|, [=list/append=]
+     |context node| to |returned nodes|. Otherwise, perform the following steps:
```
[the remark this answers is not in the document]
jimevans: I'll adjust the formatting.

index.bs
```diff
@@ -2778,6 +2831,276 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+
+      1. Let |child nodes| be an empty [=/list=] and, for each node |child| in the
+         <a spec=dom>children</a> of |context node|:
+
+        1. If |child| implements {{Element}}, [=list/append=] |child| to
+           |child nodes|.
+
+      1. If [=list/size=] of |child nodes| is equal to 0 or |max depth|
+         is equal to 0, [=list/append=] |context node| to |returned nodes|.
+         Otherwise, perform the following steps:
```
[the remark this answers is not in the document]
jimevans: Will likewise adjust formatting.

index.bs
```diff
@@ -2778,6 +2831,276 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+1. Let |context| be the result of [=trying=] to [=get a browsing context=]
+   with |context id|.
+
+1. Assert: |context| is not null.
+
+1. If |command parameters| [=map/contains=] "<code>maxReturnedNodeCount</code>", let
+   |maximum returned node count| be |command parameters|["<code>maxReturnedNodeCount</code>"].
+   Otherwise, let |maximum returned node count| be null.
+
+1. If |maximum returned node count| is not null and |maximum returned node count|
```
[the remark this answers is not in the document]
jimevans: Happy to remove the step, but is there a "fail if this doesn't match the CDDL definition" step somewhere?

index.bs
```diff
@@ -2778,6 +2831,276 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+   |serialized start node| in |start nodes parameter|:
+
+   1. Let |start node| be the result of [=trying=] to [=deserialize shared reference=] given
+      |serialized start node|, |realm| and |session|.
+
+   1. [=list/Append=] |start node| to |context nodes|.
+
+1. Assert [=list/size=] of |context nodes| is greater than 0.
+
+1. Let |result nodes| be an empty [=/list=].
```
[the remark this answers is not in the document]
jimevans: Why not? Don't we need to define `result nodes` somewhere? Or do you mean something else is needed (like the link to the definition of list)?

index.bs
```diff
@@ -2778,6 +2831,276 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+
+1. Let |type| be |locator|["<code>type</code>"].
+
+1. In the following list of conditions and associated steps, run the first set
```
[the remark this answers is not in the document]
jimevans: Fair. Fixed in next commit.

index.bs
jimevans: Fixed the assertion in next commit.

index.bs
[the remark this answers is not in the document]
jgraham: You have a branch below and each branch assigns something to result nodes. So I think that's fine from a spec point of view, you don't need to act like it's block scoped and the variable needs to be defined outside the conditional.

index.bs
jimevans: So added.

index.bs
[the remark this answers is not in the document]
jimevans: Ah. Okay, I get it now. That makes sense. Will remove.

index.bs
```diff
@@ -2778,6 +2831,276 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+1. For each |context node| in |context nodes|:
+
+  1. Let |node inner text| be the result of calling the [=innerText getter steps=] with
+     |context node| as the [=this=] value.
+
+  1. If |ignore case| is false, let |node text| be |node inner text|. Otherwise,
+     let |node text| be the result of executing [=toUppercase=] with |node inner text|
+     according to the Unicode Default Case Conversion algorithm.
+
+  1. If |match type| is "full" and |node text| [=is=] |search text|, [=list/append=]
```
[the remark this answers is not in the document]
jimevans: Should be adjusted now.

jimevans:
Beginnings of WPT tests at https://github.com/web-platform-tests/wpt/pull/42565

index.bs
```diff
@@ -2778,6 +2832,278 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+         ? serializationOptions: script.SerializationOptions,
+         ? startNodes: [ + script.SharedReference ]
+      }
+      </pre>
+   </dd>
+   <dt>Result Type</dt>
+   <dd>
+    <pre class="cddl local-cddl">
+        browsingContext.LocateNodesResult = {
+            context: browsingContext.BrowsingContext,
```
OrKoN, line 2865: do we expect result to come from a different context? I think the result context would be the same as the input context, would it?

index.bs
```diff
@@ -2778,6 +2832,278 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+            |locator|["<code>ignoreCase</code>"]. Otherwise, let |ignore case| be false.
+
+         1. If |locator| [=map/contains=] <code>matchType</code>, let |match type| be 
+            |locator|["<code>matchType</code>"]. Otherwise, let |match type| be "full".
+
+         1. Let |result nodes| be a result of [=trying=] to [=locate nodes using inner text=]
+            given |current context target|, |context nodes|, |selector|, |max depth|,
+            |match type|, |ignore case|, |maximum returned node count| and |session|.
+
+1. Assert |maximum returned node count| is null or [=list/size=] of |result nodes| is less
```
jgraham, line 3084:
```suggestion
1. Assert: |maximum returned node count| is null or [=list/size=] of |result nodes| is less
```

index.bs
```diff
@@ -2778,6 +2832,278 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
...
+   <code>script.SerializationOptions</code> production with the fields
+   set to their default values.
+
+1. If |command parameters| [=map/contains=] "<code>ownership</code>",
+   let |result ownership| be |command parameters|["<code>ownership</code>"].
+   Otherwise, let |result ownership| be "none".
+
+1. Let |serialized nodes| be the result of [=serialize as a remote value=] with
+   |result nodes|, |serialization options|, |result ownership|, a new map as
+   <var ignore>serialization internal map</var>, |realm| and |session|.
```
jgraham, line 3099: Not sure this should be in `<var>` since it's not a variable (I can believe we do this wrong elsewhere in the spec however :) )

index.bs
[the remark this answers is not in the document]
jgraham: Just to answer this: yes, see steps 6/7 in https://w3c.github.io/webdriver-bidi/#handle-an-incoming-message

index.bs
jimevans: I see other places where this type of language is required without the use of `<var>`. Corrected.

OrKoN (approved):
LGTM

jgraham (approved):

PR #552, opened 2023-09-19
Add serialization steps for proxy and generator

lutien:
***
<a href="https://pr-preview.s3.amazonaws.com/lutien/webdriver-bidi/pull/552.html" title="Last updated on Sep 19, 2023, 1:25 PM UTC (71cce1b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/552/d81292d...lutien:71cce1b.html" title="Last updated on Sep 19, 2023, 1:25 PM UTC (71cce1b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6847,10 +6847,6 @@ Issue: Should WindowProxy get attributes in a similar style to Node?
 
 Issue: Describe `script.IteratorRemoteValue` serialization.
 
-Issue: Describe `script.GeneratorRemoteValue` serialization.
-
-Issue: Describe `script.ProxyRemoteValue` serialization.
-
 Issue: handle String / Number / etc. wrapper objects specially?
 
 Values accessible from the ECMAScript runtime are represented by a mirror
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -7063,11 +7059,21 @@ an |ownership type|, a |serialization internal map|, a |realm| and a |session|:
         production in the [=local end definition=], with the <code>handle</code>
         property set to |handle id| if it's not null, or omitted otherwise.
 
+    <dt>|value| has a \[[GeneratorState]] [=internal slot=]
```
whimboo, line 7062: Similar to my other comment regarding async iterators before, it doesn't seem like this will work for `AsyncGenerator` instances given that those have a [`AsyncGeneratorState` property](https://tc39.es/ecma262/#sec-properties-of-asyncgenerator-intances).

index.bs
```diff
@@ -7063,11 +7059,21 @@ an |ownership type|, a |serialization internal map|, a |realm| and a |session|:
...
+    <dd>Let |remote value| be a [=/map=] matching the <code>script.GeneratorRemoteValue</code>
+        production in the [=local end definition=], with the <code>handle</code>
+        property set to |handle id| if it's not null, or omitted otherwise.
+
     <dt>|value| has an \[[ErrorData]] [=internal slot=]
     <dd>Let |remote value| be a [=/map=] matching the <code>script.ErrorRemoteValue</code>
         production in the [=local end definition=], with the <code>handle</code>
         property set to |handle id| if it's not null, or omitted otherwise.
 
+    <dt>|value| has a \[[ProxyHandler]] [=internal slot=] and a \[[ProxyTarget]] [=internal slot=]
```
whimboo, line 7072: Is the extra check for the `ProxyTarget` internal slot required? Similar to other object types the `ProxyHandler` should probably be enough, or?

whimboo (changes requested):

index.bs
lutien: Yeah, I think you're right, I've added the check for `AsyncGeneratorState` internal slot

index.bs
jgraham: I don't think it's currently possible to have one without the other, but checking for both seems reasonable and probably the safest assumption.

jgraham (approved):

lutien:
@whimboo, could you take a look again? James approved this PR, but it's still blocked on your review, since you requested the changes before.

index.bs
whimboo: Sounds fine to me then. Thanks.

whimboo (approved):

PR #555, opened 2023-09-25
Use range instead of `float` for angle values

jrandolf:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/555.html" title="Last updated on Sep 25, 2023, 1:28 PM UTC (64501e6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/555/4361bea...64501e6.html" title="Last updated on Sep 25, 2023, 1:28 PM UTC (64501e6)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -8846,8 +8846,10 @@ specified sequence of user input actions.
       )
 
       input.AngleProperties = (
-        ? altitudeAngle: float .default 0.0,
-        ? azimuthAngle: float .default 0.0,
+        ; 0 .. Math.PI / 2
+        ? altitudeAngle: (0.0..1.5707963267948966) .default 0.0,
+        ; 0 .. 2 * Math.PI
+        ? azimuthAngle: (0.0..6.283185307179586) .default 0.0,
       )
 
       input.TiltProperties = (


OrKoN (approved):

thiagowfx (commented):
LGTM but how many digits should we use for the decimals?

The print to PDF example was the first one where this happened, and we only used 4 decimals (rounding up).

Will implementations keep this precision? I'd suggest to round up to 4 decimals or so.

jrandolf:
> LGTM but how many digits should we use for the decimals?
> 
> The print to PDF example was the first one where this happened, and we only used 4 decimals (rounding up).
> 
> Will implementations keep this precision? I'd suggest to round up to 4 decimals or so.

Considering `6.283185307179586 === (Math.PI * 2)`, we should keep it at that precision.

thiagowfx (approved):
OK, matching JS is reasonable

index.bs
```diff
@@ -8846,8 +8846,10 @@ specified sequence of user input actions.
       )
 
       input.AngleProperties = (
-        ? altitudeAngle: float .default 0.0,
-        ? azimuthAngle: float .default 0.0,
+        ; 0 .. 2 * Math.PI
+        ? altitudeAngle: (0.0..6.283185307179586) .default 0.0,
+        ; 0 .. Math.PI / 2
+        ? azimuthAngle: (0.0..1.5707963267948966) .default 0.0,
```
juliandescottes, line 8852: I think that's the opposite? azimuthAngle is between 0 and 2π and altitudeAngle is between 0 and π/2

juliandescottes (changes requested):
Thanks but should be the other way around? (based on https://w3c.github.io/webdriver/#dfn-extract-an-action-sequence)

index.bs
OrKoN: oh right, overlooked that.

jrandolf: Indeed!

juliandescottes (approved):
thanks!

PR #557, opened 2023-09-26
Extend browsingContext.setViewport with devicePixelRatio

### LINKED ISSUE #529: Support emulation of the device pixel ratio
Originally this request was part of https://github.com/w3c/webdriver-bidi/issues/415 but then we decided to only cover width and height for the beginning. What is still left for device emulation is the device pixel ratio override.

Most likely this will become an extra argument to the `browsingContext.setViewport` command.

OrKoN:
Closes #529

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/557.html" title="Last updated on Oct 9, 2023, 6:13 AM UTC (5563672)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/557/d8df062...5563672.html" title="Last updated on Oct 9, 2023, 6:13 AM UTC (5563672)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -211,12 +211,11 @@ spec: CSS-VALUES-3; urlPrefix: https://drafts.csswg.org/css-values-3/
     text: absolute lengths; url: #absolute-lengths
 spec: CSSOM-VIEW; urlPrefix: https://drafts.csswg.org/cssom-view/
   type: dfn
+    text: viewport; url: #visual-viewport
     text: visual viewport; url: #visual-viewport
-<<<<<<< HEAD
     text: layout viewport; url: #layout-viewport
-=======
     text: device pixel ratio; url: #dom-window-devicepixelratio
->>>>>>> f7948c0 (WIP dpr)
+    text: CSS pixel; url: #dom-window-devicepixelratio
 </pre>
 
 <pre class="biblio">
@@ -3074,6 +3073,10 @@ The [=remote end steps=] with |command parameters| are:
 
 1. Let |viewport parameter| be the |command parameters|["<code>viewport</code>"] field.
 
+1. Let |device pixel ratio parameter| be the |command parameters|["<code>devicePixelRatio</code>"] field.
+
+1. If |device pixel ratio parameter| is not null and the implementation is not able to change the size of the [=CSS Pixel=] in |context|'s [=viewport=] for any reason, return [=error=] with [=error code=] [=unsupported operation=].
+
 1. If |viewport parameter| is not null:
     1. Let |width| be the |viewport parameter|["<code>width</code>"]
     1. Let |height| be the |viewport parameter|["<code>height</code>"]
@@ -3081,15 +3084,14 @@ The [=remote end steps=] with |command parameters| are:
     1. Set the [=layout viewport=] of the |context| to be |width| CSS pixels wide and |height| CSS pixels high.
 1. Else:
     1. Set the layout [[css2#viewport]] of |context| to the implementation-defined default layout viewport for |context|.
-1. Let |device pixel ratio parameter| be the |command parameters|["<code>devicePixelRatio</code>"] field.
+
 1. If |device pixel ratio parameter| is not null:
-  1. If the implementation is not able to set |device pixel ratio parameter| for |context| for any reasonm return [=error=] with [=error code=] [=unsupported operation=].
-  1. Override the [=device pixel ratio=] algorithm with the algorithm returning |device pixel ratio parameter| for |context|.
-  1. Perform implementation-specific steps to apply [=device pixel ratio=] to the page rendering for |context|.
+  1. Change the size of the [=CSS Pixel=] in |context|'s [=viewport=] such that it corresponds to |device pixel ratio parameter| device pixels.
 1. Else:
-  1. Restore the [=device pixel ratio=] algorithm for |context|.
-  1. Perform implementation-specific steps to apply [=device pixel ratio=] to the page rendering for |context|.
+  1. Set the size of the [=CSS Pixel=] in |context|'s [=viewport=] to the implementation defined default.
+
 1. Run the [[cssom-view-1#resizing-viewports]] steps.
+
 1. Return [=success=] with data null.
 
 </div>
--- a/index.bs
+++ b/index.bs
@@ -3081,11 +3081,9 @@ The [=remote end steps=] with |command parameters| are:
    with [=error code=] [=unsupported operation=]
 
 1. If |viewport parameter| is not null:
-    1. Let |width| be the |viewport parameter|["<code>width</code>"]
-    1. Let |height| be the |viewport parameter|["<code>height</code>"]
-    1. Set the [=layout viewport=] of the |context| to be |width| CSS pixels wide and |height| CSS pixels high.
+    1. Set the [=layout viewport=] of the |context| to be |viewport|["width"] CSS pixels wide and |viewport|["height"] CSS pixels high.
 1. Otherwise:
-    1. Set the [=layout viewport=] of the |context| to the implementation-defined default value.
+    1. Set the [=layout viewport=] of the |context| to the implementation defined default.
 
 1. If |device pixel ratio parameter| is not null:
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3049,6 +3051,14 @@ The [=remote end steps=] with |command parameters| are:
     1. Set the layout [[css2#viewport]] of the |context| to be |width| CSS pixels wide and |height| CSS pixels high.
 1. Else:
     1. Set the layout [[css2#viewport]] of |context| to the implementation-defined default layout viewport for |context|.
+1. Let |device pixel ratio parameter| be the |command parameters|["<code>devicePixelRatio</code>"] field.
+1. If |device pixel ratio parameter| is not null:
+  1. If the implementation is not able to set |device pixel ratio parameter| for |context| for any reasonm return [=error=] with [=error code=] [=unsupported operation=].
```
jgraham, line 3056: You want to check everything can be set upfront and fail the command if it can't rather than making it apply some settings before failing. You also potentially want to check the value is within implementation limits e.g. a device pixel ratio of 1E10 or 1E-10 should probably fail.

index.bs
```diff
@@ -3049,6 +3051,14 @@ The [=remote end steps=] with |command parameters| are:
...
+  1. Override the [=device pixel ratio=] algorithm with the algorithm returning |device pixel ratio parameter| for |context|.
```
jgraham, line 3057: "Override the algorithm" doesn't quite make sense here, since other CSS specs aren't calling this algorithm to determine what size to draw things. I think you want to say something like `"Change the size of the [=CSS Pixel=] in |context|'s [=viewport=] such that it corresponds to |device pixel ratio parameter| device pixels"`.

index.bs
```diff
@@ -3049,6 +3051,14 @@ The [=remote end steps=] with |command parameters| are:
...
+  1. Perform implementation-specific steps to apply [=device pixel ratio=] to the page rendering for |context|.
+1. Else:
+  1. Restore the [=device pixel ratio=] algorithm for |context|.
```
jgraham, line 3060: Restore to what?  I think it's OK to say something like `"Set the size of the [=CSS Pixel=] in |context|'s [=viewport=] to the implementation defined default."` or something like that.

jgraham (commented):
Added some comments. Formatting wise I think we usually have a blank line between list items.

index.bs
OrKoN: Should we write down specific limitations or is it okay to cover it via the implementation specific unsupported operation?

jgraham: I think the only limitation we should impose is that it has to be > 0. From a quick glance at the code, Gecko doesn't seem to have other specific limits on our existing pref, although I imagine we have limits elsewhere that make extreme values not work.

OrKoN:
@gsnedders I could not recall the last bits of the discussion that happened at TPAC and I wondered if you could [expand](https://github.com/w3c/webdriver-bidi/issues/529#issuecomment-1721307049) on your comment about naming of the devicePixelRatio parameter?

index.bs
OrKoN: I have added the range limit (float .gt 0.0) to the cddl type

jrandolf:
In some fuzzy sense, `browsingContext.setViewport` is related to `visualViewport`, so I think it would make sense putting `devicePixelRatio` in a new method since it doesn't exist on `visualViewport`. In particular, `devicePixelRatio` is a property on `window`, so `browsingContext.setDevicePixelRatio` makes more sense.

index.bs
```diff
@@ -3069,14 +3071,25 @@ The [=remote end steps=] with |command parameters| are:
...
     1. Let |width| be the |viewport parameter|["<code>width</code>"]
     1. Let |height| be the |viewport parameter|["<code>height</code>"]
     1. If the |width| or |height| is outside of the implementation-specific or runtime constraints, return an [=error=] with the [=error code=] [=unsupported operation=].
     1. Set the [=layout viewport=] of the |context| to be |width| CSS pixels wide and |height| CSS pixels high.
 1. Else:
     1. Set the [=layout viewport=] of the |context| to the implementation-defined default value.
+
+1. If |device pixel ratio parameter| is not null:
+  1. Change the size of the [=CSS Pixel=] in |context|'s [=layout viewport=] such that it corresponds to |device pixel ratio parameter| device pixels.
+1. Else:
```
jgraham, line 3088: We generally use "otherwise" rather than "else"

OrKoN:
Any thoughts on whether we want a separate method for configuring DPR? cc @whimboo @jgraham  (see https://github.com/w3c/webdriver-bidi/pull/557#issuecomment-1740817555)

index.bs
```diff
@@ -3069,14 +3071,25 @@ The [=remote end steps=] with |command parameters| are:
 
 1. Let |viewport parameter| be the |command parameters|["<code>viewport</code>"] field.
 
+1. Let |device pixel ratio parameter| be the |command parameters|["<code>devicePixelRatio</code>"] field.
+
+1. If |device pixel ratio parameter| is not null and the implementation is not able to change the size of the [=CSS Pixel=] in |context|'s [=layout viewport=] for any reason, return [=error=] with [=error code=] [=unsupported operation=].
```
jgraham, line 3076: I think this should say "to the specified values" or something to indicate that it's not only if you're unable to change it in general, but specifically if you're unable to change it to the parameters given. It's a bit unfortunate to say that without saying what the values represent. An alternative, but less concise, phrasing would be:
```
1. Let |viewport| be |command parameters| ["<code>viewport</code>"]

1. Let |device pixel ratio| be |command parameters|["<code>devicePixelRatio"</code>]

1. If |viewport| is not null, and the implementation is unable to set the width and height dimensions
   of the [=layout viewport=] to |viewport|["width"] and |viewport|["height"], respectively, or if 
   |device pixel ratio| is not null and the implementation is unable to set the size of the CSS pixel 
   of the [=layout viewport=] to |device pixel ratio| device pixels, or if the implementation is unable
   to adjust the [=layout viewport=] parameters for any other reason, return [=error=] 
   with [=error code=] [=unsupported operation=]
```

jgraham (changes requested):

index.bs

index.bs
OrKoN: Thanks, I have updated to the alternative phrasing you propose.

jgraham (approved):

whimboo:
> Any thoughts on whether we want a separate method for configuring DPR? cc @whimboo @jgraham (see [#557 (comment)](https://github.com/w3c/webdriver-bidi/pull/557#issuecomment-1740817555))

I'm fine with having it as part of the viewport emulation command as well.

OrKoN:
@whimboo yep, once there are no open discussions on the API structure for devicePixelRatio. Using the setViewport method sounds good to me too, to avoid roundtrip and multiple updates to the page layout.

OrKoN:
Started WPT work here https://github.com/web-platform-tests/wpt/pull/42398/files

jrandolf (approved):

index.bs
```diff
@@ -3067,16 +3069,29 @@ The [=remote end steps=] with |command parameters| are:
...
+1. Let |viewport| be the |command parameters|["<code>viewport</code>"] field.
+
+1. Let |device pixel ratio| be the |command parameters|["<code>devicePixelRatio</code>"] field.
+
+1. If |viewport| is not null, and the implementation is unable to set the width and height dimensions
+   of the [=layout viewport=] to |viewport|["width"] and |viewport|["height"], respectively, or if 
+   |device pixel ratio| is not null and the implementation is unable to set the size of the CSS pixel 
+   of the [=layout viewport=] to |device pixel ratio| device pixels, or if the implementation is unable
+   to adjust the [=layout viewport=] parameters for any other reason, return [=error=] 
+   with [=error code=] [=unsupported operation=]
```
thiagowfx, line 3081:
```suggestion
   with [=error code=] [=unsupported operation=].
```

thiagowfx (approved):

PR #561, opened 2023-09-28
Add formats to `captureScreenshot`

jrandolf:
This PR implements screenshot formats.

Closes https://github.com/w3c/webdriver-bidi/issues/383

#### Do implementations need to support all formats?
No, if the implementation is unable to capture a screenshot of a context for any reason, then return error with error code unsupported operation.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/561.html" title="Last updated on Oct 4, 2023, 11:39 AM UTC (dd89c4a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/561/0b128ab...dd89c4a.html" title="Last updated on Oct 4, 2023, 11:39 AM UTC (dd89c4a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -50,7 +50,6 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
     text: dispatch actions; url: dfn-dispatch-actions
     text: dispatch tick actions; url: dfn-dispatch-tick-actions
     text: draw a bounding box from the framebuffer; url: dfn-draw-a-bounding-box-from-the-framebuffer
-    text: encode a canvas as Base64; url: dfn-encoding-a-canvas-as-base64
     text: endpoint node; url: dfn-endpoint-node
     text: error code; url: dfn-error-code
     text: error; url: dfn-errors
@@ -160,6 +159,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
     text: 2D context creation algorithm; url: canvas.html#2d-context-creation-algorithm
     text: 2D; url: canvas.html#concept-canvas-2d
     text: a browsing context is discarded; url: window-object.html#a-browsing-context-is-discarded
+    text: a serialization of the bitmap as a file; url: canvas.html#a-serialisation-of-the-bitmap-as-a-file
     text: activation notification; url: interaction.html#activation-notification
     text: active browsing context; url: document-sequences.html#nav-bc
     text: active document; for: browsing context; url: document-sequences.html#active-document
@@ -178,6 +178,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
     text: history handling behavior; url: browsing-the-web.html#history-handling-behavior
     text: navigables; url: document-sequences.html#navigables
     text: navigation id; url: browsing-the-web.html#navigation-id
+    text: origin-clean; url: canvas.html#concept-canvas-origin-clean
     text: parent; for:navigable; url: document-sequences.html#nav-parent
     text: prompt to unload; url: browsing-the-web.html#prompt-to-unload-a-document
     text: prompt; url: timers-and-user-prompts.html#dom-prompt
@@ -2220,7 +2221,13 @@ Base64-encoded string.
 
       browsingContext.CaptureScreenshotParameters = {
         context: browsingContext.BrowsingContext,
-        ? clip: browsingContext.ClipRectangle
+        ? format: browsingContext.ImageFormat,
+        ? clip: browsingContext.ClipRectangle,
+      }
+
+      browsingContext.ImageFormat = {
+         type: string,
+         ? quality: 0.0..1.0,
       }
 
       browsingContext.ClipRectangle = (
@@ -2253,7 +2260,7 @@ Base64-encoded string.
    </dd>
 </dl>
 
-TODO: Full page screenshots, and multiple output formats
+TODO: Full page screenshots
 
 <div algorithm>
 To <dfn>normalize rect</dfn> given |rect|:
@@ -2367,6 +2374,23 @@ To <dfn>render viewport to a canvas</dfn> given |viewport| and |rect|:
 
 </div>
 
+<div algorithm>
+To <dfn>encode a canvas as Base64</dfn> given |canvas| and |format|:
+
+1. If |format| is not null, let |type| be the <code>type</code> field of
+   |format|, and let |quality| be the <code>quality</code> field of |format|.
+
+1. Otherwise, let |type| be "image/png" and let |quality| be [=undefined=].
+
+1. Let |file| be [=a serialization of the bitmap as a file=] for |canvas| with
+   |type| and |quality|. 
+
+1. Let |encoded string| be the [=forgiving-base64 encode=] of |file|.
+
+1. Return success with data |encoded string|.
+
+</div>
+
 The [=remote end steps=] with <var ignore>session</var> and |command parameters| are:
 
 1. Let |context id| be |command parameters|["<code>context</code>"].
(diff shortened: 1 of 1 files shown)


jrandolf:
@jgraham @gsnedders PTAL. We internally discussed what formats should be considered. I suggested we use the ones proposed here, but @OrKoN questioned whether they are compatible with other browsers. Considering implementations can throw unsupported operation, I think these should be okay.

Alternatively, we can make this a string and state it must be a image mime type. This is more flexible in case you folks have other requirements such as proprietary formats.

For example,

```
? format: string | browsingContext.ScreenshotFormat;
...

browsingContext.ScreenshotFormat = {
  type: string
} & Extensible
```

We could then change the algorithm to allow implementation-specific formats when the format is not one of the predefined types (webp, png, and jpeg).

OrKoN:
I think if there will be no agreement that we need to support a shared set of the formats, we can probably have a vendor prefix extension for the screenshot type

index.bs
```diff
@@ -2367,6 +2379,43 @@ To <dfn>render viewport to a canvas</dfn> given |viewport| and |rect|:
...
+
+1. Otherwise, if |format| is not <code>"png"</code>, return [=error=] with error
+   code [=invalid argument=].
+
+1. Let |file| be [=a serialization of the bitmap as a file|a serialization of
+   the |canvas| element’s bitmap as a file=] with |type| and |quality|.
+
+1. Let |data url| be a [=data url|<code>data:</code> URL=] representing |file|.
+
+1. Let |index| be the [=String.prototype.indexOf|index of=] "," in |data url|.
```
OrKoN, line 2410: We should not define it in terms of ECMAScript very likely.

jrandolf: Could you provide an alternative?

OrKoN: I believe there should be something in https://infra.spec.whatwg.org

OrKoN: e.g., https://infra.spec.whatwg.org/#strictly-split

OrKoN:
@jrandolf please link and see comments on https://github.com/w3c/webdriver-bidi/issues/383

index.bs
```diff
@@ -37,6 +37,9 @@ spec: RFC6455; urlPrefix: https://tools.ietf.org/html/rfc6455
 spec: RFC8610; urlPrefix: https://tools.ietf.org/html/rfc8610
   type: dfn
     text: match a CDDL specification; url: appendix-C
+spec: RFC2397; urlPrefix: https://tools.ietf.org/html/rfc2397
+  type: dfn
+    text: data url; url: section-2
```
thiagowfx, line 42: nit: keep sorted, move this before RFC6455

index.bs
```diff
@@ -2220,7 +2226,13 @@ Base64-encoded string.
...
       browsingContext.CaptureScreenshotParameters = {
         context: browsingContext.BrowsingContext,
-        ? clip: browsingContext.ClipRectangle
+        ? format: "png" / "webp" / browsingContext.JpegFormat .default "png",
+        ? clip: browsingContext.ClipRectangle,
+      }
+
+      browsingContext.JpegFormat = {
+         type: "jpeg",
+         ? quality: 0.0..1.0,
```
thiagowfx, line 2235: Should quality have a default?

index.bs
jrandolf: It seems infra spec only defined [substring](https://infra.spec.whatwg.org/#code-point-substring-to-the-end-of-the-string) but not _index of_.

I don't have a strong opinion here, but the current text is derived from WebDriver which uses ECMAScript to define the operations described here. Also, several parts of WebDriver and WebDriver BiDi use ECMAScript operations to define operations in algorithms.

@jgraham @gsnedders Is the terminology here fine to use? If there is push-back to the terms here, I'd appreciate some guidance on how to move forward. I think https://infra.spec.whatwg.org/ defines _some_ algorithms, but not all of them. It's also important we do not introduce additional references as WebDriver BiDi already references ECMAScript. A priori, any new reference we introduce here should be migrated to (before or after this PR).

Again, not strong opinions here, so I'm okay with what exists here.

index.bs
jrandolf: Quality has [a default set defined by the browser.](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toDataURL#:~:text=is%20not%20supported.-,encoderOptions,-Optional)

index.bs
```diff
@@ -2220,7 +2226,13 @@ Base64-encoded string.
 
       browsingContext.CaptureScreenshotParameters = {
         context: browsingContext.BrowsingContext,
-        ? clip: browsingContext.ClipRectangle
+        ? format: "png" / "webp" / browsingContext.JpegFormat .default "png",
```
jgraham, line 2229: I think if we're going to have JPEG use an object it's going to be easier and more consistent to use an object in all cases, even when we don't (currently) have adjustable parameters for the other formats.

index.bs
jgraham: I think these days we can bypass the entire issue by just using `[=forgiving base64 encode=] |file|'s data` or something, and bypass the entire "data URL" thing.

jgraham (changes requested):

index.bs
jrandolf: This example follows from `origin` in PerformActions which has `"viewport" / ElementOrigin`

index.bs
jrandolf: Resolved.

index.bs
jgraham: Yeah, although that's for compatibility with classic. Anyway I guess this is probably OK and we can extend it in the future if required.

jgraham (approved):

OrKoN (approved):

index.bs
```diff
@@ -2220,7 +2227,25 @@ Base64-encoded string.
 
       browsingContext.CaptureScreenshotParameters = {
         context: browsingContext.BrowsingContext,
-        ? clip: browsingContext.ClipRectangle
+        ? format: (
+            browsingContext.PngFormat /
+            browsingContext.WebpFormat /
+            browsingContext.JpegFormat
+          ) .default "png",
```
jgraham, line 2234: This default now doesn't match anything.

jgraham (changes requested):

thiagowfx (approved):

OrKoN (approved):
Still LGTM

index.bs
```diff
@@ -2367,6 +2374,30 @@ To <dfn>render viewport to a canvas</dfn> given |viewport| and |rect|:
 
 </div>
 
+<div algorithm>
+To <dfn>encode a canvas as Base64</dfn> given |canvas| and |format|:
+
+1. If the |canvas|’s bitmap’s [=origin-clean=] flag is set to false, return
```
jgraham, line 2380: This seems wrong? apart from anything else I don't think it should be possible since we're creating the canvas as an internal construct, and don't set the flag. But even if it was, it's unclear why it would apply to webdriver.

jgraham (changes requested):
Just one change required here, I think.

index.bs
jrandolf: This statement also exists in WebDriver, probably just copy-pasted in. Agreed that this makes no sense.

jrandolf: In fact, the bottom statement also doesn't make sense since the canvas is created with positive height and width.

jgraham (approved):

jrandolf:
@gsnedders @jgraham Is it ready to land?

whimboo:
@jrandolf are you going to add new wpt tests for this feature? CC @OrKoN.

jrandolf:
Done: https://github.com/web-platform-tests/wpt/pull/42804

PR #562, opened 2023-09-29
Remove type union for tilt and angle properties

jrandolf:
### 3rd Attempt...

`{azimuthAngle:0}` matches `TiltProperties` because it only validates whether `tiltX` and `tiltY` exist and defaults them otherwise. This implies `AngleProperties` is not matched.

This PR implements the TiltProperties and AngleProperties verbatim, according to the Webdriver spec.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/562.html" title="Last updated on Oct 4, 2023, 1:05 PM UTC (32669aa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/562/0b128ab...32669aa.html" title="Last updated on Oct 4, 2023, 1:05 PM UTC (32669aa)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -8892,17 +8892,10 @@ specified sequence of user input actions.
         ? pressure: float .default 0.0,
         ? tangentialPressure: float .default 0.0,
         ? twist: (0..359) .default 0,
-        (input.TiltProperties // input.AngleProperties)
-      )
-
-      input.AngleProperties = (
-         ? altitudeAngle: float .default 0.0,
-         ? azimuthAngle: float .default 0.0,
-      )
-
-      input.TiltProperties = (
-         ? tiltX: js-int .default 0,
-         ? tiltY: js-int .default 0,
+        ; 0 .. Math.PI / 2
+        ? altitudeAngle: (0.0..1.5707963267948966) .default 0.0,
+        ; 0 .. 2 * Math.PI
+        ? azimuthAngle: (0.0..6.283185307179586) .default 0.0,
       )
 
       input.Origin = "viewport" / "pointer" / input.ElementOrigin


jgraham:
I think you want actual prose validating that there isn't a `tiltX` or `tiltY` property because currently we pass this into the classic command that will use those properties if they exist, and IIRC we can't specify the absence of properties in CDDL.

jrandolf:
> I think you want actual prose validating that there isn't a `tiltX` or `tiltY` property because currently we pass this into the classic command that will use those properties if they exist, and IIRC we can't specify the absence of properties in CDDL.

@jgraham If `tiltX` or `tiltY` doesn't exist, then it will be set to `0`. If they do exist, they will be validated and throw if any of them are invalid.

OrKoN (approved):
LGTM

It looks like the union type meant to denote that either angle or tile properties are provided. So I am not sure if it's compatible with the current spec text. Deferring to @jgraham on that.

index.bs
```diff
@@ -8892,17 +8892,12 @@ specified sequence of user input actions.
...
-
-      input.AngleProperties = (
-         ? altitudeAngle: float .default 0.0,
-         ? azimuthAngle: float .default 0.0,
-      )
-
-      input.TiltProperties = (
-         ? tiltX: js-int .default 0,
-         ? tiltY: js-int .default 0,
+        ? tiltX: (-90..90) .default 0,
```
jgraham, line 8895: But this seems wrong.

With this definition, if I write something like:
```js
{
  [….]
  tiltX: -45,
  tiltY: 45
}
```

Then I get something with `{tiltX: 45, tiltY: -45, altitudeAngle: 0, azimuthAngle: 0}` which is nonsense because it's not what you'd get from converting the specified `tiltX`, `tiltY` values into the corresponding `altitudeAngle` / `azimuthAngle` values. As a result we'll end up emitting an event with contradictory values, even though the user didn't do anything to cause that. This is strictly worse than the (already slightly broken) situation with classic where we only pass through the supplied values, and rely on the UA to default any missing values.

The original intent here was that you can either provide the `tiltX` and `tiltY` properties, _or_ `altitudeAngle` and `azimuthAngle`, but not both. It seems like we can't specify that easily in CDDL without resorting to named enums, which seems like overkill for this case. I can think of two obvious solutions:
* Don't specify defaults in CDDL, but have prose that specifies both the defaults and which combination of properties is allowed.
* Only support `altitudeAngle` and `azimuthAngle`, and require classic users to do the coordinate conversion if they're currently using `tiltX`/`tiltY`.

jgraham (changes requested):

index.bs
OrKoN: SGTM

jgraham: We should probably also update the original WebDriver spec to specify how to handle this when generating the DOM events. But that doesn't block this.

jrandolf: Resolved.

jgraham (approved):

whimboo:
Given that both `tiltX` and `tiltY` have been removed the wpt tests need to be updated.  @jrandolf can you please fix them? Thanks.

PR #565, opened 2023-09-29
Make context/navigable id per session.

jgraham:
Also make it clear that context ids actually relate to navigables, not browsing contexts. This part is a temporary workaround while the rest of the spec continues to pass in browsing contexts rather than navigables.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/565.html" title="Last updated on Oct 2, 2023, 1:06 PM UTC (e9e85fe)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/565/0b128ab...e9e85fe.html" title="Last updated on Oct 2, 2023, 1:06 PM UTC (e9e85fe)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1808,7 +1808,7 @@ events relating to [=navigables=].
 
 Note: For historic reasons this module is called <code>browsingContext</code>
 rather than <code>navigable</code>, and the protocol uses the term
-<code>context</code> to refer to navigables, particuarly as a field in command
+<code>context</code> to refer to navigables, particularly as a field in command
 and response parameters.
 
 Issue: Update the spec to use [=navigable=] consistently and remove


index.bs
```diff
@@ -1804,7 +1804,15 @@ The [=remote end steps=] with |session| and <var ignore>command parameters</var>
 ## The browsingContext Module ## {#module-browsingContext}
 
 The <dfn export for=modules>browsingContext</dfn> module contains commands and
-events relating to browsing contexts.
+events relating to [=navigables=].
+
+Note: For historic reasons this module is called <code>browsingContext</code>
+rather than <code>navigable</code>, and the protocol uses the term
+<code>context</code> to refer to navigables, particuarly as a field in command
```
thiagowfx, line 1811:
```suggestion
<code>context</code> to refer to navigables, particularly as a field in command
```

OrKoN:
Is there an issue/discussion for this PR? I wonder if it's worth making context IDs to be session-specific: I see the value of being able to get a context id in one session and start a different session for a different task (e.g., for tracing or network monitoring) on that specific context id.

jgraham:
My assumption has been that as far as possible sessions should be isolated from each other. If you have multiple tools sharing data, they can probably also share a session id and just operate entirely within the same session. This also matches what we already do for node ids.

OrKoN:
I think it was brought up by someone at TPAC that there could be cases where you would want to use a dedicated tool (for example, a network capturing tool) that would create a session to do a specific task in the running instance. At the same it might not be possible to use an existing session because the event subscriptions might not be compatible between tools: e.g., the tool or the main program might not handle the events in the same way.

whimboo (changes requested):
I think that there is a fundamental issue here with WebDriver classic that we have to discuss and agree on first before I think that it makes sense to actually review this PR.

index.bs
```diff
@@ -1880,23 +1888,54 @@ BrowsingContextEvent = (
...
+<div algorithm>
+To <dfn>get a browsing context</dfn> given |session| and |context id|:
 
 1. If |context id| is null, return [=success=] with data null.
 
-1. If there is no browsing context with [=browsing context id=] |context id| return
-   [=error=] with [=error code=] [=no such frame=]
+1. Let |context id map| be |session|'s [=context id map=].
+
+1. For each |navigable| → |id| of |context id map|:
```
whimboo, line 1932: Note that WeakMaps do not support enumeration and as such there is currently no way to actually retrieve the navigable corresponding to the `context id` at the moment.

index.bs
```diff
@@ -1804,7 +1804,15 @@ The [=remote end steps=] with |session| and <var ignore>command parameters</var>
...
-events relating to browsing contexts.
+events relating to [=navigables=].
+
+Note: For historic reasons this module is called <code>browsingContext</code>
+rather than <code>navigable</code>, and the protocol uses the term
+<code>context</code> to refer to navigables, particularly as a field in command
+and response parameters.
+
+Issue: Update the spec to use [=navigable=] consistently and remove
+uses of [=/browsing context=].
```
OrKoN, line 1815: Should we land this part? I can extract it into a PR

PR #568, opened 2023-10-09
Implement `input.FileDialogOpened`

jrandolf:
This PR implements the event portion of file dialog flow.

Bug: https://github.com/w3c/webdriver-bidi/issues/494


OrKoN (approved):

jgraham:
Oh, sorry I missed that there is an HTML patch, I'll read that :)

jgraham:
So generally a concern I have at the moment is that we don't have any mechanism to specify the behaviour when the dialog is intercepted. I slightly worry that the default here being different to the default for alerts is confusing (with alerts we actually show the alert, even if there's an event, and rely on the subscriber sending a command to continue. If possible I think we ought to do that by default in this case as well, but also make it configurable in some way).

jrandolf:
@jgraham The problem with that approach is the blocking behavior of file dialogs, so you cannot set multiple file choosers in parallel.

IIRC, there was a consensus in TPAC to not allow that blocking behavior for file dialogs. Perhaps @OrKoN can confirm.

OrKoN:
> So generally a concern I have at the moment is that we don't have any mechanism to specify the behaviour when the dialog is intercepted. I slightly worry that the default here being different to the default for alerts is confusing (with alerts we actually show the alert, even if there's an event, and rely on the subscriber sending a command to continue. If possible I think we ought to do that by default in this case as well, but also make it configurable in some way).
> IIRC, there was a consensus in TPAC to not allow that blocking behavior for file dialogs. 

I think we touched upon that at TPAC and the idea is that file dialogs might not actually block the execution on the page so it is possible for the user to continue doing things (in the background via scripts and trigger other dialogs) while the dialog is showing that might have unpredictable results. It is different from the alerts where the loop is actually blocked while waiting for the alert resolution. We also discussed that some user agents might not implement a dialog at all.
So, the current version would be preferred from the Puppeteer's perspective. 

@jimevans @shs96c do you have insights into what would be preferred by Selenium? I feel like if the client opts into intercepting the dialogs programmatically, there might be no need to show it on screen. WDYT?

jgraham:
Conceptually file dialogs are similar to alter-type prompts, so I'm keen to make sure we don't end up with different behaviour in each case.

With alerts we currently send the client an event, and you're required to respond with a command to dismiss the alert (and provide the return value for the types where that makes sense).

The proposal here is that with file prompts opting in to getting the event means that you _don't_ get a prompt at all, but only get an event, and can provide the value asynchronously if required. Technically that does make some sense (an alert is fundamentally synchronous, whereas you can provide an input to a file dialog at any time), but it's not clear whether or not that difference is enough to justify having a totally different model compared to alerts.

I'm also not sure about _just_ subscribing to an event having side effects on browser behaviour. It is true that with e.g. network request interception we don't block the request if it matches a pattern but there isn't any session with a subscription to the relevant event. So there is _some_ precedent there. But it also seems meaningfully different that interception is itself a webdriver feature, and without someone subscribing to the event it would be possible for anyone (even a human interacting with the browser) to unblock the request.

lukewarlow:
Why does this only intercept the show picker algorithm for file inputs? Seems like it would be useful to provide a way to intercept all the show picker algorithms?

jrandolf:
> Why does this only intercept the show picker algorithm for file inputs? Seems like it would be useful to provide a way to intercept all the show picker algorithms?

For now, we are limiting the scope of this PR to file inputs.

OrKoN:
I have updated the PR to include the capability processing to reflect the latest discussion at the working group. PTAL

index.bs
```diff
@@ -1400,6 +1400,7 @@ for a session.
 <pre class="cddl remote-cddl local-cddl">
 session.CapabilityRequest = {
   ? acceptInsecureCerts: bool,
+  ? interceptFileDialogs: bool,
```
whimboo, line 1403: By having such a capability I wonder how we best separate those which belong to different protocols. Right now some of the capabilities that we have are used in both classic and bidi, but others are classic only. Now we get bidi only capabilities. All that doesn't make it easy. I wonder if we should actually re-use `unhandledPromptBehavior` for bidi as well, or if we want to make the dialog handling dependent on subscribed events.

OrKoN: I think it's already separated because the BiDi spec defines these as [additional capabilities](https://w3c.github.io/webdriver-bidi/#type-session-CapabilityRequest). Or do you mean how to separate them in the CDDL definitions?

mathiasbynens: @whimboo @jgraham What do you think? [The concept of “additional WebDriver capabilities”](https://w3c.github.io/webdriver/#dfn-additional-webdriver-capability) is defined in the Classic spec, but the WebDriver BiDi spec makes it clear that these additional WebDriver capabilities in particular are BiDi-only:

> WebDriver BiDi defines [additional WebDriver capabilities](https://w3c.github.io/webdriver/#dfn-additional-webdriver-capability). The following tables enumerates the capabilities each implementation must support for WebDriver BiDi.

Do you have a suggestion of a better way to phrase this?

jgraham: I don't think that BiDi-only capabilities are a problem in general.

Currently BiDi doesn't use `unhandledPromptBehavior` at all, but I assume that will change.

If I was designing this from scratch, I might have something like:

```
session.CapabilityRequest =  {
  …
  promptBehavior: session.PromptBehavior
}

session.PromptBehavior = {
  ?alert:  : session.PromptBehaviorOption,
  ?confirm: : session.PromptBehaviorOption,
  ?default: session.PromptBehaviorOption .default "ignore",
  ?file: : session.PromptBehaviorOption,
  ?httpAuth: : session.PromptBehaviorOption,
  ?unload: : session.PromptBehaviorOption,
}

session.PromptBehaviorOption = "accept" / "dismiss" / "ignore"
```

Then for each prompt type you could have different behaviour, depending on what you expect without needing a separate top-level capability. Classic's `unhandledPromptBehavior` would be more or less equivalent to `{default: <behavior>}`. For compatibility we could always use `unhandledPromptBehavior` as a fallback if nothing else is specified.

index.bs
```diff
@@ -9036,6 +9036,59 @@ The [=remote end steps=] given |session|, and |command parameters| are:
...
+   <dd>
+      <pre class="cddl local-cddl remote-cddl">
+         input.FileDialogOpened = (
+            method: "input.fileDialogOpened",
+            params: input.FileDialogInfo
+         )
+
+         input.FileDialogInfo = {
+            context: browsingContext.BrowsingContext,
+            element: script.SharedReference
```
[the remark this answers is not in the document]
OrKoN: I have added multiple as a flag

index.bs
OrKoN: It does not look like unhandledPromptBehavior in classic includes file dialogs (https://html.spec.whatwg.org/#user-prompts)? So it would not be exactly equivalent to the default behavior. Should we change to this structure as part of this PR? I am not sure how would the accept and dismiss work as I understood it's rather difficult to dismiss file dialogs also in Firefox (based on a thread in Matrix)

jgraham: Right, probably only "dismiss" and "ignore" work for all dialogs. "accept" works with dialogs which don't take a value or for which there's a reasonable default value. That doesn't include file or HTTP Auth.

index.bs
```diff
@@ -1444,6 +1446,34 @@ with parameter |value| is:
 
 </div>
 
+<pre class=simpledef>
+Capability: <dfn export>intercept file dialogs</dfn>
+Key: "<code>interceptFileDialogs</code>"
+Value type: boolean
+Description: Defines whether the file dialogs should be intercepted and not shown on screen.
```
sadym-chromium, line 1453: Worth mentioning the default is `true`?

index.bs
```diff
@@ -9589,6 +9619,76 @@ The [=remote end steps=] given |session|, and |command parameters| are:
...
+<dl>
+   <dt>Event Type</dt>
+   <dd>
+      <pre class="cddl local-cddl remote-cddl">
+         input.FileDialogOpened = (
+            method: "input.fileDialogOpened",
+            params: input.FileDialogInfo
+         )
+
+         input.FileDialogInfo = {
```
sadym-chromium, line 9635: Should we provide an info if the file dialog was intercepted or not? And if so, should the value be different for different sessions depending on the `intercept file dialogs` capability?

sadym-chromium (approved):
LGTM. The `isIntercepted` flag in the event can be added later on.

index.bs
OrKoN: I think the default should be false.

sadym-chromium: Then the 'matched capability serialization algorithm for the "interceptFileDialogs" capability' should return false by default.
```
1. If |value| is true, return [=success=] with data true.

1. Return [=success=] with data false.
```

index.bs
```diff
@@ -1444,6 +1446,34 @@ with parameter |value| is:
...
+
+1. Return [=success=] with data |value|.
+
+</div>
+
+<div algorithm="interceptFileDialogs capability serialization algorithm">
+The [=matched capability serialization algorithm=] for the "<code>interceptFileDialogs</code>" capability,
+with parameter |value| is:
+
+1. If |value| is false, return [=success=] with data null.
```
OrKoN, line 1471:
```suggestion
1. If |value| is true, return [=success=] with data true.
```

index.bs
```diff
@@ -1444,6 +1446,34 @@ with parameter |value| is:
...
+
+1. Return [=success=] with data true.
```
OrKoN, line 1473:
```suggestion
1. Return [=success=] with data false.
```

index.bs
OrKoN: The dialog is a shared resource so it will be intercepted if any of the active session intercepts it. I think a flag would be useful but there is no immediate use case for it.

index.bs
sadym-chromium: This algo is about matching capabilities, not the default value.

index.bs
sadym-chromium: This algo is about matching capabilities, not proviging default values.

index.bs
sadym-chromium: The actual default behavior is defined in the `set of sessions for which file dialog interception is enabled`:
```
1. If |session|'s [=intercept file dialogs=] capability is true, append |session| to |sessions|.
```
So I think it makes sense to mention the default behavior in the description.

OrKoN, line 1453:
```suggestion
Description: Defines whether the file dialogs should be intercepted and not shown on screen. Defaults to false.
```

index.bs
jgraham: I think we should introduce something that we can extend to cover more dialogs, even if for now we only support it for file dialogs. So a definition like:
```
session.CapabilityRequest =  {
  …
  promptBehavior: session.PromptBehavior
}

session.PromptBehavior = {
  ?file: : "suppress" / "dismiss" / "none",
}
```

would be fine. `"suppress"` would emit the event but not display the dialog. `"dismiss"` would cause a cancel event to be fired, as if the Cancel button was pressed, and `"none"` would not affect the dialog (although an event would still be fired if a session was subscribed). I"d even be happy to drop `"dismiss"` for now, which I think should make it functionally equivalent to what you already have.

index.bs
```diff
@@ -9589,6 +9619,76 @@ The [=remote end steps=] given |session|, and |command parameters| are:
...
+   otherwise.
+
+1. Let |body| be a [=/map=] matching the <code>input.FileDialogOpened</code>
+   production, with the <code>params</code> field set to |params|.
+
+1. For each |session| in |sessions|:
+
+  1. [=Emit an event=] with |session| and |body|.
+
+1. Let |intercepting sessions| be the [=set of sessions for which file dialog interception is enabled=].
```
jgraham, line 9672: I don't think this should use "did any session suppress the dialog". Rather if any session actually requests a specific capability value here then any later session requesting a different capability value should fail during session creation. If sessions don't request a specific value then a later session requesting a specific value can affect the behaviour (maybe later we should have a `session.CapabilityChanged` event for when a new session affects existing settings, but that's out of scope for now).

jgraham (changes requested):

whimboo:
> <jgraham> Using the userPromptOpened events
> <AutomatedTester> jgraham: that sounds great! I was hoping for that

So I assume that we will have a custom type for file open dialogs then which are emitted as `browsingContext.userPromptOpened` events?  We should then probably update the PR summary to reflect that. @sadym-chromium can you please check? Thanks.

whimboo:
> <AutomatedTester> sadym: I need to go review this and get back to the WG

@sadym-chromium would you mind to follow-up on this action item? Thanks!

index.bs
```diff
@@ -11705,6 +11710,81 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+   1. Assert |user prompt handler| is a [=map=].
+
+   1. If |user prompt handler| [=map/contains=] "<code>fileDialog</code>":
+
+      1. if |user prompt handler|["<code>fileDialog</code>"] is not equal to
+         "<code>ignore</code>", set |dismissed| to true.
+
+   1. Otherwise if |user prompt handler| [=map/contains=] "<code>default</code>" and
+      |user prompt handler|["<code>fileDialog</code>"] is not equal to
+      "<code>ignore</code>", set |dismissed| to true.
```
sadym-chromium, line 11782: Respecting default behavior makes sense, but is a breaking change, as default of "accept" and "dismiss" starts suppressing file dialog.

index.bs
```diff
@@ -2282,7 +2287,7 @@ BrowserResult = (
 
 Each [=/top-level traversable=] is associated with a single <dfn>client
 window</dfn> which represents a rectangular area containing the
-<a spec=css22>viewport</a> that will be used to render that [=/top-level
+<a spec=css2>viewport</a> that will be used to render that [=/top-level
```
sadym-chromium, line 2290: My local build didn't work with `css22` for whatever reason.

sadym-chromium:
Closing in favor of https://github.com/w3c/webdriver-bidi/pull/883

Issue #289, opened 2022-09-19
Allow creating new browsing profiles/contexts (similar to Incognito mode)


OrKoN:
Puppeteer currently allows creating incognito contexts via https://pptr.dev/api/puppeteer.browser.createincognitobrowsercontext. It works similarly to the incognito profile with the distinction that multiple profiles can be created via automation.  

The use cases:

1) The functionality is useful to isolate multiple automation tasks and run them concurrently with low overhead.
2) Ability to configure proxy server settings per profile.  

It has the following properties in Chrome:

1. The created profile has no visible windows/tabs and the window is created when the user explicitly creates a new page (that results in a new window).
2. The profile data is not saved to disk. (Not a requirement for the spec, rather an implementation detail)
3. The profile has storage isolated from other profiles. 
4. The can have its own proxy server configuration and a proxy server by pass list. 
5. Browsing contexts are associated with a profile or belong to the default profile.
6. Profile can be closed by a single command resulting in all pages belonging to it being closed. 

<img width="1225" alt="Screenshot 2022-09-19 at 11 19 31" src="https://user-images.githubusercontent.com/399150/190987168-036f98bb-8dc1-41c4-9c9a-9e6df3891adb.png">

OrKoN:
This Puppeteer API is powered by CDP's Target.createBrowserContext https://chromedevtools.github.io/devtools-protocol/tot/Target/#method-createBrowserContext

whimboo:
@OrKoN I assume this feature request is similar to the [`container` feature in Firefox](https://support.mozilla.org/en-US/kb/containers? If yes we might want to get rid of the `profile` part in the issue summary because that's causing irritation with the browser profiles.

OrKoN:
@whimboo @jgraham I have collected the properties for the profiles (called BrowserContext in Puppeteer/CDP) that Puppeteer currently relies on.

whimboo:
@OrKoN where have you noted down those properties? Maybe I miss an update somewhere? Thanks.

OrKoN:
@whimboo in the issue description

jgraham:
Out of those properties the one that worries me is the requirement that no data is written to disk. Why is that a necessary property for clients? (_maybe_ for firefox, some combination of containers and private browsing produces that property, but it seems less obviously observable to automation tools than the other properties, and arguably like an orthogonal concern)

whimboo:
[...]

PR #570, opened 2023-10-10
Support user contexts

OrKoN:
User Contexts allow basic isolation of the web content state and parallelization of test runs without paying the costs of launching a fresh/multiple browser processes. In this proposal, there is no session created for the user context: multiple user contexts are managed by the same session. The advantage is that the same session can create a fresh state for its operations. In practical terms, we want to have something that would allow us to implement https://pptr.dev/api/puppeteer.browsercontext/#example In Chrome, that corresponds to an incognito window (but you can have multiple for automation). 

Issue https://github.com/w3c/webdriver-bidi/issues/289

Related links:
- Latest discussion: https://www.w3.org/2023/09/15-webdriver-minutes.html#t02
- https://github.com/whatwg/infra/issues/400
- https://www.w3.org/2001/tag/doc/private-browsing-modes/
- https://gist.github.com/mnot/96440a5ca74fcf328d23

Open questions/TODOs:

- [ ] move to the storage domain?
- [ ] support for context names?
- [ ] define what happens to contexts if the browser is closed

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/570.html" title="Last updated on Jan 19, 2024, 12:09 PM UTC (253d3eb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/570/6462133...253d3eb.html" title="Last updated on Jan 19, 2024, 12:09 PM UTC (253d3eb)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2025,11 +2025,12 @@ The <dfn export for=commands>browser.createUserContext</dfn> command creates a [
 
 The [=remote end steps=] are:
 
-1. Let |user context| be the result of performing implementation defined steps to create an [=empty user context=].
+1. Let |user context| be the result of performing implementation-defined steps
+   to create an [=empty user context=].
 
 1. Let |result| be a [=/map=] matching the
-   <code>browser.createUserContextResult</code> production with the <code>userContext</code>
-   field set to |user context|'s [=user context id=].
+   <code>browser.createUserContextResult</code> production with the
+   <code>userContext</code> field set to |user context|'s [=user context id=].
 
 1. Return [=success=] with data |result|.
 
--- a/index.bs
+++ b/index.bs
@@ -1951,6 +1951,10 @@ BrowserCommand = (
 
 <!-- Nothing yet -->
 <pre class="cddl local-cddl">
+BrowserResult = (
+   browser.CreateUserContextResult /
+   browser.GetUserContextsResult
+)
 </pre>
 
 
--- a/index.bs
+++ b/index.bs
@@ -1294,20 +1294,20 @@ Issue: Define how this works.
 # User Contexts # {#user-contexts}
 
 A <dfn>user context</dfn> represents a collection of top-level navigables within
-a [=remote end=]. Each top-level navigable belongs to exactly one [=user
-context=], and child navigables belong to the same [=user context=] as their
-parent. A [=user context=] without navigables belonging to it is an <dfn>empty
-user context</dfn>.
+a [=remote end=]. User contexts have a <dfn>user context id</dfn>, which is a
+string set upon the user context creation. The [=default user context=]'s [=user
+context id|id=] is <code>"default"</code>. A [=remote end=] always contains at
+least one [=user context=] referred as the <dfn>default user context</dfn>.
 
-A [=remote end=] always contains at least one [=user context=] referred as the
-<dfn>default user context</dfn>.
+Each top-level navigable belongs to exactly one [=user context=], and child
+navigables belong to the same [=user context=] as their parent. Each navigable's
+[=user context=] is the [=user context=] the navigables belongs to.
+
+A [=user context=] without navigables belonging to it is an <dfn>empty
+user context</dfn>.
 
 Each [=user context=] represents an additional [=storage key=], so that no
 storage state is shared between different [=user context|user contexts=].
-
-A [=user context=] other than the [=default user context=] has a <dfn>user
-context id</dfn>, which is a string set upon context creation. The [=default
-user context=] does not have a [=user context id=].
 
 Note: The infra spec uses the term "user agent" to refer to the same concept as
 [=user context|user contexts=]. However, this is not compatible with usage of
@@ -2025,8 +2025,7 @@ The <dfn export for=commands>browser.createUserContext</dfn> command creates a
 
 The [=remote end steps=] are:
 
-1. Let |user context| be the result of performing implementation-defined steps
-   to create an [=empty user context=].
+1. Let |user context| be a new [=empty user context=].
 
 1. Let |result| be a [=/map=] matching the
    <code>browser.createUserContextResult</code> production with the
@@ -2069,14 +2068,14 @@ The [=remote end steps=] with |command parameters| are:
 1. Let |user context| be |command  parameters|["<code>userContext</code>"].
 
 1. If the [=user context=] with the [=user context id=] equal to |user context|
-   does not exist in the user agent, return [=error=] with [=error code=]
+   does not exist in the [=remote end=], return [=error=] with [=error code=]
    [=invalid argument=]
 
-1. [=Close=] any [=top-level browsing contexts=] that belong to the |user
-   context| without [=prompting to unload=].
+1. [=Close=] any [=top-level browsing contexts=] whose [=user context=]'s [=user
+   context id=] is |user context| without [=prompting to unload=].
 
-1. Perform implementation-defined steps to remove |user context| from the user
-   agent.
+1. Perform implementation-defined steps to remove |user context| from the
+   [=remote end=].
 
 1. Return [=success=] with data null.
 
@@ -2110,15 +2109,15 @@ The <dfn export for=commands>browser.getUserContexts</dfn> command returns a lis
 
 The [=remote end steps=] are:
 
-1. Let |user contexts| be the result of performing implementation defined steps
-   to get [=user contexts=] from the user agent.
+1. Let |user contexts| be the [=user contexts=] within the [=remote end=].
 
 1. Let |user context info list| be an empty [=/list=].
 
 1. For each |user context| in |user contexts|:
 
   1. Let |info| be a [=/map=] matching the <code>browser.UserContextInfo</code>
-     production with the <code>userContext</code> field set to |user context|'s [=user context id=].
+     production with the <code>userContext</code> field set to |user context|'s
+     [=user context id|id=].
 
   1. Append |info| to |user context info list|.
 
@@ -2312,10 +2311,6 @@ To <dfn>get the browsing context info</dfn> given |context|,
 
    Note: This includes the fragment component of the URL.
 
-1. Let |user context| be null if |context| belongs to the [=default user
-   context=] or the [=user context id=] of the [=user context=] which the |context|
-   belongs to otherwise.
-
 1. Let |child infos| be null.
 
 1. If |max depth| is null, or |max depth| is greater than 0:
@@ -2337,8 +2332,9 @@ To <dfn>get the browsing context info</dfn> given |context|,
    <code>browsingContext.Info</code> production with the <code>context</code>
    field set to |context id|, the <code>parent</code> field set to |parent id|
    if |is root| is <code>true</code>, or unset otherwise, the <code>url</code>
-   field set to |url|, the <code>userContext</code> field set to |user context| if it’s not null, or omitted otherwise,
-   and the <code>children</code> field set to |child infos|.
+   field set to |url|, the <code>userContext</code> field set to |context|'s
+   [=user context id=], and the <code>children</code> field set to |child
+   infos|.
 
 1. Return |context info|.
 
--- a/index.bs
+++ b/index.bs
@@ -1889,10 +1889,10 @@ managing the remote end browser process.
 
 <pre class="cddl remote-cddl">
 BrowserCommand = (
-  browser.Close,
-  browser.CreateUserContext,
-  browser.RemoveUserContext,
-  browser.GetUserContexts,
+  browser.Close //
+  browser.CreateUserContext //
+  browser.RemoveUserContext //
+  browser.GetUserContexts
 )
 </pre>
 
--- a/index.bs
+++ b/index.bs
@@ -1301,27 +1301,61 @@ Issue: Define how this works.
 
 # User Contexts # {#user-contexts}
 
-A <dfn>user context</dfn> represents a collection of top-level navigables within
-a [=remote end=]. User contexts have a <dfn>user context id</dfn>, which is a
-string set upon the user context creation. The [=default user context=]'s [=user
-context id|id=] is <code>"default"</code>. A [=remote end=] always contains one
-[=user context=] referred as the <dfn>default user context</dfn>.
+A <dfn>user context</dfn> represents a collection of zero or more
+[=/top-level traversables=] within a [=remote end=]. Each [=user context=] has
+an associated [=storage partition=], so that [=remote end=] data is not shared
+between different [=user contexts=].
 
-Each top-level navigable belongs to exactly one [=user context=], and child
-navigables belong to the same [=user context=] as their parent. Each navigable's
-[=user context=] is the [=user context=] the navigables belongs to.
-
-A [=user context=] without navigables belonging to it is an <dfn>empty
-user context</dfn>.
-
-Each [=user context=] represents an additional [=storage key=], so that no
-storage state is shared between different [=user context|user contexts=].
+Issue: Unclear that this is the best way to formally define the concept of a
+user context or the interaction with storage.
 
 Note: The infra spec uses the term "user agent" to refer to the same concept as
 [=user context|user contexts=]. However, this is not compatible with usage of
 the term "user agent" to mean the entire web client with multiple [=user
 context|user contexts=]. Although this difference is not visible to web content,
-it is observed via WebDriver, so we avoid reusing existing terminology.
+it is observed via WebDriver, so we avoid using this terminology.
+
+A [=user context=] has a <dfn>user context id</dfn>, which is a unique string
+set upon the user context creation.
+
+A [=navigable=] has an <dfn>associated user context</dfn>, which is a
+[=user context=].
+
+When a new [=/top-level traversable=] is created its [=associated user context=]
+is set to a user context in the [=set of user contexts=].
+
+Note: In some cases the user context is set by specification when the
+[=/top-level traversable=] is created, however in cases where no such
+requirements are present, the [=associated user context=] for a [=/top-level
+traversable=] is implemenation-defined.
+
+Issue: Should we specify that [=/top-level traversables=] with a non-null
+opener have the same [=associated user context=] as their opener?
+Need to check if this is something existing implementations enforce.
+
+A [=child navigable=]'s [=associated user context=] is it's
+[=navigable/parent=]'s [=associated user context=].
+
+A [=user context=] which isn't the [=associated user context=] for any
+[=/top-level traversable=] is an <dfn>empty user context</dfn>.
+
+The <dfn>default user context</dfn> is a [=user context=] with [=user context
+id=] <code>"default"</code>.
+
+An implementation has a <dfn>set of user contexts</dfn>, which is a [=/set=] of
+[=user contexts=]. Initially this contains the [=default user context=].
+
+Implementations may [=set/append=] new [=user contexts=] to the [=set of user
+contexts=] at any time, for example in response to user actions.
+
+Note: "At any time" here includes during implementation startup, so a given
+implementation might always have multiple entries in the [=set of user contexts=].
+
+Implementations may [=set/remove=] any [=empty user context=], with exception of
+the [=default user context=], from the [=set of user contexts=] at any
+time. However they are not required to remove such [=user contexts=]. [=User
+contexts=] that are not [=empty user contexts=] must not be removed from the
+[=set of user contexts=].
 
 # Modules # {#modules}
 
@@ -2032,7 +2066,9 @@ The <dfn export for=commands>browser.createUserContext</dfn> command creates a
 
 The [=remote end steps=] are:
 
-1. Let |user context| be a new [=empty user context=].
+1. Let |user context| be a new [=user context=].
+
+1. [=set/Append=] |user context| to the [=set of user contexts=].
 
 1. Let |result| be a [=/map=] matching the
    <code>browser.createUserContextResult</code> production with the
@@ -2072,22 +2108,26 @@ user context and all browsing contexts in it without running
 
 The [=remote end steps=] with |command parameters| are:
 
-1. Let |user context| be |command  parameters|["<code>userContext</code>"].
+1. Let |user context id| be |command  parameters|["<code>userContext</code>"].
 
-1. If |user context| is <code>"default"</code>, return [=error=] with [=error
+1. If |user context id| is <code>"default"</code>, return [=error=] with [=error
    code=] [=invalid argument=].
 
-1. If the [=user context=] with the [=user context id=] equal to |user context|
-   does not exist in the [=remote end=], return [=error=] with [=error code=] [=no
-   such user context=].
+1. For each |user context| in the [=set of user contexts=]:
 
-1. [=Close=] any [=top-level browsing contexts=] whose [=user context=]'s [=user
-   context id=] is |user context| without [=prompting to unload=].
+  1. If |user context|'s [=user context id=] is |user context id|:
 
-1. Perform implementation-defined steps to remove |user context| from the
-   [=remote end=].
+    1. For each [=/top-level traversable=] |navigable|:
 
-1. Return [=success=] with data null.
+      1. If |navigable|'s [=associated user context=] is |user context|:
+
+        1. [=Close=] |navigable| without [=prompting to unload=].
+
+    1. [=set/Remove=] |user context| for the [=set of user contexts=].
+
+    1. Return [=success=] with data null.
+
+1. Return [=error=] with [=error code=] [=no such user context=].
 
 </div>
 
@@ -2205,11 +2245,11 @@ To <dfn export>get a browsing context</dfn> given |context id|:
 browsingContext.InfoList = [*browsingContext.Info]
 
 browsingContext.Info = {
+  children: browsingContext.InfoList / null,
   context: browsingContext.BrowsingContext,
   url: text,
-  children: browsingContext.InfoList / null
+  userContext: browser.UserContext
   ? parent: browsingContext.BrowsingContext / null,
-  ? userContext: browser.UserContext
 }
 </pre>
 
@@ -2283,20 +2323,22 @@ To <dfn>get the browsing context info</dfn> given |context|,
 
   1. Set |child infos| to an empty [=/list=].
 
-  1. For each |context| of |child contexts|:
+  1. For each |child context| of |child contexts|:
 
     1. Let |info| be the result of [=get the browsing context info=] given
-       |context|, |child depth|, and false.
+       |child context|, |child depth|, and false.
 
     1. Append |info| to |child infos|
+
+1. Let |user context| be |context|'s [=associated user context=].
 
 1. Let |context info| be a [=/map=] matching the
    <code>browsingContext.Info</code> production with the <code>context</code>
    field set to |context id|, the <code>parent</code> field set to |parent id|
    if |is root| is <code>true</code>, or unset otherwise, the <code>url</code>
-   field set to |url|, the <code>userContext</code> field set to |context|'s
-   [=user context id=], and the <code>children</code> field set to |child
-   infos|.
+   field set to |url|, the <code>userContext</code> field set to
+   |user context|'s [=user context id=], and the <code>children</code> field
+   set to |child infos|.
 
 1. Return |context info|.
 
--- a/index.bs
+++ b/index.bs
@@ -1983,9 +1983,6 @@ The [=remote end steps=] with |session| and <var ignore>command parameters</var>
 
   1. [=Close=] any [=top-level browsing contexts=] without [=prompting to unload=].
 
-     Note: This implicitly only affects browsing contexts that were under
-     automation in the first place.
-
   1. Perform implementation defined steps to clean up resources associated with
      the [=remote end=] under automation.
 
@@ -2076,11 +2073,6 @@ The [=remote end steps=] with |command parameters| are:
 
 1. [=Close=] any [=top-level browsing contexts=] that belong to the |user
    context| without [=prompting to unload=].
-
-    Note: This implicitly only affects browsing contexts that were under
-    automation in the first place.
-
-    Issue: Should we allow closing the default context? Probably not.
 
 1. Perform implementation defined steps to remove the |user context| from the
    user agent.
--- a/index.bs
+++ b/index.bs
@@ -604,9 +604,6 @@ with the following additional codes:
   <dt><dfn>no such user context</dfn>
   <dd>Tried to reference an unknown [=user context=].
 
-  <dt><dfn>no such user context</dfn>
-  <dd>Tried to reference an unknown [=user context=].
-
   <dt><dfn>unable to close browser</dfn>
   <dd>Tried to close the browser, but failed to do so.
 
--- a/index.bs
+++ b/index.bs
@@ -1353,6 +1353,19 @@ time. However they are not required to remove such [=user contexts=]. [=User
 contexts=] that are not [=empty user contexts=] must not be removed from the
 [=set of user contexts=].
 
+<div algorithm>
+To <dfn>get user context</dfn> given |user context id|:
+
+1. For each |user context| in the [=set of user contexts=]:
+
+ 1. If |user context|'s [=user context id=] equals |user context id|:
+
+   1. Return |user context|.
+
+1. Return null.
+
+</div>
+
 # Modules # {#modules}
 
 ## The session Module ## {#module-session}
@@ -2157,21 +2170,19 @@ The [=remote end steps=] with |command parameters| are:
 1. If |user context id| is <code>"default"</code>, return [=error=] with [=error
    code=] [=invalid argument=].
 
-1. For each |user context| in the [=set of user contexts=]:
+1. Set |user context| to [=get user context=] with |user context id|.
 
-  1. If |user context|'s [=user context id=] is |user context id|:
+1. If |user context| is null, return [=error=] with [=error code=] [=no such user context=].
 
-    1. For each [=/top-level traversable=] |navigable|:
+1. For each [=/top-level traversable=] |navigable|:
 
-      1. If |navigable|'s [=associated user context=] is |user context|:
+  1. If |navigable|'s [=associated user context=] is |user context|:
 
-        1. [=Close=] |navigable| without [=prompting to unload=].
+    1. [=Close=] |navigable| without [=prompting to unload=].
 
-    1. [=set/Remove=] |user context| for the [=set of user contexts=].
+1. [=set/Remove=] |user context| for the [=set of user contexts=].
 
-    1. Return [=success=] with data null.
-
-1. Return [=error=] with [=error code=] [=no such user context=].
+1. Return [=success=] with data null.
 
 </div>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1804,6 +1804,96 @@ The [=remote end steps=] with |session| and <var ignore>command parameters</var>
...
+   </dd>
+</dl>
+
+<div algorithm="remote end steps for browser.createPartition">
+
+</div>
+
+#### The browser.closePartition Command #### {#command-browser-closePartition}
+
+The <dfn export for=commands>browser.closePartition</dfn> command closes the partition and all browsing contexts in it without running before unload handlers.
```
mathiasbynens, line 1837: For readability:

```suggestion
The <dfn export for=commands>browser.closePartition</dfn> command closes the partition and all browsing contexts in it without running <code>beforeunload</code> handlers.
```

index.bs
```diff
@@ -1804,6 +1804,96 @@ The [=remote end steps=] with |session| and <var ignore>command parameters</var>
 
 </div>
 
+#### The browser.createPartition Command #### {#command-browser-createPartition}
+
+The <dfn export for=commands>browser.createPartition</dfn> command creates a storage partition for a group of browsing contexts.
```
gsnedders, line 1809: We should absolutely cross-reference "storage partition".

That said, I don't think a storage partition is what we want, as I believe I said in the F2F, and as https://github.com/w3c/webdriver-bidi/issues/289#issuecomment-1721026858 kinda goes into.

We want something that encompasses the global partition (as is the case of ephemeral sessions and the like), whereas specs mostly use "storage partition" to refer to (e.g.) the storage addressable from example.com as the top-level site.

Based on that discussion, we want something more like "user agent" (https://infra.spec.whatwg.org/#user-agent), though the terminology does seem somewhat clumsy here, especially given how we already use "user agent" in WebDriver (Classic/BiDi).

That said, there is some variation between implementations (e.g., whether they share DNS cache results), which we may or may not want to overly specify in WebDriver.

cc @jyasskin, given I was speaking to him at TPAC about this

jyasskin: Yep, see https://github.com/privacycg/storage-partitioning and https://w3ctag.github.io/privacy-principles/#dfn-partition for definitions of "partition". [Browsing contexts](https://html.spec.whatwg.org/multipage/document-sequences.html#browsing-context) can navigate between partitions, and HTML is [recommending](https://html.spec.whatwg.org/multipage/document-sequences.html#infrastructure-for-sequences-of-documents) that specs should usually use [navigables](https://html.spec.whatwg.org/multipage/document-sequences.html#navigable), [traversibles](https://html.spec.whatwg.org/multipage/document-sequences.html#traversable-navigables), or [top-level traversibles](https://html.spec.whatwg.org/multipage/document-sequences.html#top-level-traversables) instead of browsing contexts.

With the "user agent" definition in Infra, I was trying to capture the instances of an implementation that are almost completely isolated from each other, so that existing specs that use "A user agent has a quuxinator map" correctly separate those maps. Searching in https://w3c.github.io/webdriver/, I don't see uses of "user agent" that conflict with the Infra definition, although I could be missing them, and some might be more precise as "implementation". For this command, I think the definition might be that it "creates a new [user agent](https://infra.spec.whatwg.org/#user-agent) for the same [implementation](https://infra.spec.whatwg.org/#implementation) as the current" ... "browser process"? "session"? ... ", and returns a new session inside that user agent"?

The current text isn't very clear about what implicit context there is or how you target requests to particular UAs/partitions/sessions. It seems like there's both an implicit browser process and an implicit session, and you can also create new explicit sessions, but nothing uses the `sessionId` returned when you create a new session... Adding profiles (/UAs/partitions) will need the other creation commands to be able to make their thing inside a particular one of those.

index.bs
```diff
@@ -1804,6 +1804,96 @@ The [=remote end steps=] with |session| and <var ignore>command parameters</var>
...
+   </dd>
+</dl>
+
+<div algorithm="remote end steps for browser.closePartition">
+
+</div>
+
+#### The browser.getPartitions Command #### {#command-browser-getPartitions}
+
+The <dfn export for=commands>browser.getPartitions</dfn> command returns a list of existing partitions in the browser instance.
```
jyasskin, line 1867: Do you want getPartitions() to be able to return other existing profiles that the user might have created, or only ones created by WebDriver commands? I see both "browser process" and "browser instance" here, but no definitions for them, which might clarify this.

index.bs
OrKoN: In this proposal, there is no session created for the partition: all partitions are managed by the same session. The advantage is that the same session can create a fresh state for its operations. In practical terms, we want to have something that would allows to implement https://pptr.dev/api/puppeteer.browsercontext/#example In Chrome, that corresponds to an incognito window (but you can have multiple for automation).

How exactly to name this is an open question.

OrKoN: I think specifying it as a creation of a new user agent might be what we need but at time same time we do not want to have a new session and a completely fresh browser process for this feature (if that is the case, I think it would be easier to start a new session instead). Thanks for the links and ideas, I will read up on the specs you linked and see how I can rephrase. And sorry if I missed the previous links to the discussion about it. 

> Adding profiles (/UAs/partitions) will need the other creation commands to be able to make their thing inside a particular one of those.

The current draft modifies the browsing context creation command to accept which partition the context should be created in. Afterwards, the session can address that context by using the context ID and can also determine which contexts are in which partitions based on the partition ID in `browsingContext.Info`.

index.bs
OrKoN: That part I have not clarified yet. I think it probably makes sense to only see partitions/profiles/UAs created by the WebDriver session.

index.bs
jyasskin: You have a good point that WebDriver needs to be more precise about what it means by "user agent" and "implementation" than most web specs. After all, WebDriver involves at least 2 different "user agents" and "implementations": the software that drives web browsers, and the web browsers themselves. Infra's definition is compatible with this: the key part of my change to that definition is that a single implementation can be configured to act as more than one UA. The application to browser profiles is just an example.

I do think you're going to confuse people by using "partition" to talk about browser profiles. Other web specs are tending toward "partitions" being aligned with top-level sites, and toward letting [some sorts of data](https://developer.mozilla.org/en-US/docs/Web/API/Document/requestStorageAccess) cross partition boundaries.

jgraham: Yes, I also agree that partition isn't a good fit; I used it above just to avoid introducing a new term.

the more I think about this, the more that I think the gecko-internal term of "user context" makes some sense; in that sense a single user agent (in the pre-infra sense) can be divided into multiple "user contexts" which represent a user-controlled partitioning of the browser state.

jyasskin: FYI, we're also defining "context" in https://w3ctag.github.io/privacy-principles/#context, but that's not exported, and I don't think it'll be used from enough places that you need to worry about using a different meaning in Web Driver.

index.bs
gsnedders: Historically implementations have differed here; it's possible to start GeckoDriver against a pre-existing profile, whereas SafariDriver always creates a temporary automation session which ins't persisted. I expect implementations will continue to differ, so it should maybe be implementation defined.

jgraham: I don't think "created by the WebDriver session" is the right line, but something like "which are visible to the WebDriver session" or something. Basically, if you can subscribe to events from a specific browsing context, you should be able to see the user context that's part of.

I think that's basically equivalent to making it implementation-defined, but I'm also not sure we need any special language for this feature; the constraint is just that the model has to be consistent, and it's already defacto possible for implementations to have parts of the browser that are not visible to WebDriver clients, but that can be understood at the session level.

index.bs
```diff
@@ -1219,6 +1219,15 @@ access to that data in a {{Window}} global.
 
 Issue: Define how this works.
 
+# User Contexts # {#user-contexts}
+
+A <dfn>user context</dfn> is an aggregation of browsing contexts that separates the browser state for browsing contexts in one user context from browsing contexts in another user context.
+The user agent contains at least one user context, the <dfn>default user context</dfn>. 
+
+Note: User contexts are a WebDriver-specific abstraction to allow working with private browsing modes (https://www.w3.org/2001/tag/doc/private-browsing-modes/).
```
jgraham, line 1227: This note seems inaccurate. User contexts may or may not map to private browsing (I wonder if we should consider a notion of user context flags, with "private" being one such flag? Certainly gecko containers _aren't_ like private browsing by default (in that they persist state to disk).

However we _should_ define that seperate user contexts don't share state (e.g. cookies, storage, etc.)

index.bs
```diff
@@ -1812,6 +1821,117 @@ The [=remote end steps=] with |session| and <var ignore>command parameters</var>
...
+
+The <dfn export for=commands>browser.createUserContext</dfn> command creates a [=user context=] in the user agent.
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+      <pre class="cddl remote-cddl">
+      browser.Close = (
+        method: "browser.createUserContext",
+        params: EmptyParams,
```
jgraham, line 1834: I wonder if we should allow the user to provide an optional name?

index.bs
```diff
@@ -1812,6 +1821,117 @@ The [=remote end steps=] with |session| and <var ignore>command parameters</var>
...
+
+The [=remote end steps=] are:
+
+1. Let |user contexts| be the result of implementation-specific steps to get [=user contexts=] from the user agent.
+
+1. Let |user context info list| be an empty [=/list=].
+
+1. For each |user context| in |user contexts|:
+
+  1. If the |user context| is the [=default user context=], continue.
```
jgraham, line 1920: Let's not skip the default user context here.

index.bs
```diff
@@ -2575,7 +2696,8 @@ The <dfn export for=commands>browsingContext.create</dfn> command creates a new
       browsingContext.CreateParameters = {
         type: browsingContext.CreateType,
         ? referenceContext: browsingContext.BrowsingContext,
-        ? background: bool .default false
+        ? background: bool .default false,
+        ? userContext: Browser.UserContext / null
```
jgraham, line 2700:
```suggestion
        ? userContext: browser.UserContext / null
```

index.bs
```diff
@@ -1923,6 +2043,7 @@ browsingContext.Info = {
   url: text,
   children: browsingContext.InfoList / null
   ? parent: browsingContext.BrowsingContext / null,
+  ? userContext: Browser.UserContext / null
```
jgraham, line 2046:
```suggestion
  ? userContext: browser.UserContext / null
```

index.bs
OrKoN: We currently don't support any names. Does Firefox support names?

index.bs
OrKoN: Updated based on your suggestion on chat!

index.bs
jgraham: Firefox containers are named, so I'd expect us to at least return the names for any existing containers. Since (I think) the name has to be unique, we could provide the name as the id. Of course that doesn't work if we enforce UUID (or non-string) ids, or put any other restrictions on the ids.

We don't necessarily have to support setting names, but if we don't (or if we make the names something other than human readable strings), I think gecko would likely end up extending the API to provide some way to communicate how the user contexts map onto containers.

index.bs
```diff
@@ -2688,13 +2876,23 @@ The [=remote end steps=] with |command parameters| are:
...
 
+  1. Let |user context| be the [=default user context=].
+  
+  1. If |command parameters|[<code>userContext</code>] is present and is not null:
+
+    1. Let |user context| be |command parameters|[<code>userContext</code>].
+
+    1. If the [=user context=] with the [=user context id=] equal to |user context| does 
+      not exist in the user agent, return [=error=] with [=error code=] [=invalid argument=].
+
```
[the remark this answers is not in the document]
OrKoN: Thanks, I think the updated version is better.

index.bs
```diff
@@ -1247,6 +1247,30 @@ access to that data in a {{Window}} global.
 
 Issue: Define how this works.
 
+# User Contexts # {#user-contexts}
+
+A <dfn>user context</dfn> represents a collection of top-level navigables within a remote
+end. Each top-level navigable belongs to exactly one user context, and child
```
OrKoN, line 1253:
```suggestion
A <dfn>user context</dfn> represents a collection of top-level navigables within a 
[=remote end=]. Each top-level navigable belongs to exactly one user context, and child
```

index.bs
```diff
@@ -1247,6 +1247,30 @@ access to that data in a {{Window}} global.
...
+navigables belong to the same user context as their parent. A [=user context=] without
+navigables belonging to it is an <dfn>empty user context</dfn>. 
+
+The remote end always contains one [=user context=] referred as the <dfn>default user
+context</dfn>.
+
+Issue: Should we have a default user context?
```
OrKoN, line 1260:
```suggestion
```

index.bs
```diff
@@ -1247,6 +1247,30 @@ access to that data in a {{Window}} global.
...
 Issue: Define how this works.
 
+# User Contexts # {#user-contexts}
+
+A <dfn>user context</dfn> represents a collection of top-level navigables within a remote
+end. Each top-level navigable belongs to exactly one user context, and child
+navigables belong to the same user context as their parent. A [=user context=] without
+navigables belonging to it is an <dfn>empty user context</dfn>. 
+
+The remote end always contains one [=user context=] referred as the <dfn>default user
```
OrKoN, line 1257:
```suggestion
A [=remote end=] always contains one [=user context=] referred as the <dfn>default user
```

index.bs
```diff
@@ -1840,6 +1887,141 @@ The [=remote end steps=] with |session| and <var ignore>command parameters</var>
 
 </div>
 
+#### The browser.createUserContext Command #### {#command-browser-createUserContext}
+
+The <dfn export for=commands>browser.createUserContext</dfn> command creates a [=user context=] in the user agent.
```
OrKoN, line 1892:
```suggestion
The <dfn export for=commands>browser.createUserContext</dfn> command creates a [=user context=].
```

index.bs
```diff
@@ -1840,6 +1887,141 @@ The [=remote end steps=] with |session| and <var ignore>command parameters</var>
...
+</dl>
+
+<div algorithm="remote end steps for browser.closeUserContext">
+
+The [=remote end steps=] with |command parameters| are:
+
+1. Let |user context| be |command  parameters|["<code>userContext</code>"].
+
+1. If the [=user context=] with the [=user context id=] equal to |user context|
+   does not exist in the user agent, return [=error=] with [=error code=] [=invalid
```
OrKoN, line 1959:
```suggestion
   does not exist in the [=remote end=], return [=error=] with [=error code=] [=invalid
```

index.bs
```diff
@@ -1840,6 +1887,141 @@ The [=remote end steps=] with |session| and <var ignore>command parameters</var>
...
+1. Perform implementation defined steps to remove the |user context| from the
+   user agent.
+
+1. Return [=success=] with data null.
+
+</div>
+
+#### The browser.getUserContexts Command #### {#command-browser-getUserContexts}
+
+The <dfn export for=commands>browser.getUserContexts</dfn> command returns a list of existing user contexts in the user agent.
```
OrKoN, line 1979:
```suggestion
The <dfn export for=commands>browser.getUserContexts</dfn> command returns a list of existing user contexts in a [=remote end=].
```

index.bs
```diff
@@ -1229,6 +1229,29 @@ access to that data in a {{Window}} global.
...
+
+A [=remote end=] always contains one [=user context=] referred as the <dfn>default user
+context</dfn>.
+
+Each user context represents an additional [=storage key=], so that no storage
+state is shared between different user contexts.
+
+User contexts other than the [=default user context=] have a <dfn>user context
+id</dfn>, which is a string set upon context creation. The [=default user
+context=]'s does not have a [=user context id=].
```
sadym-chromium, line 1247: What about `null` for default user context id?

OrKoN: I think it might be easier to treat the default context id as undefined.

OrKoN: Open to change that.

index.bs
```diff
@@ -2004,6 +2185,7 @@ browsingContext.Info = {
   url: text,
   children: browsingContext.InfoList / null
   ? parent: browsingContext.BrowsingContext / null,
+  ? userContext: browser.UserContext / null
```
sadym-chromium, line 2188: I don't think we use `null` at all. It's either missing of some string.
```suggestion
  ? userContext: browser.UserContext
```

index.bs
```diff
@@ -2067,6 +2249,10 @@ To <dfn>get the browsing context info</dfn> given |context|,
 
    Note: This includes the fragment component of the URL.
 
+1. Let |user context| be undefined if |context| belongs to the [=default user
```
sadym-chromium, line 2252:
```suggestion
1. Let |user context| be null if |context| belongs to the [=default user
```

index.bs
sadym-chromium: In `browsingContext.create` you treat `null` as default user context as well and I'd prefer keep using `null` for it. Not a strong opinion though.

index.bs
```diff
@@ -1889,6 +1935,141 @@ The [=remote end steps=] with |session| and <var ignore>command parameters</var>
...
+      }
+    </pre>
+   </dd>
+</dl>
+
+<div algorithm="remote end steps for browser.createUserContext">
+
+The [=remote end steps=] are:
+
+1. Let |user context| be the result of performing implementation defined steps to create an [=empty user context=].
```
jgraham, line 1966: I think you can just say `Let |user context| be the result of creating an [=empty user context=]`, or something like that. At some level this clearly must involve implementation-specific processing, but I don't think we need to call it out in this case.

index.bs
```diff
@@ -1889,6 +1935,141 @@ The [=remote end steps=] with |session| and <var ignore>command parameters</var>
...
+1. Let |user context| be |command  parameters|["<code>userContext</code>"].
+
+1. If the [=user context=] with the [=user context id=] equal to |user context|
+   does not exist in the user agent, return [=error=] with [=error code=] [=invalid
+   argument=]
+
+1. [=Close=] any [=top-level browsing contexts=] that belong to the |user
+   context| without [=prompting to unload=].
+
+    Note: This implicitly only affects browsing contexts that were under
```
jgraham, line 2013: Not if you can have user contexts created outside of automation?

index.bs
```diff
@@ -1889,6 +1935,141 @@ The [=remote end steps=] with |session| and <var ignore>command parameters</var>
...
+    automation in the first place.
+
+    Issue: Should we allow closing the default context? Probably not.
```
jgraham, line 2016: I think it's fine to close it, but it probably becomes an empty user context at that point.

index.bs
```diff
@@ -1889,6 +1935,141 @@ The [=remote end steps=] with |session| and <var ignore>command parameters</var>
...
+
+1. Perform implementation defined steps to remove the |user context| from the
```
jgraham, line 2018: I think this would be easier if we just said that an implementation has a set of active user contexts, which we add to on creation, or remove from on close, rather than throwing in "implementation-defined" everywhere.

jgraham (changes requested):

index.bs
```diff
@@ -1889,6 +1935,141 @@ The [=remote end steps=] with |session| and <var ignore>command parameters</var>
...
+   <code>browser.createUserContextResult</code> production with the <code>userContext</code>
+   field set to |user context|'s [=user context id=].
+
+1. Return [=success=] with data |result|.
+
+</div>
+
+#### The browser.closeUserContext Command #### {#command-browser-closeUserContext}
+
+The <dfn export for=commands>browser.closeUserContext</dfn> command closes a user context and all browsing contexts in it without running <code>beforeunload</code> handlers.
```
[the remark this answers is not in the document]
jrandolf: IMO, `close` implies gone. I associate `remove` with `detach`.

index.bs
jrandolf: > Or we could just give it a string name, like "default"? That makes the types simpler.

I think for consistency, I agree with @sadym-chromium. Moreover, `null`s are just `Optional<string>` in most languages, so I'm not sure if this makes types truly simpler.

index.bs
jrandolf: I think being explicit here is important. For example, I just started looking at this PR and if I didn't see "implementation-defined", I'd be questioning how to create an [=empty user context=] to matrix :P

index.bs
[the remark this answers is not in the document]
jrandolf: Although, considering the opposite of `create` in the context of the web is `remove`, I'm inclined to use `remove` as well. I've made the change.

index.bs
jrandolf: Indeed, I think we can safely remove this statement since we always know about all browsing contexts.

jrandolf: I've removed the sentence here and above (in the `browser.close` command).

index.bs
jrandolf: +1.

I think whether or not the implementation really needs to close the default context at this point is not necessary to flesh out.

In particular, for Chromium, we can just get away with emitting the appropriate events and maintaining the default context until the user context is closed.

jgraham (changes requested):
Generally I'd like us to at least to try to define a more formal model for how user contexts work, rather than just handwaving "implementation-defined" a lot. I don't think a basic model is actually that complicated, so I think we should try to at least get the scaffolding in place in this PR.

index.bs
```diff
@@ -1229,6 +1229,30 @@ access to that data in a {{Window}} global.
...
+parent. A [=user context=] without navigables belonging to it is an <dfn>empty
+user context</dfn>.
+
+A [=remote end=] always contains at least one [=user context=] referred as the
+<dfn>default user context</dfn>.
+
+Each [=user context=] represents an additional [=storage key=], so that no
+storage state is shared between different [=user context|user contexts=].
+
+A [=user context=] other than the [=default user context=] has a <dfn>user
```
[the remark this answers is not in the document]
OrKoN: Would something like that work?

```
User contexts have a <dfn>user context id</dfn>, which is a string set upon
context creation. The [=default user context=]'s [=user context
id=] is <code>"default"</code>.
```

index.bs
```diff
@@ -1889,6 +1934,140 @@ The [=remote end steps=] with |session| and <var ignore>command parameters</var>
...
+      }
+    </pre>
+   </dd>
+</dl>
+
+<div algorithm="remote end steps for browser.createUserContext">
+
+The [=remote end steps=] are:
+
+1. Let |user context| be the result of performing implementation-defined steps
```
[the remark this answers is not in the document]
OrKoN: removed "implementation-defined steps" but I am not sure if we need the "user context set"?

> And then we should say that each top-level navigable belongs to a specific user context, which is set when it's created

We already have a similar definition in the User Contexts section and it looks to be sufficient? PTAL

OrKoN:
Looks like I have messed up the branch a bit. Please do not review changes for now.

index.bs
OrKoN: > But this could be implemented entirely within the session, right?

Would that mean that other session won't see the names given to contexts by other sessions?

index.bs
```diff
@@ -1889,6 +1934,140 @@ The [=remote end steps=] with |session| and <var ignore>command parameters</var>
...
+
+The [=remote end steps=] with |command parameters| are:
+
+1. Let |user context| be |command  parameters|["<code>userContext</code>"].
+
+1. If the [=user context=] with the [=user context id=] equal to |user context|
+   does not exist in the user agent, return [=error=] with [=error code=]
+   [=invalid argument=]
+
+1. [=Close=] any [=top-level browsing contexts=] that belong to the |user
```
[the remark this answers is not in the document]
OrKoN: I have introduced the user context as a property of a navigable.

index.bs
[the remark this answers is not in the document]
OrKoN: Added `default` as the ID of the default context.

OrKoN:
I have recovered lost changes and the PR is now ready for review.

index.bs
jgraham: Yes, it's true that it depends on what guarantees we want to put in place here. If different sessions should get the same user context id (which would be the default in Firefox, and would make sense if you assume that user contexts can also be created outside of WebDriver) then we need it to be a property of the implementation. If different sessions should get the different user context ids (which might be easier to implement in Chrome) then the ids can be properties of the session alone.

In practice I think it would be trivial for different sessions to map user context ids, especially if browsing context ids are shared (which I think is an open discussion, and which worries me a bit, but which the spec currently requires).

In Firefox, given that the underlying container name is exposed in the UI, I'd be very surprised if we don't provide _some_ way to get at that name, even if we also layer on a per-session id on top of that.

OrKoN: I think we are aiming at "different sessions should get the same user context id" as it's also how it's implemented in Chrome. So I think it's the same in both browsers. With the names it'd different because the WebDriver BiDi implementation in Chrome is not part of the browser. And theoretically multiple BiDi implementation instances could talk to the same Chrome instance and they will only have access to the data that lives in the Chrome instance and will have different session data. 

Do user context names in Firefox have to be unique? Do you also have a unique ID or do you plan to use the name as the ID?

jgraham: Based on experiments, the container name in Firefox doesn't have to be unique, so I guess there's also an underlying id that's different.

whimboo: Yes, [each context has its own context id](https://searchfox.org/mozilla-central/rev/77aeaa184d59944ecb117e452ea43f5bd1556e4c/toolkit/components/contextualidentity/ContextualIdentityService.sys.mjs#264).

index.bs
```diff
@@ -1889,6 +1933,91 @@ The [=remote end steps=] with |session| and <var ignore>command parameters</var>
...
+
+<div algorithm="remote end steps for browser.removeUserContext">
+
+The [=remote end steps=] with |command parameters| are:
+
+1. Let |user context| be |command  parameters|["<code>userContext</code>"].
+
+1. If the [=user context=] with the [=user context id=] equal to |user context|
+   does not exist in the [=remote end=], return [=error=] with [=error code=]
+   [=invalid argument=]
```
juliandescottes, line 2009: Should we consider a specific error (eg `unknown user context`) ? Asking this because of the discussion at https://github.com/web-platform-tests/wpt/issues/43534#issuecomment-1859711557 , in case we prefer to keep `invalid argument` for errors which can be caught entirely by schema validation.

index.bs
```diff
@@ -2737,13 +2870,26 @@ The [=remote end steps=] with |command parameters| are:
...
   <!-- This is based on step 5 of https://w3c.github.io/webdriver/#new-window,
        but without using the "current browsing context" concept. -->
   1. Create a new [=top-level browsing context=] by running the [=window open
-     steps=] with <var ignore>url</var> set to "<code>about:blank</code>",
-     <var ignore>target</var> set to the empty string, and
-     <var ignore>features</var> set to "<code>noopener</code>". Which OS window the new [=/browsing context=]
-     is created in depends on |type| and |reference context|:
+     steps=] in the |user context| with <var ignore>url</var> set to
+     "<code>about:blank</code>", <var ignore>target</var> set to the empty
+     string, and <var ignore>features</var> set to "<code>noopener</code>".
```
juliandescottes, line 2890: For `type=window`, I think we agree that the `user_context` provided in `browsingContext.create` will only impact the initial tab of this window. If additional tabs are created in the same window with `browsingContext.create` + `type=tab`, they will not be assigned by default to the user context used to create the window.

index.bs
OrKoN: Makes sense. I guess `no such user context` would be more aligned with existing errors?

OrKoN: Added `no such user context`

index.bs
OrKoN: Good point, I think we can and should validate that explicitly. I have added a validation step.

OrKoN: Or let me think about it, maybe it's simpler to change the default.

OrKoN: I have ended up doing both improvements. PTAL

index.bs
```diff
@@ -1950,6 +2002,91 @@ The [=remote end steps=] with |session| and <var ignore>command parameters</var>
...
+
+<div algorithm="remote end steps for browser.removeUserContext">
+
+The [=remote end steps=] with |command parameters| are:
+
+1. Let |user context| be |command  parameters|["<code>userContext</code>"].
+
+1. If the [=user context=] with the [=user context id=] equal to |user context|
+   does not exist in the [=remote end=], return [=error=] with [=error code=] [=no
+   such user context=].
```
juliandescottes, line 2078: What happens in case the user tries to remove the "default" user context. Should we allow that, or should we restrict this command to only work with non-default user contexts?

For reference, in Firefox trying to delete the default user context doesn't do anything.

OrKoN: Good point, I think we probably want to prevent that given that there could be differences in implementations that are not worth unifying. I think actually in Chrome it would not be possible too.

index.bs
```diff
@@ -2801,13 +2945,29 @@ The [=remote end steps=] with |command parameters| are:
...
+    1. Let |user context id| be |command parameters|[<code>userContext</code>].
+
+    1. If the [=user context=] with the [=user context id=] equal to |user context id| does
+      not exist in the [=remote end=], return [=error=] with [=error code=] [=no such user context=].
+
+    1. Set |user context| to the [=user context=] with the [=user context id=] equal
+      to |user context id|.
+
+    1. If |reference context| is not null and |reference context|'s [=user context=] is not |user context|,
+      return [=error=] with [=error code=] [=invalid user context id=].
```
juliandescottes, line 2961: In Firefox, we can open have tabs from several user contexts (containers) next to each other in the same window. I imagine it's fine to enforce this limitation if other vendors don't have this flexibility. 

Does that mean that in Chrome, if you have 1 window with 1 tab in the default context and you use `browsingContext.create` with type=tab, it will force creating a new window on the fly? In that case it's not compliant with the following step of the type=tab section:

> The new browsing context should reuse an existing OS window, if any.

So I'm wondering if we really need to be strict and fail here. We could modify the text of the "if type is tab" section below to say we will "attempt" to reuse an existing window (containing the reference context, if relevant), but it's not guaranteed.

cc @jgraham

OrKoN: > In Firefox, we can open have tabs from several user contexts (containers) next to each other in the same window. I imagine it's fine to enforce this limitation if other vendors don't have this flexibility.

Yes, that is not supported with Chrome.

> Does that mean that in Chrome, if you have 1 window with 1 tab in the default context and you use browsingContext.create with type=tab, it will force creating a new window on the fly? In that case it's not compliant with the following step of the type=tab section:

> The new browsing context should reuse an existing OS window, if any.

Yes, it will create a tab in a new window. I think "should" allows for that.

juliandescottes: Ah maybe, I thought the "should" was just there to cover the case where there is no existing window ("if any"). But maybe this also covers the cases where existing windows are not compatible with the tab we are trying to add.

The spec about the reference context uses the same phrasing with "should":

> If reference context is not null, the new browsing context should reuse the window containing reference context, if any.

So if the "window containing the reference context" is not compatible with the tab we are trying to add, then it should be fine to add the tab in another window? 

Again, not sure we should throw here, if the spec is flexible enough to allow to pick another window if needed?

Since nothing in Firefox prevents from having tabs from several containers in the same window, it seems a bit arbitrary to throw. Imagine we have 1 window with 1 tab "tabA" owned by "contextA". If someone uses "browsingContext.create(type=tab, userContext=contextB)" with no reference context, Firefox will add this tab next to tabA. But if the user passes referenceContext=tabA, we would throw because contexts don't match?

If we really want to have a consistent behavior across all browsers, I think we need to enforce that all tabs in a window all have the same user context, but that seems difficult/impossible to implement.

jgraham: "should" in specs is defined as "unless there's a valid reason not to". To me the "should reuse the window containing the reference context" is fine; it more or less macro expands to "shall reuse the window containing the reference context, unless there's a valid reason not to", and "we don't support multiple user contexts in the same window" seems like potentially a valid reason.

I also agree that Firefox in general can't enforce the condition that each OS window only gets tabs with a single user context and therefore we can't in WebDriver. For example we might connect to a browser where the user already created multiple containers in the same window.

Given that implementation differences seem impossible to avoid here, I think it might be OK to just allow whatever behaviour the implementation does? In practice clients can always request a new window when creating a tab in a different context, so for test automation purposes it's possible to build a stable, interoperable, foundation despite the underlying differences.

index.bs
```diff
@@ -2802,13 +3038,29 @@ The [=remote end steps=] with |command parameters| are:
...
+
+    1. Let |user context id| be |command parameters|[<code>userContext</code>].
+
+    1. If the [=user context=] with the [=user context id=] equal to |user context id| does
+      not exist in the [=remote end=], return [=error=] with [=error code=] [=no such user context=].
+
+    1. Set |user context| to the [=user context=] with the [=user context id=] equal
+      to |user context id|.
+
+    1. If |reference context| is not null and |reference context|'s [=user context=] is not |user context|,
```
jgraham, line 3053: I think we should have something more generic e.g. moving the first step down and saying something like:

```
1. If the implementation is unable to create a new [=top-level traversable=] with [=associated user context=] |user context| for any reason, then return [=error=] with [=error code=] [=unsupported operation=].
```

That makes the error for the specific case of "can't create a user context in the wrong window in Chrome" a generic "unsupported operation" error rather than "invalid user context".

jgraham (changes requested):

index.bs
OrKoN: As I understand we leave out the names for now and might add it eventually?

index.bs
```diff
@@ -1951,6 +2034,149 @@ The [=remote end steps=] with |session| and <var ignore>command parameters</var>
...
+        params: EmptyParams,
+      )
+      </pre>
+   </dd>
+   <dt>Return Type</dt>
+   <dd>
+    <pre class="cddl local-cddl">
+      browser.createUserContextResult = {
+        userContext: browser.UserContext
+      }
```
whimboo, line 2057: Shouldn't it actually re-use the already existent type definition?

```suggestion
      browser.createUserContextResult = (
        browser.UserContextInfo
      )
```

OrKoN: Good point, updated the struct and also the text.

index.bs
```diff
@@ -3038,29 +3032,31 @@ The [=remote end steps=] with |command parameters| are:
...
 
     1. Let |user context id| be |command parameters|[<code>userContext</code>].
 
-    1. If the [=user context=] with the [=user context id=] equal to |user context id| does
+    1. If the [=user context=] with the [=user context id=] |user context id| does
       not exist in the [=remote end=], return [=error=] with [=error code=] [=no such user context=].
 
-    1. Set |user context| to the [=user context=] with the [=user context id=] equal
-      to |user context id|.
+    1. Set |user context| to the [=user context=] with the [=user context id=] |user context id|.
```
jgraham, line 3045: I suggest rewriting this as:

```
1. Let |found context| be false.

1. For each |target user context| in the [=set of user contexts=]:

  1. If |target context|'s [=user context id=] equals |command parameters|["<code>userContext</code>"]:

    1. Set |user context| to |target context|.
    
    1. Let |found context| be true.
    
    1. Break.
 
 1. If |found context| is false, return [=error=] with [=error code=] [=no such user context=].
 ```
 
 We could also break this out into a "get user context" algorithm that returns null if there's no context, like we do for browsing contexts.

jgraham (approved):
This looks good, thanks!

I added a suggestion for making the lookup of user context by id more explicit, but I'm happy to approve the PR since I think basically all the behaviour makes sense.

juliandescottes (approved):

index.bs
OrKoN: I have extracted the get user context algorithm

jgraham (approved):
Thanks!

PR #571, opened 2023-10-12
Differentiate unspecified and null on browsingContext.SetViewport

jrandolf:
This PR also removes some extraneous wording related to unsupported operations.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/571.html" title="Last updated on Oct 12, 2023, 1:56 PM UTC (81a5e99)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/571/a3a09be...81a5e99.html" title="Last updated on Oct 12, 2023, 1:56 PM UTC (81a5e99)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3068,8 +3068,8 @@ The <dfn export for=commands>browsingContext.setViewport</dfn> command  specific
 
       browsingContext.SetViewportParameters = {
         context: browsingContext.BrowsingContext,
-        viewport: browsingContext.Viewport / null,
-        devicePixelRatio: (float .gt 0.0) / null,
+        ? viewport: browsingContext.Viewport / null,
+        ? devicePixelRatio: (float .gt 0.0) / null,
       }
 
       browsingContext.Viewport = {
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3100,22 +3100,31 @@ The [=remote end steps=] with |command parameters| are:
 
 1. Let |device pixel ratio| be the |command parameters|["<code>devicePixelRatio</code>"] field.
 
-1. If |viewport| is not null, and the implementation is unable to set the width and height dimensions
-   of the [=layout viewport=] to |viewport|["width"] and |viewport|["height"], respectively, or if 
-   |device pixel ratio| is not null and the implementation is unable to set the size of the CSS pixel 
-   of the [=layout viewport=] to |device pixel ratio| device pixels, or if the implementation is unable
-   to adjust the [=layout viewport=] parameters for any other reason, return [=error=] 
-   with [=error code=] [=unsupported operation=].
+1. If the implementation is unable to set the [=layout viewport=] for any
```
OrKoN, line 3103: I think we want to be specific here about the operations related to the viewport. Setting the viewport is not precise enough.

OrKoN: we also only need to throw an error if the attribute is actually specified. Let's revert to the previous definition?

jrandolf: How about "If the implementation is unable to perform any operations defined in this algorithm for any reason, ..."?

jrandolf: I've phrased it to be the third part of the previous definition. I think being so specific is a bit redundant since the only two operations are setting dpr and setting viewport.

jrandolf: I've separated the previous warning and moved them into their corresponding locations.

index.bs
```diff
@@ -3068,8 +3068,8 @@ The <dfn export for=commands>browsingContext.setViewport</dfn> command  specific
 
       browsingContext.SetViewportParameters = {
         context: browsingContext.BrowsingContext,
-        viewport: browsingContext.Viewport / null,
-        devicePixelRatio: (float .gt 0.0) / null,
+        ? viewport: browsingContext.Viewport / false,
```
OrKoN, line 3071: I prefer that we use null to reset emulation (to be consistent with the rest of the spec).

OrKoN: also, changing to false is a breaking change

jrandolf: @jgraham Because we use null for `?` as well, WDYT? I think semantically `false` implies "turn of emulation", so it makes sense as well. I don't have a strong opinion here, but we would need to use different wording for optional and null state then.

jgraham: We distinguish missing and null elsewhere in the spec, so if you're not able to make that distinction you're going to have a bad time (and we need to have a broader conversation about the API conventions).

index.bs
```diff
@@ -3090,32 +3090,46 @@ The <dfn export for=commands>browsingContext.setViewport</dfn> command  specific
...
 
-1. Let |device pixel ratio| be the |command parameters|["<code>devicePixelRatio</code>"] field.
-
-1. If |viewport| is not null, and the implementation is unable to set the width and height dimensions
-   of the [=layout viewport=] to |viewport|["width"] and |viewport|["height"], respectively, or if 
-   |device pixel ratio| is not null and the implementation is unable to set the size of the CSS pixel 
-   of the [=layout viewport=] to |device pixel ratio| device pixels, or if the implementation is unable
-   to adjust the [=layout viewport=] parameters for any other reason, return [=error=] 
-   with [=error code=] [=unsupported operation=].
+1. Let |device pixel ratio| be the value of the <code>devicePixelRatio</code>
```
jgraham, line 3104: So, I prefer using the `Let |field| be |command parameters|["<code>field</code>"]` style. 

But also since it can be missing and map lookups are supposed to be infallible, you now need to first check if `|command parameters| [=map/contains=] "<code>field</code>"`.

index.bs
```diff
@@ -3090,32 +3090,46 @@ The <dfn export for=commands>browsingContext.setViewport</dfn> command  specific
...
-    1. Set the [=layout viewport=] of the |context| to be |viewport|["width"] CSS pixels wide and |viewport|["height"] CSS pixels high.
-1. Otherwise:
-    1. Set the [=layout viewport=] of the |context| to the implementation defined default.
+
+   1. If |viewport| is not false, set the width of |context|'s [=layout
+      viewport=] to be the <code>width</code> of |viewport| in CSS pixels and
+      set the height of the |context|'s [=layout viewport=] to be the
+      <code>height</code> of |viewport| in CSS pixels.
+
+      1. If the implementation is unable to perform the previous action for any
```
jgraham, line 3114: This seems to regress the earlier change that moved this checks to the start, so that the overall command is atomic.

jgraham (changes requested):

index.bs
jrandolf: Fixed.

index.bs
jrandolf: Done, however I think using field accessor notation goes counter to the previous thought that we don't want to lock into a language (like JavaScript). Fields can be accessed in other ways.

jgraham (approved):

index.bs
jrandolf: Hmm, interesting. TIL.

OrKoN (approved):

whimboo:
@jrandolf I assume that you are going to update the wpt tests?

PR #583, opened 2023-10-27
editorial: fix Script.InternalId type

Lightning00Blade:
We always define Ids as text, this is an exception, we should change this.
Note: Both Chromium and Firefox return strings atm

***
<a href="https://pr-preview.s3.amazonaws.com/Lightning00Blade/webdriver-bidi/pull/583.html" title="Last updated on Nov 6, 2023, 3:47 PM UTC (8068535)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/583/a7adbf0...Lightning00Blade:8068535.html" title="Last updated on Nov 6, 2023, 3:47 PM UTC (8068535)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7149,7 +7149,7 @@ script.RemoteValue = (
   script.WindowProxyRemoteValue
 )
 
-script.InternalId = js-uint;
+script.InternalId = text;
 
 script.ListRemoteValue = [*script.RemoteValue];
 
--- a/index.bs
+++ b/index.bs
@@ -6402,6 +6402,19 @@ the ECMAScript runtime. The handle is only valid in a specific [=Realm=].
 Each ECMAScript [=Realm=] has a corresponding <dfn>handle object map</dfn>. This is a
 strong [=/map=] from handle ids to their corresponding objects.
 
+#### The script.InternalId Type #### {#type-script-InternalId}
+
+[=Remote end definition=] and [=local end definition=]
+
+<pre class="cddl remote-cddl local-cddl">
+script.InternalId = text;
+</pre>
+
+The <code>script.InternalId</code> type represents the id of 
+a previously serialized <code>script.RemoteValue</code> during
+[=serialize as a remote value|serialization=].
+
+
 #### The script.LocalValue Type #### {#type-script-LocalValue}
 
 [=Remote end definition=]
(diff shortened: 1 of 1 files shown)


OrKoN (approved):

jgraham (changes requested):
It's defined as a unique integer in the prose, so on its own this change seems wrong (and suggests any change here requires more justification than just "editorial")

jgraham:
"Firefox and Chrome return strings" seems like a good justification BTW. But do they return strings that represent integers? If so it seems unfortunate because people will probably do `parseInt` on those strings and we might need to specify that it's the string representation of an integer.

jgraham:
If we're returning a UUID already that somewhat alleviates the concern (we can just change the prose to say "globally unique string"). My worry is that in cases where we just say "return a string" and everyone is returning (say) an integer then clients depend on it being an integer. 

I still worry a bit that we'll find cases where any client returning a string in a known, but unspecified, format will cause authors to depend on that format (e.g. Chrome returning the string representation of an integer will cause people to write tests that depend on the string representing an integer), but so far I don't think we have consensus that all ids should be UUIDs to avoid this problem.

sadym-chromium:
We are fine with making `internalId` a `UUID`.

sadym-chromium:
> `Let internal id be a unique across the internalId fields of the values of serialization internal map integer.`

Let internal id be the string representation of a UUID based on truly random, or pseudo-random numbers.

index.bs
```diff
@@ -7149,7 +7149,7 @@ script.RemoteValue = (
   script.WindowProxyRemoteValue
 )
 
-script.InternalId = js-uint;
+script.InternalId = text;
```
Lightning00Blade, line 7152: One question that I have should this definition have a separate block in Types similar to other IDs (https://w3c.github.io/webdriver-bidi/#type-script-SharedId or https://w3c.github.io/webdriver-bidi/#type-browsingContext-Browsingcontext) ?

sadym-chromium (approved):

index.bs
whimboo: This might probably be a good idea.

Lightning00Blade: Done.

sadym-chromium (approved):

whimboo (approved):
Looks good to me as well now. Thanks!

whimboo:
@jgraham can you please update your review status? Merging is blocked because you requested changes. Thanks!

PR #584, opened 2023-10-30
Add a browsingContext.traverseHistory command

jgraham:
This traverses the history by a specified `delta` (e.g. +1 for forward, or -1 for back).

For now there is no `wait` parameter; the command always returns as soon as the hsitory traversal has been enqueued (this should eventually be equivalent to setting wait to `none`).

This model differs somewhat from the CDP model where one navigates to an explicit history entry and failure is only possible if the entry id is invalid. But back/forwward only seems closer to the use cases we have, and allowing events to be traced to a specific traversal seems consistent with the way we handle other navigation-related events.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/584.html" title="Last updated on Oct 31, 2023, 1:42 PM UTC (e8fa74a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/584/38ba606...e8fa74a.html" title="Last updated on Oct 31, 2023, 1:42 PM UTC (e8fa74a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3198,7 +3198,7 @@ The [=remote end steps=] with |command parameters| are:
     1. Let |target index| be |current index| plus |delta|.
 
     1. Let |valid entry| be false if |all steps|[|target index|] does not exist,
-       or false otherwise.
+       or true otherwise.
 
     1. [=Resume=] with "<code>check history</code>", |resume id|, and
        |valid entry|.
--- a/index.bs
+++ b/index.bs
@@ -174,6 +174,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
     text: default view; url: nav-history-apis.html#dom-document-defaultview
     text: environment settings object's Realm; url: webappapis.html#environment-settings-object's-realm
     text: focused area of the document; url: document-sequences.html#focused-area-of-the-document
+    text: getting all used history steps; url:browsing-the-web.html#getting-all-used-history-steps
     text: handled; url: webappapis.html#concept-error-handled
     text: hidden; url: document-sequences.html#system-visibility-state
     text: history handling behavior; url: browsing-the-web.html#history-handling-behavior
@@ -188,6 +189,8 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
     text: run the animation frame callbacks; url: imagebitmap-and-animations.html#run-the-animation-frame-callbacks
     text: same origin domain; url: browsers.html#same-origin-domain
     text: session history; url: history.html#session-history
+    text: session history entry; url: browsing-the-web.html#session-history-entry
+    text: session history traversal queue; url: document-sequences.html#tn-session-history-traversal-queue
     text: set up a window environment settings object; url: window-object.html#set-up-a-window-environment-settings-object
     text: set up a worker environment settings object; url: workers.html#set-up-a-worker-environment-settings-object
     text: set up a worklet environment settings object; url: worklets.html#set-up-a-worklet-environment-settings-object
@@ -195,6 +198,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
     text: system visibility state; url: document-sequences.html#system-visibility-state
     text: traversable navigable; url:document-sequences.html#traversable-navigable
     text: visible; url: document-sequences.html#system-visibility-state
+    text: traverse the history by a delta; url: browsing-the-web.html#traverse-the-history-by-a-delta
     text: window open steps; url: window-object.html#window-open-steps
     text: worker event loop; url: webappapis.html#worker-event-loop-2
     text: worklet global scopes; url:worklets.html#concept-document-worklet-global-scopes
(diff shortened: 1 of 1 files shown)


jgraham:
Note that this is an extremely basic version of the command, that skips all the issues around trying to wait for the history navigation to complete (or even to start; we currently don't return an error if you try to navigate past the last position in the session history, or before the first position).

However I think it's _nearly_ enough for what Puppeteer does, the CDP's `Page.navigateToHistoryEntry` also doesn't wait or anything. The main difference is that the CDP method isn't ever called with an invalid history entry, so you know upfront if the navigation will succeed (subject to race conditions).

jgraham:
I've tried to fix the error issue by pre-checking the history entry exists. That's not perfect, but I think it's good enough for now.

OrKoN (approved):
I think it would work for Puppeteer

index.bs
```diff
@@ -3133,6 +3143,116 @@ The [=remote end steps=] with |command parameters| are:
...
+   the following steps:
+
+    1. Let |all steps| be the result of [=getting all used history steps=] for |navigable|.
+
+    1. Let |current index| be the index of |navigable|'s [=current session history
+       step=] within |all steps|.
+
+    1. Let |target index| be |current index| plus |delta|.
+
+    1. Let |valid entry| be false if |all steps|[|target index|] does not exist,
```
jimevans, line 3200: Won’t this always be false?

jimevans (changes requested):

index.bs
jgraham: Why? If I open a new tab and then try to navigate forward then `|all steps|` will have length 1, `|current index|` will be `0` and `|target index|` will be `1`, so `|all steps|[|target index|]` won't exist (also I copied the logic from HTML, so I'm fairly sure it's right).

index.bs
```diff
@@ -3133,6 +3143,116 @@ The [=remote end steps=] with |command parameters| are:
...
+       or false otherwise.
```
jgraham, line 3201:
```suggestion
       or true otherwise.
```

index.bs
jgraham: Oh, I see, the context didn't show me the incorrect line. Sorry!

jimevans (approved):

index.bs
```diff
@@ -1863,7 +1872,8 @@ BrowsingContextResult = (
   browsingContext.CreateResult /
   browsingContext.GetTreeResult /
   browsingContext.NavigateResult /
-  browsingContext.PrintResult
+  browsingContext.PrintResult /
```
whimboo, line 1875: As a sidenote: don't we have to use `//` here as well?

index.bs
```diff
@@ -3133,6 +3143,116 @@ The [=remote end steps=] with |command parameters| are:
...
+        context: browsingContext.BrowsingContext,
+        delta: js-int,
+      }
+      </pre>
+   </dd>
+   <dt>Return Type</dt>
+   <dd>
+    <pre class="cddl local-cddl">
+        browsingContext.TraverseHistoryResult = {
+        }
```
whimboo, line 3170: Should this for now be an `EmptyResult`?

index.bs
```diff
@@ -3133,6 +3143,116 @@ The [=remote end steps=] with |command parameters| are:
...
+    1. Let |target index| be |current index| plus |delta|.
+
+    1. Let |valid entry| be false if |all steps|[|target index|] does not exist,
+       or true otherwise.
+
+    1. [=Resume=] with "<code>check history</code>", |resume id|, and
+       |valid entry|.
+
+1. Let |is valid entry| be [=await=] with «"<code>check history</code>"», and
+   |resume id|.
```
whimboo, line 3207: Hm, what exactly is `check history` here? It's not actually defined while being called twice with a different set of parameters. I assume the code here awaits the queued task to be complete?

whimboo (commented):
Generally looks fine to me as well. Just some questions for clarity.

index.bs
jgraham: See https://github.com/w3c/webdriver-bidi/pull/512

index.bs
jgraham: Could do, but since we're planning to update this in the future it didn't seem helpful to change.

index.bs
jgraham: It's just a constant used in the await mechanism to identify what we're waiting for (basically, we identify wait points by name, id pairs).

index.bs
whimboo: The question is when this will happen. Also this current type is not extensible. But if all agree with this I'm fine as well.

index.bs
whimboo: Thanks. Maybe it becomes fully clear when implementing this feature.

whimboo (approved):

index.bs
```diff
@@ -3133,6 +3143,116 @@ The [=remote end steps=] with |command parameters| are:
...
+#### The browsingContext.traverseHistory Command ####  {#command-browsingContext-traverseHistory}
+
+The <dfn export for=commands>browsingContext.traverseHistory</dfn> command
+traverses the history of a given navigable by a delta.
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+      <pre class="cddl remote-cddl">
+      BrowsingContext.TraverseHistory = {
```
sadym-chromium, line 3155: Should be `browsingContext` starting with lower case `b`.

sadym-chromium: nevermind. It is fixed in #601

PR #598, opened 2023-11-09
Editorial: fix/refactor capabilities

jrandolf:
This PR moves out the proxy configuration into a separate type and splits it further for the correct semantics. It also moves "webSocketUrl" algorithms to the capabilities section.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/598.html" title="Last updated on Nov 16, 2023, 1:05 PM UTC (7ff6bc8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/598/5395e5d...7ff6bc8.html" title="Last updated on Nov 16, 2023, 1:05 PM UTC (7ff6bc8)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1458,52 +1458,37 @@ session.ProxyConfiguration = {
 }
 
 session.AutodetectProxyConfiguration = (
-   ; Indicates that the proxy to use should be detected in an
-   ; implementation-specific way.
    proxyType: "autodetect"
    Extensible
 )
 
 session.DirectProxyConfiguration = (
-   ; Indicates that the browser should not use a proxy at all.
    proxyType: "direct",
    Extensible
 )
 
 session.ManualProxyConfiguration = (
-   ; Indicates a manual proxy configuration.
    proxyType: "manual",
-   ; Defines the proxy host for FTP traffic.
    ? ftpProxy: text,
-   ; Defines the proxy host for HTTP traffic.
    ? httpProxy: text,
-   ; Defines the proxy host for encrypted TLS traffic.
    ? sslProxy: text,
-   ; Defines the proxy host for a SOCKS proxy.
    ? session.SocksProxyConfiguration
-   ; Lists the address for which the proxy should be bypassed.
    ? noProxy: [*text],
    Extensible
 )
 
 session.SocksProxyConfiguration = (
-   ; Defines the proxy host for a SOCKS proxy.
    socksProxy: text,
-   ; Defines the SOCKS proxy version.
    socksVersion: 0..255,
 )
 
 session.PacProxyConfiguration = (
-   ; Indicates that the proxy to use is defined by the proxyAutoconfigUrl.
    proxyType: "pac",
-   ; Defines the URL for a proxy auto-config file
    proxyAutoconfigUrl: text,
    Extensible
 )
 
 session.SystemProxyConfiguration = (
-   ; Indicates that the browser should use the various proxies configured
-   ; for the underlying Operating System.
    proxyType: "system"
    Extensible
 )
@@ -1603,21 +1588,12 @@ This is a [=static command=].
       session.NewResult = {
         sessionId: text,
         capabilities: {
-          ; Initially set to false, indicates the session will not implicitly
-          ; trust untrusted or self-signed TLS certificates on navigation.
           acceptInsecureCerts: bool,
-          ; ASCII Lowercase name of the user agent as a string.
           browserName: text,
-          ; The user agent version, as a string.
           browserVersion: text,
-          ; ASCII Lowercase name of the current platform as a string.
           platformName: text,
-          ; Indicates whether the remote end supports all of the resizing and
-          ; positioning commands.
           setWindowRect: bool,
-          ; Defined when the request defines a "proxy".
           ? proxy: session.ProxyConfiguration,
-          ; Defined when the request defined a "webSocketUrl"
           ? webSocketUrl: bool,
           Extensible
         }
--- a/index.bs
+++ b/index.bs
@@ -1466,7 +1466,7 @@ session.AutodetectProxyConfiguration = (
 
 session.DirectProxyConfiguration = (
    ; Indicates that the browser should not use a proxy at all.
-   proxyType: "direct"
+   proxyType: "direct",
    Extensible
 )
 
--- a/index.bs
+++ b/index.bs
@@ -1067,33 +1067,8 @@ To <dfn>close the WebSocket connections</dfn> given |session|:
 
 ## Establishing a Connection ## {#establishing}
 
-WebDriver clients opt in to a bidirectional connection by requesting a
-capability with the name "<code>webSocketUrl</code>" and value
-true.
-
-This specification defines an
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1428,23 +1403,113 @@ session.CapabilityRequest = {
...
+session.AutodetectProxyConfiguration = (
+   ; Indicates that the proxy to use should be detected in an
+   ; implementation-specific way.
+   proxyType: "autodetect"
+   Extensible
+)
+
+session.DirectProxyConfiguration = (
+   ; Indicates that the browser should not use a proxy at all.
+   proxyType: "direct"
```
OrKoN, line 1469:
```suggestion
   proxyType: "direct",
```

?

OrKoN (approved):

index.bs
```diff
@@ -1538,21 +1603,22 @@ This is a [=static command=].
...
-            ? sslProxy: text,
-            ? socksProxy: text,
-            ? socksVersion: 0..255,
-          },
+          ; Indicates whether the remote end supports all of the resizing and
+          ; positioning commands.
           setWindowRect: bool,
+          ; Defined when the request defines a "proxy".
+          ? proxy: session.ProxyConfiguration,
+          ; Defined when the request defined a "webSocketUrl"
```
jrandolf, line 1620:
```suggestion
          ; Defined when the request defines a "webSocketUrl"
```

index.bs
```diff
@@ -1428,23 +1403,113 @@ session.CapabilityRequest = {
...
+   session.ManualProxyConfiguration //
+   session.PacProxyConfiguration //
+   session.SystemProxyConfiguration //
+}
+
+session.AutodetectProxyConfiguration = (
+   ; Indicates that the proxy to use should be detected in an
+   ; implementation-specific way.
+   proxyType: "autodetect"
+   Extensible
```
jgraham, line 1464: I thought we decided at some point that `Extensible` didn't really do anything?

index.bs
```diff
@@ -1428,23 +1403,113 @@ session.CapabilityRequest = {
...
+)
+
+session.DirectProxyConfiguration = (
+   ; Indicates that the browser should not use a proxy at all.
+   proxyType: "direct",
+   Extensible
+)
+
+session.ManualProxyConfiguration = (
+   ; Indicates a manual proxy configuration.
```
jgraham, line 1474: I don't love having lots of comments in the CDDL; I'd rather we use prose in the specification to explain the format. That avoids mixing normative CDDL with informative descriptions (plus the comments make it quite hard to read, especially without syntax highlighting).

index.bs
```diff
@@ -1428,23 +1403,113 @@ session.CapabilityRequest = {
...
+   proxyType: "manual",
+   ; Defines the proxy host for FTP traffic.
+   ? ftpProxy: text,
+   ; Defines the proxy host for HTTP traffic.
+   ? httpProxy: text,
+   ; Defines the proxy host for encrypted TLS traffic.
+   ? sslProxy: text,
+   ; Defines the proxy host for a SOCKS proxy.
+   ? session.SocksProxyConfiguration
+   ; Lists the address for which the proxy should be bypassed.
```
jgraham, line 1484: Does this definitely not apply to other proxy types?

jgraham (changes requested):

index.bs
jrandolf: Indeed, it does not. See https://www.w3.org/TR/webdriver2/#proxy

index.bs
jrandolf: Comments are good because downstream generated code can use them. If you really want, we can remove them, but I suggest we keep them for downstream generators.

index.bs
jrandolf: 1. "With Extensible" => "The object may have other properties"
2. "Without Extensible" => "The object may or may not have other properties"

For (2), it's up to the implementation to define semantics there because CDDL is not explicit.

jgraham (changes requested):

index.bs
jrandolf: CDDL has an extension mechanism using `//=`. This allows anyone to extend CDDL without tampering with the original code, so in practice, we can remove Extensible without repercussions.

I'll still keep it here, but we can discuss in another PR whether we should remove all `Extensible`s.

index.bs
jgraham: To be clear: as such we need to remove the comments to land this, unless there's something missing in the above reasoning.

index.bs
```diff
@@ -1538,21 +1588,22 @@ This is a [=static command=].
       session.NewResult = {
         sessionId: text,
         capabilities: {
+          ; Initially set to false, indicates the session will not implicitly
```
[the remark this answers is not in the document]
jrandolf: Sorry, missed this one.

jgraham (approved):

index.bs
```diff
@@ -1542,17 +1592,9 @@ This is a [=static command=].
...
-            ? ftpProxy: text,
-            ? httpProxy: text,
-            ? noProxy: [*text],
-            ? sslProxy: text,
-            ? socksProxy: text,
-            ? socksVersion: 0..255,
-          },
           setWindowRect: bool,
+          ? proxy: session.ProxyConfiguration,
+          ? webSocketUrl: bool,
```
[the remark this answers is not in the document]
jrandolf: The boolean is saying "we return support web socket". Naming is kinda poor, we could have just said "webSocket". See the `webSocketUrl capability serialization algorithm`.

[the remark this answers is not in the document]
whimboo: But in WebDriver classic this is not a boolean for the new session response. So isn't that a conflict?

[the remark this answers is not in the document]
jrandolf: Can you point to the place in the spec where it says this? I don't believe `webSocketUrl` exists in WebDriver classic.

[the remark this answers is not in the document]
whimboo: Well, it doesn't exist in the spec but is used for the BiDi (WebSocket) opt-in. See the steps 6 and 7 of https://w3c.github.io/webdriver-bidi/#establishing.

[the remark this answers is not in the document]
jgraham: Sorry, @whimboo is correct that this is a bug, and the return type is a string (for HTTP+BiDi sessions; it's a boolean for BiDi-only sessions). Apologies for not catching that. It's also true that the spec is confusing.

The key algorithm is https://w3c.github.io/webdriver/#ref-for-dfn-remote-end-steps-6

In step 4 we process capabilities and (assuming there's a match and we're starting a websocket connection), end up with an object that contains `{webSocketUrl: true}`.

In step 9 we call into https://w3c.github.io/webdriver-bidi/#establishing which _updates_ the `capabilities` object so that it contains the actual URL as the value of the `webSocketUrl` property.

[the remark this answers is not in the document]
jrandolf: Oh my, indeed that is very confusing. Here is the fix: https://github.com/w3c/webdriver-bidi/pull/606

PR #599, opened 2023-11-09
(cddl): remove brackets from range

christian-bromann:
I can't really read of the CDDL spec whether or not extra brackets around type primitives, e.g. ranges, can be ignored or not but if we assume not, an opening bracket (especially as part of a key assignment) creates a group that has to consist of a key value pair and can't be a single type.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/599.html" title="Last updated on Nov 9, 2023, 7:34 PM UTC (95a979a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/599/d327638...95a979a.html" title="Last updated on Nov 9, 2023, 7:34 PM UTC (95a979a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3222,7 +3222,7 @@ PDF document represented as a Base64-encoded string.
         ? orientation: ("portrait" / "landscape") .default "portrait",
         ? page: browsingContext.PrintPageParameters,
         ? pageRanges: [*(js-uint / text)],
-        ? scale: (0.1..2.0) .default 1.0,
+        ? scale: 0.1..2.0 .default 1.0,
         ? shrinkToFit: bool .default true,
       }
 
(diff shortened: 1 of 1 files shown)


thiagowfx (approved):

OrKoN:
I recall there was some ambiguity without the brackets. @jrandolf PTAL

thiagowfx (changes requested):
Undo approval

christian-bromann:
This is why I mentioned that the spec is not very clear about this. One could argue that the `0.1..2.0` part within `0.1..2.0 .default 1.0` can be seen as a single type (a range) and resolves into `[range] .default [float]`. That said, I've seen the following example while skimming through the CDDL spec:

```
label = bstr .size (1..63)
```

which proves your assumption to be correct. Therefor I will go ahead and close this.

Just fyi: I am working on a [cddl parser](https://github.com/christian-bromann/cddl) where these nit picky details make a big difference. Thanks for your feedback!

jrandolf:
@christian-bromann You may want to check out https://github.com/anweiss/cddl. It's already fully compliant and we use it as our backend for code generation (https://github.com/google/cddlconv).

PR #600, opened 2023-11-09
(cddl): merging groups

christian-bromann:
In 2.1 of the [cddl spec](https://datatracker.ietf.org/doc/rfc8610/) it is defined how groups can be merged together:

> Groups can be used to factor out common parts of structs, e.g.,
   instead of writing specifications in copy/paste style, such as in
   Figure 5, one can factor out the common subgroup, choose a name for
   it, and write only the specific parts into the individual maps
   (Figure 6).

While the `.and` type describes the same:

>    A ".and" control on a type indicates that the data item matches both
   the left-hand-side type and the type given as the right-hand side.
   (Formally, the resulting type is the intersection of the two types
   given.)

      "type1 .and type2"

I don't think it can be used in the context when a group is assigned to a variable especially as we can just use the format described in 2.1.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/600.html" title="Last updated on Nov 9, 2023, 7:29 PM UTC (2785795)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/600/d327638...2785795.html" title="Last updated on Nov 9, 2023, 7:29 PM UTC (2785795)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7313,14 +7313,16 @@ script.FunctionRemoteValue = {
 }
 
 script.RegExpRemoteValue = {
+  script.RegExpLocalValue,
   ? handle: script.Handle,
   ? internalId: script.InternalId,
-} .and script.RegExpLocalValue
+}
 
 script.DateRemoteValue = {
+  script.DateLocalValue,
   ? handle: script.Handle,
   ? internalId: script.InternalId,
-} .and script.DateLocalValue
+}
 
 script.MapRemoteValue = {
   type: "map",


OrKoN:
@jrandolf PTAL

jrandolf (changes requested):
This is incorrect. See https://github.com/w3c/webdriver-bidi/pull/461

jrandolf:
To be more specific, `script.RegExpLocalValue` (and similarly the other) is a map therefore it define _types_, not groups, so what you wrote is a keyless value in a map (which is valid syntax, but does not spread.

The easiest way to think the difference between groups and types is the behavior in an array.

Example 1:

```
b = text
a = [b]
```

Means array with a single string.

Example 2:

```
b = (text, text)
a = [b]
```

Means array with two strings.

Example 3:

```
b = [text, text]
a = [b]
```

Means array with a single array with two strings.

Note for arrays, the following are equivalent:

```
a = [(foo: text, bar: text)]
a = [(text, text)]
a = [text, text]
```

For maps, the following is _syntactically_ allowed, but is considered illegal in CDDL:

```
a = {(text, text)}
```

which means a map with two entries (but the key is not defined, so it doesn't make sense which is why it's illegal.)

jrandolf:
Closing this as it's incorrect.

Issue #448, opened 2023-06-14
Support emulation of the User-Agent


sadym-chromium:

whimboo:
Commands for this could be `browsingContext.setUserAgent` and `browsingContext.getUserAgent`. Passing `null` as user agent should reset the formerly set custom user agent.

thiagowfx:
Isn't this a duplicate of https://github.com/w3c/webdriver-bidi/issues/446? cc @OrKoN

OrKoN:
Now it is not: https://github.com/w3c/webdriver-bidi/issues/446 is about getting the original non-emulated user agent from a browser binary without a need to have browsing contexts.

mathiasbynens:
I wanted to bring up [User-Agent Client Hints](https://wicg.github.io/ua-client-hints/). Although it’s a separate feature, it’s somewhat related to User-Agent string emulation, and as a user you’d probably want to emulate both the User-Agent string + the Client Hints accordingly (since otherwise it’s trivial for the target page to detect emulation).

Our options are:

A. include Client Hints support as part of this API
B. provide a separate API for Client Hints
C. support User-Agent emulation without Client Hints emulation

An example of A would be [CDP’s `Emulation.setUserAgentOverride`](https://chromedevtools.github.io/devtools-protocol/tot/Emulation/#method-setUserAgentOverride) which accepts an optional `userAgentMetadata` param that configures the Client Hints.

thiagowfx:
Let's start with C).

From the IRC log above:

> <AutomatedTester> orkon: Let's drop CH as that is out of scope. UA changes can be good for browsers vendors but not for the average tester

It seems that we do not want to pursue client hints. @OrKoN could you confirm this?

thiagowfx:
I am intending to add this command as part of the "browsingContext" module (e.g. instead of a new"emulation" module).

christian-bromann:
Within the WebdriverIO framework we are using `script.addPreloadScript` to emulate this Web API, e.g.

```ts
await this.scriptAddPreloadScript({
    functionDeclaration: /*js*/`() => {
        Object.defineProperty(navigator, 'userAgent', {
            value: ${JSON.stringify(options)}
        })
    }`
})
```

Is there any advantage to have this handled through the protocol?

thiagowfx:
> Is there any advantage to have this handled through the protocol?

[...]

PR #602, opened 2023-11-14
DRAFT: Introduce "browsingContext.setUserAgent" command

thiagowfx:
Bug: #448


index.bs
```diff
@@ -6530,7 +6584,7 @@ strong [=/map=] from handle ids to their corresponding objects.
 script.InternalId = text;
 </pre>
 
-The <code>script.InternalId</code> type represents the id of 
+The <code>script.InternalId</code> type represents the id of
```
OrKoN, line 6587: let's split editorial changes into a separate CL for easier review

thiagowfx: This is just my text editor removing whitespace on save. This keeps coming up because this repository has no linter. I'll look into adding pre-commit here.

For now, could you just ignore this diff? It's really small. I am happy to split it before submission.

index.bs
```diff
@@ -3406,6 +3408,58 @@ The [=remote end steps=] with |command parameters| are:
...
+   <dd>
+    <pre class="cddl remote-cddl">
+      browsingContext.SetUserAgent = (
+        method: "browsingContext.setUserAgent",
+        params: browsingContext.SetUserAgentParameters
+      )
+
+      browsingContext.SetUserAgentParameters = {
+        context: browsingContext.BrowsingContext,
+        ? userAgent: text,
```
OrKoN, line 3427:
```suggestion
        userAgent: text / null,
```

index.bs
```diff
@@ -3406,6 +3408,58 @@ The [=remote end steps=] with |command parameters| are:
...
+
+1. Let |context| be the result of [=trying=] to [=get a browsing context=] with
+   |context id|.
+
+1. Let |user agent| be the value of the <code>userAgent</code> field of |command
+   parameters| if present, or null otherwise.
+
+1. If |user agent| is not null:
+
+   1. Set |context|'s [=user agent=] to |user agent|.
```
OrKoN, line 3454: I am not sure if that would give use desired side effects? I am not sure that is really per-context because it is listed under "System state and capabilities". Also, the default User Agent value is used in https://fetch.spec.whatwg.org/#default-user-agent-value directly. I am not sure if we can redefine it from the BiDi spec or if we need some hooks into other spec. Perhaps we can use https://w3c.github.io/webdriver-bidi/#patches for that?

index.bs
thiagowfx: I was a bit unsure of this. @jgraham could you double check here? Is the preferred pattern to use `?` or `/ null`?

index.bs
thiagowfx: I made some experiments and the user agent seems to be individual per context.

For example, given a page with an iframe, changing the userAgent in the iframe does not change it in the top-level context. And vice-versa.

OrKoN: Hm, what do you mean by changing? CDP implementation is per-context but I don't think this spec change would specify that.

thiagowfx: Changing the user agent = calling https://chromedevtools.github.io/devtools-protocol/tot/Emulation/#method-setUserAgentOverride

index.bs
OrKoN: The prose checks for null so it needs to be null to match the current text. In general, we tend to use null if we unset the value to the default and undefined (?) if we want it to remain unchanged (e.g., we applied this to the setViewport command).

thiagowfx: Done.

index.bs
thiagowfx: https://github.com/w3c/webdriver-bidi/pull/603

index.bs
thiagowfx: I'm not quite sure what is the AI here. Should I just change the prose to mention `fetch`?

Or should we restrict it to top-level contexts only?

thiagowfx: @jgraham what do you think of the approach proposed by @OrKoN? Should I look into it?

jgraham: So FWIW it looks like gecko _does_ restrict to the top level context. Obviously that could be changed if there's a use case for it.

I think the patch that would make sense would be to have a UA override defined on a (top-level) navigable. Then when asked to get the user agent in a specific Window you get the corresponding navigable and then (depending on the semantics we want) either go directly to it's top-level traversable and check if that has an override set, or maybe walk the parent tree looking for something with an override set. If there's no override you fall back to the default.

I think that works for requests that are kicked off from navigations or scripts running in a window. I don't know what implementations currently do for scripts running in workers, especially service workers or shared workers which could be associated with > 1 (top-level) navigable.

whimboo: I would find it hard to differentiate which override applies to all browsing contexts of a tab or just a specific one when we have more emulation. Maybe we should consider to only set it per top-level browsing context and that the setting bubbles down into frames as well? Here I don't see a reason why a frame should have a different user agent. Do you have an example for such a requirement?

thiagowfx:
Example extension: https://github.com/w3c/webdriver-bidi/pull/538/

Issue #587, opened 2023-11-03
Umbrella / Meta: Browser Permissions as an extension module


thiagowfx:
Permissions spec umbrella issue: https://github.com/w3c/permissions/issues/424

- [x] [(optionally) add a recommendation about how external specifications should be authored (e.g., defining complete modules). I think we can leave the text about implementation extensions as is. It looks like there is nothing blocking on the WebDriver BiDi spec side](https://github.com/w3c/permissions/pull/425#discussion_r1386603368): https://github.com/w3c/webdriver-bidi/pull/605

"Interacting with Permissions for Powerful Features"

The [**WebDriver BiDi spec**](https://w3c.github.io/webdriver-bidi/) spec should support interaction with **browser permissions** as an [extension module](https://w3c.github.io/webdriver-bidi/#extension-modules). These permissions represent a user's choice to allow or deny access to "powerful features" of the platform. For developers, the specification standardizes an API to query the permission state of a powerful feature, and be notified if a permission to use a powerful feature changes state.

The Permissions spec already supports [WebDriver classic](https://www.w3.org/TR/permissions/#set-permission-command). This bug is about making it support [WebDriver BiDi](https://w3c.github.io/webdriver-bidi/) as well.

References:

- W3C Permissions spec: https://www.w3.org/TR/permissions/
- W3C Permissions repo: https://github.com/w3c/permissions/

Soft references:

- https://github.com/w3c/webdriver-bidi/issues/506
- https://github.com/w3c/webdriver-bidi/pull/523

whimboo:
> The Browser Testing and Tools Working Group just discussed `Permission extensions`.
> The full IRC log of that discussion

And there was one more comment from @jgraham that actually sneaked into the logs for https://github.com/w3c/webdriver-bidi/issues/287#issuecomment-1802375543:

> <jgraham_> (don't want to extend this topic further, but I always assumed BiDi would move to an event-based model for permissions i.e. you'd get a permissions.Request event with some information)

thiagowfx:
The recommendation was added (https://github.com/w3c/webdriver-bidi/pull/605), and the umbrella issue in the permissions repo was closed (https://github.com/w3c/permissions/issues/424), therefore this can be closed.

I'll just note that there's one missing AI from https://github.com/w3c/permissions/issues/424:

>  TODO: Add `user context` to `SetPermissionParameters`

@OrKoN you may want to open a separate bug/FR for that, if still relevant

PR #605, opened 2023-11-16
Introduce recommendations for external specification modules

thiagowfx:
Bug: #506, #587

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/605.html" title="Last updated on Nov 21, 2023, 7:56 PM UTC (ddbe5b3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/605/a827e75...ddbe5b3.html" title="Last updated on Nov 21, 2023, 7:56 PM UTC (ddbe5b3)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -474,8 +474,17 @@ An implementation may define <dfn export>extension modules</dfn>. These must hav
 [=module name=] that contains a single colon "<code>:</code>" character. The
 part before the colon is the prefix; this is typically the same for all
 [=extension modules=] specific to a given implementation and should be unique for a
-given implementation. Such modules extend the [=local end definition=] and [=remote end definition=]
-providing additional groups as choices for the defined [=commands=] and [=events=].
+given implementation.
+Such modules extend the [=local end definition=] and [=remote end definition=]
+providing additional groups as choices for the defined [=commands=], [=events=] and types.
+
+<dfn export>External specifications</dfn> may define their own modules that
+extend the protocol.
+The recommendation is that such modules should have a [=module name=] that does
+not contain a colon "<code>:</code>" character, and that it is preferred
+to add new modules or commands rather than extending existing ones.
+Such modules may extend the [=local end definition=] and [=remote end definition=] providing
+additional groups as choices for the defined [=commands=], [=events=] and types.
 
 ## Commands ## {#commands}
 
--- a/index.bs
+++ b/index.bs
@@ -478,12 +478,13 @@ given implementation.
 Such modules extend the [=local end definition=] and [=remote end definition=]
 providing additional groups as choices for the defined [=commands=], [=events=] and types.
 
-External specifications may define their own modules that extend the protocol.
-The recommendation is that such modules should have a [=module name=] that does
-not contain a colon "<code>:</code>" character, and that it is preferred
-to add new modules rather than extending existing ones.
-Such modules may extend the [=local end definition=] and [=remote end definition=] providing
-additional groups as choices for the defined [=commands=], [=events=] and types.
+Other specifications may define their own WebDriver-BiDi modules that define the protocol.
+Such modules must not not have a name which contains a colon (<code>:</code>) character,
+not must they define command names, event names, or property names that contain that character.
+
+Authors of external specifications are encouraged to to add new modules rather than extending
+existing ones. Where it is desired to extend an existing module, it is preferred to integrate the
+extension directly into the specification containing the original module definition.
 
 ## Commands ## {#commands}
 
--- a/index.bs
+++ b/index.bs
@@ -480,7 +480,7 @@ providing additional groups as choices for the defined [=commands=], [=events=] 
 
 Other specifications may define their own WebDriver-BiDi modules that define the protocol.
 Such modules must not not have a name which contains a colon (<code>:</code>) character,
-not must they define command names, event names, or property names that contain that character.
+nor must they define command names, event names, or property names that contain that character.
 
 Authors of external specifications are encouraged to to add new modules rather than extending
 existing ones. Where it is desired to extend an existing module, it is preferred to integrate the
--- a/index.bs
+++ b/index.bs
@@ -475,8 +475,6 @@ An implementation may define <dfn export>extension modules</dfn>. These must hav
 part before the colon is the prefix; this is typically the same for all
 [=extension modules=] specific to a given implementation and should be unique for a
 given implementation.
-Such modules extend the [=local end definition=] and [=remote end definition=]
-providing additional groups as choices for the defined [=commands=], [=events=] and types.
 
 Other specifications may define their own WebDriver-BiDi modules that define the protocol.
 Such modules must not not have a name which contains a colon (<code>:</code>) character,
--- a/index.bs
+++ b/index.bs
@@ -478,7 +478,7 @@ given implementation.
 
 Other specifications may define their own WebDriver-BiDi modules that define the protocol.
 Such modules must not not have a name which contains a colon (<code>:</code>) character,
-nor must they define command names, event names, or property names that contain that character.
+nor must they define [=command names=], [=event names=], or property names that contain that character.
 
 Authors of external specifications are encouraged to to add new modules rather than extending
 existing ones. Where it is desired to extend an existing module, it is preferred to integrate the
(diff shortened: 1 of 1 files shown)


thiagowfx:
@OrKoN I'm not quite sure what we really want to add here...

(from the discussions on https://github.com/w3c/permissions/pull/425)

Thoughts?

(Corresponding WG meeting discussion: https://github.com/w3c/webdriver-bidi/issues/587#issuecomment-1802371733)

OrKoN (commented):
I think it is not what we want. Let me come back to this tomorrow.

jgraham:
There are two different things here:
* Open specifications, which can in principle define whatever they want. However we can suggest they they _should_ avoid extending existing commands or modules where possible (i.e. prefer to add new modules; for extending existing modules updating the source specification of the existing module is preferred), and _never_ define names including a colon.
* Implementations. For implementations we want to define that they _must not_ add additional modules or extend new modules _unless_ the added modules, commands, events, or properties have a correctly prefixed name (i.e. one that contains a `:`).

thiagowfx:
Previously:

> When defining [=extension modules=], implementers should consider the
possibility of future standardization of the module. If the module is
likely to be standardized in the future, implementers should use a
[=module name=] that does not contain a colon "<code>:</code>" character.

index.bs
```diff
@@ -474,8 +474,22 @@ An implementation may define <dfn export>extension modules</dfn>. These must hav
...
+
+<dfn export>External specifications</dfn> may define their own modules that
+extend the protocol.
+The recommendation is that such modules should have a [=module name=] that does
+not contain a colon "<code>:</code>" character, and that it is preferred
+to add new modules or commands rather than extending existing ones.
+Such modules may extend the [=local end definition=] and [=remote end definition=] providing
+additional groups as choices for the defined [=commands=], [=events=] and types.
+
+When defining [=extension modules=], implementers should consider the
```
OrKoN, line 489: I think this paragraph should be removed because it is in conflict with the naming constraint of the implementation extension modules.

OrKoN (changes requested):

index.bs
thiagowfx: I agree, this was an oversight, thought I had removed it. Done.

index.bs
```diff
@@ -474,8 +474,17 @@ An implementation may define <dfn export>extension modules</dfn>. These must hav
 [=module name=] that contains a single colon "<code>:</code>" character. The
 part before the colon is the prefix; this is typically the same for all
 [=extension modules=] specific to a given implementation and should be unique for a
-given implementation. Such modules extend the [=local end definition=] and [=remote end definition=]
-providing additional groups as choices for the defined [=commands=] and [=events=].
+given implementation.
+Such modules extend the [=local end definition=] and [=remote end definition=]
+providing additional groups as choices for the defined [=commands=], [=events=] and types.
+
+<dfn export>External specifications</dfn> may define their own modules that
```
thiagowfx, line 481: The only purpose of this `dfn` is to have a place to link to from external specifications. Open to other suggestions.

index.bs
```diff
@@ -474,8 +474,17 @@ An implementation may define <dfn export>extension modules</dfn>. These must hav
 [=module name=] that contains a single colon "<code>:</code>" character. The
 part before the colon is the prefix; this is typically the same for all
 [=extension modules=] specific to a given implementation and should be unique for a
-given implementation. Such modules extend the [=local end definition=] and [=remote end definition=]
-providing additional groups as choices for the defined [=commands=] and [=events=].
+given implementation.
+Such modules extend the [=local end definition=] and [=remote end definition=]
+providing additional groups as choices for the defined [=commands=], [=events=] and types.
```
thiagowfx, line 479: types or properties?

index.bs
```diff
@@ -474,8 +474,17 @@ An implementation may define <dfn export>extension modules</dfn>. These must hav
...
+
+<dfn export>External specifications</dfn> may define their own modules that
+extend the protocol.
+The recommendation is that such modules should have a [=module name=] that does
+not contain a colon "<code>:</code>" character, and that it is preferred
+to add new modules or commands rather than extending existing ones.
```
OrKoN, line 485:
```suggestion
to add new modules rather than extending existing ones.
```

index.bs
OrKoN: I think the definition is not needed

thiagowfx: How do we link to this part from other PRs then?

OrKoN: I don't think we need to, we just link to the whole spec.

thiagowfx: OK, removed.

index.bs
```diff
@@ -474,8 +474,16 @@ An implementation may define <dfn export>extension modules</dfn>. These must hav
...
+given implementation.
+Such modules extend the [=local end definition=] and [=remote end definition=]
+providing additional groups as choices for the defined [=commands=], [=events=] and types.
+
+External specifications may define their own modules that extend the protocol.
+The recommendation is that such modules should have a [=module name=] that does
+not contain a colon "<code>:</code>" character, and that it is preferred
+to add new modules rather than extending existing ones.
+Such modules may extend the [=local end definition=] and [=remote end definition=] providing
+additional groups as choices for the defined [=commands=], [=events=] and types.
```
jgraham, line 486:
```suggestion
Other specifications may define their own WebDriver-BiDi modules that define the protocol.
Such modules must not not have a name which contains a colon (<code>:</code>) character,
not must they define command names, event names, or property names that contain that character.

Authors of external specifications are encouraged to to add new modules rather than extending
existing ones. Where it is desired to extend an existing module, it is preferred to integrate the
extension directly into the specification containing the original module definition.
```

Or something.

jgraham (changes requested):

index.bs
```diff
@@ -474,8 +474,16 @@ An implementation may define <dfn export>extension modules</dfn>. These must hav
 [=module name=] that contains a single colon "<code>:</code>" character. The
 part before the colon is the prefix; this is typically the same for all
 [=extension modules=] specific to a given implementation and should be unique for a
-given implementation. Such modules extend the [=local end definition=] and [=remote end definition=]
-providing additional groups as choices for the defined [=commands=] and [=events=].
+given implementation.
+Such modules extend the [=local end definition=] and [=remote end definition=]
```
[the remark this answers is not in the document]
thiagowfx: OK, I'll remove this paragraph.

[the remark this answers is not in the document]
thiagowfx: Done.

In hindsight though, this is pre-existing prose. Please double check if deleting it is indeed desired.

index.bs
```diff
@@ -474,8 +474,15 @@ An implementation may define <dfn export>extension modules</dfn>. These must hav
 [=module name=] that contains a single colon "<code>:</code>" character. The
 part before the colon is the prefix; this is typically the same for all
 [=extension modules=] specific to a given implementation and should be unique for a
-given implementation. Such modules extend the [=local end definition=] and [=remote end definition=]
-providing additional groups as choices for the defined [=commands=] and [=events=].
+given implementation.
+
+Other specifications may define their own WebDriver-BiDi modules that define the protocol.
+Such modules must not not have a name which contains a colon (<code>:</code>) character,
```
OrKoN, line 480:
```suggestion
Such modules must not have a name which contains a colon (<code>:</code>) character,
```

index.bs
```diff
@@ -474,8 +474,15 @@ An implementation may define <dfn export>extension modules</dfn>. These must hav
 [=module name=] that contains a single colon "<code>:</code>" character. The
 part before the colon is the prefix; this is typically the same for all
 [=extension modules=] specific to a given implementation and should be unique for a
-given implementation. Such modules extend the [=local end definition=] and [=remote end definition=]
-providing additional groups as choices for the defined [=commands=] and [=events=].
+given implementation.
+
+Other specifications may define their own WebDriver-BiDi modules that define the protocol.
```
jgraham, line 479: Two "define"s in the same sentence.

thiagowfx: This was your suggestion applied as-is :P 

I suppose the intent here is to replace the second "define" with "extend"

thiagowfx: Done. Now it reads:

> Other specifications may define their own WebDriver-BiDi modules that extend the protocol.

jgraham (approved):

PR #610, opened 2023-11-22
Normative: reject unwanted partial innerText match

jugglinmike:
***
<a href="https://pr-preview.s3.amazonaws.com/bocoup/webdriver-bidi/pull/610.html" title="Last updated on Nov 22, 2023, 9:17 PM UTC (bfcbd1f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/610/829057e...bocoup:bfcbd1f.html" title="Last updated on Nov 22, 2023, 9:17 PM UTC (bfcbd1f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3051,7 +3051,8 @@ and |session|:
       1. If |match type| is <code>"full"</code> and |node text| [=is=] |search text|,
          [=list/append=] |context node| to |returned nodes|.
 
-      1. Otherwise, [=list/append=] |context node| to |returned nodes|.
+      1. Otherwise, if |match type| is <code>"partial"</code>, [=list/append=]
+         |context node| to |returned nodes|.
 
     1. Otherwise, perform the following steps:
 


jimevans (approved):

OrKoN (approved):

index.bs
```diff
@@ -3051,7 +3051,8 @@ and |session|:
       1. If |match type| is <code>"full"</code> and |node text| [=is=] |search text|,
          [=list/append=] |context node| to |returned nodes|.
 
-      1. Otherwise, [=list/append=] |context node| to |returned nodes|.
+      1. Otherwise, if |match type| is <code>"partial"</code>, [=list/append=]
```
jimevans, line 3054: Since `match type` can only be “partial” or “full”, this if statement seems redundant. Help me understand where the algorithm fails without the “if” clause?

jimevans (changes requested):

index.bs
jugglinmike: If "match type" is "full" and "node text" is not "search text". For example, if we're searching strictly for "hello" in `<div>hello world</div>`.

jimevans: That’s reasonable. Thanks for the clarification.

PR #620, opened 2023-12-14
Clarify `script.addPreloadScript` with duplicate `context`

### LINKED ISSUE #617: Ignore duplicated `contexts` entries in `script.addPreloadScript`, `session.subscribe` and `session.unsubscribe`
Currently, if the same `context` is set twice in the `contexts` param of the `script.addPreloadScript` command, the required scripts is added twice to that context. I wonder if this is correct and expected behavior, or we should throw exception / ignore the duplicate?

Lightning00Blade:
Closes #617.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/620.html" title="Last updated on Dec 18, 2023, 3:55 PM UTC (cfbea2d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/620/9ddaeb4...cfbea2d.html" title="Last updated on Dec 18, 2023, 3:55 PM UTC (cfbea2d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -8292,15 +8292,19 @@ The [=remote end steps=] given |session| and |command parameters| are:
 1. Let |arguments| be the <code>arguments</code> field of |command
    parameters| if present, or an empty [=/list=] otherwise.
 
-1. Let the |contexts| be the value of the <code>contexts</code> field of |command parameters| if it is present or null if it isn’t.
+1. If the value of the <code>contexts</code> field of |command parameters| if it is present:
 
-1. If |contexts| is not null:
+   1. Let |contexts| be an empty [=/set=].
 
-   1. For each |context id| of |contexts|:
+   1. For each |context id| of |command parameters|["<code>contexts</code>"]
 
       1. Let |context| be the result of [=trying=] to [=get a browsing context=] with |context id|.
 
-      1. If |context| is not a [=top-level browsing context=], return [=error=] with [=error code=] [=invalid argument=].
+      1. If |context| is a [=top-level browsing context=] append |context| to |contexts|.
+
+      1. Otherwise return [=error=] with [=error code=] [=invalid argument=].
+
+1. Otherwise let |contexts| to null.
 
 1. Let |sandbox| be the value of the "<code>sandbox</code>" field in |command
    parameters|, if present, or null otherwise.
--- a/index.bs
+++ b/index.bs
@@ -8300,9 +8300,9 @@ The [=remote end steps=] given |session| and |command parameters| are:
 
       1. Let |context| be the result of [=trying=] to [=get a browsing context=] with |context id|.
 
-      1. If |context| is a [=top-level browsing context=] append |context| to |contexts|.
+      1. If |context| is not a [=top-level browsing context=], return [=error=] with [=error code=] [=invalid argument=].
 
-      1. Otherwise return [=error=] with [=error code=] [=invalid argument=].
+      1. Append |context| to |contexts|.
 
 1. Otherwise let |contexts| to null.
 
--- a/index.bs
+++ b/index.bs
@@ -8292,7 +8292,7 @@ The [=remote end steps=] given |session| and |command parameters| are:
 1. Let |arguments| be the <code>arguments</code> field of |command
    parameters| if present, or an empty [=/list=] otherwise.
 
-1. Let |contexts| to null.
+1. Let |contexts| be null.
 
 1. If the <code>contexts</code> field of |command parameters| is present:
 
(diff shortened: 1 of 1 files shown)


whimboo:
As I've mentioned on the issue I wonder if we should do the same for `session.subscribe` and `session.unsubscribe` as well given that those two methods suffer from the same issue.

index.bs
```diff
@@ -8292,15 +8292,19 @@ The [=remote end steps=] given |session| and |command parameters| are:
 1. Let |arguments| be the <code>arguments</code> field of |command
    parameters| if present, or an empty [=/list=] otherwise.
 
-1. Let the |contexts| be the value of the <code>contexts</code> field of |command parameters| if it is present or null if it isn’t.
+1. If the value of the <code>contexts</code> field of |command parameters| if it is present:
```
sadym-chromium, line 8295:
```suggestion
1. If the <code>contexts</code> field of |command parameters| is present:
```

sadym-chromium:
> As I've mentioned on the issue I wonder if we should do the same for `session.subscribe` and `session.unsubscribe` as well given that those two methods suffer from the same issue.

In I'm general agree with your proposal to fix the `subscribe` / `unsubscribe` as well, but let's split the changes.

Lightning00Blade:
Agree that we should change it there as well so I would keep the issue open till then, but let's do it in separate PRs.

index.bs
```diff
@@ -8292,15 +8292,19 @@ The [=remote end steps=] given |session| and |command parameters| are:
 1. Let |arguments| be the <code>arguments</code> field of |command
    parameters| if present, or an empty [=/list=] otherwise.
 
-1. Let the |contexts| be the value of the <code>contexts</code> field of |command parameters| if it is present or null if it isn’t.
+1. If the <code>contexts</code> field of |command parameters| is present:
 
-1. If |contexts| is not null:
+   1. Let |contexts| be an empty [=/set=].
```
sadym-chromium, line 8297: I'm not sure if the declaration can be inside "if" condition.

sadym-chromium (approved):

index.bs
Lightning00Blade: Not sure either, maybe a better thing would be `Let contexts be null` and then set it to a Set in the if statement.

sadym-chromium: I don't think this is important though )

index.bs
```diff
@@ -8292,15 +8292,19 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
 
-   1. For each |context id| of |contexts|:
+   1. For each |context id| of |command parameters|["<code>contexts</code>"]
 
       1. Let |context| be the result of [=trying=] to [=get a browsing context=] with |context id|.
 
-      1. If |context| is not a [=top-level browsing context=], return [=error=] with [=error code=] [=invalid argument=].
+      1. If |context| is a [=top-level browsing context=] append |context| to |contexts|.
+
+      1. Otherwise return [=error=] with [=error code=] [=invalid argument=].
```
jgraham, line 8305: Style nit, but I'd prefer an early return if null, to avoid the if/otherwise which is a bit awkward in specs.

jgraham (approved):

whimboo:
@Lightning00Blade would you be so kind and add a wpt test for the multiple usage of `context`?

index.bs
```diff
@@ -8292,16 +8292,20 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
-   1. For each |context id| of |contexts|:
+   1. For each |context id| of |command parameters|["<code>contexts</code>"]
 
       1. Let |context| be the result of [=trying=] to [=get a browsing context=] with |context id|.
 
       1. If |context| is not a [=top-level browsing context=], return [=error=] with [=error code=] [=invalid argument=].
 
+      1. Append |context| to |contexts|.
+
+1. Otherwise let |contexts| to null.
```
whimboo, line 8307: I assume this is a mistake?

```suggestion
1. Otherwise set |contexts| to null.
```

whimboo (approved):

whimboo:
> @Lightning00Blade would you be so kind and add a wpt test for the multiple usage of `context`?

Err, this will actually be part of https://github.com/web-platform-tests/wpt/pull/41932.

index.bs
Lightning00Blade: It was not mistake as we had may or may not have had it created before. 
Did not like it so changed the structure to:
1. Set to null
2. Check if param exist 
2.1. Set to Empty Set
2.2. Loop logic

jrandolf (approved):

index.bs
whimboo: This got refactored and doesn't exist anymore.

whimboo (approved):
Looks good. Thanks!

PR #621, opened 2023-12-14
Implement "owners" in workers

jrandolf:
Every dedicated worker has an [outside setting](https://html.spec.whatwg.org/multipage/workers.html#set-up-a-worker-environment-settings-object) that it's created under. This info is needed to understand the tree structure of dedicated workers.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/621.html" title="Last updated on Dec 19, 2023, 11:54 AM UTC (af0ceaf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/621/ffbf706...af0ceaf.html" title="Last updated on Dec 19, 2023, 11:54 AM UTC (af0ceaf)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -191,17 +191,17 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
     text: report an error; url: webappapis.html#report-the-error
     text: run the animation frame callbacks; url: imagebitmap-and-animations.html#run-the-animation-frame-callbacks
     text: same origin domain; url: browsers.html#same-origin-domain
-    text: session history; url: history.html#session-history
     text: session history entry; url: browsing-the-web.html#session-history-entry
     text: session history traversal queue; url: document-sequences.html#tn-session-history-traversal-queue
+    text: session history; url: history.html#session-history
     text: set up a window environment settings object; url: window-object.html#set-up-a-window-environment-settings-object
     text: set up a worker environment settings object; url: workers.html#set-up-a-worker-environment-settings-object
     text: set up a worklet environment settings object; url: worklets.html#set-up-a-worklet-environment-settings-object
     text: shared worker; url: workers.html#shared-workers
     text: system visibility state; url: document-sequences.html#system-visibility-state
     text: traversable navigable; url:document-sequences.html#traversable-navigable
-    text: visible; url: document-sequences.html#system-visibility-state
     text: traverse the history by a delta; url: browsing-the-web.html#traverse-the-history-by-a-delta
+    text: visible; url: document-sequences.html#system-visibility-state
     text: window open steps; url: window-object.html#window-open-steps
     text: worker event loop; url: webappapis.html#worker-event-loop-2
     text: worklet global scopes; url:worklets.html#concept-document-worklet-global-scopes
@@ -7021,17 +7021,20 @@ script.WindowRealmInfo = {
 
 script.DedicatedWorkerRealmInfo = {
   script.BaseRealmInfo,
-  type: "dedicated-worker"
+  type: "dedicated-worker",
+  owners: [script.Realm]
 }
 
 script.SharedWorkerRealmInfo = {
   script.BaseRealmInfo,
-  type: "shared-worker"
+  type: "shared-worker",
+  owners: [+script.Realm]
 }
 
 script.ServiceWorkerRealmInfo = {
   script.BaseRealmInfo,
   type: "service-worker"
+  owners: [*script.Realm]
 }
 
 script.WorkerRealmInfo = {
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -7119,9 +7120,11 @@ To <dfn>get the realm info</dfn> given |environment settings|:
 
     <dt>|global object| is a {{DedicatedWorkerGlobalScope}} object
     <dd>
+      1. Let |parent realm info| be the result of [=get the realm info=] given [=set up a
```
jgraham, line 7123: I don't understand how this is supposed to work. We can't retrospectively pull out the argument that would have been passed to `set up a worker environment environment settings object` like this.

jgraham (commented):
This seems fine in principle, but I don't think the implementation works as written.

index.bs
jrandolf: This isn't retrospectively though because there is a statement saying _immediately prior to returning the settings
object_. This implies this is all part of the `set up a worker environment environment settings object` algorithm.

jrandolf: Alternatively, in the `realmCreated` remote end event trigger, we can write after

> 1. Let |environment settings| be the newly created [=environment settings
   object=].
   
1. If the invoked algorithm is not [=set up a window environment settings object=], let |outside settings| be the [=outside settings=] parameter of the invoked algorithm.

jrandolf: `outside settings` will always exist in other algorithms.

jrandolf: Done.

index.bs
```diff
@@ -7119,9 +7121,16 @@ To <dfn>get the realm info</dfn> given |environment settings|:
 
     <dt>|global object| is a {{DedicatedWorkerGlobalScope}} object
     <dd>
+      1. Let |owners| be an empty [=/list=].
+      1. For each |owner| in the |global object|'s associated [=owner set=]:
+         1. Let |owner environment settings| be the [=environment settings object=] whose
+            [=relevant global object=] is |owner|.
+         1. Let |owner realm info| be the result of [=get the realm info=] given |owner environment settings|.
+         1. Append |owner realm info| to |owners|.
```
jgraham, line 7129: Do we want all the owner realm info, or just the realm id? In `getRealms` providing all the information is going to lead to duplicate entries for the owners, and the CDDL suggests that you just wanted the id.

index.bs
```diff
@@ -184,24 +184,25 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
     text: navigables; url: document-sequences.html#navigables
     text: navigation id; url: browsing-the-web.html#navigation-id
     text: origin-clean; url: canvas.html#concept-canvas-origin-clean
+    text: owner set; url: workers.html#concept-WorkerGlobalScope-owner-set
```
jgraham, line 187: I think we already link to owner set just as `[=owner set=]`, so I think this is unnecessary.

jgraham (changes requested):
Any specific reason we're only doing this for dedicated workers? I think the same approach also works for shared workers.

index.bs
jrandolf: Indeed, we just want the ID.

index.bs
jrandolf: Indeed, you're right.

jrandolf:
> Any specific reason we're only doing this for dedicated workers? I think the same approach also works for shared workers.

@jgraham So I'm a little conflicted right now about exposing all owners. What I really want to expose is the "initial owner", not necessarily all of them. If we want to expose all of them, then we'd need an event like `realm.ownerAdded` and `realm.ownerRemoved` to really capture the entire lifecycle of ownership.

What do you think about extending this to that length?

jgraham:
I think dedicated workers only ever have one owner, so for that case the events don't seem necessary. For shared workers it makes more sense to me to expose the owners in the same way we do for dedicated workers, but not add lifecycle events until there's clear user demand that it does to not expose the owners because we don't also have lifecycle events.

So, for now, I'd just add the field and information to all worker types, so that we at least have a consistent basis to build from, and consider extending the support later if necessary.

jrandolf:
> I think dedicated workers only ever have one owner, so for that case the events don't seem necessary. For shared workers it makes more sense to me to expose the owners in the same way we do for dedicated workers, but not add lifecycle events until there's clear user demand that it does to not expose the owners because we don't also have lifecycle events.
> 
> So, for now, I'd just add the field and information to all worker types, so that we at least have a consistent basis to build from, and consider extending the support later if necessary.

Done.

index.bs
```diff
@@ -7074,6 +7077,26 @@ To <dfn>get the browsing context</dfn> with given |realm|:
...
+<div algorithm>
+To <dfn>get the worker's owners</dfn> with given |global object|:
+
+1. Assert: |global object| is a {{WorkerGlobalScope}} object.
+
+1. Let |owners| be an empty [=/list=].
+
+1. For each |owner| in the |global object|'s associated [=owner set=]:
+
+   1. Let |owner environment settings| be the [=environment settings object=]
```
jgraham, line 7090: I know this is copied from elsewhere in the spec, but looking at HTML now it seems like one should use https://html.spec.whatwg.org/multipage/webappapis.html#concept-settings-object-global (since relevant global object is only really defined for platform objects, and environment settings objects aren't platform objects).

jgraham (commented):
This basically looks fine, but I think there's a technical error copied from elsewhere in the spec.

index.bs
jrandolf: Done.

jgraham (approved):

PR #624, opened 2023-12-19
Specify a locator for accessible role

jugglinmike:
I've referenced [the definition of "role" in WAI-ARIA](https://w3c.github.io/aria/#introroles) because [that's how WebDriver explains the computation of a role given an element](https://w3c.github.io/webdriver/#get-computed-role), but [computedrole from CORE-AAM](https://www.w3.org/TR/core-aam-1.2/#roleMappingComputedRole) seems more appropriate.

***
<a href="https://pr-preview.s3.amazonaws.com/bocoup/webdriver-bidi/pull/624.html" title="Last updated on Dec 19, 2023, 3:30 AM UTC (ba4ce76)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/624/46d0bc1...bocoup:ba4ce76.html" title="Last updated on Dec 19, 2023, 3:30 AM UTC (ba4ce76)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -246,6 +246,9 @@ spec: UNICODE; urlPrefix: https://www.unicode.org/versions/Unicode15.0.0/
   type: dfn
     text: Unicode Default Case Conversion algorithm; url: ch03.pdf#G34944
     text: toUppercase; url: ch03.pdf#G34078
+spec: WAI-ARIA; urlPrefix:https://w3c.github.io/aria/
+  type: dfn
+    text: WAI-ARIA role; url: #introroles
 </pre>
 
 <pre class="biblio">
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3076,6 +3085,41 @@ and |session|:
 
 </div>
 
+<div algorithm="locate nodes using role">
+To <dfn>locate nodes using role</dfn> with given |context nodes|, |selector|,
```
OrKoN, line 3089: Puppeteer would currently need both accessible name and role as an input to the query. Would it make sense to include both into this locator?

jugglinmike: I believe the plan is to specify a distinct locator for accessible name.

OrKoN: Those are different though: they are for getting the values and not for searching. 

I think it'd would be beneficial to have a single locator to search by a11y attributes (so it's easier to extend the list of attributes). Currently, the locators not composable so the client like Puppeteer would be required to query twice and merge results client-side.

OrKoN: Related https://github.com/WICG/aom/issues/197

OrKoN:
@jugglinmike should I take over this PR?

OrKoN:
I have opened a draft based on this PR https://github.com/w3c/webdriver-bidi/pull/660 (we will discuss it internally first)

PR #628, opened 2023-12-19
Clarify `session.subscribe` and `session.unsubscribe`

Lightning00Blade:
Closes https://github.com/w3c/webdriver-bidi/issues/617.


index.bs
```diff
@@ -1506,8 +1506,8 @@ session.SystemProxyConfiguration = (
 
 <pre class="cddl remote-cddl">
 session.SubscriptionRequest = {
-  events: [*text],
-  ? contexts: [*browsingContext.BrowsingContext],
+  events: [+text],
+  ? contexts: [+browsingContext.BrowsingContext],
```
Lightning00Blade, line 1510: If we encounter `events: []` or `contexts: []` we should throw errors.

juliandescottes (commented):
For session subscribe/unsubscribe, the `update the event map` algorithm is already making sure to skip duplicated events and contexts. It also has additional logic to deal with the fact that we only subscribe to top level context. 

So I don't think we have to switch to sets to skip duplicates. Since it adds some complexity, maybe we should reconsider this?

PR #632, opened 2024-01-05
Remove "worker" and "worklet" realm types

jrandolf:
These types are never created since their specialized types are already enumerated as realm types.

***

### :boom: Error: connect ETIMEDOUT 128.30.52.141:443 :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jan 5, 2024, 11:27 AM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [HTML Diff Service](http://services.w3.org/htmldiff) - The HTML Diff Service is used to create HTML diffs of the spec changes suggested in a pull request.

:link: [Related URL](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fpr-preview.s3.amazonaws.com%2Fw3c%2Fwebdriver-bidi%2Fpull%2F632%2Fd3ec8cb.html&doc2=https%3A%2F%2Fpr-preview.s3.amazonaws.com%2Fw3c%2Fwebdriver-bidi%2Fpull%2F632.html)

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/webdriver-bidi%23632.)._
</details>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7017,10 +7017,8 @@ script.RealmInfo = (
   script.DedicatedWorkerRealmInfo /
   script.SharedWorkerRealmInfo /
   script.ServiceWorkerRealmInfo /
-  script.WorkerRealmInfo /
   script.PaintWorkletRealmInfo /
-  script.AudioWorkletRealmInfo /
-  script.WorkletRealmInfo
+  script.AudioWorkletRealmInfo
 )
 
 script.BaseRealmInfo = (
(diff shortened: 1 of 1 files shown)


OrKoN (commented):
I recall a discussion about that and there was a seemingly good reason to keep the base types: to allow for forward compatibility if new specialized classes of workers or worklets arise. Has anything changed in this regard?

PR #637, opened 2024-01-11
Make `network.Cookie` extensible

sadym-chromium:
In order to be able using generated types and provide some vendor-specific fields (like `goog:sameParty`), it would be very useful to have the `cookies` type explicitly extensible.


whimboo (approved):
Looks like that this was a miss given that `storage.PartialCookie` already contains `Extensible`.

PR #641, opened 2024-01-17
Jgraham/partitions

jgraham:


index.bs
```diff
@@ -2204,11 +2293,11 @@ To <dfn export>get a browsing context</dfn> given |context id|:
 browsingContext.InfoList = [*browsingContext.Info]
 
 browsingContext.Info = {
+  children: browsingContext.InfoList / null
```
OrKoN, line 2296:
```suggestion
  children: browsingContext.InfoList / null,
```

it looks like the comma was missing originally

index.bs
```diff
@@ -2031,7 +2065,9 @@ The <dfn export for=commands>browser.createUserContext</dfn> command creates a
 
 The [=remote end steps=] are:
 
-1. Let |user context| be a new [=empty user context=].
+1. Let |user context| be a new [=user context=].
```
OrKoN, line 2068: Should we keep the empty here to make sure there are no top-level navigables in the freshly created context?

index.bs
```diff
@@ -1300,27 +1300,60 @@ Issue: Define how this works.
...
+between different [=user contexts=].
 
 Note: The infra spec uses the term "user agent" to refer to the same concept as
 [=user context|user contexts=]. However, this is not compatible with usage of
 the term "user agent" to mean the entire web client with multiple [=user
 context|user contexts=]. Although this difference is not visible to web content,
-it is observed via WebDriver, so we avoid reusing existing terminology.
+it is observed via WebDriver, so we avoid using this terminology.
+
+Issue: Unclear if this is the correct way to define this.
```
OrKoN, line 1314: Does this refer to the entire definition or the next sentence?

index.bs
jgraham: I don't think it's necessary. Whether or not a user context is empty is a computed property of the context, so talking about creating one as if that's directly controllable doesn't make sense. Moreover, at the moment a user context is created it can't have any navigables in it because all existing navigables must belong to a different user context, and nothing has created one in this user context, so I think it has to be empty.

index.bs
jgraham: The entire definition.

OrKoN (approved):

PR #651, opened 2024-01-31
Remove `owners` for shared and service workers

jrandolf:
This PR removes the `owners` property for shared and service workers.

For shared and service workers, the `owners` property makes little sense since the set of owners is dynamic. In particular, to properly implement the `owners` property, we would need another lifecycle method such as `script.RealmUpdated` which contains updates to the list of owners. As of now, we've never come across the need inspecting ownership of service and shared worker owners, so removing this and adding it when the use-case arises may be more beneficial for implementors.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/651.html" title="Last updated on Jan 31, 2024, 10:21 PM UTC (760fbd7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/651/6462133...760fbd7.html" title="Last updated on Jan 31, 2024, 10:21 PM UTC (760fbd7)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7120,14 +7120,12 @@ script.DedicatedWorkerRealmInfo = {
 
 script.SharedWorkerRealmInfo = {
   script.BaseRealmInfo,
-  type: "shared-worker",
-  owners: [+script.Realm]
+  type: "shared-worker"
 }
 
 script.ServiceWorkerRealmInfo = {
   script.BaseRealmInfo,
   type: "service-worker"
-  owners: [*script.Realm]
 }
 
 script.WorkerRealmInfo = {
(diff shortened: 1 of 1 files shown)


OrKoN (approved):

jgraham:
I don't think I agree with the arguments here in general; we have other cases where information is dynamic and we don't have corresponding lifecycle events, and I think there are clear use cases for questions like "why is this service worker active", which knowing its owner set helps answer.

But I don't object to the idea of removing this for now, if no one is actively planning to implement it. I would at least like us to make it easy to reland this feature once we do have interest.

jrandolf:
> I don't think I agree with the arguments here in general; we have other cases where information is dynamic and we don't have corresponding lifecycle events, and I think there are clear use cases for questions like "why is this service worker active", which knowing its owner set helps answer.
> 
> But I don't object to the idea of removing this for now, if no one is actively planning to implement it. I would at least like us to make it easy to reland this feature once we do have interest.

I should rephrase; there are clear use-cases, but these use-cases are low priority and has hardly ever been requested, so with our current roadmap, I suggest we remove this feature so we can focus implementing the core features and reland if there is interest as you've stated.

jgraham (approved):

PR #652, opened 2024-02-07
Add a capability for returning the default user agent header value

### LINKED ISSUE #446: Ability to get the user agent
In CDP, `Browser.getVersion` returns `userAgent` fields that are used by Puppeteer for  https://pptr.dev/api/puppeteer.browser.useragent which allows getting the user agent without a need for any browsing contexts to exist.

Does it make sense to add it to WebDriver BiDi? (perhaps, as the result of session.new command)

OrKoN:
This PR would allow client to get the original default User-Agent header value. In Puppeteer, it would allow supporting the `Browser.userAgent()` API.

Closes #446

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/652.html" title="Last updated on Mar 7, 2024, 2:33 PM UTC (f57ecd1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/652/f7e2538...f57ecd1.html" title="Last updated on Mar 7, 2024, 2:33 PM UTC (f57ecd1)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -269,9 +269,6 @@ spec: UNICODE; urlPrefix: https://www.unicode.org/versions/Unicode15.0.0/
   type: dfn
     text: Unicode Default Case Conversion algorithm; url: ch03.pdf#G34944
     text: toUppercase; url: ch03.pdf#G34078
-spec: FETCH; urlPrefix: https://fetch.spec.whatwg.org/
-  type: dfn
-    text: default User-Agent value; url: #default-user-agent-value
 </pre>
 
 <pre class="biblio">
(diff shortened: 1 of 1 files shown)


jrandolf (changes requested):
The CDDL also needs to be updated.

jrandolf (approved):

whimboo (changes requested):
We discussed this yesterday and actually wonder if it wouldn't even be better to add the `userAgent` capability to the WebDriver classic specification. That way we wouldn't have to make this capability BiDi specific.

If you agree this PR would have to be changed to only update the CDDL definition.

OrKoN:
@whimboo is there a requirement for that from the classic users?

jgraham:
So, in detail, the proposal as specified here is that you'd get an additional capability if the implementation _supports_ BiDi irrespective of whether BiDi is requested or enabled (I think that's different to what I said in the previous discussion).

I still think that doesn't make any sense, and we should just add the capability directly to classic.

whimboo (approved):
Looks good to me. Thanks.

Issue #443, opened 2023-06-06
Accessibility module in WebDriver BiDi?


OrKoN:
Puppeteer exposes the information about the browser's a11y tree in a couple of ways:

1) via [Accessibility.getFullAXTree](https://chromedevtools.github.io/devtools-protocol/tot/Accessibility/#method-getFullAXTree) to generate a [snapshot of the a11y tree](https://pptr.dev/api/puppeteer.accessibility.snapshot/)
2) via [Accessibility.queryAXTree](https://chromedevtools.github.io/devtools-protocol/tot/Accessibility/#method-queryAXTree) to query elements by the accessible name and role. This allows Puppeteer to implement element selectors based on the a11y tree data.

Additionally, there were requests recently to include A11y tree information into WebDriver https://goo.gle/chromedriver-a11ytree. I think we should discuss if we want to add these capabilities to WebDriver BiDi.

christian-bromann:
One could argue that a very popular use case is to fetch an element via a11y name. WebdriverIO currently computes some complex xPath to make a good guess about it according to the spec but it [doesn't work all time](https://github.com/webdriverio/webdriverio/issues/8826) due to xPath limitations. It the would help a lot if this could be a primitive, e.g. a selector strategy.

clshortfuse:
I would love for this to be supported. Had a request posted in the interop issue with reasonings for it:

> Currently, I would like to ensure a Web Component presents a role at a certain point in the layout tree. For example:
> 
> ```
> <x-button role=none>
>   #shadowroot
>   <input type=button aria-labelledby=slot>
>   <slot id=slot></slot>
> </x-button>
> ```
> 
> And
> 
> ```
> <x-button role=button>
>   #shadowroot
>   <slot id=slot></slot>
> </x-button>
> ```
> 
> ```
> <x-button> 
>   #shadowroot
>   <slot id=slot></slot>
> </x-button>
> ```
> 
> It would make sense that the next "meaningful" ARIA role starting from `<x-button>` is `button`. But there are multiple ways to achieve this, including hiding the attribute via `ElementInternals`. I think I might be able to iterate the DOM and test each element manually, but I'm worried about closed shadowRoots. I have concerns with just iterating the DOM because I'm not really sure what's being exposed by the browsers to screen-readers. There might be groups or hidden nodes. The DOM and AXTree isn't guaranteed to be `1:1`.
> 
[...]

PR #660, opened 2024-02-13
Add an ability to locate nodes using computed accessibility attributes

OrKoN:
Based on: https://github.com/w3c/webdriver-bidi/pull/624
Issue: https://github.com/w3c/webdriver-bidi/issues/443

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/660.html" title="Last updated on Apr 15, 2024, 12:44 PM UTC (b0ee307)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/660/efb27e8...b0ee307.html" title="Last updated on Apr 15, 2024, 12:44 PM UTC (b0ee307)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -269,6 +269,9 @@ spec: UNICODE; urlPrefix: https://www.unicode.org/versions/Unicode15.0.0/
   type: dfn
     text: Unicode Default Case Conversion algorithm; url: ch03.pdf#G34944
     text: toUppercase; url: ch03.pdf#G34078
+spec: WAI-ARIA; urlPrefix:https://w3c.github.io/aria/
+  type: dfn
+    text: WAI-ARIA role; url: #introroles
 </pre>
 
 <pre class="biblio">
@@ -2501,6 +2504,7 @@ To <dfn>await a navigation</dfn> given |context|, |request|, |wait condition|, a
 browsingContext.Locator = (
    browsingContext.CssLocator /
    browsingContext.InnerTextLocator /
+   browsingContext.RoleLocator /
    browsingContext.XPathLocator
 )
 
@@ -2515,6 +2519,11 @@ browsingContext.InnerTextLocator = {
    ? ignoreCase: bool
    ? matchType: "full" / "partial",
    ? maxDepth: js-uint,
+}
+
+browsingContext.RoleLocator = {
+   type: "role",
+   value: text
 }
 
 browsingContext.XPathLocator = {
@@ -3406,6 +3415,41 @@ and |session|:
 
 </div>
 
+<div algorithm="locate nodes using role">
+To <dfn>locate nodes using role</dfn> with given |context nodes|, |selector|,
+|maximum returned node count|, and |session|:
+
+1. If |selector| is the empty string, return [=error=] with [=error code=]
+   [=invalid selector=].
+
+1. Let |returned nodes| be an empty [=/list=].
+
+1. For each |context node| in |context nodes|:
+
+  1. Let |role| be the result of computing the [=WAI-ARIA role=] of |context node|.
+
+  1. If |selector| [=is=] |role|:
+
+    1. [=list/Append=] |context node| to |returned nodes|.
+
+  1. Let |child nodes| be an empty [=/list=] and, for each node |child| in the
+     <a spec=dom>children</a> of |context node|:
+
+    1. If |child| implements {{Element}}, [=list/append=] |child| to |child nodes|.
+
+  1. Let |child node matches| be the result of [=locate nodes using role=]
+     with |child nodes|, |selector|, |maximum returned node count|, and |session|.
+
+  1. [=list/Extend=] |returned nodes| with |child node matches|.
+
+1. If |maximum returned node count| is not null, [=list/remove=] all entries
+   in |returned nodes| with an index greater than or equal to |maximum returned
+   node count|.
+
+1. Return [=success=] with data |returned nodes|.
+
+</div>
+
 <div algorithm="remote end steps for browsingContext.locateNodes">
 The [=remote end steps=] with |session| and |command parameters| are:
 
@@ -3481,6 +3525,14 @@ The [=remote end steps=] with |session| and |command parameters| are:
          1. Let |result nodes| be a result of [=trying=] to [=locate nodes using inner text=]
             given |context nodes|, |selector|, |max depth|, |match type|,
             |ignore case|, |maximum returned node count| and |session|.
+
+      <dt>|type| is the string "<code>role</code>"
+      <dd>
+         1. Let |selector| be |locator|["<code>value</code>"].
+
+         1. Let |result nodes| be a result of [=trying=] to [=locate nodes using role=]
+            given |context nodes|, |selector|, |maximum returned node count| and
+            |session|.
 
 1. Assert: |maximum returned node count| is null or [=list/size=] of |result nodes| is less
    than or equal to |maximum returned node count|.
--- a/index.bs
+++ b/index.bs
@@ -3452,9 +3452,7 @@ To <dfn>collect nodes using accessibility attributes</dfn> with given |context n
   1. If |match| is true:
 
     1. If |maximum returned node count| is not null and [=list/size=] of |returned
-      nodes| is greater or equal to |maximum returned node count|:
-
-      1. Return |returned nodes|.
+      nodes| is equal to |maximum returned node count|, [=break=].
 
     1. [=list/Append=] |context node| to |returned nodes|.
 
@@ -3465,9 +3463,6 @@ To <dfn>collect nodes using accessibility attributes</dfn> with given |context n
 
   1. [=Try=] to [=collect nodes using accessibility attributes=] with |child nodes|,
      |selector|, |maximum returned node count|, and |returned nodes|.
-
-  1. If |maximum returned node count| is not null and [=list/size=] of |returned
-      nodes| is greater or equal to |maximum returned node count|, break.
 
 1. Return |returned nodes|.
 
--- a/index.bs
+++ b/index.bs
@@ -3454,7 +3454,7 @@ To <dfn>collect nodes using accessibility attributes</dfn> with given |context n
     1. If |maximum returned node count| is not null and [=list/size=] of |returned
       nodes| is greater or equal to |maximum returned node count|:
 
-      1. Return [=success=] with data |returned nodes|.
+      1. Return |returned nodes|.
 
     1. [=list/Append=] |context node| to |returned nodes|.
 
@@ -3466,7 +3466,10 @@ To <dfn>collect nodes using accessibility attributes</dfn> with given |context n
   1. [=Try=] to [=collect nodes using accessibility attributes=] with |child nodes|,
      |selector|, |maximum returned node count|, and |returned nodes|.
 
-  1. Return [=success=] with data |returned nodes|.
+  1. If |maximum returned node count| is not null and [=list/size=] of |returned
+      nodes| is greater or equal to |maximum returned node count|, break.
+
+1. Return |returned nodes|.
 
 </div>
 
--- a/index.bs
+++ b/index.bs
@@ -3441,7 +3441,7 @@ To <dfn>collect nodes using accessibility attributes</dfn> with given |context n
 
       1. Set |match| to false.
 
-  1. If |selector|'s <code>name</code> is present:
+  1. If |selector| [=map/contains=] "<code>name</code>":
 
     1. Let |name| be the [=accessible name=] of |context node|.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2203,6 +2210,15 @@ browsingContext.InnerTextLocator = {
    ? maxDepth: js-uint,
 }
 
+browsingContext.AccessibilityLocator = {
+   type: "accessibility",
+   value: {
+    ? role: text,
+    ? name: text,
+    Extensible
```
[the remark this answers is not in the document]
OrKoN: Extensible is needed to allow implementation extension attributes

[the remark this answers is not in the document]
sadym-chromium: In sucha case we have to extend the *locate nodes using accessibility attributes* algorithm with mentioning that. Current specification version says that the filter should respect only `name` and `role`, and any other fields would be ignored. If the intention is to allow some vendor-specific filtering, I guess it should be mentioned here.

OrKoN, line 2218:
```suggestion
```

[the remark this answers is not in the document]
OrKoN: Removed extensible for now.

sadym-chromium (approved):

OrKoN:
@jrandolf PTAL

index.bs
```diff
@@ -2203,6 +2210,14 @@ browsingContext.InnerTextLocator = {
    ? maxDepth: js-uint,
 }
 
+browsingContext.AccessibilityLocator = {
```
whimboo, line 2213: nit: Please move up before `CssLocator` to keep the alphabetical order. Same for `browsingContext.Locator` as well.

index.bs
```diff
@@ -3076,6 +3091,57 @@ and |session|:
...
+1. If |selector|'s <code>role</code> and |selector|'s <code>name</code> are missing, return [=error=] with [=error code=]
+   [=invalid selector=].
+
+1. Let |returned nodes| be an empty [=/list=].
+
+1. For each |context node| in |context nodes|:
+
+  1. Let |match| be true.
+
+  1. If |selector|'s <code>role</code> is present:
```
whimboo, line 3107: Not sure I understand that logic here. Shouldn't we first run some kind of selector to find specific nodes below the start nodes? Note that `context nodes` are these start nodes. And right now we would just return these which have the wanted role and name.

So shouldn't it be similar to the CSS selector but by using the `[aria-label=%label%]`.

OrKoN: I believe iteration happens here:

```
  1. Let |child nodes| be an empty [=/list=] and, for each node |child| in the
     <a spec=dom>children</a> of |context node|:

    1. If |child| implements {{Element}}, [=list/append=] |child| to |child nodes|.
```

so this algorithm returns all nodes that match the predicate (role OR name matches). aria-label is not a computed attribute and, therefore, does not query in the same way.

index.bs
```diff
@@ -2187,6 +2193,7 @@ To <dfn>await a navigation</dfn> given |context|, |request|, |wait condition|, a
 browsingContext.Locator = (
    browsingContext.CssLocator /
    browsingContext.InnerTextLocator /
+   browsingContext.AccessibilityLocator /
```
whimboo, line 2196: Please move this line up as well.

index.bs
```diff
@@ -2203,6 +2210,14 @@ browsingContext.InnerTextLocator = {
    ? maxDepth: js-uint,
 }
 
+browsingContext.AccessibilityLocator = {
+   type: "accessibility",
+   value: {
+    ? role: text,
+    ? name: text,
```
whimboo, line 2217:
```suggestion
    ? name: text,
    ? role: text,
```

Beside that one more question... In WebDriver classic we have [`Get Computed Label`](https://w3c.github.io/webdriver/#get-computed-label), so should we use `label` instead of `name`?

index.bs
whimboo: You are right. Interesting that I missed that.

Beside that one question. For the `locate nodes using inner text` locator we also have a `maxDepth` optional argument to limit the search on the DOM. I wonder if we should have it here as well given that computing the role and label might be expensive as well, or is that not the case?

index.bs
```diff
@@ -3076,6 +3091,57 @@ and |session|:
...
+
+  1. Let |child node matches| be the result of [=locate nodes using accessibility attributes=]
+     with |child nodes|, |selector|, |maximum returned node count|, and |session|.
+
+  1. [=list/Extend=] |returned nodes| with |child node matches|.
+
+1. If |maximum returned node count| is not null, [=list/remove=] all entries
+   in |returned nodes| with an index greater than or equal to |maximum returned
+   node count|.
+
```
whimboo, line 3140: When seeing these lines again I actually wonder if that behavior is correct. Consider a node with 2 children whereby the first children has a lot of children as well. If we have collected more matching nodes as allowed, the 2nd child of the original context node wouldn't be included.

I wonder if we should flatten the array at the very end from top to bottom to not actually loose important already collected nodes. It's not blocking this PR from landing but we should at least file a new issue to further discuss (if necessary) and fix.

whimboo (changes requested):

index.bs
OrKoN: I am not sure: for now it was not an issue in Puppeteer but we could add it later if it becomes a problem (usually those attributes are cached for a11y purposes anyway). There is `maximum returned node count` to limit the number of matches returned and abort early and that helps with the expensiveness of the search.

index.bs
OrKoN: I think you are right and it looks like the innerText selector has the same issue, let me look over it again

index.bs
OrKoN: I am not sure why it is called label since it references accessible name in the spec (the spec name is "Accessible Name and Description Computation 1.1")

index.bs
OrKoN: Rewrote this part.

index.bs
whimboo: I can see: 

> The purpose of the computation is to create a [perceivable](https://www.w3.org/TR/accname-1.1/#dfn-perceivable) label or description for alternative presentations, in the form of a flat string of space separated textual tokens.

@jgraham what's your take on that?

index.bs
whimboo: I reopened this issue for awaiting the results of the check.

OrKoN: I have rewrote that part also addressing this issue (marked wrong thread).

sadym-chromium (approved):

index.bs
juliandescottes: I would be in the favor of name as well, since this is about accessible name. The accessible name might be derived from an aria-label, but not necessarily (can come from title etc...).
Also, both Firefox and Chrome DevTools expose this as "name" in the UI, so it could be helpful for users writing tests if it matches our APIs.

whimboo: Ah good to know. Aligning what DevTools has might actually be also a good idea and will help web developers and testers to align

index.bs
```diff
@@ -3405,6 +3420,56 @@ and |session|:
 
 </div>
 
+<div algorithm="locate nodes using accessibility attributes">
+To <dfn>locate nodes using accessibility attributes</dfn> with given |context nodes|, |selector|,
+|maximum returned node count|, |session|, and |returned nodes|:
+
+1. If |selector|'s <code>role</code> and |selector|'s <code>name</code> are missing, return [=error=] with [=error code=]
```
jgraham, line 3427: It technically doesn't matter, but I don't love this organisation where we recursively call into the algorithm and re-perform steps that are actually redundant.

I think I'd prefer that we pull out the parts that actually make sense to run in each recursive call into their own algorithm and just call that.

index.bs
```diff
@@ -3405,6 +3420,56 @@ and |session|:
...
+
+    1. Let |name| be the result of computing the [=accessible name=] of |context node|.
+
+    1. If |selector|'s <code>name</code> [=is|is not=] |name|:
+
+      1. Set |match| to false.
+
+  1. If |match| is true:
+
+    1. [=list/Append=] |context node| to |returned nodes|.
```
jgraham, line 3461: So we can end up with more than `|maximum returned node count|` nodes being returned? That seems to violate the assert in the calling steps. I think we either need to check each time we append, or truncate the list at the end.

jgraham (changes requested):

christian-bromann (approved):
This will be very helpful for WebdriverIO not have to create weird xpath selectors that aren't 100% correct all the time 

👍

lutien:
@OrKoN, are you planning to add wpt tests for it, once the spec is merged?

OrKoN:
@lutien yes

index.bs
```diff
@@ -3406,6 +3421,68 @@ and |session|:
...
+To <dfn>collect nodes using accessibility attributes</dfn> with given |context nodes|, |selector|,
+|maximum returned node count|, and |returned nodes|:
+
+1. If |returned nodes| is null:
+
+  1. Set |returned nodes| to an empty [=/list=].
+
+1. For each |context node| in |context nodes|:
+
+  1. Let |match| be true.
```
whimboo, line 3434: Should this by default really be true? It would mean that if a selector is used that doesn't contain any `name` or `role` we would match all the nodes in the document. This could be kinda invasive for huge documents. Or can a computed role / name never be `undefined`?

Nevertheless isn't it better to set it to `false` here by default and to `true` only when a match has found?

index.bs
```diff
@@ -3406,6 +3421,68 @@ and |session|:
...
+
+  1. If |selector|'s <code>role</code> is present:
+
+    1. Let |role| be the [=computed role=] of |context node|.
+
+    1. If |selector|'s <code>role</code> [=is|is not=] |role|:
```
whimboo, line 3440: Not sure how complex the computation of both the role and name are, but we won't have to do it when the given key hasn't been specified, right?

jgraham (changes requested):

index.bs
OrKoN: That is checked in the beginning of the main algorithm, so at least one attribute will be provided. Setting default to false makes the text more verbose as it cannot be flipped to true based on a single attribute check.

index.bs
OrKoN: yes, I believe the key is checked above at line 3436

index.bs
```diff
@@ -3406,6 +3421,68 @@ and |session|:
...
+    1. If |selector|'s <code>name</code> [=is|is not=] |name|:
+
+      1. Set |match| to false.
+
+  1. If |match| is true:
+
+    1. If |maximum returned node count| is not null and [=list/size=] of |returned
+      nodes| is greater or equal to |maximum returned node count|:
+
+      1. Return [=success=] with data |returned nodes|.
```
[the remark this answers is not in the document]
jgraham: I mean that it shouldn't say "is greater [than] or equal to" if it's an invariant that it can't be greater than, which the later assert requires that it is. So I think we should remove "if greater or".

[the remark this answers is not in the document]
OrKoN: Also, I think a break would continue the recursion so it has to be a return to prevent redundant iterations?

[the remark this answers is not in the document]
jgraham: Unless I'm misunderstanding the semantics there's only one loop here? And neither break nor return will propogate through recursive calls. Anyway either is fine.

index.bs
```diff
@@ -3406,6 +3421,69 @@ and |session|:
...
+
+  1. Let |child nodes| be an empty [=/list=] and, for each node |child| in the
+     <a spec=dom>children</a> of |context node|:
+
+    1. If |child| implements {{Element}}, [=list/append=] |child| to |child nodes|.
+
+  1. [=Try=] to [=collect nodes using accessibility attributes=] with |child nodes|,
+     |selector|, |maximum returned node count|, and |returned nodes|.
+
+  1. Return [=success=] with data |returned nodes|.
```
jgraham, line 3469: This is indented too much, right? i.e. the return should be after the loop, otherwise we only ever process the first context node and the first element children below it.

jgraham (changes requested):

index.bs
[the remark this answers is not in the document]
OrKoN: Yeah, I have misread the indentation

index.bs
```diff
@@ -3482,6 +3558,13 @@ The [=remote end steps=] with |session| and |command parameters| are:
             given |context nodes|, |selector|, |max depth|, |match type|,
             |ignore case|, |maximum returned node count| and |session|.
 
+      <dt>|type| is the string "<code>accessibility</code>"
+      <dd>
+         1. Let |selector| be |locator|["<code>value</code>"].
+
+         1. Let |result nodes| be a result of [=trying=] to [=locate nodes using accessibility attributes=]
```
jgraham, line 3565:
```suggestion
         1. Let |result nodes| be [=locate nodes using accessibility attributes=]
```

Since we're no longer returning a result type, we can't use `[=trying=]`.

jgraham (approved):
But note the small fixup required.

Issue #516, opened 2023-08-23
Extend `browsingContext.Info` with `opener`


Lightning00Blade:
In CDP there is a concept `opener` exits related to new Windows/Tab being open from another Window/Tab.
Puppeteer also supports this for CDP via the `page.on('popup')`.
Ref [CDP TargeInfo](https://chromedevtools.github.io/devtools-protocol/tot/Target/#type-TargetInfo)

The proposed solution would be add a property on the `browsingContext.Info Type` called `opener`, that will be `browsingContext.BrowsingContext / null`  of either the browsing context or the top level browsing context.

Example of such scenarios:

-  evaluation of `window.open('about:blank')`.
- clicking a anchor `<a target=_blank href="/">open</a>`.

OrKoN:
Sounds reasonable to me.

OrKoN:
cc @jgraham @whimboo @juliandescottes do you have any concerns about this?

jgraham:
This seems fine.

OrKoN:
I think the difficultly here is that when the HTML spec calls the algorithm to emit the browsing context create event, there is no original opener ID available in that algorithm and I am not sure what is the best way to make it available. Note that Puppeteer requires an opener ID even if the browsing context is loaded with noopener=true. There is a place for saving the original opener after the context is created but it is too late for including it in the created event: https://github.com/w3c/webdriver-bidi/pull/664

jgraham:
In gecko this is easy to implement in the case that we have `window.opener` but hard in cases where `window.opener` is null (in that case we discard knowledge of the original opener early, like the spec). So a question is how much of the use case here depends on always having opener, vs only having it in cases where `window.opener` exists.

Lightning00Blade:
The PR for this has already landed, closing.

PR #664, opened 2024-02-20
Add opener to browsing context info

OrKoN:
Issue https://github.com/w3c/webdriver-bidi/issues/516
HTML spec https://github.com/whatwg/html/pull/10402

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/664.html" title="Last updated on Jun 11, 2024, 11:41 AM UTC (ee59b1f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/664/5614cf1...ee59b1f.html" title="Last updated on Jun 11, 2024, 11:41 AM UTC (ee59b1f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2335,7 +2335,7 @@ browsingContext.InfoList = [*browsingContext.Info]
 browsingContext.Info = {
   children: browsingContext.InfoList / null,
   context: browsingContext.BrowsingContext,
-  opener: browsingContext.BrowsingContext / null,
+  originalOpener: browsingContext.BrowsingContext / null,
   url: text,
   userContext: browser.UserContext,
   ? parent: browsingContext.BrowsingContext / null,
@@ -2430,7 +2430,7 @@ To <dfn>get the browsing context info</dfn> given |context|,
    field set to |context id|, the <code>parent</code> field set to |parent id|
    if |is root| is <code>true</code>, or unset otherwise, the <code>url</code>
    field set to |url|, the <code>userContext</code> field set to
-   |user context|'s [=user context id=], <code>opener</code>
+   |user context|'s [=user context id=], <code>originalOpener</code>
    field set to |opener id|, and the <code>children</code> field
    set to |child infos|.
 
--- a/index.bs
+++ b/index.bs
@@ -231,6 +231,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
     text: window open steps; url: window-object.html#window-open-steps
     text: worker event loop; url: webappapis.html#worker-event-loop-2
     text: worklet global scopes; url:worklets.html#concept-document-worklet-global-scopes
+    text: the rules for choosing a navigable; url:document-sequences.html#the-rules-for-choosing-a-navigable
 spec: RESOURCE-TIMING; urlPrefix: https://w3c.github.io/resource-timing/
   type: dfn
     text: convert fetch timestamp; url: dfn-convert-fetch-timestamp
@@ -2307,6 +2308,10 @@ same as the [=window handle=].
 Each [=/browsing context=] also has an <dfn>associated storage partition</dfn>,
 which is the [=storage partition=] it uses to persist data.
 
+Each [=/browsing context=] also has an <dfn>original opener</dfn>, which
+is a [=/browsing context=] that opened the browsing context, initially
+set to null.
+
 <div algorithm>
 To <dfn export>get a browsing context</dfn> given |context id|:
 
@@ -2331,8 +2336,9 @@ browsingContext.InfoList = [*browsingContext.Info]
 browsingContext.Info = {
   children: browsingContext.InfoList / null,
   context: browsingContext.BrowsingContext,
+  opener: browsingContext.BrowsingContext / null,
   url: text,
-  userContext: browser.UserContext
+  userContext: browser.UserContext,
   ? parent: browsingContext.BrowsingContext / null,
 }
 </pre>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2287,6 +2288,10 @@ same as the [=window handle=].
 Each [=/browsing context=] also has an <dfn>associated storage partition</dfn>,
 which is the [=storage partition=] it uses to persist data.
 
+Each [=/browsing context=] also has an <dfn>original opener</dfn>, which
```
Lightning00Blade, line 2291: While reading this paragraph, it makes you think that there is always an opener, which is not the case.
Something like below should be more accurate 
```
is a [=/browsing context=] that opened the browsing context or null.

**Review by Lightning00Blade — APPROVED** [review]
LGTM + small comment.

File: index.bs [hunk 2]
```diff
@@ -2287,6 +2288,10 @@ same as the [=window handle=].
...
+is a [=/browsing context=] that opened the browsing context, initially
```
OrKoN, line 2292:
```suggestion
is a [=/browsing context=] that opened the browsing context or null, initially
```

index.bs
```diff
@@ -4039,6 +4050,11 @@ When the [=create a new browsing context=] algorithm is invoked, after the
 
 1. Let |context| be the newly created browsing context.
 
+1. Set |context|'s [=original opener=] to |context|'s [=opener browsing
+   context=], as if the |context|'s [=opener browsing context=] was
+   created according to [=the rules for choosing a navigable=] but with
+   <code>noopener</code> always set to false.
```
OrKoN, line 4056: @jgraham WDYT about defining the original opener like this? Another way to specify it would be to update the HTML-spec but it would require many changes since the opener is discarded before the algorithm descends to notify WebDriver BiDi about the browsing context creation.

OrKoN:
@whimboo @jgraham could you please review? Do you have suggestions for a better spec text?

OrKoN:
@whimboo could you please check if the implementation of this is feasible in Firefox? If it is, I will proceed with a PR to the HTML spec to store the original opener ID.

whimboo:
> @whimboo could you please check if the implementation of this is feasible in Firefox? If it is, I will proceed with a PR to the HTML spec to store the original opener ID.

We discussed with DOM folks and as of now there is no such ability in Gecko. I've filed https://bugzilla.mozilla.org/show_bug.cgi?id=1891670 to get it added but it might take some time.

whimboo:
> We discussed with DOM folks and as of now there is no such ability in Gecko. I've filed https://bugzilla.mozilla.org/show_bug.cgi?id=1891670 to get it added but it might take some time.

There is a patch available that will hopefully land soon. So at least from our side it should no longer be a blocker to get this feature added to WebDriver BiDi.

OrKoN:
@whimboo @jgraham I have updated the PR so that it relies on the direct integration with the HTML spec. PTAL and if it looks good I can start finalizing the HTML spec change.

index.bs
```diff
@@ -4160,10 +4170,13 @@ To <dfn>Emit a context created event</dfn> given |session| and |context|:
 
 The [=remote end event trigger=] is
 the <dfn export>WebDriver BiDi navigable created</dfn> steps given
```
sadym-chromium, line 4172: FYI this export link is not correct

index.bs
```diff
@@ -2307,6 +2307,10 @@ same as the [=window handle=].
 Each [=/browsing context=] also has an <dfn>associated storage partition</dfn>,
 which is the [=storage partition=] it uses to persist data.
 
+Each [=/browsing context=] also has an associated <dfn>original
+opener</dfn>, which is a [=/browsing context=] that caused the creation
```
sadym-chromium, line 2311: NIT: we don't have a rule for this, but it would be really helpful to avoid line breaks in the tags. Otherwise editing spec become more chellenging.

index.bs
```diff
@@ -4160,10 +4170,13 @@ To <dfn>Emit a context created event</dfn> given |session| and |context|:
 
 The [=remote end event trigger=] is
 the <dfn export>WebDriver BiDi navigable created</dfn> steps given
-|navigable|:
+|navigable| and |opener navigable|:
 
 1. Let |context| be |navigable|'s [=active browsing context=].
 
+1. Set |context|'s [=original opener=] to |opener navigable|'s [=active
+   browsing context=].
```
sadym-chromium, line 4178: NIT: Same line break here

sadym-chromium (approved):

index.bs
OrKoN: WDYM?

OrKoN: The link works from here https://html.spec.whatwg.org/#:~:text=The%20following%20terms%20are%20defined%20in%20WebDriver%20BiDi%3A%20%5BWEBDRIVERBIDI%5D

index.bs
```diff
@@ -2331,8 +2335,9 @@ browsingContext.InfoList = [*browsingContext.Info]
 browsingContext.Info = {
   children: browsingContext.InfoList / null,
   context: browsingContext.BrowsingContext,
+  opener: browsingContext.BrowsingContext / null,
```
jgraham, line 2338: Could we call this `originalOpener` perhaps, to make it clearer it's not the same as `window.opener`?

jgraham (approved):

Yooch3 (approved):

Lightning00Blade:
@whimboo I am currently working on the test for this feature.

PR #666, opened 2024-02-20
chore: remove pointer defaults

jrandolf:
We enforce the default pointer properties even when the pointer properties may already be known from previous commands (such as `pointerdown`). We should only use the default for `pointerdown` events.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/666.html" title="Last updated on Feb 20, 2024, 1:50 PM UTC (fbf070f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/666/42abefe...fbf070f.html" title="Last updated on Feb 20, 2024, 1:50 PM UTC (fbf070f)">Diff</a>


OrKoN:
@jrandolf is it in the spec that the properties can be inferred?

jrandolf:
> @jrandolf is it in the spec that the properties can be inferred? (could you please add a link)

This is implementation-defined. See https://www.w3.org/TR/webdriver2/#dfn-dispatch-a-pointerdown-action

OrKoN (approved):
LGTM (as the classic spec does not require actual default values). @juliandescottes @jgraham PTAL

jgraham:
The defaults match the defaults if you contrust an event in the DOM API, per: https://www.w3.org/TR/pointerevents/#pointer-events-and-interfaces. I think that's what's intended by classic, and saying "this is just implementation defined" is not OK (authors should be able to depend on event properties being consistent between browsers for a given input).

jrandolf:
> https://www.w3.org/TR/pointerevents/#pointer-events-and-interfaces

The defaults don't really make sense for a lot of events. For example, for pointer, the default pressure is always 0.5 (try it on any browser)

jgraham:
That seems like a reason to update those specs, rather than a reason to throw our hands up and say "implementation defined".

jrandolf:
> https://www.w3.org/TR/pointerevents/#pointer-events-and-interfaces

`PointerEventInit` is used for custom events built from the DOM. It does not reflective what is expected from the automation side. We could take your arguments and apply them to keyboard events and etc, but this wouldn't benefit an automation user. In particular, keyboard events, e.g., allow setting the `ctrl` state, but we know for automation they problem want it to integrate with the browser state.

As an alternative, we can try to specify the defaults within WebDriver in detail.

PR #668, opened 2024-02-22
Throw for setFiles on single inputs

jrandolf:


OrKoN (approved):

jgraham (approved):

jgraham:
Although, I just noticed that this entire bit assumes we can access properties on elements like `|element|["<code>property</code>"]`, which we just totally can't.

PR #669, opened 2024-02-22
Update property access in setFiles

jrandolf:


OrKoN (approved):

index.bs
```diff
@@ -10505,12 +10505,13 @@ The [=remote end steps=] given |session| and |command parameters| are:
 1. If |element| doesn't implement {{Element}}, return [=error=] with [=error code=]
    [=no such element=].
 
-1. If |element| doesn't implement {{HTMLInputElement}},
-   |element|["<code>type</code>"] is not "file", or
-   |element|["<code>disabled</code>"] is true, return [=error=] with [=error
-   code=] [=unable to set file input=].
+1. If |element| doesn't implement {{HTMLInputElement}}, |element|'s
+   <{input/type}> attribute is not "<code>file</code>", or |element|'s <{input/disabled}>
```
jgraham, line 10509: For disabled I think we should use https://html.spec.whatwg.org/#concept-fe-disabled

jgraham, line 10509: For the type attribute, we should maybe use https://html.spec.whatwg.org/#file-upload-state-(type=file) i.e. `|element is not in the [=file upload state=]`

index.bs
```diff
@@ -10505,12 +10505,13 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+   attribute is true, return [=error=] with [=error code=] [=unable to set file
+   input=].
 
-1. If the [=set/size=] of |files| is greater than 1 and |element|["multiple"] is
+1. If the [=set/size=] of |files| is greater than 1 and |element|'s
+   <{input/multiple}> attribute is
```
jgraham, line 10514: Instead of true/false I think we should use set/not set to match HTML.

jgraham (changes requested):

index.bs
```diff
@@ -10505,13 +10507,13 @@ The [=remote end steps=] given |session| and |command parameters| are:
 1. If |element| doesn't implement {{Element}}, return [=error=] with [=error code=]
    [=no such element=].
 
-1. If |element| doesn't implement {{HTMLInputElement}},
-   |element|["<code>type</code>"] is not "file", or
-   |element|["<code>disabled</code>"] is true, return [=error=] with [=error
-   code=] [=unable to set file input=].
+1. If |element| doesn't implement {{HTMLInputElement}}, |element|'s
+   <{input/type}> is in the [=File Upload state=], or |element| is [=disabled=],
```
jgraham, line 10511:
```suggestion
   <{input/type}> is not in the [=File Upload state=], or |element| is [=disabled=],
```

jgraham (approved):
Approved with change

Issue #635, opened 2024-01-09
Clarify `script.Target` behavior when both `realm` and `context` are provided  


Lightning00Blade:
The spec defines this behavior [here](https://w3c.github.io/webdriver-bidi/#get-a-realm-from-a-target)
Reading though it we first check for `context + ?sandbox` then for `realm` but as a user I may send both `context + realm`.
The can be done as an oversight as in sending `script.Source` for the `target` in `script.evaluate/script.callFunction`.
Currently Firefox fails while Chromium works when both are provided.
We should align this on the spec level with correct type of Error (invalid argument ?).

lutien:
For the context, there was already some discussion of this topic in this issue: https://github.com/w3c/webdriver-bidi/issues/274.

whimboo:
@jgraham do you think that the idea from my last comment to have an optional `type` field for now and which will be changed to required later could work?

PR #680, opened 2024-03-07
Fix order of groups in script.Target

OrKoN:
The [spec text defines that script.ContextTarget is matched first](https://w3c.github.io/webdriver-bidi/#get-a-realm-from-a-target). The CDDL spec also [picks groups in order](https://datatracker.ietf.org/doc/html/rfc8610#appendix-C:~:text=The%20choice%20uses%20PEG%20semantics%20as%0A%20%20%20discussed%20in%20Appendix%20A%3A%20the%20first%20choice%20that%20matches%20wins). Therefore, the ContextTarget needs to be defined first in the group.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/680.html" title="Last updated on Mar 7, 2024, 2:25 PM UTC (ebf8cbd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/680/f7e2538...ebf8cbd.html" title="Last updated on Mar 7, 2024, 2:25 PM UTC (ebf8cbd)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -8614,8 +8614,8 @@ script.ContextTarget = {
 }
 
 script.Target = (
-  script.RealmTarget /
-  script.ContextTarget
+  script.ContextTarget /
+  script.RealmTarget
 );
 </pre>
 


whimboo:
@OrKoN I would still propose that we fix it as suggested on https://github.com/w3c/webdriver-bidi/issues/635 by for now having an optional `type` field added.

Having the processing in order of definition might require us to also check other CDDL definitions if those are valid.

jrandolf (approved):

OrKoN:
> @OrKoN I would still propose that we fix it as suggested on #635 by for now having an optional `type` field added.
> 
> Having the processing in order of definition might require us to also check other CDDL definitions if those are valid.

I am afraid there is no way around processing groups in order since it is defined by the CDDL spec? I think this change does not block the more general fix by using the type attribute and it does not affect implementations that do not rely strictly on CDDL.

OrKoN:
@whimboo it also makes the order alphabetical :)

whimboo (approved):

PR #681, opened 2024-03-08
Prompt handler

jgraham:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/681.html" title="Last updated on Jun 24, 2024, 9:11 AM UTC (2ba6dce)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/681/a86a0c7...2ba6dce.html" title="Last updated on Jun 24, 2024, 9:11 AM UTC (2ba6dce)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1667,12 +1667,9 @@ session.SystemProxyConfiguration = (
 [=Remote end definition=] and [=local end definition=]
 
 <pre class="cddl remote-cddl local-cddl">
-session.UserPromptHandler = session.UserPromptHandlerType /
-                            session.UserPromptHandlerConfiguration;
-
 session.UserPromptHandlerType = "accept" / "dismiss" / "ignore";
 
-session.UserPromptHandlerConfiguration = {
+session.UserPromptHandler = {
   ? alert: session.UserPromptHandlerType,
   ? beforeUnload: session.UserPromptHandlerType,
   ? confirm: session.UserPromptHandlerType,
--- a/index.bs
+++ b/index.bs
@@ -2658,6 +2658,17 @@ browsingContext.ReadinessState = "none" / "interactive" / "complete"
 The <code>browsingContext.ReadinessState</code> type represents the stage of
 document loading at which a navigation command will return.
 
+#### The browsingContext.UserPrompt Type #### {#type-browsingContext-UserPrompt}
+
+[=Remote end definition=] and [=local end definition=]
+
+<pre class="cddl remote-cddl local-cddl">
+browsingContext.UserPromptType = "alert" / "confirm" / "prompt" / "beforeunload";
+</pre>
+
+The <code>browsingContext.UserPrompt</code> type represents the possible user
+prompt types.
+
 ### Commands ### {#module-browsingContext-commands}
 
 #### The browsingContext.activate Command ####  {#command-browsingContext-activate}
@@ -4563,7 +4574,7 @@ failed</dfn> steps given |context| and |navigation status|:
         browsingContext.UserPromptClosedParameters = {
           context: browsingContext.BrowsingContext,
           accepted: bool,
-          type: "alert" / "confirm" / "prompt" / "beforeunload",
+          type: browsingContext.UserPromptType,
           ? userText: text
         }
       </pre>
@@ -4614,7 +4625,7 @@ closed</dfn> steps given |window|, |type|, |accepted| and optional |user text|
           context: browsingContext.BrowsingContext,
           handler: "accept" / "dismiss" / "ignore",
           message: text,
-          type: "alert" / "confirm" / "prompt" / "beforeunload",
+          type: browsingContext.UserPromptType,
           ? defaultValue: text
         }
       </pre>
--- a/index.bs
+++ b/index.bs
@@ -90,6 +90,7 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
     text: no such element; url: dfn-no-such-element
     text: no such frame; url: dfn-no-such-frame
     text: parse a page range; url: dfn-parse-a-page-range
+    text: handler; for: prompt handler configuration; url: dfn-handler
     text: process capabilities; url: dfn-processing-capabilities
     text: readiness state; url: dfn-readiness-state
     text: remote end steps; url: dfn-remote-end-steps
@@ -231,6 +232,9 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
     text: window open steps; url: window-object.html#window-open-steps
     text: worker event loop; url: webappapis.html#worker-event-loop-2
     text: worklet global scopes; url:worklets.html#concept-document-worklet-global-scopes
+spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
+  type: dfn
+    text: convert a JSON-derived JavaScript value to an Infra value; url: convert-a-json-derived-javascript-value-to-an-infra-value
 spec: RESOURCE-TIMING; urlPrefix: https://w3c.github.io/resource-timing/
   type: dfn
     text: convert fetch timestamp; url: dfn-convert-fetch-timestamp
@@ -1567,7 +1571,7 @@ session.CapabilityRequest = {
   ? browserVersion: text,
   ? platformName: text,
   ? proxy: session.ProxyConfiguration,
-  ? webSocketUrl: bool,
+  ? unhandledPromptBehavior: session.UserPromptHandler,
   Extensible
 };
 </pre>
@@ -1657,6 +1661,28 @@ session.SystemProxyConfiguration = (
 )
 
 </pre>
+
+#### The session.UserPromptHandler Type #### {#type-session-UserPromptHandler}
+
+[=Remote end definition=] and [=local end definition=]
+
+<pre class="cddl remote-cddl local-cddl">
+session.UserPromptHandler = session.UserPromptHandlerType /
+                            session.UserPromptHandlerConfiguration;
+
+session.UserPromptHandlerType = "accept" / "dismiss" / "ignore";
+
+session.UserPromptHandlerConfiguration = {
+  ? alert: session.UserPromptHandlerType,
+  ? beforeUnload: session.UserPromptHandlerType,
+  ? confirm: session.UserPromptHandlerType,
+  ? default: session.UserPromptHandlerType,
+  ? prompt: session.UserPromptHandlerType,
+};
+</pre>
+
+The <code>session.UserPromptHandler</code> type represents the configuration of
+the user prompt handler.
 
 #### The session.SubscriptionRequest Type #### {#type-session-SubscriptionRequest}
 
@@ -1758,6 +1784,7 @@ This is a [=static command=].
           setWindowRect: bool,
           userAgent: text,
           ? proxy: session.ProxyConfiguration,
+          ? unahandledPromptBehavior: session.UserPromptHandler,
           ? webSocketUrl: text,
           Extensible
         }
@@ -1777,8 +1804,11 @@ The [=remote end steps=] given |session| and |command parameters| are:
 
 1. Let |flags| be a [=/set=] containing "<code>bidi</code>".
 
-1. Let |capabilities| be the result of [=trying=] to [=process capabilities=]
+1. Let |capabilities json| be the result of [=trying=] to [=process capabilities=]
    with |command parameters| and |flags|.
+
+1. Let |capabilities| be [=convert a JSON-derived JavaScript value to an Infra
+   value=] with |capabilities json|.
 
 1. Let |session| be the result of [=trying=] to [=create a session=] with
    |capabilities| and |flags|.
@@ -4583,8 +4613,9 @@ closed</dfn> steps given |window|, |accepted| and optional |user text|
 
         browsingContext.UserPromptOpenedParameters = {
           context: browsingContext.BrowsingContext,
-          type: "alert" / "confirm" / "prompt" / "beforeunload",
+          handler: "accept" / "dismiss" / "ignore",
           message: text,
+          type: "alert" / "confirm" / "prompt" / "beforeunload",
           ? defaultValue: text
         }
       </pre>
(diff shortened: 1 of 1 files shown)


jgraham:
https://github.com/whatwg/html/pull/10189 is the HTML side of this change. It also depends on landing https://github.com/w3c/webdriver/pull/1791

index.bs
```diff
@@ -1550,7 +1554,7 @@ session.CapabilityRequest = {
   ? browserVersion: text,
   ? platformName: text,
   ? proxy: session.ProxyConfiguration,
-  ? webSocketUrl: bool,
+  ? userPromptHandler: session.UserPromptHandler,
```
whimboo, line 1557: The capability is named `unhandledPromptBehavior`.

index.bs
```diff
@@ -1641,6 +1645,28 @@ session.SystemProxyConfiguration = (
 
 </pre>
 
+#### The session.UserPromptHandler Type #### {#type-session-UserPromptHandler}
+
+[=Remote end definition=] and [=local end definition=]
+
+<pre class="cddl remote-cddl local-cddl">
+session.UserPromptHandler = session.UserPromptHandlerType /
```
whimboo, line 1653: We should also keep support for the string type capability given that `accept and notify` and `dismiss and notify` are also valid options, and those don't match with the `UserPromptHandlerConfiguration` type.

index.bs
```diff
@@ -1741,6 +1767,7 @@ This is a [=static command=].
           setWindowRect: bool,
           userAgent: text,
           ? proxy: session.ProxyConfiguration,
+          ? userPromptHandler: session.UserPromptHandler,
```
whimboo, line 1770: This also needs to be `unhandledPromptBehavior` similar to the above mentioned case.

index.bs
```diff
@@ -4422,8 +4452,9 @@ closed</dfn> steps given |window|, |accepted| and optional |user text|
 
         browsingContext.UserPromptOpenedParameters = {
           context: browsingContext.BrowsingContext,
-          type: "alert" / "confirm" / "prompt" / "beforeunload",
+          handler: "accept" / "dismiss" / "ignore",
           message: text,
+          type: "alert" / "confirm" / "prompt" / "beforeunload",
```
whimboo, line 4457: Was there actually a reason why we didnt' add the `type` to the `browsingContext.UserPromptClosedParameters` as well? I think it would be good to know for the client which type of alert was closed.

index.bs
```diff
@@ -4422,8 +4452,9 @@ closed</dfn> steps given |window|, |accepted| and optional |user text|
 
         browsingContext.UserPromptOpenedParameters = {
           context: browsingContext.BrowsingContext,
-          type: "alert" / "confirm" / "prompt" / "beforeunload",
+          handler: "accept" / "dismiss" / "ignore",
```
whimboo, line 4455: Should we also inform the client about the `notify` flag?

whimboo (changes requested):
Sorry for the delay @jgraham. But I decided to wait a bit with further feedback until the general handling of the capability has finished. I've a couple of questions that I've added inline. Maybe you can have a look at these? Thanks.

index.bs
jgraham:
```suggestion
  ? unhandledPromptBehavior: session.UserPromptHandler,
```

index.bs
jgraham: The `and notify` variants are just for the `Unexpected User Prompt Open` error and so don't have any effect on BiDi-only sessions. You can still subscribe to the events and if there's a prompt handler installed you'll get events for the prompt opening and being dismissed. Therefore I don't think this makes sense.

index.bs
jgraham:
```suggestion
          ? unahandledPromptBehavior: session.UserPromptHandler,
```

index.bs
jgraham: Again, I don't think it makes sense. For a HTTP session you'll get an error. For a BiDi-only session you are being notified by the event.

index.bs
jgraham: We could. In theory you can know as long as you subscribe to both events, since the prompt that's closed is always the last prompt opened for that context (as long as we don't support multiple prompts). I don't really know why you'd subscribe to the close event but not the open event, but we can make things a bit easier for anyone who does that.

index.bs
whimboo: So what should happen if no event handler is set for a BiDi-only session, which means that prompts are not handled in the case of the `ignore` setting. Does it mean we always raise an error? But when? None of the commands for BiDi check if a prompt is still open and potentially would fail with some random error depending on what the command is trying to accomplish.

index.bs
whimboo: Sure, but one option I could imagine is that you have set the handler to `accept` or `dismiss` to automatically close those prompts, but just want to know which kind of prompts were open at which times. But that might not be used that often.

whimboo:
@jgraham as it looks like you missed to specify the [steps for handling a user prompt](https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/681.html#command-browsingContext-handleUserPrompt).

index.bs
jgraham: It means you get events and have to handle the prompts yourself using `browsingContext.handleUserPrompt`.

I don't think other commands should error, but they might be blocked on the prompt if they need to interact with the js event loop. Some commands don't (e.g. those that can run entirely in the parent process). It's probably technically specified how this works, but I kind of doubt the specifications match browsers in practice. 

We might want to tighten this up in the longer term, but I don't know what we could say in the short term. Unless we have some evidence that say CDP takes a specific approach here that we'd like to copy.

whimboo: Ok, as it sounds like it worth a follow-up issue then, also given that other prompt types will trigger the exact same situation and it's not just related to `beforeunload` prompts.

index.bs
```diff
@@ -1641,6 +1645,28 @@ session.SystemProxyConfiguration = (
...
+session.UserPromptHandler = session.UserPromptHandlerType /
+                            session.UserPromptHandlerConfiguration;
+
+session.UserPromptHandlerType = "accept" / "dismiss" / "ignore";
+
+session.UserPromptHandlerConfiguration = {
+  ? alert: session.UserPromptHandlerType,
+  ? beforeUnload: session.UserPromptHandlerType,
+  ? confirm: session.UserPromptHandlerType,
+  ? default: session.UserPromptHandlerType,
```
OrKoN, line 1662: If we want to extend it to file dialogs and auth prompts (it was mentioned previously), we probably should not have a default handler and require defining prompt handlers per type since we won't be able to support 'accept' for file and auth prompts.

index.bs
```diff
@@ -4457,6 +4494,10 @@ opened</dfn> steps given |window|, |type|, |message|, and optional |default valu
 
   1. [=Emit an event=] with |session| and |body|.
 
+1. If |handler| is "<code>ignore</code>", set handler to "<code>none</code>".
+
+1. Return |handler|
```
OrKoN, line 4499:
```suggestion
1. Return |handler|.
```

index.bs
jgraham: We could remove it from BiDi, but we'd still need to specify it for classic because historically that's only had a default handler, not per-type handlers. 

For `alert` `accept` and `dismiss` already do the same thing, so one option for `file` would be to make both of them act like `dismiss`, or make `accept` act like `ignore`. For HTTP Auth arguably `accept` and `dismiss` both make sense, although `accept` sending empty credentials does seem like it has a high chance of ending up in a loop.

OrKoN: Perhaps it makes sense to only allow `dismiss` and `ignore` (as though seem to be supported by any prompt type) for the default UserPromptHandlerType in BiDi? I think it is perhaps not very important but my concern is that we will eventually be defining 'accept'  handlers where it makes no much sense just to be compatible with the list of possible default values.

OrKoN (approved):
LGTM

index.bs
jgraham: I think we'll need to define them anyway for as long as we still care about classic, because classic implementations previously only allowed a single string value as the prompt handler. So if we start invoking these algorithms for new prompt types (as we have for `beforeUnload`) we need to specify what happens with those existing string values. Passing an object with only the `default` key set is defined to be exactly the same as defining a single string value.

For `beforeUnload` the current decision is that the `default` value is ignored: if you don't provide an explicit `beforeUnload` key then the prompt is auto-dismissed. That matches the default behavior of classic which is to always close `beforeUnload` prompts.

By the same token we could ignore `default` for new prompt types and always use `ignore` unless a prompt-specific default is supplied. That would also match the legacy behaviour where we just don't handle these other modals at all.

On the other hand, in the long run, this approach of making `default` be "what would classic do?" is itself quite confusing for users or clients that don't care about the classic behaviour. 

So, just thinking about BiDi, we have several options:
* We can continue to support `default` and single string handler values, aligning the behaviour with whatever classic does in those cases. This has the disadvantage of being confusing e.g. it is unclear why `beforeUnload` doesn't respect `default`.
* We can have different semantics for `default` between classic and BiDi. For example we could make `beforeUnload` actually respect `default` for BiDi only sessions. This has the disadvantage of adding even more complexity in transitioning between classic and BiDi, and more complexity in the implementations, plus it doesn't really solve the question of how to handle `"default": "accept"` in the `file` case.
* We could subset the values allowed as default to just `dismiss` and `ignore`, assuming that they make sense for every prompt type, and force writing out any prompts you want to accept longhand e.g. `{"default": "dismiss", "confirm": "accept", "beforeUnload": "accept"}`. We'd also need a default if `default` isn't provided, which basically has the same problem that the legacy default for `alert`, `confirm` and `prompt` is `dismiss`, but `beforeUnload` is `accept` and `file` and `httpAuth` are effectively `ignore`.
* We could remove `default` (and the single-string values) altogether and force writing out everything longhand. That means something like `{"alert": "accept", "prompt": "accept", "confirm": "accept", "beforeUnload": "accept"}`. This is basically like the option above except the "what if default isn't specified" question only occurs when you don't specify a value at all (which is unfortunately probably the most common case).
* We could find some way to rename `default` that better specifies that it doesn't apply to all prompt types (but that doesn't help with string-only values).

I think if we were only concerned with BiDi we probably wouldn't have a default value, but explain that if no prompt handler is supplied the behaviour depends on the modal type, and that the keys override the default for that prompt type.

With the fact that classic has previously _only_ had the default override, and it didn't apply to all modals, it's less clear to me what the right option is.

OrKoN: I would prefer this option:

> We can have different semantics for default between classic and BiDi. For example we could make beforeUnload actually respect default for BiDi only sessions. 

combined with

> We could subset the values allowed as default to just dismiss and ignore

But, in my opinion, the current proposal also looks good although I would find it confusing if we have eventually add more prompts that do not have the auto-accept behavior. Perhaps we could extend "accept" with some with extra parameters, for example, `{"alertParams": {"userText": "default value for accept"}, "fileParams": {"file": "/dev/null"}}`. Perhaps it also would be an option to allow prompt handler to be either `ignore` or an instance of `HandleUserPromptParameters` but I am not sure what the implications for that would be w.r.t to the classic compatibility.

So I think the current proposal looks good for now but I want @sadym-chromium to review this as well.

jgraham: "Different semantics" is my least favourite option, because it means if you transition from HTTP+BiDi to BiDi-only then things mysteriously break without even an error message. For example if you were sending `"dismiss"` as the default prompt handler then suddenly any navigation with a `beforeUnload` prompt will be silently cancelled. Subsetting the allowed values at least makes it clear as soon as you start the session what the problem is.

index.bs
```diff
@@ -1550,7 +1553,7 @@ session.CapabilityRequest = {
   ? browserVersion: text,
   ? platformName: text,
   ? proxy: session.ProxyConfiguration,
-  ? webSocketUrl: bool,
```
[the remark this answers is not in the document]
whimboo: I just want to note that this is not blocking the merge. Given that it hasn't been addressed yet lets just resolve this comment.

index.bs
whimboo: Just one more question in regards of multiple session... When we have a classic session with `dismiss and notify`, and a BiDi client wants to create another session will that actually work with capability matching when it specifies `dismiss` as handler? The `dismiss and notify` will not work as you said above.

jgraham: https://w3c.github.io/webdriver/#dfn-check-user-prompt-handler-matches says that we fail to create the session in that case, but I think you're correct that we could relax that to not care about the "notify" part, since that can only be set by HTTP sessions, doesn't affect, BiDi, and you can only have a single HTTP session.

whimboo (approved):
I think it looks fine now. If I detect something not working as expected we can do it in a follow-up PR. 

Tests will be added as part of my work on https://bugzilla.mozilla.org/show_bug.cgi?id=1824220

OrKoN:
@jgraham so to summarize we will move on with allowing "accept" in the default values or do we still want to completely decouple from classic and drop "accept" in the default configuration and for prompts where it does not make sense? I am not sure if the PR is already updated with the latest discussions? I will take another look if it is ready.

index.bs
```diff
@@ -4514,6 +4541,7 @@ failed</dfn> steps given |context| and |navigation status|:
         browsingContext.UserPromptClosedParameters = {
           context: browsingContext.BrowsingContext,
           accepted: bool,
+          type: "alert" / "confirm" / "prompt" / "beforeunload",
```
OrKoN, line 4544: nit: perhaps it makes sense to extract this into a separate type and re-use between UserPromptClosedParameters and UserPromptOpenedParameters to make sure they are always the same?

OrKoN (approved):

index.bs
sadym-chromium: Note: as discussed in https://github.com/w3c/webdriver-bidi/pull/681#issuecomment-2163474276, the changing `default` semantics is not a concern from the user side. So we decided to go with: "BiDi `default` covers all the prompts".

index.bs
```diff
@@ -4536,6 +4574,7 @@ failed</dfn> steps given |context| and |navigation status|:
         browsingContext.UserPromptClosedParameters = {
           context: browsingContext.BrowsingContext,
           accepted: bool,
+          type: browsingContext.UserPromptType,
```
sadym-chromium, line 4577: This field seems to be tricky to implement on our side. Can we move this part out of the scope of this PR and discuss it separately?

sadym-chromium: UPD: as long as there can be not more than 1 open prompt, it's not an issue.

sadym-chromium (approved):

sadym-chromium (approved):

OrKoN (approved):

Issue #684, opened 2024-03-11
Consider removing `ownership` and `sandbox` from `browsingContext.locateNodes`


sadym-chromium:
`browsingContext.locateNodes` returns an array of `script.NodeRemoteValue`, and `sharedId` can be used instead of `handle` throughout all the sandboxes.

whimboo:
I think (if I remember correctly) we originally added both because of a request from Google. From our side we do not have a special use-case given as you say the `sharedId` is stable as long as the process exists.

sadym-chromium:
> I think (if I remember correctly) we originally added both because of a request from Google. From our side we do not have a special use-case given as you say the `sharedId` is stable as long as the process exists.

@whimboo If we talk about a generic script serialization, we cannot provide a stable `sharedId` for a generic ecmascript objects, only for DOM. But the `browsingContext.locateNodes` returns a list of nodes, and we can do that.

sadym-chromium:
@jimevans so the idea is to serialize with `ownership: none`, and never provide a handle. If so, we don't need a sandbox anymore. The [*serialize as a remote value*](https://w3c.github.io/webdriver-bidi/#serialize-as-a-remote-value) requires realm only to provide handles in 
>**step 4** *Let handle id be the handle for an object with realm, ownership type and value.*

If we don't have a major concerns here, I will prepare a PR.

sadym-chromium:
Here we go: https://github.com/w3c/webdriver-bidi/pull/688

sadym-chromium:
Implemented in https://github.com/w3c/webdriver-bidi/pull/688.

PR #688, opened 2024-03-12
Default `ownership` to `none` in `browsingContext.locateNodes`

sadym-chromium:
Addressing https://github.com/w3c/webdriver-bidi/issues/684.

Support only `ownership: none` in `browsingContext.locateNodes`. For that, _"get a realm from a browsing context"_ had to be extracted from _"get a realm from a target"_.

The algorithm uses the default browsing context's realm instead of sandbox. This does not have any observable effects, as the sandbox specific `handle` is not provided in case of `ownership: none`.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/688.html" title="Last updated on Mar 12, 2024, 1:19 PM UTC (77fe9d0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/688/92650cd...77fe9d0.html" title="Last updated on Mar 12, 2024, 1:19 PM UTC (77fe9d0)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -8643,9 +8643,13 @@ To <dfn>get a realm from a target</dfn> given |target|:
 
 1. If |target| matches the <code>script.ContextTarget</code> production:
 
+  1. Let |sandbox| be null.
+
+  1. If |target| [=map/contains=] "<code>sandbox</code>", set |sandbox| to
+     |target|["<code>sandbox</code>"].
+
   1. Let |realm| be [=get a realm from a browsing context=] with
-    |target|["<code>context</code>"] and |target|["<code>sandbox</code>"] if present,
-    or an null otherwise.
+     |target|["<code>context</code>"] and |sandbox|.
 
 1. Otherwise:
 
--- a/index.bs
+++ b/index.bs
@@ -3258,8 +3258,6 @@ list of all nodes matching the specified locator.
          context: browsingContext.BrowsingContext,
          locator: browsingContext.Locator,
          ? maxNodeCount: (js-uint .ge 1),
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -8626,28 +8613,39 @@ This is useful in cases where a context identifier can stand in for the realm
...
+To <dfn>get a realm from a target</dfn> given |target|:
 
-    1. Let |realm| be |environment settings|' [=realm execution context=]'s
-       Realm component.
+1. If |target| matches the <code>script.ContextTarget</code> production:
 
-  1. Otherwise: let |realm| be result of [=trying=] to
-     [=get or create a sandbox realm=] given |target|["<code>sandbox</code>"] and
-     |context|.
+  1. Let |realm| be [=get a realm from a browsing context=] with
```
jgraham, line 8646: I'd prefer to split this into two steps, first getting the sandbox and then calling `get a realm from a browsing context`, just to avoid the awkward parsing of the "if present, or null otherwise" part.

jgraham (approved):

index.bs
sadym-chromium: Agree, done

jgraham (approved):

sadym-chromium:
Merging, as I assume @jimevans in the [comment](https://github.com/w3c/webdriver-bidi/issues/684#issuecomment-1988853789) approved this approach as well.

whimboo:
@sadym-chromium are you going to update the wpt tests and the BiDi client browsing_context module?

PR #689, opened 2024-03-15
Add an optional step to validate if the files which a client tries to set with input.setFiles command exist.

### LINKED ISSUE #677: Add a step to validate if the files which a client tries to set with `input.setFiles` command exist
Similar to WebDriver Classic actions (see [here](https://w3c.github.io/webdriver/#dfn-dispatch-actions-for-a-string) point 8.5), for `input.setFiles` command we could add a step to validate if paths specified in `files` argument point to existing files.

lutien:
Closes #677


index.bs
```diff
@@ -10522,6 +10526,9 @@ The [=remote end steps=] given |session| and |command parameters| are:
 1. Let |files| be the value of the |command parameters|["<code>files</code>"]
    field.
 
+1. If one of the |files| doesn't exist and the user agent can not handle such case,
```
jgraham, line 10529: I think we should phrase this more like 

```
1. If, for any reason, the remote end is unable the [=selected files=] of |element| to the files with paths given in |files|, return error with [=error code=] [=unsupported operation=].

   Note: For example remote ends may be unable to set [=selected paths=] to files that do not currently exist on the filesystem.
````

We should also move it below the cancel steps, although I guess that will become moot once we change how cancellation works.

lutien: Ok, maybe then below the part where update the files? Because I'm not sure if it's fine to put it before the `Otherwise`. I've updated it like this now

OrKoN (approved):

index.bs
whimboo: @jgraham would you mind having a 2nd look so that we can merge this PR if you are fine?

lutien:
We will update Firefox implementation to throw the right error type with: https://bugzilla.mozilla.org/show_bug.cgi?id=1887644

Issue #537, opened 2023-09-12
Describe `script.IteratorRemoteValue` serialization


sadym-chromium:
https://w3c.github.io/webdriver-bidi/#issue-4eaad8c4

Should any object with callable property `next` be serialized as an iterator?

Lightning00Blade:
Closing as won't fix per https://github.com/w3c/webdriver-bidi/pull/691

PR #691, opened 2024-03-20
Remove IteratorRemoteValue from the spec

OrKoN:
Currently, how IteratorRemoteValue is produced is not specified as per https://github.com/w3c/webdriver-bidi/issues/537

Additionally, a general iterator is any object with a next method as per https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-iterator-interface and it is an interface that any object can implement.

Since there is currently no use case requiring serialization support of iterators, we propose removing IteratorRemoteValue from the spec.

The following use cases have been considered:

1) Evaluating an iterator from the client code (In puppeteer code):

```
сonst it = await page.evaluateHandle(() => produce an iterator)
await it.evaluate(it => it.next())
```

Conclusion: serialization as an object is sufficient, since the user knows if they are dealing with an iterator and additionally they can check for the presence of the next method if needed.

2) Logging an iterator object to console

We believe that serializing an iterator as an object would be sufficient.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/691.html" title="Last updated on Mar 20, 2024, 10:32 AM UTC (7d6750f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/691/efb27e8...7d6750f.html" title="Last updated on Mar 20, 2024, 10:32 AM UTC (7d6750f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7721,7 +7721,6 @@ script.RemoteValue = (
   script.SetRemoteValue /
   script.WeakMapRemoteValue /
   script.WeakSetRemoteValue /
-  script.IteratorRemoteValue /
   script.GeneratorRemoteValue /
   script.ErrorRemoteValue /
   script.ProxyRemoteValue /
@@ -7796,12 +7795,6 @@ script.WeakMapRemoteValue = {
 
 script.WeakSetRemoteValue = {
   type: "weakset",
-  ? handle: script.Handle,
-  ? internalId: script.InternalId,
-}
-
-script.IteratorRemoteValue = {
-  type: "iterator",
   ? handle: script.Handle,
   ? internalId: script.InternalId,
 }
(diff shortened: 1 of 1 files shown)


sadym-chromium (approved):

sadym-chromium:
@jgraham @jimevans WDYT?

OrKoN:
Does anyone have any use cases in mind that would require special handing of iterators during serialzation?

jgraham (approved):
Since this is unused it's technically a noop change. But also I agree that since Iterator is just an interface we don't need a specific type.

whimboo:
@OrKoN can you please remove or better fix all the wpt tests for iterator serialization? Those are failing at the moment on wpt.fyi. Thanks.

OrKoN:
@whimboo which tests do you mean? I have updated tests in https://github.com/web-platform-tests/wpt/pull/45217 and they all pass for Firefox.

whimboo:
Oh wait. I was looking at eg [this test](https://wpt.fyi/results/webdriver/tests/bidi/script/evaluate/remote_values.py?label=master&label=experimental&product=firefox&product=firefox_android&product=chrome&product=chrome_android&product=safari&aligned&view=subtest). But yeah it passes now for Firefox and as it looks like requires a bidi mapper release for Chrome to pass the test?

PR #693, opened 2024-03-26
Update handling of "devicePixelRatio" argument in "browsingContext.setViewport" command.

lutien:
Since the implementation of changing the CSS pixels is problematic for at least Chrome and Firefox (probably Safari as well?) and likely not required, the suggestion here is to rather emulate certain parts, e.g. image source set selection and media queries.

***
<a href="https://pr-preview.s3.amazonaws.com/lutien/webdriver-bidi/pull/693.html" title="Last updated on Apr 19, 2024, 1:32 PM UTC (8563aea)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/693/88f2def...lutien:8563aea.html" title="Last updated on Apr 19, 2024, 1:32 PM UTC (8563aea)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3961,9 +3961,6 @@ The [=remote end steps=] with |command parameters| are:
 
             Note: This will take an effect because of the patch of [[#patchs-determine-the-device-pixel-ratio]].
 
-         1. Run [=evaluate media queries and report changes=] for [=document=] currently loaded
-            in a specified |context|.
-
       1. Otherwise:
 
          1. When the [=select an image source from a source set=] steps are run, use the implementation's default behavior,
@@ -3974,8 +3971,8 @@ The [=remote end steps=] with |command parameters| are:
 
          1. [=map/Remove=] |navigable| from [=device pixel ratio overrides=].
 
-         1. Run [=evaluate media queries and report changes=] for [=document=] currently loaded
-            in a specified |context|.
+      1. Run [=evaluate media queries and report changes=] for [=document=] currently loaded
+         in a specified |context|.
 
 1. Return [=success=] with data null.
 
--- a/index.bs
+++ b/index.bs
@@ -194,6 +194,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
     text: create a new browsing context; url: browsers.html#creating-a-new-browsing-context
     text: default classic script fetch options; url: webappapis.html#default-classic-script-fetch-options
     text: default view; url: nav-history-apis.html#dom-document-defaultview
+    text: descendant navigables; utl: document-sequences.html#descendant-navigables
     text: environment settings object's Realm; url: webappapis.html#environment-settings-object's-realm
     text: focused area of the document; url: document-sequences.html#focused-area-of-the-document
     text: getting all used history steps; url:browsing-the-web.html#getting-all-used-history-steps
@@ -3939,7 +3940,7 @@ The [=remote end steps=] with |command parameters| are:
    1. Let |device pixel ratio| be the |command
       parameters|["<code>devicePixelRatio</code>"].
 
-   1. For the |context| and all descendant traversables:
+   1. For the |context| and all [=descendant navigables=]:
 
       1. If |device pixel ratio| is not null:
 
--- a/index.bs
+++ b/index.bs
@@ -203,6 +203,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
     text: history handling behavior; url: browsing-the-web.html#history-handling-behavior
     text: innerText getter steps; url:dom.html#dom-innertext
     text: input type; url: input.html#dom-input-type
+    text: navigable; for:window; url: document-sequences.html#window-navigable
     text: navigables; url: document-sequences.html#navigables
     text: navigation id; url: browsing-the-web.html#navigation-id
     text: origin-clean; url: canvas.html#concept-canvas-origin-clean
@@ -1328,7 +1329,7 @@ it is observed via WebDriver, so we avoid using this terminology.
 A [=user context=] has a <dfn export for="user context">user context id</dfn>,
 which is a unique string set upon the user context creation.
 
-A [=navigable=] has an <dfn export>associated user context</dfn>, which is a [=user
+A [=/navigable=] has an <dfn export>associated user context</dfn>, which is a [=user
 context=].
 
 When a new [=/top-level traversable=] is created its [=associated user context=]
@@ -2279,6 +2280,9 @@ BrowsingContextEvent = (
   browsingContext.UserPromptOpened
 )
 </pre>
+
+A [=remote end=] has a <dfn>device pixel ratio overrides</dfn> which is a weak map
+between [=navigables=] and device pixel ratio overrides. It is initially empty.
 
 ### Types ### {#module-browsingcontext-types}
 
@@ -3942,6 +3946,9 @@ The [=remote end steps=] with |command parameters| are:
 
    1. For the |context| and all [=descendant navigables=]:
 
+      1. Let |navigable| be the [=/navigable=] whose [=navigable/active document=] is
+         |context|'s [=browsing context/active document=].
+
       1. If |device pixel ratio| is not null:
 
          1. When the [=select an image source from a source set=] are run, act as if
@@ -3950,7 +3957,7 @@ The [=remote end steps=] with |command parameters| are:
          1. For the purposes of the [=resolution media feature=], act as if
             the implementation's resolution is |device pixel ratio| dppx scaled by the page zoom.
 
-         1. When the [=determine the device pixel ratio=] steps are run, return |device pixel ratio|.
+         1. [=map/Set=] [=device pixel ratio overrides=][|navigable|] to |device pixel ratio|.
 
          1. Run [=evaluate media queries and report changes=] for [=document=] currently loaded
             in a specified |context|.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3844,19 +3850,35 @@ The [=remote end steps=] with |command parameters| are:
...
+
+         1. For the purposes of the [=resolution media feature=], act as if
+            the implementation's resolution is |device pixel ratio| dppx scaled by the page zoom.
+
+         1. Run [=evaluate media queries and report changes=] for [=document=] currently loaded
+            in a specified |context|.
+
+      1. Otherwise:
+
+         1. Reset special behavior for the [=select an image source from a source set=],
```
jgraham, line 3875: I don't love the "Reset special behavior" formulation; it's more or less clear in context what it must mean, but "special behaviour" isn't defined. Maybe we can say something more like "`When the [=select an image source from a source set=] steps are run, use the implementation's default behavior, without any changes made by previous invocations of these steps.` or something like that. It's still not really clear, but short of actually defining how this stuff works, it might be the best we can do. Similarly below.

jgraham (commented):
I agree that this matches what the gecko implementation does. It needs review from other implementors to ensure it can really be implemented across different engines (and that the spec, though necessarily handwavy, is nevertheless clear enough to understand).

index.bs
lutien: Yeah, I also didn't like this so much, but didn't come up with anything better.
Updated now, as you suggested.

lutien:
@OrKoN, @sadym-chromium, could you take a look if that's somewhat similar to what CDP is doing
@gsnedders, could you also assess if that's something Safari could potentially implement

index.bs
```diff
@@ -3844,19 +3850,36 @@ The [=remote end steps=] with |command parameters| are:
...
+
 1. If |command parameters| [=map/contains=] the <code>devicePixelRatio</code> field:
 
    1. Let |device pixel ratio| be the |command
       parameters|["<code>devicePixelRatio</code>"].
 
-   1. If |device pixel ratio| is not null, change the size of the [=CSS Pixel=]
-      in |context|'s [=layout viewport=] such that it corresponds to |device
-      pixel ratio| in device pixels.
+   1. For the |context| and all descendant traversables:
```
OrKoN, line 3860: > all descendant traversables

this means iframes, right?

index.bs
```diff
@@ -3844,19 +3850,36 @@ The [=remote end steps=] with |command parameters| are:
...
 
-   1. Otherwise, set the size of the [=CSS Pixel=] in |context|'s [=layout
-      viewport=] to the implementation-defined default.
+      1. If |device pixel ratio| is not null:
```
OrKoN, line 3862: I think we also need to cover the case of:

1) window.devicePixelRatio
2) screenshots being generated using the emulated devicePixelRatio

index.bs
lutien: Ah, that what you meant, here yes, it means all the children iframes.
(I thought yesterday you were talking about changing dpr only in an isolated iframe).

index.bs
lutien: I've added the step for `window.devicePixelRatio` with patching `determine the device pixel ratio` algorithm, and I think it should cover capturing a screenshot, since in [the render document to a canvas algorithm](https://w3c.github.io/webdriver-bidi/#render-document-to-a-canvas) we retrieve `ratio` from `determine the device pixel ratio` (point 1).

whimboo:
> <jgraham> jgraham: The spec change was based on what Firefox's RDM mode uses, but we need someone to check if that is the same behaviour that other browsers have.

That's actually not accurate. The Firefox RDM implementation also doesn't use `fullZoom` but exact the same mode as CDP does by setting the `overrideDPPX` property on the browsing context which then gets distributed through all child contexts as well. 

For details see:
https://searchfox.org/mozilla-central/rev/058ab60e5020d7c5c98cf82d298aa84626e0cd79/dom/chrome-webidl/BrowsingContext.webidl#163-176

As such using the behavior from CDP would be equivalent to RDM and that is what we are proposing here.

AutomatedTester:
I wonder if we need to have anything to `devicePixelRatio` at all because CDP does have it but if you then send over other commands(interaction commands as example) they will not work as intended and the wrong elements will get the command leading to `element click intercepted` errors being thrown.

If we leave it in then I think we need to have meaningful tests added to show that interactions work as expected (which they don't currently)

index.bs
```diff
@@ -269,6 +271,10 @@ spec: UNICODE; urlPrefix: https://www.unicode.org/versions/Unicode15.0.0/
   type: dfn
     text: Unicode Default Case Conversion algorithm; url: ch03.pdf#G34944
     text: toUppercase; url: ch03.pdf#G34078
+spec: MEDIAQUARIES4; urlPrefix: https://drafts.csswg.org/mediaqueries-4/
```
juliandescottes, line 274:
```suggestion
spec: MEDIAQUERIES4; urlPrefix: https://drafts.csswg.org/mediaqueries-4/
```

index.bs
lutien: @jgraham, what do you think?

lutien: > The HTML spec says:
> 
> > Currently, all [traversable navigables](https://html.spec.whatwg.org/multipage/document-sequences.html#traversable-navigable) are [top-level traversables](https://html.spec.whatwg.org/multipage/document-sequences.html#top-level-traversable). Future proposals envision introducing non-top-level traversables.
> 
> so if I read it right it cannot be iframes. Perhaps, we should link to https://html.spec.whatwg.org/multipage/document-sequences.html#descendant-navigables here?

I've checked again, and I think you're right, so I've replaced it to `descendant navigables`.

index.bs
```diff
@@ -3927,19 +3933,40 @@ The [=remote end steps=] with |command parameters| are:
...
+      1. If |device pixel ratio| is not null:
 
-1. Run the [[cssom-view-1#resizing-viewports]] steps.
+         1. When the [=select an image source from a source set=] are run, act as if
+            the implementation's pixel density was set to |device pixel ratio| when selecting an image.
+
+         1. For the purposes of the [=resolution media feature=], act as if
+            the implementation's resolution is |device pixel ratio| dppx scaled by the page zoom.
+
+         1. When the [=determine the device pixel ratio=] steps are run, return |device pixel ratio|.
```
jgraham, line 3953: Can we at least file an issue to upstream this change? It seems like CSSOM could either call into WebDriver or itself define an override device pixel ratio map that WebDriver would update at this step. I don't like totally monkey patching the algorithm in this case where it's not falling back to implementation-defined behavior.

index.bs
```diff
@@ -3927,19 +3933,40 @@ The [=remote end steps=] with |command parameters| are:
...
+
+      1. Otherwise:
+
+         1. When the [=select an image source from a source set=] steps are run, use the implementation's default behavior,
+            without any changes made by previous invocations of these steps.
+
+         1. For the purposes of the [=resolution media feature=], use the implementation's default behavior,
+            without any changes made by previous invocations of these steps.
+
+         1. When the [=determine the device pixel ratio=] steps are run, return the implemenation's default value.
```
jgraham, line 3966: This seems to be yet another new algorithm; you actually want to say "run the original algorithm".

Maybe the right way to do this one is actually to define the algorithm we want in the "patches to other specifications" section, something like:

```
Insert the following steps at the start of the [=determine the device pixel ratio=] algorithm:

1. If [=device pixel ratio overrides=] [=map/contains=] |window|'s [=window/navigable=], return [=device pixel ratio overrides=][|window|=].
```

Then in this section you'd need to define `device pixel ratio overrides` as a weak map between navigables and the dpr override, and set / remove entries at the appropriate time.

lutien: Alright, I've updated this part. Let me know how it looks

index.bs
lutien: Do we still have to file the issue, since we specified the patch for spec? (sorry, I don't know what the right way to do such things).

index.bs
```diff
@@ -3927,19 +3937,43 @@ The [=remote end steps=] with |command parameters| are:
...
-1. Run the [[cssom-view-1#resizing-viewports]] steps.
+      1. If |device pixel ratio| is not null:
+
+         1. When the [=select an image source from a source set=] are run, act as if
+            the implementation's pixel density was set to |device pixel ratio| when selecting an image.
+
+         1. For the purposes of the [=resolution media feature=], act as if
+            the implementation's resolution is |device pixel ratio| dppx scaled by the page zoom.
+
+         1. [=map/Set=] [=device pixel ratio overrides=][|navigable|] to |device pixel ratio|.
```
jgraham, line 3960: Maybe add a note linking to the spec patch? Just so it's clear why this has an effect.

jgraham (approved):

index.bs
lutien: Sure, I've added the note.

I've pushed the changes and all jobs are green, but looks like the preview bot didn't update the links in the header 😕 Not sure why

index.bs
```diff
@@ -3927,19 +3937,45 @@ The [=remote end steps=] with |command parameters| are:
...
+
+            Note: This will take an effect because of the patch of [[#patchs-determine-the-device-pixel-ratio]].
+
+         1. Run [=evaluate media queries and report changes=] for [=document=] currently loaded
```
OrKoN, line 3964: Nit: looks like this step can be de-deduplicates since it happens in both branches?

OrKoN (approved):

index.bs
lutien: Right! I've moved this step out of the conditions, so we have it only once

PR #695, opened 2024-04-04
editorial: replace occurrences of "undefined"

### LINKED ISSUE #391: Replace `?sandbox: text` with `?sandbox: text / null` in the spec?
Follow-up-of: #292

Could you let me know if the following is an error in the spec?

The spec says:

> 1. Let |sandbox| be the value of the "<code>sandbox</code>" field in |command
>  parameters|, if present, or null otherwise.

.i.e. `null` is possible. However the type specification is:

```
?sandbox: text
```

Should it have been

```
?sandbox: text / null
```

...instead? Other places in the spec have `/ null` to denote explicit nullability possible.

The alternative would be to eliminate the explicit `or null otherwise` language. Here I am specifically thinking of `undefined` vs `null` in TS/JS.

I am happy to send a PR if you confirm such a fix is indeed desirable.

thiagowfx:
Be precise whether it refers to the ECMA value of undefined, or to signify lack of value (null).

Closes #391

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/695.html" title="Last updated on Jun 26, 2024, 8:35 AM UTC (4c97af7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/695/05a283e...4c97af7.html" title="Last updated on Jun 26, 2024, 8:35 AM UTC (4c97af7)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1154,7 +1154,7 @@ with parameters |session|, |capabilities|, and |flags| is:
 1. Let |webSocketUrl| be the result of [=getting a property=] named
    "<code>webSocketUrl</code>" from |capabilities|.
 
-1. If |webSocketUrl| is null or false, return.
+1. If |webSocketUrl| is [=undefined=] or false, return.
 
 1. [=Assert=]: |webSocketUrl| is true.
 
@@ -7142,7 +7142,7 @@ To <dfn>deserialize local value</dfn> given |local protocol value|, |realm| and
       1. Let |pattern| be the value of the <code>pattern</code> field of |local protocol value|.
 
       1. Let |flags| be the value of the <code>flags</code> field of |local protocol value| or
-         null if no such a field.
+         [=undefined=] if no such a field.
 
       1. Let |regex_result| be [=Regexp=](|pattern|, |flags|). If this throws exception, return [=error=]
          with [=error code=] [=invalid argument=].
@@ -10574,6 +10574,6 @@ end:
 Other specifications can define <dfn>console steps</dfn>.
 
 1. At the point when the [=Printer=] operation is called with
-   arguments |name|, |printerArgs| and |options| (which is null if the
+   arguments |name|, |printerArgs| and |options| (which is undefined if the
    argument is not provided), call any [=console steps=] defined in
    external specification with arguments |name|, |printerArgs|, and |options|.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -7142,7 +7142,7 @@ To <dfn>deserialize local value</dfn> given |local protocol value|, |realm| and
       1. Let |pattern| be the value of the <code>pattern</code> field of |local protocol value|.
 
       1. Let |flags| be the value of the <code>flags</code> field of |local protocol value| or
-         undefined if no such a field.
+         null if no such a field.
```
jgraham, line 7145: This is calling the ecma262 RegExp constructor, which takes undefined, not null, for missing flags, so this is wrong.

index.bs
```diff
@@ -7090,10 +7090,10 @@ To <dfn>deserialize local value</dfn> given |local protocol value|, |realm| and
 1. If |local protocol value| matches the <code>script.ChannelValue</code> production,
    return [=create a channel=] with |session|, |realm| and |local protocol value|.
 
-1. Let |type| be the value of the <code>type</code> field of |local protocol value| or undefined if
+1. Let |type| be the value of the <code>type</code> field of |local protocol value| or null if
```
jgraham, line 7093: If you're going to rewrite this, I'd start by rewriting to use the infra-based language we've adopted elsewhere e.g. start with something like `If |local protocol value| [=map/contains=] "<code>type</code>", let |type| be |local protocol value|["<code>type</code>"]. Otherwise let |type| be null.`

But actually reading the rest of the algorithm, it seems like it's always an error to not have "`type`" and I don't see a case that makes sense without `"value"`, so we could either return early if they're null or, if we're confident that the callers handle validation (e.g. all the data is coming from CDDL that enforces these values), we could assert that the fields exist.

jgraham (changes requested):

index.bs
```diff
@@ -10574,6 +10574,6 @@ end:
 Other specifications can define <dfn>console steps</dfn>.
 
 1. At the point when the [=Printer=] operation is called with
-   arguments |name|, |printerArgs| and |options| (which is undefined if the
+   arguments |name|, |printerArgs| and |options| (which is null if the
```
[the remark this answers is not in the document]
thiagowfx: Oops. Reverted.

thiagowfx:
@jgraham friendly ping

jgraham (approved):

jgraham:
You need to join the WG to make the IPR bot happy.

thiagowfx:
It seems I can only join the group if I am an Invited Expert. Closing.

PR #702, opened 2024-04-24
Do not compute accessible role and name for non-Element nodes in the accessibility locator

OrKoN:
This PR updates the accessibility locator algorithm to avoid getting accessible name and role on non-element nodes while allowing to descend into non-element nodes' subtrees.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/702.html" title="Last updated on Apr 30, 2024, 6:14 AM UTC (22f61d4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/702/f3f93f5...22f61d4.html" title="Last updated on Apr 30, 2024, 6:14 AM UTC (22f61d4)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3384,19 +3384,6 @@ To <dfn>locate nodes using inner text</dfn> with given |context nodes|,
 
 1. For each |context node| in |context nodes|:
 
-  1. If |context node| implements {{Document}} or {{DocumentFragment}}:
-
-    1. Let |child nodes| be an empty [=/list=].
-
-    1. For each node |child| in the <a spec=dom>children</a> of |context node|.
-
-      1. [=list/Append=] |child| to |child nodes|.
-
-    1. [=Extend=] |returned nodes| with the result of [=trying=] to [=locate nodes using inner text=]
-      with |child nodes|, |selector|, |max depth|, |match type|, |ignore case|, and |maximum returned node count|.
-
-  1. If |context node| does not implement {{HTMLElement}} then [=continue=].
-
   1. Let |node inner text| be the result of calling the [=innerText getter steps=] with
      |context node| as the [=this=] value.
 
@@ -3410,7 +3397,7 @@ To <dfn>locate nodes using inner text</dfn> with given |context nodes|,
     1. Let |child nodes| be an empty [=/list=] and, for each node |child| in the
        <a spec=dom>children</a> of |context node|:
 
-      1. [=list/Append=] |child| to |child nodes|.
+      1. If |child| implements {{Element}}, [=list/append=] |child| to |child nodes|.
 
     1. If [=list/size=] of |child nodes| is equal to 0 or |max depth| is equal to 0,
        perform the following steps:
--- a/index.bs
+++ b/index.bs
@@ -3384,7 +3384,18 @@ To <dfn>locate nodes using inner text</dfn> with given |context nodes|,
 
 1. For each |context node| in |context nodes|:
 
-  1. If |context node| does not implement {{HTMLElement}}, [=break=].
+  1. If |context node| implements {{Document}} or {{DocumentFragment}}:
+
+    1. Let |child nodes| be an empty [=/list=].
+
+    1. For each node |child| in the <a spec=dom>children</a> of |context node|.
+
+      1. [=list/Append=] |child| to |child nodes|.
+
+    1. [=Extend=] |returned nodes| with the result of [=trying=] to [=locate nodes using inner text=]
+      with |child nodes|, |selector|, |max depth|, |match type|, |ignore case|, and |maximum returned node count|.
+
+  1. If |context node| does not implement {{HTMLElement}} then [=continue=].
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3442,21 +3444,25 @@ To <dfn>collect nodes using accessibility attributes</dfn> with given |context n
 
   1. Let |match| be true.
 
-  1. If |selector| [=map/contains=] "<code>role</code>":
+  1. If |context node| implements {{Element}}:
```
OrKoN, line 3447: Should it be HTMLElement here too?

index.bs
```diff
@@ -3384,6 +3384,8 @@ To <dfn>locate nodes using inner text</dfn> with given |context nodes|,
 
 1. For each |context node| in |context nodes|:
 
+  1. If |context node| does not implement {{HTMLElement}}, [=break=].
```
OrKoN, line 3387: The innerText algorithm seem to recurse only if there is a match? this is probably not sufficient since I imagine we want to recurse for Document and DocumentFragment nodes at least

index.bs
jgraham: No, I think this works for any kind of Element.

index.bs
jgraham: I think you want to continue, not break? And yes, I think you want a specific step for `Document` and `DocumentFragment`. So overall something like:
```
1. If |context node| implements {{Document}} or {{DocumentFragment}}:
 
  1. Let |child nodes| be an empty [=/list=]
  
  1. For each node |child| in the [=children=] of |context node|
  
    1. [=list/Append=] |child| to |children|.

  1. [=Extend=] |returned nodes| with the result of [=trying=] to [=locate nodes using inner text=] with |child nodes|, |selector|, |max depth|, |match type|, |ignore case|, and |maximum returned node count|.
  
1. If |context node| does not implement {{HTMLElement}} then [=continue=].
```

Then further down you could eliminate the Element checks before calling the algorithm recursively.

I also note that in the recursive calls we need to use `[=trying=]` because it returns a result type (even though recursive calls are actually infallible).

jgraham (changes requested):

index.bs
```diff
@@ -3384,6 +3384,19 @@ To <dfn>locate nodes using inner text</dfn> with given |context nodes|,
 
 1. For each |context node| in |context nodes|:
 
+  1. If |context node| implements {{Document}} or {{DocumentFragment}}:
```
sadym-chromium, line 3387: Note: so we traverse in-depth only html elements, documents and document fragments, while ignoring any other non-element nodes.

index.bs
```diff
@@ -3384,6 +3384,19 @@ To <dfn>locate nodes using inner text</dfn> with given |context nodes|,
...
+
+    1. Let |child nodes| be an empty [=/list=].
+
+    1. For each node |child| in the <a spec=dom>children</a> of |context node|.
+
+      1. [=list/Append=] |child| to |child nodes|.
+
+    1. [=Extend=] |returned nodes| with the result of [=trying=] to [=locate nodes using inner text=]
+      with |child nodes|, |selector|, |max depth|, |match type|, |ignore case|, and |maximum returned node count|.
```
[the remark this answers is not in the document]
OrKoN: looks like we should!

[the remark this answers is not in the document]
OrKoN: Is max depth handling not correct currently? https://w3c.github.io/webdriver-bidi/#locate-nodes-using-inner-text it looks like it would be decremented on each iteration on context nodes not reflecting the actual depth?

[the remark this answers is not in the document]
jimevans: @OrKoN I wouldn't put it past me to have totally botched the `max depth` calculation. Feel free to fix as appropriate.

OrKoN:
I am gonna split the changes required for a11y locator from changes to innerText (which requires more changes)

index.bs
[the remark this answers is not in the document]
OrKoN: Opened https://github.com/w3c/webdriver-bidi/pull/702 to address changes to innerText

index.bs
```diff
@@ -3442,21 +3442,25 @@ To <dfn>collect nodes using accessibility attributes</dfn> with given |context n
...
 
-    1. If |selector|["<code>name</code>"] [=is|is not=] |name|:
+      1. Let |name| be the [=accessible name=] of |context node|.
 
-      1. Set |match| to false.
+      1. If |selector|["<code>name</code>"] [=is|is not=] |name|:
+
+        1. Set |match| to false.
+
+  1. Otherwise, set |match| to false.
```
sadym-chromium, line 3463: For `document`, the `match` will be `false`, and no children will be traversed

sadym-chromium: UPD: My bad, the children will still be traversed

lutien (approved):
Looks good to me as well

jgraham (approved):

PR #705, opened 2024-04-30
Make innerText locator work with non-element nodes

OrKoN:


index.bs
```diff
@@ -3384,6 +3384,21 @@ To <dfn>locate nodes using inner text</dfn> with given |context nodes|,
...
 
+  1. If |context node| implements {{Document}} or {{DocumentFragment}}:
+
+    1. Let |child nodes| be an empty [=/list=].
+
+    1. For each node |child| in the <a spec=dom>children</a> of |context node|.
+
+      1. [=list/Append=] |child| to |child nodes|.
+
+    1. Let |child max depth| be null if |max depth| is null, or |max depth| - 1 otherwise:
```
sadym-chromium, line 3395:
```suggestion
    1. Let |child max depth| be null if |max depth| is null, or |max depth| - 1 otherwise.
```

index.bs
```diff
@@ -3410,12 +3425,10 @@ To <dfn>locate nodes using inner text</dfn> with given |context nodes|,
 
     1. Otherwise, perform the following steps:
 
-      1. If |max depth| is not null:
-
-        1. Set |max depth| to |max depth| - 1.
+      1. Let |child max depth| be null if |max depth| is null, or |max depth| - 1 otherwise:
```
OrKoN, line 3428:
```suggestion
      1. Let |child max depth| be null if |max depth| is null, or |max depth| - 1 otherwise.
```

index.bs
```diff
@@ -3384,6 +3384,21 @@ To <dfn>locate nodes using inner text</dfn> with given |context nodes|,
 
 1. For each |context node| in |context nodes|:
 
+  1. If |context node| implements {{Document}} or {{DocumentFragment}}:
+
+    1. Let |child nodes| be an empty [=/list=].
+
+    1. For each node |child| in the <a spec=dom>children</a> of |context node|.
```
sadym-chromium, line 3391: Maybe use node's [`childNodes`](https://dom.spec.whatwg.org/#dom-node-childnodes) instead, as we are not interested in other types of children?

sadym-chromium: Related issue in serialization: https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/705.html#issue-4a5a93fc

sadym-chromium: Can be out of scope of the PR of course

sadym-chromium (approved):

index.bs
OrKoN: This is pre-existing text, so I am not sure. Do you want to file an issue?

OrKoN: It would useful to know how exactly childNodes and children are different https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/705.html#issue-4a5a93fc ? here we link not to the DOM property though, but to the concept of children nodes so it looks fine to me in this place.

sadym-chromium:
Do we have any blockers for this PR?

OrKoN:
@sadym-chromium Mozilla's review I believe

sadym-chromium:
There is still an open question how to deal with request with document as a start node, and maxDepth equals zero:
```
{
    "type": "innerText",
    "maxDepth": 0,
    "value": "BAR"
}
```
As long as document does not have `innerText`, we can:
1. Traverse to the nearest child html element(s) and perform search there
2. Return an empty array. 
3. Do not decrease maxDepth on non-html elements.

OrKoN:
@sadym-chromium I think #2 would be something I would expect. I.e., the document does match and the max depth is 0, therefore, no recursive calls.

sadym-chromium:
> @sadym-chromium I think #2 would be something I would expect. I.e., the document does match and the max depth is 0, therefore, no recursive calls.

Currently spec says in this case the search should go to all the depth.

jgraham:
I feel like users are probably going to expect not decreasing depth on document elements i.e. passing in `document` is equivalent to `document.documentElement`, but we should at least be consistent between different selector types. That would also be inconsistent with the way that node serialization works, but maybe that is less important?

whimboo:
@OrKoN are you going to create / update wpt tests for this change?

sadym-chromium:
> I feel like users are probably going to expect not decreasing depth on document elements i.e. passing in `document` is equivalent to `document.documentElement`, but we should at least be consistent between different selector types. That would also be inconsistent with the way that node serialization works, but maybe that is less important?

Here is the spec change for this option: https://github.com/w3c/webdriver-bidi/pull/713

sadym-chromium:
* Spec: https://github.com/w3c/webdriver-bidi/pull/713
* WPT: https://github.com/web-platform-tests/wpt/pull/46345

PR #706, opened 2024-05-06
Add an autofill module

yoavweiss:
This is a PR similar to https://github.com/w3c/webdriver/pull/1797, but on the BiDi side.
It adds a `save` and `trigger` autofill commands.

/cc @sadym-chromium

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -10651,6 +10651,152 @@ The [=remote end steps=] given |session| and |command parameters| are:
 
 </div>
 
+## The autofill Module ## {#module-autofill}
+
+The <dfn export for=modules>autofill</dfn> module contains functionality for
+saving and triggering autofill.
+
+A <a>browsing context</a> has a <dfn>autofill store for testing</dfn>, an
+<a>ordered map</a>.
+
+### Definition ### {#module-autofill-definition}
+
+[=remote end definition=]
+
+<pre class="cddl remote-cddl">
+
+InputCommand = (
+  autofill.save //
+  autofill.trigger
+)
+</pre>
+
+### Commands ### {#module-autofill-commands}
+
+#### The autofill.save Command ####  {#command-autofill-save}
+
+The <dfn export for=commands>autofill.save</dfn> command saves a set of field
+name values to the <a>browsing context</a>'s [=autofill store for testing=].
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl remote-cddl">
+      autofill.save = (
+        method: "autofill.save",
+        params: autofill.saveParameters
+      )
+
+      autofill.saveParameters = {
+        context: browsingContext.BrowsingContext,
+        fields: [*autofill.field]
+      }
+
+      autofill.field = (
+        autofill.fieldName /
+        autofill.fieldValue
+      )
+    </pre>
+   </dd>
+   <dt>Return Type</dt>
+   <dd>
+    <pre class="cddl">
+     EmptyResult
+    </pre>
+   </dd>
+</dl>
+
+<div algorithm="remote end steps for autofill.save">
+
+The [=remote end steps=] with |session| and |command parameters| are:
+
+1. Let |context id| be the value of the <code>context</code> field of
+   |command parameters|.
+
+1. Let |context| be the result of [=trying=] to [=get a browsing context=]
+   with |context id|.
+
+1. Let |store| be |context|'s [=autofill store for testing=].
+
+1. Let |fields| be the value of the <code>fields</code> field of |command
+   parameters|.
+
+1. For each |field| in |fields|:
+
+   1. Let |name| be |field|'s |fieldName|.
+
+   1. Let |value| be |field|'s |fieldValue|.
+
+   1. Set |store|[|fieldName|] to |fieldValue|.
+
+1. Return [=success=] with data null.
+</div>
+
+#### The autofill.trigger Command ####  {#command-autofill-trigger}
+
+The <dfn export for=commands>autofill.trigger</dfn> command triggers autofill
+on a particular form field using the <a>browsing context</a>'s [=autofill
+store for testing=].
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl remote-cddl">
+      autofill.trigger = (
+        method: "autofill.trigger",
+        params: autofill.triggerParameters
+      )
+
+      autofill.triggerParameters = {
+        context: browsingContext.BrowsingContext,
+        element: script.SharedReference
+      }
+    </pre>
+   </dd>
+   <dt>Return Type</dt>
+   <dd>
+    <pre class="cddl">
+     EmptyResult
+    </pre>
+   </dd>
+</dl>
+
+<div algorithm="remote end steps for autofill.trigger">
+
+The [=remote end steps=] with |session| and |command parameters| are:
+
+1. Let |context id| be the value of the <code>context</code> field of
+   |command parameters|.
+
+1. Let |context| be the result of [=trying=] to [=get a browsing context=]
+   with |context id|.
+
+1. Let |document| be |context|'s [=active document=].
+
+1. Let |environment settings| be the [=environment settings object=] whose
+   [=relevant global object=]'s <a>associated <code>Document</code></a> is
+   |document|.
+
+1. Let |realm| be |environment settings|' [=realm execution context=]'s
+  Realm component.
+
+1. Let |element id| be the value of the <code>element</code> field of
+   |command parameters|.
+
+1. Let |element| be the result of [=trying=] to [=deserialize remote reference=]
+   with the |element id|, |realm|, and |session|.
+
+1. Let |store| be |context|'s [=autofill store for testing=].
+
+1. If |element| is not [=form associated=] or |element|'s [=form owner=] is
+   null, return an [=error=] with [=error code=] [=invalid element state=].
+                                                     
+1. The [=user agent=] should [=autofill=] |element| and |element|'s [=form
+   owner=], while taking into account the contents of |context|'s [=autofill
+   store for testing=].
+
+1. Return [=success=] with data null.
+</div>
 # Patches to Other Specifications # {#patches}
 
 This specification requires some changes to external specifications to provide the necessary


jimevans (commented):
@yoavweiss Does this proposal depend on the user agent (the browser) supporting autofill? What if the browser doesn't support this feature? Is it dependent on the [autofill portion of the HTML Living Standard](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill)? Is it expected that the property names in the map must match the [autofill field name tokens](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill-field)?

index.bs
```diff
@@ -10651,6 +10657,155 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+
+The <dfn export for=commands>autofill.save</dfn> command saves a set of field
+name values to the [=/browsing context=]'s [=autofill store for testing=].
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl remote-cddl">
+      autofill.save = (
+        method: "autofill.save",
```
OrKoN, line 10692: I think it would be preferred to save the data by exercising the browser logic based on a filled out form instead of adding autofill fields to a test only storage.

index.bs
```diff
@@ -10651,6 +10657,155 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+   <dd>
+    <pre class="cddl remote-cddl">
+      autofill.trigger = (
+        method: "autofill.trigger",
+        params: autofill.triggerParameters
+      )
+
+      autofill.triggerParameters = {
+        context: browsingContext.BrowsingContext,
+        element: script.SharedReference
```
OrKoN, line 10760: Potentially, based on attributes in the storage, multiple suggestions could be available in autofill. I think the protocol should either allow defining the entire data that is being triggered or some sort of an event to notify what autofill options are available based on the storage. WDYT?

index.bs
OrKoN: For example, if a form is split across multiple iframes, the saving might work or not work in the same way

index.bs
yoavweiss: Interesting!

Do you have in mind a `getAutofillOptions` method that would then allow the caller to call `trigger` with one of the options?

index.bs
OrKoN: > I'm guessing this would be multiple forms that act as one (through JS calls across the same-origin iframes). Do I understand you correctly?

implementations might also use cross-origin iframes and from autofill perspective that would be one form too.

I guess that if the goal of the proposed save method is not to actually test the saving functionality of a given autofill implementation and just to seed the implementation storage with test data, the current proposal works too.

yoavweiss: In that case, do we need to provide a list of forms and iframes as part of the API?



> I guess that if the goal of the proposed save method is not to actually test the saving functionality of a given autofill implementation and just to seed the implementation storage with test data, the current proposal works too.

That's indeed my initial goal, but it'd be good if the API enabled testing of the saving functionality as well. (as presumably, this is also something that sites can do correctly or incorrectly)

index.bs
```diff
@@ -10651,6 +10657,155 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+
+1. Let |element| be the result of [=trying=] to [=deserialize remote reference=]
+   with the |element id|, |realm|, and |session|.
+
+1. Let |store| be |context|'s [=autofill store for testing=].
+
+1. If |element| is not a [=form-associated element=] or |element|'s [=form owner=] is
+   null, return an [=error=] with [=error code=] [=invalid element state=].
+
+1. The [=user agent=] should [=autofill=] |element| and |element|'s [=form
```
OrKoN, line 10802: should  [=autofill=] link to something in HTML spec?

index.bs
OrKoN: > That's indeed my initial goal, but it'd be good if the API enabled testing of the saving functionality as well. (as presumably, this is also something that sites can do correctly or incorrectly)

I guess this is interesting but perhaps to a lesser extend: it is very likely that the data was stored by filling out forms previously on other websites or entering the data manually into the browser settings.

OrKoN: > In that case, do we need to provide a list of forms and iframes as part of the API?

I think we can just provide a SharedReference to the form element that is to be saved to autofill?

index.bs
yoavweiss: I thought it'd link to https://html.spec.whatwg.org/#autofill. I'll take a look

index.bs
martinLechnerShopify: @OrKoN @yoavweiss @theindra throwing in another potential approach here, to avoid an explicit `save` command: What if we passed in the expected state of the autofill "Storage" on the implementation on every `trigger` call? Something like (pseudo-code, don't focus on naming of the variables :-) ):
`autofill.trigger(onMyForm, withAutofillState)`
where `onMyForm` is the form I want to see the fields being filled, and `withAutofillState` being a key/value hash that describes the current state of autofill for the test.

This more functional approach makes it also very clear that the state of autofill is only expected to be preserved during the context in which the test runs, i.e. no sideeffects from non-cleared autofill storage state compared to an explicit `save()` that kind of implies a `clear()` after the test ran.

OrKoN: Yes, this is the approach I [mentioned](https://github.com/w3c/webdriver-bidi/pull/706#discussion_r1613177353) and we went this direction in CDP https://chromedevtools.github.io/devtools-protocol/tot/Autofill/#method-trigger

index.bs
```diff
@@ -10651,6 +10657,155 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+   <dd>
+    <pre class="cddl remote-cddl">
+      autofill.save = (
+        method: "autofill.save",
+        params: autofill.saveParameters
+      )
+
+      autofill.saveParameters = {
+        context: browsingContext.BrowsingContext,
+        fields: [*autofill.field]
```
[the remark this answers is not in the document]
martinLechnerShopify: @OrKoN trying to make sure I fully understand the concern here:

> what I worry about is that the users would need to know which set of attributes actually can be successfully saved in a given implementation

I'm nor super familiar with the descriptive language of webdriver-bidi in this file, is there any way to define an enumeration that specifies which keys are allowed? These keys would 1:1 map to the keys in the autofill specification, and hence would kind of mimic the form-based approach, just without specifying the form. Kind of a "shortcut".

>what to do if the fields from of incompatible items are being sent at once such as fields from an address and from a credit card

I'm trying to understand where this would cause problems? If the setup state of autofill for the test is to have credit card and address information stored, and then the test only creates forms for addresses and only asserts on addresses and passes, wouldn't this be good and actually real world conditions, as I'm sure I have way more autofill entries stored than a page renders.

>  it seems like saving step can be left out and the fields can be sent directly to trigger. An explicit save step would also need some way to clean up the data (e.g., between tests).

I think we are on the same page re. the explicit `save` vs passing in state for every test, as I suggested [here](https://github.com/w3c/webdriver-bidi/pull/706/files#r1615990960). I believe our 2 proposals are essentially the same, right?

index.bs
martinLechnerShopify: I'd also see value in both having a "shortcut" to pass explicit autofill state into the test (when testing filling forms), as well as creating a form to test the saving functionality of autofill. For me, saving properly, and filling properly, are 2 different things, and if a test fails for my implementation, being able to clearly distinguish whether saving failed, or filling failed, makes sense.

I can imaging the test suite testing filling the fields uses the "pass state into the test", to be independent of the saving functionality. When testing saving, we'd obviously have to use the form approach. This would clearly separate the 2, and would highlight clearly what part failed.

martinLechnerShopify: @OrKoN I've changed the spec to not have `autofill.save` anymore, and instead pass in the autofill fields for each test. Curious to get your feedback on it.

index.bs
martinLechnerShopify: with explicitly passing in the autofill storage for each test, do you still deem this relevant?
I'm thinking that we could make it a requirement that the keys in the autofill storage need to be unique, or alternatively say if there are multiple values for the same key, take the first one?

Unless we wanted to specifically include the option to test such a selection process, which would then be a different story. I'm not clear though if this is a useful test in this context, would appreciate some more background in this case.

index.bs
yoavweiss: Fixed

index.bs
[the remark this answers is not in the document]
yoavweiss: @martinLechnerShopify now pushed the parameters onto `trigger`. @OrKoN - PTAL?

index.bs
```diff
@@ -10696,6 +10703,95 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+
+The <dfn export for=modules>autofill module</dfn> contains functionality for triggering autofill.
+
+### Definition ### {#module-autofill-definition}
+
+[=remote end definition=]
+
+<pre class="cddl remote-cddl">
+
+InputCommand = (
```
OrKoN, line 10716:
```suggestion
AutofillCommand = (
```

OrKoN, line 10716: InputCommand needs to be added to CommandData as well

index.bs
```diff
@@ -10696,6 +10703,95 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+#### The autofill.trigger Command ####  {#command-autofill-trigger}
+
+The <dfn export for=commands>autofill.trigger</dfn> command triggers autofill
+on a particular form field using the provided fields.
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl remote-cddl">
+      autofill.trigger = (
```
OrKoN, line 10732:
```suggestion
      autofill.Trigger = (
```

index.bs
```diff
@@ -10696,6 +10703,95 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+        method: "autofill.trigger",
+        params: autofill.triggerParameters
+      )
+
+      autofill.triggerParameters = {
```
OrKoN, line 10737:
```suggestion
      autofill.TriggerParameters = {
```

index.bs
```diff
@@ -10696,6 +10703,95 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+        context: browsingContext.BrowsingContext,
+        element: script.SharedReference,
+        fields: [*autofill.field]
```
OrKoN, line 10740:
```suggestion
        fields: [*autofill.Field]
```

index.bs
```diff
@@ -10696,6 +10703,95 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+      }
+
+      autofill.field = (
```
OrKoN, line 10743:
```suggestion
      autofill.Field = (
```

index.bs
```diff
@@ -10696,6 +10703,95 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+        name: tstr,
```
OrKoN, line 10744:
```suggestion
        name: text,
```

OrKoN, line 10744: I think we would probably need a named type for the field name and a definition what that type represents (linking to the HTML spec).

index.bs
```diff
@@ -10696,6 +10703,95 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+1. Let |element id| be the value of the <code>element</code> field of
+   |command parameters|.
+
+1. Let |element| be the result of [=trying=] to [=deserialize remote reference=]
+   with the |element id|, |realm|, and |session|.
+
+1. Let |fields| be the value of the <code>fields</code> field of |command
+   parameters|.
+
+1. If |element| is not a [=form-associated element=] or |element|'s [=form owner=] is
```
OrKoN, line 10785: I think we should also define something like if implementation is not able to autofill the provided set of fields, return invalid argument error (or perhaps unsupported error?)

index.bs
martinLechnerShopify: added AutofillCommand to CommandData, please let me know if this is how it was intended

index.bs
martinLechnerShopify: good point, I think unsupported operation would fit better than invalid argument. It could be that the argument is technically valid, just the implementation cannot support it (yet).

index.bs
martinLechnerShopify: good point. I've tried my best to define a named type and link to the respective section in the autofill documentation to define what values are allowed. Would appreciate some more pair of eyes over it to make sure the type is correct and valid.

martinLechnerShopify:
@OrKoN I've addressed all your comments, would you be able to give it another pass? Thanks in advance!

index.bs
```diff
@@ -282,6 +286,9 @@ spec: CORE-AAM; urlPrefix:https://www.w3.org/TR/core-aam-1.2
 spec: MEDIAQUERIES4; urlPrefix: https://drafts.csswg.org/mediaqueries-4/
   type: dfn
     text: resolution media feature; url: #resolution
+spec:infra; urlPrefix: https://infra.spec.whatwg.org/
```
OrKoN, line 289: I think we should not reference infra for the user agent since it has a different meaning there. Let's just say `user agent` without a link to the infra spec.

OrKoN (commented):
Generally, the approach LGTM but I wonder what @jgraham @whimboo @gsnedders think.

index.bs
yoavweiss: Done!

index.bs
yoavweiss: Resolving as this is done.

martinLechnerShopify:
Hi @OrKoN , coming back from a 2-weeks vacation, I wanted to check in and see where we stand with this PR - is this ok to be merged, so we can move on to the implementation phase? Thanks in advance for a quick update!

OrKoN:
@martinLechnerShopify we need the approval from other browser vendors as well to merge this

martinLechnerShopify:
@OrKoN thanks, is there some kind of regular "sync" or a timeline when other browser vendors are usually commenting/approving? I'd like to make sure we can push this over the finish line together in a reasonable timeframe.
Thanks in advance!

OrKoN:
The meeting happens on the second Wednesday of the month. @AutomatedTester is the chair and the organizer of the meeting. @yoavweiss already presented in the proposal in one of the previous meetings https://www.w3.org/2024/05/08-webdriver-minutes.html

shs96c (commented):
~~Could we please include a link to the spec or rfc this is interacting with?~~ I see that this has happened automatically. My apologies.

yoavweiss:
Thanks all for the TPAC discussion. The presentation link above is the wrong one tho. Here's the [actual one](https://drive.google.com/file/d/1SYxBKtdPBPlM0KnXNPnE-YPozdmIYV-2/view?usp=sharing).

jgraham:
So did the design here get any updates in response to the TPAC discussion? IIRC correctly a concern in that session was that this API ends up both specifying the autofill data and also requesting that data to be filled, so it was unclear that it was leaving enough of the original browser codepaths to be meaningful. In particular it wouldn't allow checking that the expected data is saved/provided by the data given a prior user interaction e.g. a form submission.

martinLechnerShopify:
@jgraham sorry for the delayed response as I was involved with a few high prio projects at Shopify.

Thanks for getting back to us! As to your points, if I recall TPAC discussions correctly, I think we said there is value in having both the explicit tests on filling correctly (i.e. the proposal we currently have), and the actual data saving part. We could discuss if it makes sense to also combine both for some kind-of e2e tests.

This would leave us with 3 possible kinds of autofill tests:
1. just the filling part as per the specification
2. just the saving part
3. a combined saving + filling

Do you consider 1 and 2 valuable in itself, do you think it only makes sense to go with 3, or do you see all 3 options viable?

Also, looking a bit more closely into option #3, I am thinking that the test flow would be like:
1. create an test-data-autofill form in the test, fill it with values
2. submit the form (and with this, implicitly assume the browser has saved the autofill values)
3. create a new test-autofill form
4. let the browser suggest autofill values
5. select an option
6. let the browser autofill and assess validity of autofilled data

wdyt?
cc also @yoavweiss @theindra @OrKoN

OrKoN:
I think testing the filling part (no. 1) is the most important as there are many ways data might end up being saved into the autofill storage (updated over time, profile sync etc). The web authors probably care most if a specific form(s) can be auto-filled on their website (given it is in storage) and not so much if browser's autofill storage can be populated using their website specifically. Also, not every user agent might be saving data from forms and given that this storage part is largely non-standard, focusing on the filling part seems to be most beneficial. The spec that includes the saving part should also define how the data is stored and for how long (probably should not survive the WebDriver session), how the autofill options are presented to the client and how they can be selected, how sensitive data like credit cards could be saved and loaded.

martinLechnerShopify:
@jgraham any thoughts on the above? Would you be OK with a phased approach?

index.bs
```diff
@@ -11984,6 +11993,118 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+        element: script.SharedReference,
+        field: autofill.Field
+      }
+
+      autofill.Field = {
+        name: autofill.FieldName,
+        value: text
+      }
+
+      autofill.setAddress = (
```
OrKoN, line 12039: I believe each command should have its own section

index.bs
```diff
@@ -11984,6 +11993,118 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+        method: "autofill.setAddress",
```
sadym-chromium, line 12040: I'm a bit confused with this API:
1. `autofill.trigger` has a single param `element` with `field {name, value}`. Does it mean it triggers the autofill only on a single field? What about multi-field addresses?
2. What is the point of `autofill.setAddress`, if the only way to trigger that autofill is `"autofill.trigger`, which provides `values`?

index.bs
```diff
@@ -13616,6 +13622,121 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+The <dfn export for=modules>autofill module</dfn> contains functionality for triggering autofill.
+
+### Definition ### {#module-autofill-definition}
+
+{^remote end definition^}
+
+<pre class="cddl" data-cddl-module="remote-cddl">
+
+AutofillCommand = (
+  autofill.trigger
```
```suggestion
  autofill.Trigger
```

index.bs
```diff
@@ -13616,6 +13622,121 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+#### The autofill.trigger Command ####  {#command-autofill-trigger}
+
+The <dfn export for=commands>autofill.trigger</dfn> command triggers autofill
+on a particular form field using the provided fields.
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl" data-cddl-module="remote-cddl">
+      autofill.trigger = (
```
```suggestion
      autofill.Trigger = (
```

index.bs
```diff
@@ -13616,6 +13622,121 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+        method: "autofill.trigger",
+        params: autofill.TriggerParameters
+      )
+
+      autofill.TriggerParameters = {
+        context: browsingContext.BrowsingContext,
+        element: script.SharedReference,
+        card: autofill.Card,
+        address: autofill.Address
```
```suggestion
        ? card: autofill.Card,
        ? address: autofill.Address
```

index.bs
```diff
@@ -13616,6 +13622,121 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+  Realm component.
+
+1. Let |element id| be the value of the <code>element</code> field of
+   |command parameters|.
+
+1. Let |element| be the result of [=trying=] to [=deserialize remote reference=]
+   with the |element id|, |realm|, and |session|.
+
+1. Let |field| be the value of the <code>field</code> field of |command
+   parameters|.
```
sadym-chromium: Please validate and parse the right values (`card` or `address`).

index.bs
```diff
@@ -13616,6 +13622,121 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+
+1. If |element| is not a [=form-associated element=] or |element|'s [=form owner=] is
+   null, return an [=error=] with [=error code=] [=invalid element state=].
+
+1. If [=/user agent=] cannot autofill |field|, return an [=error=] with
+   [=error code=] [=unsupported operation=].
```
```suggestion
1. If, for any reason, the remote end is unable to [=autofill=], return error with [=error code=] [=unsupported operation=].
```

index.bs
yoavweiss: Done

index.bs
yoavweiss: Done

index.bs
yoavweiss: Done

index.bs
yoavweiss: Done

index.bs
yoavweiss: Done

index.bs
yoavweiss: We've solved this by moving away from the previous shape into a single `Trigger` call that accepts a card or an address.

index.bs
```diff
@@ -11816,6 +11825,118 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+     EmptyResult
+    </pre>
+   </dd>
+</dl>
+
+<div algorithm="remote end steps for autofill.trigger">
+
+The [=remote end steps=] with |session| and |command parameters| are:
+
+1. Let |navigable id| be the value of the <code>navigable</code> field of
```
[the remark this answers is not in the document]
yoavweiss: Done

index.bs
```diff
@@ -13616,6 +13622,132 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+1. Let |element id| be the value of the <code>element</code> field of
+   |command parameters|.
+
+1. Let |element| be the result of [=trying=] to [=deserialize remote reference=]
+   with the |element id|, |realm|, and |session|.
+
+1. If |element| is not a [=form-associated element=] or |element|'s [=form owner=] is
+   null, return an [=error=] with [=error code=] [=invalid element state=].
+
+1. Let |card| be the value of the <code>card</code> field of |command parameters|.
```
```suggestion
1. Let |card| be the value of the <code>card</code> field of |command parameters| if present, or null otherwise.
```

index.bs
```diff
@@ -13616,6 +13622,132 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+
+1. Let |address| be the value of the <code>address</code> field of |command parameters|.
```
```suggestion
1. Let |address| be the value of the <code>address</code> field of |command parameters| if present, or null otherwise.
```

index.bs
```diff
@@ -13616,6 +13622,132 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+
+1. If, for any reason, the remote end is unable to [=autofill=], return error with
+   [=error code=] [=unsupported operation=].
+
+1. If both |card| and |address| are not empty, return error with [=error code=]
```
```suggestion
1. If both |card| and |address| are not null, return error with [=error code=]
```

index.bs
```diff
@@ -13616,6 +13622,132 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+   [=invalid argument=].
+
+1. If both |card| and |address| are empty, return error with [=error code=]
```
```suggestion
1. If both |card| and |address| are null, return error with [=error code=]
```

index.bs
```diff
@@ -13616,6 +13622,132 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+   owner=], while taking into account the contents of |parsed card| and |parsed address|.
+
+1. Return [=success=] with data null.
+
+</div>
+
+#### The autofill.FieldName Type #### {#type-autofill-FieldName}
+
+<pre class="cddl" data-cddl-module="remote-cddl,local-cddl">
+   autofill.FieldName = text;
```
sadym-chromium: I believe it should be one of the specified autofill names: https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill-detail-tokens

index.bs
```diff
@@ -13616,6 +13622,132 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+   [=invalid argument=].
+
+1. If both |card| and |address| are empty, return error with [=error code=]
+   [=invalid argument=].
+
+1. Let |parsed card| be the result of parsing |card| as an <code>autofill.Card</code>.
+
+1. Let |parsed address| be the result of parsing |card| as an <code>autofill.Address</code>.
+
+1. The [=/user agent=] should [=autofill=] |element| and |element|'s [=form
```
sadym-chromium, line 13736: The term [=autofill=] is too vague. The specification should explicitly state who exactly the user agent must use the |parsed card| or |parsed address| to populate the form fields.

index.bs
yoavweiss: I state that right below, but not sure what's the right way to include that in the CDDL. Should we redefine that as a CDDL type? Something else?

index.bs
yoavweiss: Can you elaborate on exactly what you want to see defined here? Mapping of address/card data to the relevant input fields?

index.bs
sadym-chromium: It can be specified in cddl like [ErrorCode](https://www.w3.org/TR/webdriver-bidi/#cddl-type-errorcode)

yoavweiss: OK, done!

OrKoN (approved):
LGTM (% we could add a note that the intention of the method is to test the actual implementation of the autofill component in the browser as realistically as possible and not to implement a different WebDriver-BiDi-specific form filling method)

index.bs
sadym-chromium: As discussed, we want to test the actual browser's autofill experience, so I believe pointing to "autofill" part of HTML spec is enough.

sadym-chromium:
@yoavweiss please wait for at least one other vendor to approve the PR

sadym-chromium:
@jgraham could you please take a look?

yoavweiss:
@rmondello - FYI this is the autofill testing infra I was talking about during the breakout.

jgraham:
I think @burg had concerns about the model here, and I'm not sure if they've been addressed.

AIUI the proposal here is that WebDriver gains a way to load moaxk data into the autofill database, and then trigger a fill regardless of whether any user action could populate the database with that data, or trigger a fill on the current page. It's not clear to me that this will test enough of the browser codepaths to be confident that autofill is actually going to work for the user in the way the test implies.

yoavweiss:
> AIUI the proposal here is that WebDriver gains a way to load moaxk data into the autofill database

The data is not necessarily loaded into the database, as it's only used for this particular test. This is based on feedback from @sadym-chromium (IIRC), who preferred to avoid the risk of the testing infra polluting the autofill DB.



> then trigger a fill regardless of whether any user action could populate the database with that data

This API proposal doesn't try to address the "populate the database" part of autofill. We can probably be more explicit about that.



> trigger a fill on the current page

Yeah, this proposal will replace the user action needed to trigger autofill.

> It's not clear to me that this will test enough of the browser codepaths to be confident that autofill is actually going to work for the user in the way the test implies.

I believe that it would exercise enough of the codepaths to give us greater assurance that *when* autofill works (with the provided data), we be able to make sure the form in question gets filled correctly.

Is there anything specific you're concerned with? Any improvements you have in mind?

yoavweiss:
WPT mocks that demonstrate what's likely to be tested with this: https://github.com/web-platform-tests/wpt/pull/44329

index.bs
```diff
@@ -13616,6 +13623,189 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+
+      autofill.Field = {
+        name: autofill.FieldName,
+        value: text
+      }
+
+      autofill.Card = {
+        number: text,
+        name: text,
+        expiryMonth: text,
```
burg, line 13672: Why is there no credit card type field here, but it's in the FieldToken list?

I'm concerned that these fields don't have an expected format.. is this `%MM`, `%mm`, or `%M` in printf parlance? What should the remote end do if it can't parse the value?

index.bs
```diff
@@ -13616,6 +13623,189 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+   <dd>
+    <pre class="cddl" data-cddl-module="remote-cddl">
+      autofill.Trigger = (
+        method: "autofill.trigger",
+        params: autofill.TriggerParameters
+      )
+
+      autofill.TriggerParameters = {
+        context: browsingContext.BrowsingContext,
+        element: script.SharedReference,
```
burg, line 13659: Autofill (consisting of form/field classification, data selection, and filling) all happens for the entire form, not one element. Many of these field tokens are commonly split across multiple form fields, and sometimes across multiple forms. (Also, \<input\> with no \<form\>.) So I'm not sure what to say about this parameter except it's unnecessary IMO.

index.bs
```diff
@@ -13616,6 +13623,189 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+        ? card: autofill.Card,
+        ? address: autofill.Address
```
burg, line 13661: What about separate billing and shipping addresses? I'm hesitant to associate specific datas with the 'trigger' request because of complications like this. As far as I can tell, there is no way in the autofill module flow, where a test would simulate filling two separate addresses in a meaningful way.

index.bs
```diff
@@ -13616,6 +13623,189 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+
+      autofill.Card = {
+        number: text,
+        name: text,
+        expiryMonth: text,
+        expiryYear: text,
+        cvc: text
+      }
+
+      autofill.Address = {
```
burg, line 13677: The fields that an address consists of depends on the locale of the website, form, and user agent. How do you plan to deal with this? I'm thinking that Japanese has both romanized and traditional address formats, but also Hindi and others.

index.bs
```diff
@@ -13616,6 +13623,189 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+#### The autofill.FieldName Type #### {#type-autofill-FieldName}
+
+<pre class="cddl" data-cddl-module="local-cddl">
+AutofillToken = "name" /
+                "honorific-prefix" /
+                "given-name" /
+                "additional-name" /
+                "family-name" /
+                "honorific-suffix" /
+                "nickname" /
```
burg, line 13752: Is the intent here to be exhaustive? I feel like one-time-code, nickname, organization-title, bday, impp are not relevant if we are focusing on autofill for shopping cart checkout pages.

index.bs
```diff
@@ -13616,6 +13623,189 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+
+1. If both |card| and |address| are not null, return error with [=error code=]
+   [=invalid argument=].
+
+1. If both |card| and |address| are null, return error with [=error code=]
+   [=invalid argument=].
+
+1. Let |parsed card| be the result of parsing |card| as an <code>autofill.Card</code>.
+
+1. Let |parsed address| be the result of parsing |card| as an <code>autofill.Address</code>.
```
burg, line 13734: I'd like to humbly propose that we have separate commands that add the mock data, which will make it easier to split out parsing failures from autofill itself not working due to issues in form/field classification or field-filling.

`autofill.registerMockData(type, {field: value})`
`autofill.clearMockData()`

and so on.

Another suggestion would be to add events for autofill domain similar to userPromptHandler. This would make it possible to have tests go through the following flow:

1. Set up MockDatas to be used -> get back MockDataIDs
2. set up the form/page state being tested
3. call `autofill.trigger(browsingContext)` // --> causes steps 4-7
4. `autofill.candidatesPromptOpened(candidateIDs)` // candidateIDs established in Step 1
5. call `autofill.selectAutoFillCandidate(candidateID)`or similar
6. `autofill.candidatesPromptClosed() event`
7. `autofill.didFillFormElements(mockDatas)` // say which mockData was used, but not which fields?

yoavweiss:
@burg - https://github.com/web-platform-tests/wpt/pull/57171 demonstrates how the API can be used to help developers test autofill on their forms.

Beyond those, https://wicg.github.io/autofill-event/ is an API that would be testable by this proposal.

PR #708, opened 2024-05-07
Define algorithms for the HTML spec to call when a navigable is created or destroyed

OrKoN:
This PR defines hooks for the HTML spec to use instead of patching the HTML spec. The hooks are defined for navigables as this more closely matches how the WebDriver BiDi spec uses the term browsing context. The destroyed hook gets the browsing context explicitly as the navigable might be disassociated from the active browser context at this point.

Issue: https://github.com/whatwg/html/issues/6194
HTML spec: https://github.com/whatwg/html/pull/10329

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/708.html" title="Last updated on Jun 4, 2024, 7:22 AM UTC (78e1bb1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/708/c29a684...78e1bb1.html" title="Last updated on Jun 4, 2024, 7:22 AM UTC (78e1bb1)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -10663,27 +10663,6 @@ as part of the standards process.
 
 ## HTML ##  {#patches-html}
 
-The [=a browsing context is discarded=] algorithm is modified to read as follows:
-
-<div algorithm>
-To discard a browsing context |browsingContext|, run these steps:
-
-1. If this is not a recursive invocation of this algorithm, call any <dfn export>browsing context
-   tree discarded</dfn> steps defined in other applicable specifications with |browsingContext|.
-
-1. Discard all {{Document}} objects for all the entries in |browsingContext|'s
-   [=session history=].
-
-1. If |browsingContext| is a [=top-level browsing context=], then [=remove a browsing context=]
-   |browsingContext|.
-
-</div>
-
-Issue: The actual patch might be better to split the algorithm into an outer algorithm that is
-called by external callers and an inner algorithm that's used for recursive calls. That's quite hard
-to express as a patch to the specification since it requires changing multiple parts.
-
-
 The [=report an error=] algorithm is modified with an additional step at the
 end:
 
--- a/index.bs
+++ b/index.bs
@@ -4136,14 +4136,14 @@ To <dfn>Emit a context created event</dfn> given |session| and |context|:
 
 </div>
 
+
 <div algorithm="remote end event trigger for browsingContext.contextCreated">
 
-The [=remote end event trigger=] is:
+The [=remote end event trigger=] is
+the <dfn export>WebDriver BiDi navigable created</dfn> steps given
+|navigable|:
 
-When the [=create a new browsing context=] algorithm is invoked, after the
-[=active document=] of the browsing context is set, run the following steps:
-
-1. Let |context| be the newly created browsing context.
+1. Let |context| be |navigable|'s [=active browsing context=].
 
 1. Let |related browsing contexts| be a [=/set=] containing |context|.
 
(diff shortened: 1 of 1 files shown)


OrKoN:
@jgraham what do you think about adding navigable lifecycle hooks into the HTML spec and correspondingly to WebDriver BiDi? I want to try to define the original opener ID via the created hook later.

jgraham:
Well in general I agree we want to avoid monkeypatching other specs, but have them call directly in to webdriver algorithms when we need to emit an event or similar.

sadym-chromium (approved):

sadym-chromium:
I believe with this change the section [**8.1. HTML**](https://w3c.github.io/webdriver-bidi/#patches-html) is not needed anymore.

sadym-chromium:
I just figured out: this PR changes the [`browsingContext.contextDestroyed`](https://w3c.github.io/webdriver-bidi/#event-browsingContext-contextDestroyed) event behavior. It used to be emitted only for the most-top-level browsing context to be destroyed. [Step 1](https://w3c.github.io/webdriver-bidi/#patches-html ):
> If this is **not a recursive invocation** of this algorithm, call any browsing context tree discarded steps defined in other applicable specifications with browsingContext.

With this change, the event will be emitted for all the nested navigables, effectively emitting the "destroyed" event for all the children.

We can either update the HTML spec (as part of your [PR](https://github.com/whatwg/html/pull/10329)) with a flag if the destroy is recursive (I'm not sure if this will actually work out), or we can accept the changed `browsingContext.contextDestroyed` logic.

OrKoN:
@sadym-chromium see the discussion in https://github.com/whatwg/html/pull/10329#discussion_r1625527017

OrKoN:
@sadym-chromium @jgraham could you please review this CL?

sadym-chromium:
> @sadym-chromium see the discussion in [whatwg/html#10329 (comment)](https://github.com/whatwg/html/pull/10329#discussion_r1625527017)

This assumption is incorrect. The events are not emitted recursively for the nested navigables.

> ~With this change, the event will be emitted for all the nested navigables, effectively emitting the "destroyed" event for all the children.~

sadym-chromium (approved):

sadym-chromium:
@OrKoN for my understanding of this spec version, the `browsingContext.contextDestroyed` event should not be emitted for the iframes when the parent navigable is navigated: https://github.com/web-platform-tests/wpt/pull/46562/files

index.bs
```diff
@@ -4183,9 +4183,9 @@ The [=remote end event trigger=] is:
...
 <div algorithm="remote end event trigger for browsingContext.contexDestroyed">
 
-Define the following [=browsing context tree discarded=] steps:
-
-1. Let |context| be the browsing context being discarded.
+The [=remote end event trigger=] is
+the <dfn export>WebDriver BiDi navigable destroyed</dfn> steps given
+<var ignore>navigable</var> and |context|:
 
 1. Let |params| be the result of [=get the browsing context info=], given
```
jgraham, line 4190: `|context|` now seems to be undefined here.

jgraham (changes requested):

index.bs
OrKoN: I think it is defined on line 4188 or do you mean something else?

jgraham: Oh, sorry, I misread it.

jgraham (approved):

PR #711, opened 2024-05-13
Add validtion for headers and method in network.continueRequest

juliandescottes:


juliandescottes:
I have only updated the parameters for continueRequest here, once we agree on the general approach I can do the same for

OrKoN:
@Lightning00Blade could you please check how we validate headers and the method in CDP?

index.bs
```diff
@@ -592,6 +596,12 @@ WebDriver BiDi extends the set of [=error codes=] from [[WEBDRIVER|WebDriver]]
 with the following additional codes:
 
 <dl>
+  <dt><dfn for=errors export>invalid header</dfn>
+  <dd>Tried to use an invalid [=/header].
+
+  <dt><dfn for=errors export>invalid method</dfn>
+  <dd>Tried to use an invalid [=/method].
```
whimboo, line 603: Do we need these new errors or could we just as usual report an `invalid argument` error instead?

juliandescottes: As discussed, the idea behind having specific errors was to help with https://github.com/puppeteer/puppeteer/issues/12412 

This way Puppeteer could simply check for the Invalid header error message, and that would be cross browser.
We can use invalid argument, and make sure to mention "headers" in the message. @whimboo also suggested updating the spec for `invalid argument` to always contain the name of the argument in the error data.

@jgraham @OrKoN thoughts about using invalid argument vs introducing new errors?

jgraham: I think @whimboo is right, that this is usually the kind of thing that we use the generic `invalid argument` error for.

juliandescottes:
@Lightning00Blade @OrKoN any concern or remark about this change? I'm close to landing the corresponding changes for Firefox for info.

OrKoN (approved):
LGTM (we probably want header validation for the response methods as well=

juliandescottes:
> LGTM (we probably want header validation for the response methods as well=

Thanks, I'll follow up quickly with a PR for the response methods!

PR #721, opened 2024-05-29
Add network.setCacheBehavior command

jgraham:
This currently only allows setting the request cache mode to 'no-store' which bypasses the cache entirely. If we have future use cases for setting different cache modes, this appraoch would be extensible to those use cases.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/721.html" title="Last updated on Jul 3, 2024, 6:12 PM UTC (331e65f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/721/05a283e...331e65f.html" title="Last updated on Jul 3, 2024, 6:12 PM UTC (331e65f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1446,9 +1446,9 @@ To <dfn>cleanup remote end state</dfn>.
 
 1. [=map/Clear=] the [=before request sent map=].
 
-1. Set the [=default cache bypass=] to false.
+1. Set the [=default cache behavior=] to "<code>default</code>".
 
-1. [=set/Empty=] the [=navigable cache bypass set=].
+1. [=map/Clear=] the [=navigable cache behavior map=].
 
 </div>
 
@@ -4733,12 +4733,12 @@ A [=remote end=] has a <dfn>before request sent map</dfn> which is initially an
 empty map.  It's used to track the network events for which a
 <code>network.beforeRequestSent</code> event has already been sent.
 
-A [=remote end=] has a <dfn>default cache bypass</dfn> which is a boolean. It is
-initially false.
+A [=remote end=] has a <dfn>default cache behavior</dfn> which is a string. It is
+initially "<code>default</code>".
 
-A [=remote end=] has a <dfn>navigable cache bypass set</dfn> which is initially an
-empty weak set. It's used to track the [=/top-level traversables=] in which
-network caches are bypassed.
+A [=remote end=] has a <dfn>navigable cache behavior map</dfn> which is a weak
+map between [=/top-level traversables=] and strings representing cache
+behavior. It is initially empty.
 
 ### Network Intercepts ### {#network-intercepts}
 
@@ -6577,22 +6577,22 @@ requests will be affected.
 
 </div>
 
-#### The network.setCacheBypass Command ####  {#command-network-setCacheBypass}
+#### The network.setCacheBehavior Command ####  {#command-network-setCacheBehavior}
 
-The <dfn export for=commands>network.setCacheBypass</dfn> command bypasses the
-network cache for certain requests.
+The <dfn export for=commands>network.setCacheBehavior</dfn> command configures
+the network cache behavior for certain requests.
 
 <dl>
    <dt>Command Type</dt>
    <dd>
     <pre class="cddl remote-cddl">
-      network.SetCacheBypass = (
-        method: "network.setCacheBypass",
-        params: network.SetCacheBypassParameters
+      network.SetCacheBehavior = (
+        method: "network.setCacheBehavior",
+        params: network.SetCacheBehaviorParameters
       )
 
-      network.SetCacheBypassParameters = {
-        bypass: bool,
+      network.SetCacheBehaviorParameters = {
+        cacheBehavior: "default" / "bypass",
         ? contexts: [+browsingContext.BrowsingContext]
       }
     </pre>
@@ -6606,7 +6606,7 @@ network cache for certain requests.
 </dl>
 
 <div algorithm>
-The <dfn export>WebDriver BiDi bypass cache</dfn> steps given |request| are:
+The <dfn export>WebDriver BiDi cache behavior</dfn> steps given |request| are:
 
 1. Let |context| be null.
 
@@ -6618,53 +6618,78 @@ The <dfn export>WebDriver BiDi bypass cache</dfn> steps given |request| are:
      settings|' [=environment settings object/global object=], set |context| to
      the [=top-level browsing context=] for that browsing context.
 
-1. If |context| is not null and [=navigable cache bypass set=] [=set/contains=]
-   |context|, return true.
+1. If |context| is not null and [=navigable cache behavior map=] [=set/contains=]
+   |context|, return [=navigable cache behavior map=][|context|].
 
-1.  Return [=default cache bypass=].
+1. Return [=default cache behavior=].
 
 </div>
 
-<div algorithm="remote end steps for network.setCacheBypass">
+<div algorithm="remote end steps for network.setCacheBehavior">
 The [=remote end steps=] given <var ignore>session</var> and |command parameters| are:
 
-1. Let |bypass| be |command parameters|["<code>bypass</code>"].
+1. Let |behavior| be |command parameters|["<code>cacheBehavior</code>"].
 
 1. If |command parameters| does not [=map/contain=] "<code>contexts</code>":
 
- 1. Set the [=default cache bypass=] to |bypass|.
+  1. Set the [=default cache behavior=] to |behavior|.
 
- 1. If |bypass| is true, perform implementation-defined steps to disable any
-    implementation-specific resource caches for network requests. Otherwise
-    re-enable any implementation-specific resource caches for all contexts.
+  1. [=map/Clear=] [=navigable cache behavior map=].
 
- 1. Return [=success=] with data null.
+  1. Switch on the value of behavior:
+    <dl>
+      <dt>"<code>bypass</code>"
+      <dd>Perform implementation-defined steps to disable any
+        implementation-specific resource caches for all future
+        network requests.
+      <dt>"<code>default</code>"
+      <dd>Perform implementation-defined steps to enable any
+        implementation-specific resource caches that are usually enabled in the
+        current [=remote end=] configuration for all future network requests.
+    </dl>
+
+  1. Return [=success=] with data null.
 
 1. Let |contexts| be an empty [=/set=].
 
 1. For each |context id| of |command parameters|["<code>contexts</code>"]:
 
- 1. Let |context| be the result of [=trying=] to [=get a browsing context=]
+  1. Let |context| be the result of [=trying=] to [=get a browsing context=]
     with |context id|.
 
- 1. If |context| is not a [=top-level browsing context=], return [=error=]
-    with [=error code=] [=invalid argument=].
+  1. If |context| is not a [=top-level browsing context=], return [=error=]
+     with [=error code=] [=invalid argument=].
 
- 1. [=list/Append=] |context| to |contexts|.
+  1. [=list/Append=] |context| to |contexts|.
 
 1. For each |context| in |contexts|:
 
-  1. If |bypass| is true:
+    1. If [=navigable cache behavior map=] [=map/contains=] |context|, and
+       [=navigable cache behavior map=][|context|] is equal to |behavior| then
+       continue.
 
-    1. [=set/append=] |context| to [=navigable cache bypass set=].
+    1. Switch on the value of behavior:
+      <dl>
+        <dt>"<code>bypass</code>"
+        <dd>Perform implementation-defined steps to disable any implementation-specific
+          resource caches for network requests originating from any browsing
+          context for which |context| is the [=top-level browsing context=].
+        <dt>"<code>default</code>"
+        <dd>Perform implementation-defined steps to enable any
+          implementation-specific resource caches that are usually enabled in the
+          current [=remote end=] configuration for network requests
+          originating from any browsing context for which |context| is the
+          [=top-level browsing context=].
+      </dl>
 
-    1. Perform implementation-defined steps to disable any implementation-specific
-       resource caches for network requests originating from |context|.
+    1. If |behavior| is equal to [=default cache behavior=]:
 
-  1. Otherwise, if [=navigable cache bypass set=] [=set/contains=]
-     |context|, [=set/remove=] |context| from [=navigable cache bypass set=] and
-     re-enable any implementation-specific resource caches for network requests
-     originating from |context|.
+      1. If [=navigable cache behavior map=] [=map/contains=] |context|,
+         [=map/remove=] [=navigable cache behavior map=][|context|].
+
+    1. Otherwise:
+
+      1. Set [=navigable cache behavior map=][|context|] to |behavior|.
 
 1. Return [=success=] with data null.
 
--- a/index.bs
+++ b/index.bs
@@ -1450,6 +1450,10 @@ To <dfn>cleanup remote end state</dfn>.
 
 1. [=map/Clear=] the [=navigable cache behavior map=].
 
+1. 1. Perform implementation-defined steps to enable any
+   implementation-specific resource caches that are usually enabled in the
+   current [=remote end=] configuration.
+
 </div>
 
 <div algorithm>
@@ -6640,12 +6644,11 @@ The [=remote end steps=] given <var ignore>session</var> and |command parameters
     <dl>
       <dt>"<code>bypass</code>"
       <dd>Perform implementation-defined steps to disable any
-        implementation-specific resource caches for all future
-        network requests.
+        implementation-specific resource caches.
       <dt>"<code>default</code>"
       <dd>Perform implementation-defined steps to enable any
         implementation-specific resource caches that are usually enabled in the
-        current [=remote end=] configuration for all future network requests.
+        current [=remote end=] configuration.
     </dl>
 
   1. Return [=success=] with data null.
--- a/index.bs
+++ b/index.bs
@@ -4706,7 +4706,7 @@ NetworkCommand = (
   network.FailRequest //
   network.ProvideResponse //
   network.RemoveIntercept //
-  network.SetCacheBehavior //
+  network.SetCacheBehavior
 )
 
 </pre>
--- a/index.bs
+++ b/index.bs
@@ -6577,7 +6577,7 @@ the network cache behaviour for certain requests.
 
       network.SetCacheModeParameters = {
         mode: "no-store" / null,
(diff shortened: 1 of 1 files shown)


jgraham:
Draft Fetch PR: https://github.com/whatwg/fetch/pull/1756

index.bs
```diff
@@ -6438,6 +6445,90 @@ requests will be affected.
...
+   <dd>
+    <pre class="cddl remote-cddl">
+      network.SetCacheMode = (
+        method: "network.setCacheMode",
+        params: network.SetCacheModeParameters
+      )
+
+      network.SetCacheModeParameters = {
+        mode: "no-store" / null,
+        ? contexts: [browsingContext.BrowsingContext]
```
Lightning00Blade, line 6464: Let's make the list have at least 1 element if provided, like other commands.
```suggestion
        ? contexts: [+browsingContext.BrowsingContext]
```

index.bs
```diff
@@ -6438,6 +6445,90 @@ requests will be affected.
 
 </div>
 
+#### The network.setCacheMode Command ####  {#command-network-setCacheMode}
+
+The <dfn export for=commands>network.removeSetCacheMode</dfn> command overrides
```
whimboo, line 6450:
```suggestion
The <dfn export for=commands>network.setCacheMode</dfn> command overrides
```

index.bs
```diff
@@ -6438,6 +6445,90 @@ requests will be affected.
...
+  1. Let |environment settings| be |request|'s [=request/window=]
+
+  1. If there is a [=/browsing context=] whose [=active window=] is |environment
+     settings|' [=environment settings object/global object=], set |context| to
+     the [=top-level browsing context=] for that context.
+
+1. If |context| is not null:
+
+  1. If [=cache mode override map=] [=map/contains=] |context|, set
+     |request|&apos;s [=request/cache mode=] to [=cache mode override
```
whimboo, line 6493:
```suggestion
     |request|'s [=request/cache mode=] to [=cache mode override
```

index.bs
```diff
@@ -6438,6 +6445,90 @@ requests will be affected.
...
+
+1. If |command parameters| does not [=map/contain=] "<code>contexts</code>", set
+   the [=default cache mode override=] to |mode| and return [=success=] with
+   data null.
+
+1. Let |contexts| be an empty [=/list=].
+
+1. For each |context id| of |command parameters|["<code>contexts</code>"]:
+
+  1. Let |context| be the result of [=trying=] to [=get a browsing context=]
```
whimboo, line 6514: This should be [similar to what we already have for `network.addIntercept`](https://w3c.github.io/webdriver-bidi/#command-network-addIntercept) so that we use a `Set` and can filter out identical browsing contexts.

index.bs
```diff
@@ -6438,6 +6445,90 @@ requests will be affected.
...
+     with |context id|.
+
+  1. If |context| is not a [=top-level browsing context=], return [=error=]
+     with [=error code=] [=invalid argument=].
+
+  1. [=list/Append=] |context| to |contexts|.
+
+1. For each |context| in |contexts|:
```
whimboo, line 6522: We run the loop twice. Should we combine steps 4 and 5?

index.bs
```diff
@@ -6438,6 +6445,90 @@ requests will be affected.
...
+   <dt>Return Type</dt>
+   <dd>
+    <pre class="cddl">
+      EmptyResult
+    </pre>
+   </dd>
+</dl>
+
+<div algorithm>
+The <dfn export>WebDriver BiDi update request cache mode</dfn> steps given
```
whimboo, line 6477: I assume we have to land these changes first and afterward the fetch ones?

whimboo (changes requested):

index.bs
jgraham: Yes.

index.bs
```diff
@@ -6438,6 +6445,90 @@ requests will be affected.
...
+     the [=top-level browsing context=] for that context.
+
+1. If |context| is not null:
+
+  1. If [=cache mode override map=] [=map/contains=] |context|, set
+     |request|&apos;s [=request/cache mode=] to [=cache mode override
+     map=][|context|] and return.
+
+1. If [=default cache mode override=] is not null, set
+     |request|&apos;s [=request/cache mode=] to [=default cache mode override=].
```
jgraham, line 6497:
```suggestion
     |request|'s [=request/cache mode=] to [=default cache mode override=].
```

index.bs
jgraham: It doesn't really matter (here, or afaict in `addIntercept`) because it's idempotent, so processing the same context twice doesn't change the outcome. Indeed the `addIntercept` text wrongly suggests it's possible to get an empty list of contexts (probably because of a CDDL change). Anyway, unless we were planning to change things so that passing a context twice is actually an error, I'd prefer to leave things as they are.

index.bs
jgraham: We can't. We first need to check that we have only valid context ids and then perform updates. Otherwise a contexts list like `[valid context, invalid context]` would update the state for `valid context` before throwing an error.

whimboo: Good point! That would be indeed an unwanted behavior.

index.bs
whimboo: Oh and two more questions (which might also answer the fetch spec but maybe you could tell right away):

1. This also affects normal navigation (including reload)?
2. What if no network interception is enabled? It has to work then as well, right?

jgraham: Yes, this should affect all network requests, including reloads, whether or not network interception is enabled.  

I think we might want to also allow setting the cache mode when intercepting at the `beforeRequestSent` stage, in case you really only care about certain requests, but we can handle that in a separate PR.

whimboo: That is a good idea, could you at least file an issue for that? Thanks.

whimboo (approved):
It looks fine to me. Lets have @OrKoN or @Lightning00Blade to have another look.

OrKoN:
@sadym-chromium would you mind reviewing?

index.bs
```diff
@@ -6438,6 +6445,90 @@ requests will be affected.
...
+  1. Let |environment settings| be |request|'s [=request/window=]
+
+  1. If there is a [=/browsing context=] whose [=active window=] is |environment
+     settings|' [=environment settings object/global object=], set |context| to
+     the [=top-level browsing context=] for that context.
+
+1. If |context| is not null:
+
+  1. If [=cache mode override map=] [=map/contains=] |context|, set
+     |request|'s [=request/cache mode=] to [=cache mode override
```
OrKoN, line 6493: so one difference to the CDP implementation and what Puppeteer relied on is that in CDP this command also disabled the memory cache which does not seem to be specified https://github.com/whatwg/fetch/issues/1400 there are some definitions for the memory image cache https://html.spec.whatwg.org/multipage/images.html#the-list-of-available-images so perhaps we should include it at least?

OrKoN, line 6493: Another difference I see is how the CORS preflight cache is handled https://fetch.spec.whatwg.org/#cors-preflight-cache CDP implementation also disables it resulting in pre-flight requests being sent every time.

juliandescottes:
Had a quick look at how Firefox behaves when disabling cache from DevTools (which we intend to reuse for BiDi).
- for CORS preflight requests, Firefox will also send the preflight requests every time
- for the various image/stylesheet caches, Firefox will also bypass them and re-issue the requests
- however for requests handled by a service worker, Firefox will not disable the cache used by the service worker

jgraham:
So I'm not really sure what to do with this.

It seems like talking in terms of the cache mode has made an interop difference visible: Firefox neither reads from nor writes to the cache, but Chrome does write to it. That would affect what happens if you do a no-cache request followed by a cached request.

However for the non-HTTP caches, working purely in terms of the cache mode on the request seems like it maybe doesn't work well? Per fetch it seems like requests use the CORS preflight cache irrespective of their cache-mode. So we could have a webdriver-specific rule that we should clear the preflight cache for a request if the override mode is, say, `no-store` or `reload`, but then the WebDriver API has different behaviour from the DOM API. Or we could have a separate flag to bypass the CORS cache (i.e. something like network.setCacheMode({cacheMode, contexts, bypassCORS})`, but that seems potentially unwieldy.

This also applies to all the possible in-memory caches, but even more so since the details may differ between browsers.

Or the other option is to just support the bare "bypass all caches" use case instead of specifying a specific cache mode i.e. make the command `network.setCacheBypass({bypass, contexts})`. That seems close to what CDP has, but it's rather inflexible if we want to offer more control (similar to existing DOM APIs) in the future.

index.bs
```diff
@@ -6554,6 +6577,98 @@ requests will be affected.
...
+</div>
+
+<div algorithm="remote end steps for network.setCacheBypass">
+The [=remote end steps=] given <var ignore>session</var> and |command parameters| are:
+
+1. Let |bypass| be |command parameters|["<code>bypass</code>"].
+
+1. If |command parameters| does not [=map/contain=] "<code>contexts</code>":
+
+ 1. Set the [=default cache bypass=] to |bypass|.
```
lutien, line 6635:
```suggestion
  1. Set the [=default cache bypass=] to |bypass|.
```
I think here and another to points in this list, there should be one more space in the beginning, at the moment it's not align correctly in the generated html

index.bs
```diff
@@ -6554,6 +6577,98 @@ requests will be affected.
...
+   <dt>Return Type</dt>
+   <dd>
+    <pre class="cddl">
+      EmptyResult
+    </pre>
+   </dd>
+</dl>
+
+<div algorithm>
+The <dfn export>WebDriver BiDi bypass cache</dfn> steps given |request| are:
```
OrKoN, line 6609: The fetch spec draft is not updated yet, are these steps going to be used there?

index.bs
```diff
@@ -6554,6 +6577,98 @@ requests will be affected.
...
+
+1. Let |bypass| be |command parameters|["<code>bypass</code>"].
+
+1. If |command parameters| does not [=map/contain=] "<code>contexts</code>":
+
+ 1. Set the [=default cache bypass=] to |bypass|.
+
+ 1. If |bypass| is true, perform implementation-defined steps to disable any
+    implementation-specific resource caches for network requests. Otherwise
+    re-enable any implementation-specific resource caches for all contexts.
```
whimboo, line 6639: What about the status of the `navigable cache bypass set`. I assume we should update it here as well for both scenarios.

whimboo (changes requested):
Resetting my review given that some more work is needed.

index.bs
```diff
@@ -1435,10 +1435,23 @@ To <dfn>cleanup the session</dfn> given |session|:
...
+<div algorithm>
+To <dfn>cleanup remote end state</dfn>.
+
+1. [=map/Clear=] the [=before request sent map=].
+
+1. Set the [=default cache bypass=] to false.
+
+1. [=set/Empty=] the [=navigable cache bypass set=].
+
+</div>
```
juliandescottes, line 1453: For multiple sessions, it could be an issue to clear the flag + set when one session ends?

lutien: I've read again, and I think it actually implies that we should run these steps when/if all the sessions are closed, so should be fine with multiple sessions, I guess.

juliandescottes: Oh right indeed, thanks for checking.

index.bs
```diff
@@ -6554,6 +6577,98 @@ requests will be affected.
...
+
+    1. [=set/append=] |context| to [=navigable cache bypass set=].
+
+    1. Perform implementation-defined steps to disable any implementation-specific
+       resource caches for network requests originating from |context|.
+
+  1. Otherwise, if [=navigable cache bypass set=] [=set/contains=]
+     |context|, [=set/remove=] |context| from [=navigable cache bypass set=] and
+     re-enable any implementation-specific resource caches for network requests
+     originating from |context|.
```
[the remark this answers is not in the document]
OrKoN: I think if there is a global default state it should apply to newly created contexts as well. Currently, it seems like only the spec-defined caches would be affected but not the implementation-defined caches?  Note that for Puppeteer there is currently no use case for bypassing the cache for the whole browser so we can also exclude the global default state for now.

index.bs
sadym-chromium: The current version states the default "cache bypass" should not override specific browsing context's ones. @whimboo suggests to override it. I agree with @whimboo that `SetCacheBypassParameters` without `contexts` should set it to all the contexts regardless of their `cacheBypass` state.

index.bs
```diff
@@ -6554,6 +6577,98 @@ requests will be affected.
...
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl remote-cddl">
+      network.SetCacheBypass = (
+        method: "network.setCacheBypass",
+        params: network.SetCacheBypassParameters
+      )
+
+      network.SetCacheBypassParameters = {
+        bypass: bool,
```
sadym-chromium, line 6595: Would it make the logic more clear if we change `bypass: bool` to `mode: "bypass" | "default"`?

sadym-chromium: I believe the `bypass: false` actually means not _"do not bypass cache"_, but rather ''remove custom cache behavior"

index.bs
OrKoN: `implementation-specific` caches are likely the ones that are not specified by the fetch spec (or other specs). The specified caches should be handled via the fetch spec change and make use of the `navigable cache bypass set` IIUC.

index.bs
OrKoN: I like this proposal: I think it would make it easier to extend the behavior for bypassing the cache mode if needed.

OrKoN (changes requested):
The summary of changes that are needed:

1) The fetch spec PR needs an update to reflect latest changes.
2) Remove the ability to set global cache bypass or define how it works for contexts created after the global override was set. 
3) @sadym-chromium's suggested change to use an enum instead of a boolean so that it could be extended later 
4) indentation issue pointed out by @lutien

index.bs
```diff
@@ -4689,7 +4705,8 @@ NetworkCommand = (
   network.ContinueWithAuth //
   network.FailRequest //
   network.ProvideResponse //
-  network.RemoveIntercept
+  network.RemoveIntercept //
+  network.SetCacheBypass //
```
OrKoN, line 4709:
```suggestion
  network.SetCacheBehavior //
```

index.bs
```diff
@@ -4689,7 +4705,8 @@ NetworkCommand = (
   network.ContinueWithAuth //
   network.FailRequest //
   network.ProvideResponse //
-  network.RemoveIntercept
+  network.RemoveIntercept //
+  network.SetCacheBehavior //
```
whimboo, line 4709:
```suggestion
  network.SetCacheBehavior
```

index.bs
```diff
@@ -6554,6 +6577,123 @@ requests will be affected.
...
+<div algorithm="remote end steps for network.setCacheBehavior">
+The [=remote end steps=] given <var ignore>session</var> and |command parameters| are:
+
+1. Let |behavior| be |command parameters|["<code>cacheBehavior</code>"].
+
+1. If |command parameters| does not [=map/contain=] "<code>contexts</code>":
+
+  1. Set the [=default cache behavior=] to |behavior|.
+
+  1. [=map/Clear=] [=navigable cache behavior map=].
```
sadym-chromium, line 6637: Note: the global cache behavior erases all the specific browsing context's cache behavior previously set.

sadym-chromium (approved):

OrKoN (approved):

index.bs
```diff
@@ -6554,6 +6577,123 @@ requests will be affected.
...
+        <dt>"<code>bypass</code>"
+        <dd>Perform implementation-defined steps to disable any implementation-specific
+          resource caches for network requests originating from any browsing
+          context for which |context| is the [=top-level browsing context=].
+        <dt>"<code>default</code>"
+        <dd>Perform implementation-defined steps to enable any
+          implementation-specific resource caches that are usually enabled in the
+          current [=remote end=] configuration for network requests
+          originating from any browsing context for which |context| is the
+          [=top-level browsing context=].
```
juliandescottes, line 6682: It's not clear to me if this takes precedence over the global setting or not?

For instance if I first setCacheBehavior "bypass" globally, and then set it to "default" for one context, what should happen?

I believe that the intent is that it should override whatever is the default setting, but since the global usage of the command is supposed to enable/disable implementation specific resource caches "for all future requests", it's hard to tell which one will prevail?

The actual mechanism for enabling/disabling resource caches is not clearly specified so I imagine it's difficult to formalize, but maybe we need a note here to clarify what happens in such a case?

whimboo (changes requested):
Looks like something is wrong given that the preview fails to build. @jgraham can you please check?

index.bs
jgraham: Yes, I think that's the same as the way that enabling an event globally works; if you subscribe for a specific context and then subscribe for all contexts and then unsubscribe for all contexts you don't end up with the event subscription for the original set of contexts.

index.bs
jgraham: I've tried to make this explicit by actually adding a step when we create a context we explicitly disable the caches if it's has a top level context where the caches are explicitly disabled, or if the default is to disable the caches. That assumes that in the absence of that step we'd get the default behavior. I hope that solves the problem well enough for now.

juliandescottes: That works for me, thanks!

juliandescottes (approved):

index.bs
```diff
@@ -6554,6 +6585,134 @@ requests will be affected.
...
+
+1. If |command parameters| does not [=map/contain=] "<code>contexts</code>":
+
+  1. Set the [=default cache behavior=] to |behavior|.
+
+  1. [=map/Clear=] [=navigable cache behavior map=].
+
+  1. Switch on the value of behavior:
+    <dl>
+      <dt>"<code>bypass</code>"
```
whimboo, line 6661: @jgraham one more question regarding the PR of the Fetch specification. Would it be helpful to actually export the two cache modes so that we don't have to use string values?

jgraham: Maybe, although we still have to use a string, it just becomes a string we can link to something. Fetch itself doesn't seem to consistently do that e.g. it doesn't for cache-mode, but we could. In any case since that doesn't have any normative impact we can easily do it in a followup.

whimboo: Sounds fine with a follow-up if needed.

whimboo (approved):
Thanks it looks fine to me now as well.

whimboo:
I'm going to update the [existing tentative tests](https://github.com/web-platform-tests/wpt/pull/46935) (which failed to merge upstream) via https://bugzilla.mozilla.org/show_bug.cgi?id=1906100.

PR #725, opened 2024-06-04
Always return a valid value for the "webSocketUrl" capability

whimboo:
Per CDDL it's not allowed to return a boolean value for `webSocketUrl`. As such we need to check the existence and value of this capability first so that it can be removed if it's not `true`.

@jgraham can you please review? Thanks.


index.bs
```diff
@@ -1166,14 +1167,13 @@ WebDriver clients opt in to a bidirectional connection by requesting the
 The [=WebDriver new session algorithm=] defined by this specification,
 with parameters |session|, |capabilities|, and |flags| is:
 
-1. If |flags| contains "<code>bidi</code>", return.
-
 1. Let |webSocketUrl| be the result of [=getting a property=] named
    "<code>webSocketUrl</code>" from |capabilities|.
 
-1. If |webSocketUrl| is undefined or false, return.
+1. If |webSocketUrl| is undefined or false, calling
```
jgraham, line 1173: It isn't actually possible for the value to be anything other than missing or `true`. That's because `capabilities` here is given by the output of https://w3c.github.io/webdriver/#dfn-matching-capabilities which calls the matched capability deserialization algorithm in https://w3c.github.io/webdriver-bidi/#type-session-CapabilityRequest and doesn't set a property when the return value is null, which it is when the input is anything other than `true`. So I think you can just remove `or false` here and revert the rest of the change.

jgraham (changes requested):

index.bs
whimboo: Oh that's a good point. Given that we do not have matching implemented yet in our BiDi code this was missing. I'll revert and check geckodriver because we get `false` as value in Marionette.

whimboo:
@jgraham please re-review. Thanks.

jgraham (approved):

PR #726, opened 2024-06-07
Rename |is root| to |include parent id|

OrKoN:
|is root| makes it look like it indicates whether the context is a root of the context tree (parent id === null) but it actually indicates whether the |parent id| needs to be included into the resulting info object.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/726.html" title="Last updated on Jun 11, 2024, 2:23 PM UTC (a77d336)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/726/5614cf1...a77d336.html" title="Last updated on Jun 11, 2024, 2:23 PM UTC (a77d336)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2379,7 +2379,7 @@ TODO: make this return a list in document order
 
 <div algorithm>
 To <dfn>get the browsing context info</dfn> given |context|,
-|max depth| and |is root|:
+|max depth| and |include parent id|:
 
 1. Let |context id| be the [=browsing context id=] for |context|.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2419,7 +2419,7 @@ To <dfn>get the browsing context info</dfn> given |context|,
 1. Let |context info| be a [=/map=] matching the
    <code>browsingContext.Info</code> production with the <code>context</code>
    field set to |context id|, the <code>parent</code> field set to |parent id|
-   if |is root| is <code>true</code>, or unset otherwise, the <code>url</code>
+   if |is root| is <code>false</code>, or unset otherwise, the <code>url</code>
```
OrKoN, line 2422: why do we need |is root|? it |is root| being true the same as |parent id| being null?

sadym-chromium (approved):

index.bs
OrKoN: ah, that clears this up then, thanks!

whimboo (approved):
This change looks fine and indeed makes it much easier to understand what this flag is actually used for.

Can you please also update the WebDriver BiDi client and the wpt tests for the updated name? Thanks.

PR #730, opened 2024-06-14
Change originalOpener to be optional

OrKoN:
Since the opener Id would not be present on each context, we think it makes sense to actually define it as an optional non-nullable attribute.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/730.html" title="Last updated on Jun 14, 2024, 12:47 PM UTC (4302f62)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/730/a86a0c7...4302f62.html" title="Last updated on Jun 14, 2024, 12:47 PM UTC (4302f62)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2344,7 +2344,7 @@ browsingContext.InfoList = [*browsingContext.Info]
 browsingContext.Info = {
   children: browsingContext.InfoList / null,
   context: browsingContext.BrowsingContext,
-  originalOpener: browsingContext.BrowsingContext / null,
+  ? originalOpener: browsingContext.BrowsingContext,
   url: text,
   userContext: browser.UserContext,
   ? parent: browsingContext.BrowsingContext / null,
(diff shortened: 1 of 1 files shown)


Lightning00Blade (approved):

OrKoN:
@jgraham @whimboo as a follow-up to the opener ID change, we actually think it will be more convenient to declare it as optional so that null values are not sent around. WDYT?

whimboo:
> @jgraham @whimboo as a follow-up to the opener ID change, we actually think it will be more convenient to declare it as optional so that null values are not sent around. WDYT?

So only top-level browsing context will get an opener, right? Or are there also situations when the opener can be there `null` as well? Probably not, but given that we use the same `BrowsingContextInfo` struct for all browsing contexts it would make sense to set it as optional. I'm going to give @jgraham the final last word here.

OrKoN:
> So only top-level browsing context will get an opener, right? Or are there also situations when the opener can be there null as well?

null would also mean there is no opener, so not setting the attribute in that case sounds fine to me.

whimboo (approved):
Forgot to approve the changes by last week.

OrKoN:
As discussed offline, we could return unset for child navigables but we have no strong arguments for doing it and would like to keep the spec as is.

Issue #502, opened 2023-07-20
Add browsingContext event to monitor history API usage


juliandescottes:
Current browsingContext events only cover regular page navigation or fragment navigation ([navigationStarted](https://w3c.github.io/webdriver-bidi/#event-browsingContext-navigationStarted), [fragmentNavigated](https://w3c.github.io/webdriver-bidi/#event-browsingContext-fragmentNavigated), ...)

For parity with https://chromedevtools.github.io/devtools-protocol/tot/Page/#event-navigatedWithinDocument we could also expose additional events for history API usage.

jgraham:
We previously had https://github.com/w3c/webdriver-bidi/issues/402 in which I suggested `fragmentNavigated` should cover this use case. But APIs like `pushState()` don't end up going through the navigation path in HTML and so aren't covered by that event as specified at present. So there are two options here:
* Change HTML so that we emit `fragmentNavigated` for anything that changes the position in the session history without changing the active document.
* Add a new event, specifically for the case where there isn't an actual navigation, but something is using one of the DOM APIs to create new history entries.

OrKoN:
Option #1 sounds good to me.

OrKoN:
Option #2 would require some implementation effort on our side but if there is a use case that is currently not covered, it sounds good too.

OrKoN:
> speaking with my selenium hat on I think having a single event type is much easier DX for our users

sounds like a good argument for the option #1

jrandolf:
I'd suggest renaming the `fragmentNavigated` event to something along the lines of `sameDocumentNavigated` and doing (1). If the use-case arises that we need to separate history and fragment navigation. We can add a type of the navigation.

whimboo:
Removing or renaming an existing API disrupts our backward compatibility model with WebDriver. If we intend to introduce a new name such as `sameDocumentNavigated`, we may need to retain `frameNavigated` for the time being. We could consider marking it as deprecated and potentially remove it later once we are confident that clients have been updated?

whimboo:
@OrKoN is there any more work to do or can we close the issue as well?

OrKoN:
I think we can close it. There are more things that could be done such as adding the indication of the method that modifies the history or the serialized history state but I think it would make sense to wait for some client requests for those.

whimboo:
Sounds fine. Thanks a lot for working on it!

PR #740, opened 2024-07-02
Add browsingContext.historyUpdated event

OrKoN:
This change adds `browsingContext.historyUpdated` event to notify the client when a history update happens.

Issue: https://github.com/w3c/webdriver-bidi/issues/502
HTML spec: https://github.com/whatwg/html/pull/10587

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/740.html" title="Last updated on Aug 29, 2024, 2:34 PM UTC (d1f691f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/740/287e9c6...d1f691f.html" title="Last updated on Aug 29, 2024, 2:34 PM UTC (d1f691f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2333,6 +2333,7 @@ BrowsingContextEvent = (
   browsingContext.DomContentLoaded //
   browsingContext.DownloadWillBegin //
   browsingContext.FragmentNavigated //
+  browsingContext.HistoryUpdated //
   browsingContext.Load //
   browsingContext.NavigationAborted //
   browsingContext.NavigationFailed //
@@ -4360,6 +4361,38 @@ navigated</dfn> steps given |navigable| and |navigation status|:
 
 1. For each |session| in the [=set of sessions for which an event is enabled=]
    given "<code>browsingContext.fragmentNavigated</code>" and |related navigable|:
+
+  1. [=Emit an event=] with |session| and |body|.
+
+</div>
+
+#### The browsingContext.historyUpdated Event #### {#event-browsingContext-historyUpdated}
+
+<dl>
+   <dt>Event Type</dt>
+   <dd>
+      <pre class="cddl local-cddl">
+        browsingContext.HistoryUpdated = (
+         method: "browsingContext.historyUpdated",
+         params: {
+            url: text,
+         }
+        )
+      </pre>
+   </dd>
+</dl>
+
+<div algorithm>
+The [=remote end event trigger=] is the <dfn export>WebDriver BiDi history updated</dfn> steps given |navigable|:
+
+1. Let |body| be a [=/map=] matching the
+    <code>browsingContext.HistoryUpdated</code> production, with the
+    <code>url</code> field set to |navigable|'s [=active browsing context=]'s [=active document=]'s <a spec=DOM>URL</a>.
+
+1. Let |related browsing contexts| be a [=/set=] containing |navigable|'s [=active browsing context=].
+
+1. For each |session| in the [=set of sessions for which an event is enabled=]
+   given "<code>browsingContext.historyUpdated</code>" and |related browsing contexts|:
 
   1. [=Emit an event=] with |session| and |body|.
 
--- a/index.bs
+++ b/index.bs
@@ -4388,10 +4388,10 @@ navigated</dfn> steps given |navigable| and |navigation status|:
 <div algorithm>
 The [=remote end event trigger=] is the <dfn export>WebDriver BiDi history updated</dfn> steps given |navigable|:
 
-1. Let |url| be the result of running the [=URL serializer=], 
+1. Let |url| be the result of running the [=URL serializer=],
    given |navigable|'s [=active browsing context=]'s [=active document=]'s <a spec=dom>URL</a>.
 
-1. Let |params| be a [=/map=] matching the <code>browsingContext.HistoryUpdatedParameters</code> production, 
+1. Let |params| be a [=/map=] matching the <code>browsingContext.HistoryUpdatedParameters</code> production,
    with the <code>url</code> field set to |url| and
    the <code>context<code> field set to |navigable|'s [=navigable id=].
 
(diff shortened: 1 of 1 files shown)


whimboo (changes requested):
Thank you Alex for getting started on this event! For now I have two questions. See inline.

index.bs
```diff
@@ -4377,6 +4378,38 @@ navigated</dfn> steps given |context| and |navigation status|:
...
+         params: {
+            url: text,
+         }
+        )
+      </pre>
+   </dd>
+</dl>
+
+<div algorithm>
+The [=remote end event trigger=] is the <dfn export>WebDriver BiDi history updated</dfn> steps given |navigable|:
```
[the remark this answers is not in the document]
OrKoN: Navigables renaming is not required since this is for html spec that is already written in terms of navigables. Navigation status does not apply here since history update is not a navigation.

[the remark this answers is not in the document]
OrKoN: Fragment navigated is a navigation:

1) fragment navigations go through the navigate algorithm https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigate-fragid

2) non fragment “navigations” (in quotes) do not go through the navigate algorithm https://html.spec.whatwg.org/#navigate-non-frag-sync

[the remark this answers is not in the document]
OrKoN: After a discussion, we changed the PR to emit the event in https://html.spec.whatwg.org/#shared-history-push/replace-state-steps as it seems to better match the use case. The previously considered integration point in `URL and history update steps` seems to be used internally by the HTML spec for regular navigation types.

sadym-chromium (approved):
LGTM

index.bs
```diff
@@ -4377,6 +4378,38 @@ navigated</dfn> steps given |context| and |navigation status|:
...
+
+1. Let |body| be a [=/map=] matching the
+    <code>browsingContext.HistoryUpdated</code> production, with the
+    <code>url</code> field set to |navigable|'s [=active browsing context=]'s [=active document=]'s <a spec=DOM>URL</a>.
```
juliandescottes, line 4402: If `url` is text, I imagine this should rather be 

>  url field set to the result of running the [URL serializer](https://url.spec.whatwg.org/#concept-url-serializer), given navigable's active document’s [URL](https://dom.spec.whatwg.org/#concept-document-url).

Or something similar?

juliandescottes, line 4402: We usually have a type for the params, so I'm not sure if works exactly the same here, but `url` is a field under body.params, whereas the algorithm makes it a field of body.

index.bs
```diff
@@ -4377,6 +4378,38 @@ navigated</dfn> steps given |context| and |navigation status|:
...
+
+<dl>
+   <dt>Event Type</dt>
+   <dd>
+      <pre class="cddl local-cddl">
+        browsingContext.HistoryUpdated = (
+         method: "browsingContext.historyUpdated",
+         params: {
+            url: text,
+         }
```
juliandescottes, line 4391: We should at least have the `context` id where the history update happened? I know this is not a navigation but a `NavigationInfo` with navigation=null (which is supported) would probably be a good fit?

juliandescottes (changes requested):
Thanks, I feel like we should at least have the context info in the event payload, otherwise it will be hard to use?

index.bs

index.bs
OrKoN: Updated. History updates are handled differently than navigations so I do not think NavigationInfo would be a good fit.

index.bs
OrKoN: updated

OrKoN:
> Thanks, I feel like we should at least have the context info in the event payload, otherwise it will be hard to use?

thanks for reviewing, totally! It was a while since I started this PR draft 😅

juliandescottes (approved):
Thanks for the update, looks good to me!

index.bs
```diff
@@ -4365,6 +4366,48 @@ navigated</dfn> steps given |navigable| and |navigation status|:
 
 </div>
 
+#### The browsingContext.historyUpdated Event #### {#event-browsingContext-historyUpdated}
+
```
jgraham, line 4370: I know we don't have this for the other events, but it could be useful to have some informative text indicating that this is only fired for `pushState` and friends, not for history updates caused by navigation.

index.bs
```diff
@@ -4365,6 +4366,48 @@ navigated</dfn> steps given |navigable| and |navigation status|:
...
+<dl>
+   <dt>Event Type</dt>
+   <dd>
+      <pre class="cddl local-cddl">
+        browsingContext.HistoryUpdated = (
+          method: "browsingContext.historyUpdated",
+          params: browsingContext.HistoryUpdatedParameters
+        )
+
+        browsingContext.HistoryUpdatedParameters = {
```
jgraham, line 4380: I wonder if we should also include a string indicating the type of the history update to distinguish `pushState` from `replaceState`?

jgraham (approved):
A couple of comments, but no blockers.

index.bs
OrKoN: I think we could add it eventually. So far we didn't expose it and didn't have a use case for it so adding it would involve extra implementation work. Also we could add the serialized history state to be able to distinguish multiple updates if there are use cases.

OrKoN:
There have been additional discussions on the HTML spec PR resulting in the call being moved to a different place in the HTML spec. PTAL and let me know if you have concerns about it.

index.bs
OrKoN: The discussion on the HTML spec resulted in this event being sent whenever there is a user-observable history update according to the HTML spec so I think we probably do not need a note but I can add it as a follow-up.

OrKoN:
I am going to land this PR as there seem to be no additional comments. If there are still any concerns, please let me know.

Issue #585, opened 2023-11-01
Ability to detect active tab


Danp2:
In Webdriver classic, we have [get-window-handle](https://www.w3.org/TR/webdriver/#get-window-handle), but there doesn't appear to be an equivalent command in the current BiDi spec. See recent discussion [here](https://matrix.to/#/!PbfmTydoOIZebaNppn:mozilla.org/$6a4tHIXvkx4vaRCfzt59s88-0pwOmY7-OG0nZsbqruU?via=mozilla.org&via=matrix.org&via=archlinux.org). 

Would it be feasible to add a field to `browsingContext.getTree` that would indicate the active tab?

jimevans:
@Danp2 So just to clarify and be a bit pedantic, what do you mean by "active" tab? There is nothing in the WebDriver Classic spec that defines the "active tab" to be the one with the "focus" in the operating system/window manager. All it indicates is the tab against which commands should be directed. While it's true that the window handle returned by the "get window handle" command often _is_ the one with the OS focus, it's not _required_. In point of fact, it's not even required by the spec to be the "active tab" in a browser that supports tabbed browsing. If you're looking for something to tell you definitively which tab/window has the OS focus, that probably won't be in scope for the WebDriver specification.

Danp2:
@jimevans Good point. In WD Classic, [Switch to Window ](https://w3c.github.io/webdriver/#switch-to-window) includes the following note --

> Switching window will select the current top-level browsing context used as the target for all subsequent commands. In a tabbed browser, this will typically make the tab containing the browsing context the selected tab. 

I am the maintainer of a set of AutoIt UDFs that I would like to eventually support WD BiDi. For backward compatibility, I was hoping to mimic the above behavior by using the "target" BC for BiDi requests if one isn't specified by the calling routine.

whimboo:
So I think that we actually have to differentiate about which browsing context we talk about. What the request basically wants is that we can check which of the open tabs (whether in which window) is the active one. And hereby we most likely want to set the flag for that top-level browsing context that is part of the top-most window. 

Regarding WebDriver classic we indeed always select and focus the tab that the `Switch To Window` command switches to. Only for internal Firefox testing we have an option to keep the current tab the focused (active) one. Please remember that if we would run tests in a background tab those would have to cope with throttling, which is not wanted.

[...]

PR #752, opened 2024-07-18
Support setting client window state

jgraham:
Client windows here are the desktop environment windows containing the browser viewport.

This adds two commands:

* browser.getClientWindows that returns a list of all client windows
* browser.setClientWindowState that sets the state (fullscreen/normal/minimized/maximized) of a specific client window, and also supports resizing of a specific client window.

It also adds the client window id when getting browsing context info.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/752.html" title="Last updated on Sep 19, 2024, 10:06 AM UTC (4f890ca)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/752/712c17f...4f890ca.html" title="Last updated on Sep 19, 2024, 10:06 AM UTC (4f890ca)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -89,6 +89,7 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
     text: no such alert; url: dfn-no-such-alert
     text: no such element; url: dfn-no-such-element
     text: no such frame; url: dfn-no-such-frame
+    text: no such window; url: dfn-no-such-window
     text: parse a page range; url: dfn-parse-a-page-range
     text: handler; for: prompt handler configuration; url: dfn-handler
     text: process capabilities; url: dfn-processing-capabilities
@@ -253,6 +254,7 @@ spec: CSSOM-VIEW; urlPrefix: https://drafts.csswg.org/cssom-view/
     text: visual viewport page left; url: #dom-visualviewport-pageleft
     text: visual viewport page top; url: #dom-visualviewport-pagetop
     text: visual viewport; url: #visual-viewport
+    text: web-exposed screen area; url: #web-exposed-screen-area
 spec: DOM; urlPrefix: https://dom.spec.whatwg.org/
   type: dfn
     text: root; url: #concept-tree-root
@@ -260,6 +262,11 @@ spec: DOM; urlPrefix: https://dom.spec.whatwg.org/
     text: evaluate; url: #dom-xpathevaluatorbase-evaluate
     text: ORDERED_NODE_SNAPSHOT_TYPE; url: #dom-xpathresult-ordered_node_snapshot_type
     text: snapshotItem; url: #dom-xpathresult-snapshotitem
+spec: FULLSCREEN; urlPrefix: https://fullscreen.spec.whatwg.org/
+  type: dfn
+    text: fullscreen an element; url: #fullscreen-an-element
+    text: fullscreen is supported; url: #fullscreen-is-supported
+    text: fully exit fullscreen; url: #fully-exit-fullscreen
 spec: SELECTORS4; urlPrefix: https://drafts.csswg.org/selectors-4/
   type: dfn
     text: match a selector against a tree; url: #match-a-selector-against-a-tree
@@ -2013,8 +2020,10 @@ managing the remote end browser process.
 BrowserCommand = (
   browser.Close //
   browser.CreateUserContext //
+  browser.GetClientWindows //
   browser.GetUserContexts //
-  browser.RemoveUserContext
+  browser.RemoveUserContext //
+  browser.SetClientWindowState //
 )
 </pre>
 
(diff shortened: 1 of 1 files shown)


OrKoN:
@sadym-chromium could you please review?

whimboo (changes requested):
Thank you @jgraham! Overall this looks great. I've just some comments with questions and corrections.

index.bs
```diff
@@ -2028,9 +2037,225 @@ BrowserResult = (
...
+"<code>hidden</code>" and |window|'s [=client window/width=] and [=client
+window/height=] are both as close as possible to 0. If either of these options
+are supported then <dfn>minimize client window is supported</dfn>.
+
+To <dfn>restore the client window</dfn> |window| an implementation should ensure that it's
+neither in a platform-defined maximized state, nor in a platform-defined
+minimized state, and that if there is one or more [=/top-level traversable=]
+associated with |window|, at least one of those has an [=active document=] in
+the "<code>visible</code>" state. If this is supported then <dfn>restore client
+window is supported</dfn>.
```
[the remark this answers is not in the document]
jgraham: Fullscreen is different, and even browsers that don't support the "restore" operation are generally expected to support fullscreen (c.f. mobile).

index.bs
```diff
@@ -2028,9 +2037,225 @@ BrowserResult = (
...
+  1. Let |document| be |traversable|'s [=active document=].
+
+  1. Append |document| to |documents|.
+
+1. If |documents| is [=list/empty=] return [=error=] with [=error code=] [=no
+   such window=].
+
+1. If |current state| is "<code>fullscreen</code>":
+
+  1. For each |document| in |documents|:
```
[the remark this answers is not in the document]
jgraham: Well we don't know which documents have fullscreen elements. AFAICT there's nothing spec-wise that guarantees that different documents in the same client window have the same fullscreen state.

index.bs
```diff
@@ -2028,9 +2037,225 @@ BrowserResult = (
...
+#### The browser.ClientWindowInfo Type #### {#type-browser-ClientWindowInfo}
+
+<pre class="cddl remote-cddl local-cddl">
+browser.ClientWindowInfo = {
+  clientWindow: browser.ClientWindow,
+  state: "fullscreen" / "maximized" / "minimized" / "normal",
+  width: js-uint,
+  height: js-uint,
+  x: js-int,
+  y: js-int,
```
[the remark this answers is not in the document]
jgraham: We can do this for consistency, but my sense is that it makes things much harder to read in this case.

index.bs
```diff
@@ -89,6 +89,7 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
     text: no such alert; url: dfn-no-such-alert
     text: no such element; url: dfn-no-such-element
     text: no such frame; url: dfn-no-such-frame
+    text: no such window; url: dfn-no-such-window
```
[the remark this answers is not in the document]
jgraham: In we do kind of reuse `no such window` in classic (if the current browsing context's window is closed that the error you'll get), but I guess it makes more sense to have a different error for a different kind of handle. Added `no such client window`.

whimboo:
@jgraham while you are on the implementation it would be great to also have an `active` or `focus` flag on the window state type which will indicate the currently active window. There is also https://github.com/w3c/webdriver-bidi/issues/585 for the browsing context.

index.bs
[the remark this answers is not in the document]
whimboo: Hm, but below we break right after setting the first document into fullscreen mode. Why don't we have to do that for all the other documents in the window as well?

index.bs
```diff
@@ -2028,9 +2039,235 @@ BrowserResult = (
...
+</pre>
+
+The <code>browser.ClientWindow</code> uniquely identifies a [=client window=].
+
+#### The browser.ClientWindowInfo Type #### {#type-browser-ClientWindowInfo}
+
+<pre class="cddl remote-cddl local-cddl">
+browser.ClientWindowInfo = {
+  clientWindow: browser.ClientWindow,
+  hasFocus: bool,
```
whimboo, line 2230: We already have the command `browsingContext.activate` which activates (sets the focus to) a tab. Maybe we should stick with `active` here to reduce confusion?

index.bs
```diff
@@ -2028,9 +2039,235 @@ BrowserResult = (
...
+
+1. Let |state| be [=get the client window state=] with |client window|.
+
+1. Let |has system focus| be false.
+
+1. For each [=/top-level traversable=] |traversable|:
+
+  1. If |traversable|'s [=client window=] is not |client window| then continue.
+
+  1. If |traversable| has [=system focus=], set |has system focus| to true and break.
```
whimboo, line 2255: Should we actually get this information from [`taversable's document's hasFocus()`](https://html.spec.whatwg.org/#dom-document-hasfocus)?

index.bs
```diff
@@ -2263,6 +2556,92 @@ The [=remote end steps=] with |command parameters| are:
...
+   <dd>
+      <pre class="cddl remote-cddl">
+      browser.SetClientWindowState = (
+        method: "browser.setClientWindowState",
+        params: browser.SetClientWindowStateParameters
+      )
+
+      browser.SetClientWindowStateParameters = {
+        clientWindow: browser.ClientWindow,
+        browser.ClientWindowNamedState // browser.ClientWindowRectState
```
whimboo, line 2575: Should we also add the option to focus a given window by having a separate property here? Otherwise we could have it similar for navigables as a separate command.

whimboo:
@jimevans, @titusfortner and @pujagani would you mind having a look at this PR given that it's a requested feature from the Selenium side? Thanks.

index.bs
```diff
@@ -2393,6 +2772,7 @@ browsingContext.InfoList = [*browsingContext.Info]
 
 browsingContext.Info = {
   children: browsingContext.InfoList / null,
+  clientWindow: browser.ClientWindow,
```
jimevans, line 2775: This is very, very good, and necessary.

jimevans:
From an API/protocol design perspective, I'm pretty happy with what I see here. I would echo @whimboo's suggestion to either have an option in `setClientWindowState` to (try to, the platform may not let the window have the focus under certain circumstances) give the window the OS focus, or an additional method to do so.

jimevans (approved):

index.bs
```diff
@@ -2028,9 +2039,235 @@ BrowserResult = (
...
+
+1. Return "<code>normal</code>".
+
+</div>
+
+<div algorithm>
+
+To <dfn>set the client window state</dfn> given |window| and |state|:
+
+1. Let |current state| be [=get the client window state| with |window|.
```
sadym-chromium, line 2144:
```suggestion
1. Let |current state| be [=get the client window state=] with |window|.
```

index.bs
sadym-chromium: +

index.bs
```diff
@@ -2028,9 +2039,235 @@ BrowserResult = (
...
+
+   <dl>
+    <dt>"<code>fullscreen</code>"
+    <dd>
+      1. For each |document| in |documents|:
+
+        1. If |document|'s [=visibility state=] is "<code>visible</code>",
+           [=fullscreen an element=] with |document|'s [=document element=].
+
+        1. [=Break=].
```
sadym-chromium, line 2200: Iterate through all the documents?
```suggestion
```

jgraham: That doesn't seem obvious? If you have more than one visible document in a specific window then you still can't really have more than one fullscreen element. At the moment we'd pick the first one (albeit in an arbitary order), with your proposed change we'd fullscreen an element from each document in turn. I don't quite see if anything in the fullscreen spec says that you have to unfullscreen any existing documents that are already fullscreen (maybe it's assumed that user activation requirements mean that you can't get into that situation?), but it at least feels confusing.

index.bs
jgraham: Technically yes, for example during navigation when the document itself isn't fully active.

You can also have the case where the window has focus, but none of the documents do (e.g. when the focus is on the address bar).

Per HTML something can be "have system focus" but not be visible, whereas the `activate` command tries to ensure that the browsing context is actually visible. 

https://html.spec.whatwg.org/#data-model explains the terminology here.

I guess it's OK to use `active` here, but it does mean ignoring some of the edge cases.

index.bs
jgraham: Actually I think that technically neither is correct and we should define our own thing (see the example about the address bar having focus; in that case the client window has focus, but no traversable does).

index.bs
jgraham: I'm inclined to leave this to a followup. You can always focus a specific traversable using `BrowsingContext.activate` and that has the side effect of changing the system focus. A command just to focus the client window without specifying a specific browsing context doesn't have such obvious use cases.

sadym-chromium (approved):

index.bs
whimboo: As long as we know which window contains which elements I would agree that we should use the `browsingContext.activate` command instead to focus/activate a given window. Lets drop it here.

whimboo (approved):

PR #758, opened 2024-07-30
Replace HTML's "report the exception" with "report an exception".

jeremyroman:
The former is deprecated. Currently, the environment settings object's realm's global is used (which is consistent with what the spec would previously have said before). It's possible this ought to use the sandbox's realm, but that realm is not currently used for the function evaluation itself right now -- and that's the function whence the error comes.

This is part of whatwg/html#10516.

***
<a href="https://pr-preview.s3.amazonaws.com/jeremyroman/webdriver-bidi/pull/758.html" title="Last updated on Jul 30, 2024, 9:09 PM UTC (c1a8934)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/758/da20dfc...jeremyroman:c1a8934.html" title="Last updated on Jul 30, 2024, 9:09 PM UTC (c1a8934)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7123,6 +7123,9 @@ To <dfn export>run WebDriver BiDi preload scripts</dfn> given |environment setti
        |navigable|. Otherwise let |realm| be |environment settings|'
        [=realm execution context=]'s Realm component.
 
+    1. Let |exception reporting global| be be |environment settings|'
+       [=realm execution context=]'s Realm component's [=realm/global object=].
+
     1. Let |arguments| be |preload script|'s <code>arguments</code>.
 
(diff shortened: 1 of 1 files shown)


jeremyroman:
@sadym-chromium are you an appropriate reviewer for this? The sandbox stuff seems extremely underspecified here so there's not really any way for this to mesh with it perfectly (the two realms are kinda both used at the moment), but this doesn't make it worse and reflects an upstream change in whatwg/html.

jgraham (approved):
Good point about the bug in where sandboxed preload scripts run; I'll file an issue for that.

PR #768, opened 2024-09-06
Try to ensure sandboxed preload scripts run in the correct realm

jgraham:
Previously we were passing the environment settings object for the Window realm into `evaluate function body`, which is incorrect.

Instead update `environment settings` to point to the environment settings object for the sandboxed realm.

Also have a go at defining the steps to create a sandboxed realm and ensure that it has an environment settings object that we can use.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1257,8 +1257,55 @@ To <dfn>get or create a sandbox realm</dfn> given |name| and |navigable|:
 <div algorithm>
 To <dfn>create a sandbox realm</dfn> with |window|:
 
-Issue: Define creation of sandbox realm. This is going to return a
-{{SandboxWindowProxy}} wrapping |window|.
+1. Let |window environment settings| be the [=environment settings object=] whose
+   [=relevant global object=] is |window|.
+
+1. Let |global| be a new {{SandboxWindowProxy}} wrapping |window|.
+
+   Issue: This is the key step but it's basically totally undefined.
+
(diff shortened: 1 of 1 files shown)


juliandescottes (approved):

index.bs
```diff
@@ -1257,8 +1257,55 @@ To <dfn>get or create a sandbox realm</dfn> given |name| and |navigable|:
 <div algorithm>
 To <dfn>create a sandbox realm</dfn> with |window|:
 
-Issue: Define creation of sandbox realm. This is going to return a
-{{SandboxWindowProxy}} wrapping |window|.
+1. Let |window environment settings| be the [=environment settings object=] whose
```
OrKoN, line 1260: By looking at call sites of "create a sandbox realm" the argument should be navigable, so probably we need the env settings object of |navigable|'s [=active window=].

index.bs
```diff
@@ -1257,8 +1257,55 @@ To <dfn>get or create a sandbox realm</dfn> given |name| and |navigable|:
...
+
+1. Let |settings object| be a new [=environment settings object=] whose algorithms are defined as follows:
+
+  <dl>
+    <dt>The [=realm execution context=].
+    <dd>Return |execution context|.
+
+    <dt>The [=environment settings object/module map=]
+    <dd>Return |global|'s module map.
+        Issue: Need to define module map for {{SandboxWindowProxy}} objects.
```
OrKoN, line 1282: Looks like this issue is not rendered properly? Perhaps it needs newlines or needs to be outside of dl.

Issue #548, opened 2023-09-15
Web Extensions: install / uninstall


sadym-chromium:

whimboo:
This topic has been discussed at TPAC 2023. The minutes can be found at:
https://www.w3.org/2023/09/15-webdriver-minutes.html#t04

whimboo:
Maybe the definition of extensions in the capabilities could look like the following:

```
{
  "extensions": [
    "<base64_encoded_extension>",
    {"extension": "<base64_encoded_extension>", "installTemporary": true, "allowPrivateBrowsing": true},
  ],
}
```

That means by default extensions could be listed as usual as a list of base64 encoded strings, or if more options are needed for the install process an object could be passed with additional flags - all of them optional and default to false?

OrKoN:
CDP now supports loading unpacked extensions at runtime https://chromedevtools.github.io/devtools-protocol/tot/Extensions/#method-loadUnpacked (with some restrictions such as a special flag + pipe connection)

whimboo:
Fixed by https://github.com/w3c/webdriver-bidi/pull/778.

PR #778, opened 2024-09-18
webExtension module

chrmod:
addresses https://github.com/w3c/webdriver-bidi/issues/548

We aim to provide a capability and a command to install a web-extension in the session. 

This is a pre-requisite to allow WedDriver Bidi to interact with the extension background pages and service workers.

***
<a href="https://pr-preview.s3.amazonaws.com/ghostery/webdriver-bidi/pull/778.html" title="Last updated on Nov 28, 2024, 4:39 PM UTC (0c770e2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/778/46306e8...ghostery:0c770e2.html" title="Last updated on Nov 28, 2024, 4:39 PM UTC (0c770e2)">Diff</a>


index.bs
```diff
@@ -1718,6 +1719,31 @@ session.SubscriptionRequest = {
...
+Extension = text
+</pre>
+
+The <code>Extension</code> type is base64 encoded web-extension archive (xpi/crx/zip),
+an absolute path to the web-extension archive or absolute path to the unpacked web-extension.
+
+<pre class="cddl remote-cddl local-cddl">
+ExtensionDetails = {
+   extension: Extension,
+   ? installTemporary: bool .default true
```
jgraham, line 1734: I think we might be better off only supporting temporary installs at first. I don't really have an automation use case in mind where it makes sense to permanently install an extension in a specific profile, and we can always add it if there is one.

index.bs
```diff
@@ -2001,6 +2027,68 @@ The [=remote end steps=] with |session| and |command parameters| are:
...
+
+     session.InstallWebExtensionParameters = {
+       extension: session.Extension
+     }
+     </pre>
+   </dd>
+   <dt>Result Type</dt>
+   <dd>
+      <pre class="cddl">
+      session.InstallWebExtensionResult = (
```
jgraham, line 2052: You don't need to manually define success and failure types; you can just define an error code and either `return [=success=] with data [map matching success type]` or `return error with error code [new code]` and that will handle this.

index.bs
```diff
@@ -2001,6 +2027,68 @@ The [=remote end steps=] with |session| and |command parameters| are:
...
+   </dd>
+</dl>
+
+<div algorithm="remote end steps for session.installWebExtension">
+The [=remote end steps=] with |session| and |command parameters| are:
+
+1. Let the |extension| be the value of the <code>extension</code> field of
+   |command parameters|.
+
+1. [=Try=] to install web-extension with |session| and |extension|.
```
jgraham, line 2076: So there's quite a few bits missing here:
* It's possible this command isn't supported, so you need to say something like "If installing |extension| isn't supported for any reason return error with error code unsupported operation".
* You need to explictly deal with the different input types, so that end up with some abstract representation of the set of files that make up the extension.
* It would probably be good to make some checks explicit here. For example if there's no `manifest.json` file in the root that should return an error.
* You need to handle the other data passed in e.g. the private browsing mode flag.
* You need a bit more detail on what it means to install a web extension. It's going to be "implementation defined steps", but it should probably describe what the outcome is supposed to be. This will still need to be entirely fallible (i.e. there's a possibility of the install failing for implemenation-specific reasons), as in the current text.

index.bs
```diff
@@ -1588,6 +1588,7 @@ session.CapabilityRequest = {
   ? platformName: text,
   ? proxy: session.ProxyConfiguration,
   ? unhandledPromptBehavior: session.UserPromptHandler,
+  ? extensions: [*session.Extension],
```
jgraham, line 1591: I'd recommend making this a separate PR, because you need different supporting infrastructure if we want to allow a new session request to install extensions.

jgraham (changes requested):
Thanks for the PR! Quite a few comments, but this does look like it's heading in the right direction, so don't be put off :)

Rob--W:
FYI: There is prior context in the WECG to discuss the desired characteristics of extension loading functionality. For reference, here are the meeting notes:

- https://github.com/w3c/webextensions/blob/main/_minutes/2023-09-12-wecg-tpac.md#:~:text=Issue%20441
- https://github.com/w3c/webextensions/blob/main/_minutes/2024-08-01-wecg.md#:~:text=run%20tests

Related issues:
- https://github.com/w3c/webextensions/issues/19
- https://github.com/w3c/webextensions/issues/441

chrmod:
> Thanks for the PR! Quite a few comments, but this does look like it's heading in the right direction, so don't be put off :)

@jgraham thank you very much for the review. I will need a moment to unpack all comments, but sure to receive progress soon.

whimboo:
> @jgraham thank you very much for the review. I will need a moment to unpack all comments, but sure to receive progress soon.

Hi Krzysztof. I wanted to check back if you already had some time to distill the feedback from James. If not no worries, but let us know in case you don't have the time to continue on this PR. Thanks!

chrmod:
@jgraham @whimboo it took me a while to learn more on how to write such documents, sorry for keeping you waiting. I'm pretty sure that the current result is still not merge ready. Please guide me on how it should be improved.

index.bs
```diff
@@ -11302,6 +11306,147 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+extensions.ExtensionBase64Encoded = {
+   type: "base64",
+   value: text,
+}
+</pre>
+
+The <code>extensions.Extension</code> type represents an extension with its additional properties.
+
+<div algorithm>
+To <dfn>expand a install path spec</dfn> given |install path spec|:
```
chrmod, line 11367: not sure if this should be a separate algorithm or something that should be inlined with the install command

index.bs
```diff
@@ -11302,6 +11306,147 @@ The [=remote end steps=] given |session| and |command parameters| are:
 </div>
 
 
+## The extensions Module ## {#module-extensions}
```
chrmod, line 11309: Wonder if the module should not be called `WebExtensions` to avoid potential confusion with the `Extensions` section of the `Webdriver` spec.

index.bs
```diff
@@ -11302,6 +11306,147 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+<div algorithm>
+To <dfn>expand a install path spec</dfn> given |install path spec|:
+
+1. Let |path| be a value of <code>path</code> field of |install path spec| if the value of <code>type</code> of the |install path spec| is equal to <code>path</code>,
+   otherwise perform any implementation-defined steps to create a temporary folder and make its path to be the value of |path|.
+
+1. Let |archive path| be a value of <code>path</code> field of |install path spec| if the value of <code>type</code> of the |install path spec| is equal to <code>archive-path</code>,
+   otherwise if the value of <code>type</code> of the |install path spec| is equal to <code>archive-path</code> perform any implementation-defined steps to create a temporary file
+   being a result of Base64 decodeing of the value of <code>value</code> filed of |install path spec| and make its path to be the value of |archive path|,
+   otherwise |archive path| is null.
```
chrmod, line 11375: I do realize this is overly complicated, but wasn't sure how to express it otherwise. Please advise.

index.bs
```diff
@@ -11302,6 +11306,147 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+
+1. If installing |extension| isn't supported return [=error=] with error code [=unsupported operation=].
+
+1. Let |install path spec| be a value of <code>installPath</code> field of the |extension|.
+
+1. Let |extension archive| be the result of [=trying=] to [=expand a install path spec=] with |install path spec|.
+
+1. Let |extension files| be a value of <code>files<code> field of |extension archive|.
+
+1. If |extension files| does not include entry <code>manifest.json</code> return [=error=] with code [=invalid extension=]
```
oliverdunk, line 11432: I would lean towards dropping this, and in the subsequent step having something that allows the browser to return an error if anything goes wrong. There are many ways an extension can be invalid beyond the manifest.json not being present (e.g an unsupported `manifest_version`) and I don't think we want to list them all here.

index.bs
```diff
@@ -11302,6 +11306,147 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+
+1. Perform any implementation-defined steps to install the extension into |session|.
+
+1. Let |extension id| be a unique identifier of the |extension| for the browser decided by the implementation-defined steps.
+
+1. Let |allow private browsing| be a value of <code>allowPrivateBrowsing</code> field of |command parameters| if present, or false otherwise.
+
+1. Perform any implementation-defined steps to enable extension to be run in private browsing based on the valu of |allow private browsing|.
```
oliverdunk, line 11440:
```suggestion
1. Perform any implementation-defined steps to enable extension to be run in private browsing based on the value of |allow private browsing|.
```

index.bs
```diff
@@ -11302,6 +11306,147 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+
+1. Let |result| be a [=/map=] matching the
```
oliverdunk, line 11442: Is there a reason this is a map and we don't just return a single `InstallResult`? I can't see a case where we would ever install more than one extension.

index.bs
oliverdunk: I'm confused by the flow of this, since each step has something along the lines of "If A do this, otherwise if A do that". I would have expected something like:

1. If the type is "path", return the path.
2. Otherwise, if the type is "archive-path", extract the archive and return the path to the temporary directory.
3. Otherwise, if the type is "base64", decode this as a zip and return the path to the temporary directory.

oliverdunk (changes requested):
Thanks for this! It's really great to see an external contribution and it lines up nicely with our goal in the WebExtensions Community Group to write cross-browser tests for extension APIs.

I've left some feedback but please prioritize addressing anything from @jgraham who is the one with spec experience :)

chrmod:
@oliverdunk thank you for all the suggestions! Happy to contribute to WECG goals.

index.bs
```diff
@@ -11302,6 +11306,153 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+<pre class="cddl local-cddl">
+extensions.ExtensionArchive = {
+   files: [+text],
+}
+</pre>
+
+<pre class="cddl remote-cddl local-cddl">
+extensions.Extension = {
+   installPath: extensions.ExtensionPath / extensions.ExtensionArchivePath / extensions.ExtensionBase64Encoded,
+   ? allowPrivateBrowsing: bool .default false,
```
OrKoN, line 11345: @oliverdunk do you know if we have an equivalent of what allowPrivateBrowsing is expressing here?

index.bs
```diff
@@ -11302,6 +11306,153 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+
+<pre class="cddl local-cddl">
+extensions.ExtensionArchive = {
+   files: [+text],
+}
+</pre>
+
+<pre class="cddl remote-cddl local-cddl">
+extensions.Extension = {
+   installPath: extensions.ExtensionPath / extensions.ExtensionArchivePath / extensions.ExtensionBase64Encoded,
```
OrKoN, line 11344: In Chrome, we currently only allow installing extensions if the client is local to the browser (i.e., connected via pipes instead of the WebSocket). We could probably relax this based on a command line argument but I wonder if it makes sense to start with the extensions.ExtensionPath only?

cc @oliverdunk for additional thoughts

index.bs
OrKoN: @jgraham would a temporary install be something new defined by the webdriver bidi spec or do you already have a notion of "temporary" extensions in Firefox (I guess they auto-uninstall)?

@oliverdunk do we have something like this within extensions?

index.bs
OrKoN: I believe it is a map because InstallResult is meant to be a map (object). I think it should be defined like 

```
extensions.InstallResult = {
        extensionId: text
}
```

(`{}` instead of `()`)

OrKoN:
I noticed the IPR bot message: 

> PR has contribution issues. The following users were not in the repository's groups: @chrmod.

I believe you plan to join the group @chrmod?

Or is there a different workaround?

index.bs
chrmod: In Chromium this is called "Allow in Incognito"
Screenshot 2024-10-09 at 21 17 41

chrmod:
> I noticed the IPR bot message:
> 
> > PR has contribution issues. The following users were not in the repository's groups: @chrmod.
> 
> I believe you plan to join the group @chrmod?
> 
> Or is there a different workaround?

My plan was to join the group and I've made a request to W3C. In result I was added to WECG group, but not to webdriver. Not sure whom to ask fix it. Can you recommend a contact?

chrmod:
re https://github.com/w3c/webdriver-bidi/pull/778#discussion_r1793931128
@OrKoN in Firefox, Temporary Extension are similar to Chrome's unpacked ones - they can be loaded from a path or base64. The point of temporary is that they wont be stored in the profile and will be removed when the browsing session ends.

index.bs
OrKoN: I see, thanks! I wonder if alternatively we could specify a user context ID to install an extension to? (I would need to check if it's implementable though)

index.bs
chrmod: Could that be managed by the webdriver? Archives (zip/crx) and Base64 strings are can effectively be unpacked to a temporary folder, and then run as from any other path.

OrKoN:
> re [#778 (comment)](https://github.com/w3c/webdriver-bidi/pull/778#discussion_r1793931128) @OrKoN in Firefox, Temporary Extension are similar to Chrome's unpacked ones - they can be loaded from a path or base64. The point of temporary is that they wont be stored in the profile and will be removed when the browsing session ends.

thanks for details so it's different from unpacked extensions in Chrome because they persist in the profile.

index.bs
chrmod: Is it better now? or you would rather remove the intermediate data structure?

OrKoN:
> > I noticed the IPR bot message:
> > > PR has contribution issues. The following users were not in the repository's groups: @chrmod.
> > 
> > 
> > I believe you plan to join the group @chrmod?
> > Or is there a different workaround?
> 
> My plan was to join the group and I've made a request to W3C. In result I was added to WECG group, but not to webdriver. Not sure whom to ask fix it. Can you recommend a contact?

I think this link should lead to a form to join the browser testing and tools working group https://www.w3.org/groups/wg/browser-tools-testing/join

index.bs
chrmod: @OrKoN Definition fixed. Thank you for pointing it out. 

Where can I find a specification for this syntax? Currently I'm learning from examples in this doc.

index.bs
OrKoN: the driver process and the browser might not be running on the same machine so, in a general case, we probably should not handle it in the driver. I think a remote-end path to an archive or to an unpacked folder is less of a concern but installing an extension from base64 would mean writing something to the filesystem that was not present there originally.

index.bs
OrKoN: I still seem to see `()` instead of `{}`. The CDDL spec is here https://datatracker.ietf.org/doc/rfc8610/

index.bs
oliverdunk: I think having this set to `false` would be the current behavior - unpacked extensions stay installed between restarts as long as you use the same profile. I can imagine how `true` might work but I'm not sure I'm convinced of the value of that, at least for the MVP.

index.bs
oliverdunk: The extraction process is much clearer now, thanks - I'm happy with that as long as we are comfortable having a switch vs a series of if/else if (cc/ @jgraham). I'm still a little unclear why we need |extension files| and |extension archive|. Is there a reason for that?

index.bs
oliverdunk: @OrKoN, I'm fairly certain Selenium currently supports loading an extension from base64 in WebDriver. Do you have any idea how that was implemented?

index.bs
jgraham: A separate algorithm is fine.

For the sake of concreteness, I suggest that we try to make this return a https://fs.spec.whatwg.org/#directory which is interpreted as the root of the extension.

index.bs
```diff
@@ -11302,6 +11306,153 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+     extensions.InstallParameters = {
+       extension: extensions.Extension,
+     }
+     </pre>
+   </dd>
+   <dt>Result Type</dt>
+   <dd>
+      <pre class="cddl local-cddl">
+      extensions.InstallResult = (
+        extensionId: text
```
jgraham, line 11419: We don't use the `fooId` pattern in general; wherever possible we just use `foo`, so this should just be `extension`.

jgraham (changes requested):

index.bs
chrmod: Extensions can be quite big. I think AdblockPlus is around 60mb while Ghostery is around 12mb.

index.bs
```diff
@@ -11302,6 +11306,153 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+
+1. Let the |extension| be the value of the <code>extension</code> field of |command parameters|
+
+1. If installing |extension| isn't supported return [=error=] with error code [=unsupported operation=].
+
+1. Let |install path spec| be a value of <code>installPath</code> field of the |extension|.
+
+1. Let |extension archive| be the result of [=trying=] to [=expand a install path spec=] with |install path spec|.
+
+1. Let |extension files| be a value of <code>files<code> field of |extension archive|.
```
OrKoN, line 11436:
```suggestion
1. Let |extension files| be a value of <code>files</code> field of |extension archive|.
```

index.bs
chrmod: I've renamed it to the `extensions` but must say that `extension id` is something that will be well understood by the extension developers. Each browser extension has own ID:
 
Screenshot 2024-10-11 at 16 39 46

and if Firefox the ID can be specified in the manifest.json

```json
"browser_specific_settings": {
  "gecko": {
    "id": "firefox@ghostery.com",
    "strict_min_version": "112.0"
  }
}
```

How about we rename it from `extension` to `id`?

jgraham: `id` would be much worse :)

The point is that within bidi we don't generally use the `Id` suffix when passing back a handle to something. We just name the thing e.g. `context` rather than `contextId`, `request` rather than `requestId`, etc. There are three exceptions I can see: script objects get both `sharedId` and `internalId`, and the session itself gets `sessionId` for compatibility with WebDriver classic. So by the same token a handle identifying a specific web extension should just be called `extension` in the protocol.

That said I don't really feel strongly, so if others think this is a case where it's worth breaking the convention, I won't object.

index.bs
chrmod: I've change the wrong place. Fixed now. Thank you for the spec!

index.bs
```diff
@@ -11302,6 +11306,153 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+
+1. If |extension files| does not include entry <code>manifest.json</code> return [=error=] with code [=invalid extension=]
+
+1. Perform any implementation-defined steps to install the extension into |session|.
+
+1. Let |extension id| be a unique identifier of the |extension| for the browser decided by the implementation-defined steps.
```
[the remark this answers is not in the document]
chrmod: done

index.bs
jimevans: I, for one, would prefer to continue with convention, if only for the sake of consistency.

index.bs
```diff
@@ -11302,6 +11306,153 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+1. Switch on the value of the <code>type</code> field of |install path spec|:
+
+   <dl>
+    <dt>"<code>path</code>"
+    <dd>Let |path| be a value of <code>path</code> field of |install path spec|.
+
+    <dt>"<code>archive-path</code>"
+    <dd>
+      1. Let |archive path| be a value of <code>path</code> field of |install path spec|.
+      1. Perform any implementation-defined steps to extract the archive from <code>archive path</code> to a temporary directory and let |path| be the path to that temporary directory.
```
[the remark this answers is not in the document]
chrmod: Not quite sure how to achieve that. Can you please check the latest version to verify if I go into right direction?

index.bs
whimboo: Actually my last answer was wrong. Temporary actually means that the web extension is not persistent and can be installed even when it isn't signed (like for developers), and that is indeed a feature that we support in Firefox. 

But we do not yet support the `private browsing` mode installation. So both things are different. 

@OrKoN How does Chrome handle the installation of unsigned web extensions?

index.bs
```diff
@@ -11302,6 +11306,136 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+#### The extensions.Extension Type #### {#type-webExtensions-Extension}
+
+<div algorithm>
+To <dfn>expand a extension data spec</dfn> given |extension data spec|:
+
+1. Let |installation type| be a value of <code>type</code> field of |extension data spec|.
+
+1. If installing |installation type| isn't supported return [=error=] with [=error code=] [=unsupported operation=].
+
+1. Switch on the value of the <code>installation type</code>:
```
whimboo, line 11343:
```suggestion
1. In the following list of conditions and associated steps, run the first set of steps for which the associated condition is true:

<code>type</code> is the string "css":
```

index.bs
```diff
@@ -11302,6 +11306,136 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+
+   <dl>
+    <dt>"<code>path</code>"
+    <dd>
+      1. Let |path| be a value of <code>path</code> field of |extension data spec|.
```
whimboo, line 11348:
```suggestion
      1. Let |path| be |extension data spec|["<code>path</code>"].
```

This applies to other steps as well.

index.bs
```diff
@@ -11302,6 +11306,136 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+      1. Let |entry| be a [=directory entry=] represeting the directory with that |path|.
+
+    <dt>"<code>archive-path</code>"
+    <dd>
+      1. Let |archive path| be a value of <code>path</code> field of |extension data spec|.
+      1. Perform any implementation-defined steps to extract ZIP from |archive path| and let |entry| by a [=directory entry=] representing the extraction result, if this fails return [=error=] with [=error code=] [=invalid web-extension=].
```
whimboo, line 11354: May this change be appropriate?

```suggestion
      1. Perform any implementation-defined steps to extract the archive from |archive path| by using the ZIP algorithm, and let |entry| by a [=directory entry=] representing the extraction path, if this fails return [=error=] with [=error code=] [=invalid web-extension=].
```

Do we also have to add that the remote end is responsible for removing the path to not leave the extracted data behind?

index.bs
```diff
@@ -11302,6 +11306,136 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+
+    <dt>"<code>base64</code>"
+    <dd>
+      1. Let |archive path| be a value of <code>path</code> field of |extension data spec|.
```
whimboo, line 11358: Given that the extraction is done in memory we do not need this step:

```suggestion
```

index.bs
```diff
@@ -11302,6 +11306,136 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+<div algorithm="remote end steps for webExtensions.install">
+The [=remote end steps=] with |command parameters| are:
+
+1. If installing extensions isn't supported return [=error=] with error code [=unsupported operation=].
+
+1. Let |extension data spec| be a value of <code>extensionData</code> field of the |command parameters|.
+
+1. Let |extension directory entry| be the result of [=trying=] to [=expand a extension data spec=] with |extension data spec|.
+
+1. If |extension directory entry| does not include file <code>manifest.json</code> return [=error=] with [=error code=] [=invalid web-extension=].
```
whimboo, line 11423:
```suggestion
1. If |extension directory entry| does not contain a file named <code>manifest.json</code> return [=error=] with [=error code=] [=invalid web-extension=].
```

index.bs
```diff
@@ -11302,6 +11306,136 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+</dl>
+
+<div algorithm="remote end steps for webExtensions.install">
+The [=remote end steps=] with |command parameters| are:
+
+1. If installing extensions isn't supported return [=error=] with error code [=unsupported operation=].
+
+1. Let |extension data spec| be a value of <code>extensionData</code> field of the |command parameters|.
+
+1. Let |extension directory entry| be the result of [=trying=] to [=expand a extension data spec=] with |extension data spec|.
```
whimboo, line 11421: Should the `directory entry` part reference the other spec?

index.bs
```diff
@@ -11302,6 +11306,136 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+
+1. If |extension directory entry| does not include file <code>manifest.json</code> return [=error=] with [=error code=] [=invalid web-extension=].
+
+1. Perform implementation defined steps to install |extension directory entry|. If this fails, return [=error=] with [=error code=] [=invalid web-extension=]. Otherwise let |extension id| be the unique identifier of the newly installed extension.
```
whimboo, line 11425:
```suggestion
1. Perform implementation defined steps to install a web extension from the path specified by |extension directory entry|. If this fails, return [=error=] with [=error code=] [=invalid web-extension=]. Otherwise let |extension id| be the unique identifier of the newly installed web extension.
```

index.bs
```diff
@@ -11302,6 +11306,136 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+
+1. Let |allow private browsing| be a value of <code>allowPrivateBrowsing</code> field of |command parameters| if present, or false otherwise.
+
+1. Perform any implementation-defined steps to enable extension to be run in private browsing based on the value of |allow private browsing|.
```
whimboo, line 11429:
```suggestion
1. Perform any implementation-defined steps to enable the web extension to be run in private browsing mode based on the value of |allow private browsing|.
```

Also we should define what will happen if that feature is not supported, which could mean that the browser doesn't support private browsing mode at all, or cannot yet allow web extensions to be enabled in that mode. We probably want to fail as well?

index.bs
```diff
@@ -11302,6 +11306,136 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+      <pre class="cddl remote-cddl">
+      webExtensions.Install = (
+         method: "webExtensions.install",
+         params: webExtensions.InstallParameters
+      )
+
+      webExtensions.InstallParameters = {
+         extensionData: webExtensions.ExtensionPath / webExtensions.ExtensionArchivePath / webExtensions.ExtensionBase64Encoded,
+         ? allowPrivateBrowsing: bool .default false,
+      }
```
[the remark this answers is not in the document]
oliverdunk: I left some thoughts on this above. I don't think we should add a `temporary` property that controls the loading of unpacked extensions since that couples two unrelated concepts.

index.bs
```diff
@@ -11302,6 +11306,136 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+
+1. Return [=success=] with data |entry|.
+
+</div>
+
+### Commands ### {#module-extensions-commands}
+
+#### The webExtensions.install Command #### {#command-extensions-install}
+
+The <dfn export for=commands>webExtensions.install</dfn> command installs an extension for the session.
```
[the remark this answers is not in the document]
chrmod: I'm not quite sure how this works. Aren't sessions separate profiles running in the same browser?

index.bs
chrmod: isn't it enough if `|expand a extension data spec|` returns a `[=directory entry=]`?

chrmod:
@whimboo thank you for another great review. I think I've addressed all of the comments, but honestly, due to the number of suggestions, I'm not quite sure if handled them all. 

Is it fine if we focus on the suggestion still present `Files changed` tab? 

That may mean, you may have to re-post some of the comments that got hidden due to code changes. Hope that it wont be much of an inconvenience.

index.bs
```diff
@@ -11302,6 +11306,181 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+   <dd>
+      <pre class="cddl remote-cddl">
+      webExtensions.Install = (
+         method: "webExtensions.install",
+         params: webExtensions.InstallParameters
+      )
+
+      webExtensions.InstallParameters = {
+         extensionData: webExtensions.ExtensionData,
+         ? allowPrivateBrowsing: bool .default false,
```
OrKoN, line 11360: as was suggested in other comments, let's drop allowPrivateBrowsing and related steps as this concept is not well defined in the specification and might not be supported by all implementations.

index.bs
```diff
@@ -11302,6 +11306,181 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+      </pre>
+   </dd>
+</dl>
+
+<div algorithm="remote end steps for webExtensions.uninstall">
+The [=remote end steps=] with |command parameters| are:
+
+1. Let |extension| be a value of <code>extension</code> field of the |command parameters|.
+
+1. If the browser has no web extensionwith an id equal to |extension|, return [=error=] with [=error code=] [=invalid argument=].
```
OrKoN, line 11476:
```suggestion
1. If the browser has no web extension with an id equal to |extension|, return [=error=] with [=error code=] [=invalid argument=].
```

OrKoN (changes requested):

OrKoN:
> I think I've addressed all of the comments, but honestly, due to the number of suggestions, I'm not quite sure if handled them all.

You can use the "Resolve conversation" button to track which items have been resolved.

OrKoN (approved):
LGTM

index.bs
```diff
@@ -11302,6 +11306,174 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+   <code>extension</code> field set to |extension id|.
+
+1. Return [=success=] with data |result|.
+
+</div>
+
+<div algorithm>
+To <dfn>expand a web extension data spec</dfn> given |extension data spec|:
+
+1. Let |type| be |extension data spec["type"]|.
```
OrKoN, line 11416:
```suggestion
1. Let |type| be |extension data spec|["type"].
```

index.bs
```diff
@@ -603,6 +603,9 @@ WebDriver BiDi extends the set of [=error codes=] from [[WEBDRIVER|WebDriver]]
 with the following additional codes:
 
 <dl>
+  <dt><dfn for=errors export>invalid web extension</dfn>
+  <dd>Tried to install invalid a web extension.
```
kiaraarose, line 607: `Tried to install an invalid web extension.`

kiaraarose, line 607:
```suggestion
  <dd>Tried to install an invalid web extension.
```

oliverdunk (approved):
Thanks for all of the work here!

index.bs
```diff
@@ -11302,6 +11306,174 @@ The [=remote end steps=] given |session| and |command parameters| are:
 </div>
 
 
+## The webExtension Module ## {#module-webExtension}
```
whimboo, line 11309: Hm, not exactly sure what we actually agreed on. Should the module's name be singular or plural? For other modules we only use singular, so that  all the references below would have to be updated as well.

index.bs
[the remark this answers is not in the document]
whimboo: Sounds fine when not all browsers support this type of installation. We will then add such a field for Firefox only.

index.bs
```diff
@@ -11302,6 +11306,174 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+1. Let |result| be a [=/map=] matching the
+   <code>webExtensions.InstallResult</code> production with the
+   <code>extension</code> field set to |extension id|.
+
+1. Return [=success=] with data |result|.
+
+</div>
+
+<div algorithm>
+To <dfn>expand a web extension data spec</dfn> given |extension data spec|:
```
whimboo, line 11414: Question: Compared to other commands we have the remote end steps following at the end and other algorithms coming first. I think that we should follow that style.

index.bs
```diff
@@ -11302,6 +11306,174 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+
+   <dl>
+    <dt>|type| is the string "<code>path</code>"
+    <dd>
+      1. Let |path| be |extension data spec|["<code>path</code>"].
+      1. Let |entry| be a [=directory entry=] representing the directory with that |path|.
+
+    <dt>|type| is the string "<code>archivePath</code>"
+    <dd>
+      1. Let |archive path| be a value of <code>path</code> field of |extension data spec|.
```
whimboo, line 11430:
```suggestion
      1. Let |archive path| be |extension data spec|["<code>path</code>"].
```

index.bs
```diff
@@ -11302,6 +11306,174 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+      )
+
+      webExtensions.InstallParameters = {
+         extensionData: webExtensions.ExtensionData,
+      }
+
+      webExtensions.ExtensionData = (
+         webExtensions.ExtensionPath /
+         webExtensions.ExtensionArchivePath /
+         webExtensions.ExtensionBase64Encoded
```
whimboo, line 11365:
```suggestion
         webExtensions.ExtensionArchivePath /
         webExtensions.ExtensionBase64Encoded /
         webExtensions.ExtensionPath
```

index.bs
```diff
@@ -11302,6 +11306,174 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+      1. Perform any implementation-defined steps to extract ZIP from |archive path| and let |entry| be a [=directory entry=] representing the extraction result, if this fails return [=error=] with [=error code=] [=invalid web extension=].
+
+    <dt>|type| is the string "<code>base64</code>"
+    <dd>
+      1. Let |bytes| be [=forgiving-base64 decode=] |extension data spec|["<code>path</code>"].
+      1. Perform any implementation-defined steps to extract the archive from |bytes| by using the ZIP algorithm, and let |entry| by a [=directory entry=] representing the extraction path, if this fails return [=error=] with [=error code=] [=invalid web extension=].
+
+   </dl>
+
+1. Return [=success=] with data |entry|.
```
whimboo, line 11440:
```suggestion
1. Return |entry|.
```

index.bs
```diff
@@ -11302,6 +11306,136 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+         method: "webExtensions.install",
+         params: webExtensions.InstallParameters
+      )
+
+      webExtensions.InstallParameters = {
+         extensionData: webExtensions.ExtensionPath / webExtensions.ExtensionArchivePath / webExtensions.ExtensionBase64Encoded,
+         ? allowPrivateBrowsing: bool .default false,
+      }
+
+      webExtensions.ExtensionPath = {
```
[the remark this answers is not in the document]
whimboo: I assume we only move definitions to `types` when we actually have to share them. Otherwise they may end-up only here. So this seems to be fine.

index.bs
```diff
@@ -11302,6 +11306,174 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+         extension: webExtensions.Extension,
+      }
+      </pre>
+   </dd>
+</dl>
+
+<div algorithm="remote end steps for webExtensions.uninstall">
+The [=remote end steps=] with |command parameters| are:
+
+1. Let |extension| be a value of <code>extension</code> field of the |command parameters|.
```
whimboo, line 11467:
```suggestion
1. Let |extension| be |command parameters|["<code>extension</code>"].
```

index.bs
```diff
@@ -11302,6 +11306,174 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+      </pre>
+   </dd>
+</dl>
+
+<div algorithm="remote end steps for webExtensions.install">
+The [=remote end steps=] with |command parameters| are:
+
+1. If installing web extensions isn't supported return [=error=] with error code [=unsupported operation=].
+
+1. Let |extension data spec| be a value of <code>extensionData</code> field of the |command parameters|.
```
whimboo, line 11399:
```suggestion
1. Let |extension data spec| be |command parameters|["<code>extensionData</code>"].
```

index.bs
```diff
@@ -11302,6 +11306,174 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+      </pre>
+   </dd>
+</dl>
+
+<div algorithm="remote end steps for webExtensions.uninstall">
+The [=remote end steps=] with |command parameters| are:
+
+1. Let |extension| be a value of <code>extension</code> field of the |command parameters|.
+
+1. If the browser has no web extension with an id equal to |extension|, return [=error=] with [=error code=] [=invalid argument=].
```
whimboo, line 11469: Invalid argument we usually only throw when the type is invalid or the value not in a list of known values or a range. Should we add a separate error here as well like `no such web extension`?

index.bs
```diff
@@ -11302,6 +11306,174 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+
+1. Perform any implementation-defined steps to remove the web extension from the browser. If this fails, return [=error=] with [=error code=] [=unknown error=].
```
whimboo, line 11471: Above we talk about the session and here the browser. I assume no browser can actually install the extension for a webdriver session only. So maybe we should re-phrase the initial description for both commands?

whimboo (changes requested):
Thank you for all the updates! It looks way better now. 

I've checked all the changes again and noticed quite a few remaining issues. Lets get those fixed, and hopefully @jgraham can do his review as well after the update.

index.bs
OrKoN: The statement about the invalid argument does not seem to be true to me (looking at the usages in the WebDriver BiDi spec).

OrKoN: A separate error sgtm to me though!

whimboo:
@chrmod I wanted to check if you have time to continue on this PR. Let us know if you need some help. Thanks!

index.bs
chrmod: updated

index.bs
chrmod: WebExtension are installed in a browser profile. [It is advisable that remote ends create a new profile when creating a new session](https://w3c.github.io/webdriver/#privacy). So should we reference a profile instead of a browser or session?

index.bs
```diff
@@ -11302,6 +11310,175 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+
+      webExtension.InstallParameters = {
+         extensionData: webExtension.ExtensionData,
+      }
+
+      webExtension.ExtensionData = (
+         webExtension.ExtensionArchivePath /
+         webExtension.ExtensionBase64Encoded /
+         webExtension.ExtensionPath
+      )
```
whimboo, line 11370: Hm, I didn't really notice it before but shouldn't it just be the following?

```suggestion
      webExtension.InstallParameters = {
         webExtension.ExtensionData,
      }

      webExtension.ExtensionData = (
         webExtension.ExtensionArchivePath /
         webExtension.ExtensionBase64Encoded /
         webExtension.ExtensionPath
      )
```

I don't think that we need this extra level of abstraction, and could still add other fields like `temporary` later on to `webExtension.InstallParameters`. @jgraham can you please have a look?

index.bs
```diff
@@ -11302,6 +11310,175 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+
+The <dfn export for=commands>webExtension.uninstall</dfn> command uninstalls a web extension for the session.
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+      <pre class="cddl remote-cddl">
+      webExtension.Uninstall = (
+         method: "webExtension.uninstall",
+         params: webExtension.InstallParameters
```
whimboo, line 11459:
```suggestion
         params: webExtension.UninstallParameters
```

index.bs
whimboo: I would be fine by using `browser`.

whimboo (changes requested):

jgraham (changes requested):
Just a few small cleanups, but generally this looks good to me.

index.bs
whimboo: Alright then lets leave it as is.

index.bs
```diff
@@ -11302,6 +11310,175 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+
+    <dt>|type| is the string "<code>archivePath</code>"
+    <dd>
+      1. Let |archive path| be |extension data spec|["<code>path</code>"].
+      1. Perform any implementation-defined steps to extract ZIP from |archive path| and let |entry| be a [=directory entry=] representing the extraction result, if this fails return [=error=] with [=error code=] [=invalid web extension=].
+
+    <dt>|type| is the string "<code>base64</code>"
+    <dd>
+      1. Let |bytes| be [=forgiving-base64 decode=] |extension data spec|["<code>path</code>"].
+      1. Perform any implementation-defined steps to extract the archive from |bytes| by using the ZIP algorithm, and let |entry| by a [=directory entry=] representing the extraction path, if this fails return [=error=] with [=error code=] [=invalid web extension=].
```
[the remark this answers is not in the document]
chrmod: Had to add a step to read bytes from the archive, but now it looks much clear!

whimboo:
@chrmod would you mind rebasing against main which includes a fix for the test / build issue? Thanks!

chrmod:
@whimboo done and tests are green now.

index.bs
```diff
@@ -630,6 +633,9 @@ with the following additional codes:
   <dt><dfn for=errors export>no such user context</dfn>
   <dd>Tried to reference an unknown [=user context=].
 
+  <dt><dfn for=errors export>no such web extension</dfn>
+  <dd>Tried to interact with not installed web extension.
```
whimboo, line 637: Maybe we can change the error message so that it can more easily be re-used for other commands later as well?

```suggestion
  <dd>Tried to reference an unknown web extension.
```

index.bs
```diff
@@ -11346,6 +11354,187 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+
+### Types ### {#module-webExtension-types}
+
+#### The webExtension.Extension Type #### {#type-webExtension-Extension}
+
+<pre class="cddl remote-cddl local-cddl">
+webExtension.Extension = text
+</pre>
+
+The <code>webExtension.Extension</code> type represents a web extension id within a browser.
```
whimboo, line 11389: @jgraham should we actually use `remote end` for all those cases instead of `browser` to be consistent?

whimboo (approved):
Overall this looks fine to me. Just two nits where I would also like to see what @jgraham thinks about. 

@chrmod thanks a lot for working on this PR!

index.bs
jgraham: I think remote end is marginally better, but arguably only an actual browser can have an extension. I think honestly we're not 100% consistent already (e.g. with the `browser` module)

jgraham (changes requested):
This is nearly ready; I tried to make some fixups, but apparently you didn't grant write access to the branch so I can't. But I suggest applying the following patch:
```
diff --git a/index.bs b/index.bs
--- index.bs
+++ index.bs
@@ -633,9 +633,9 @@
   <dt><dfn for=errors export>no such user context</dfn>
   <dd>Tried to reference an unknown [=user context=].
 
   <dt><dfn for=errors export>no such web extension</dfn>
-  <dd>Tried to interact with not installed web extension.
+  <dd>Tried to reference and unknown web extension.
 
   <dt><dfn for=errors export>unable to close browser</dfn>
   <dd>Tried to close the browser, but failed to do so.
 
@@ -11385,15 +11385,15 @@
 webExtension.Extension = text
 </pre>
 
-The <code>webExtension.Extension</code> type represents a web extension id within a browser.
+The <code>webExtension.Extension</code> type represents a web extension id within a [=remote end=].
 
 ### Commands ### {#module-webExtension-commands}
 
 #### The webExtension.install Command #### {#command-webExtension-install}
 
-The <dfn export for=commands>webExtension.install</dfn> command installs a web extension for the browser.
+The <dfn export for=commands>webExtension.install</dfn> command installs a web extension in the [=remote end=].
 
 <dl>
    <dt>Command Type</dt>
    <dd>
@@ -11462,20 +11462,25 @@
    <dl>
     <dt>|type| is the string "<code>path</code>"
     <dd>
       1. Let |path| be |extension data spec|["<code>path</code>"].
-      1. Let |entry| be a [=directory entry=] representing the directory with that |path|.
+      1. Let |locator| be a [=directory locator=] with [=file system locator/path=] |archive path| and [=file system locator/root=] corresponding to the root of the file system.
+      1. Let |entry| be [=locate an entry=] given |locator|.
 
     <dt>|type| is the string "<code>archivePath</code>"
     <dd>
       1. Let |archive path| be |extension data spec|["<code>path</code>"].
-      1. Perform implementation defined steps to read |bytes| from a file located at |archive path|.
+      1. Let |locator| be a [=file locator=] with [=file system locator/path=] |archive path| and [=file system locator/root=] corresponding to the root of the file system.
+      1. Let |archive entry| be [=locate an entry=] given |locator|.
+      1. If |archive entry| is null, return null.
+      1. Let |bytes| be |archive entry|'s [=file entry/binary data=].
       1. Let |entry| be the result of [=trying=] to [=extract a zip archive=] given |bytes|.
 
     <dt>|type| is the string "<code>base64</code>"
     <dd>
       1. Let |bytes| be [=forgiving-base64 decode=] |extension data spec|["<code>path</code>"].
       1. Let |entry| be the result of [=trying=] to [=extract a zip archive=] given |bytes|.
+      1. If |entry| is failure, return null.
 
    </dl>
 
 1. Return |entry|.
@@ -11490,8 +11495,10 @@
 1. Let |extension data spec| be |command parameters|["<code>extensionData</code>"].
 
 1. Let |extension directory entry| be the result of [=trying=] to [=expand a web extension data spec=] with |extension data spec|.
 
+1. If extension directory entry| is null, return [=error=] with [=error code=] [=invalid web extension=].
+
 1. Perform implementation defined steps to install a web extension from |extension directory entry|. If this fails, return [=error=] with [=error code=] [=invalid web extension=]. Otherwise let |extension id| be the unique identifier of the newly installed web extension.
 
 1. Let |result| be a [=/map=] matching the
    <code>webExtension.InstallResult</code> production with the
@@ -11526,11 +11533,11 @@
 The [=remote end steps=] with |command parameters| are:
 
 1. Let |extension| be |command parameters|["<code>extension</code>"].
 
-1. If the browser has no web extension with an id equal to |extension|, return [=error=] with [=error code=] [=no such web extension=].
+1. If the [=local end=] has no web extension with id equal to |extension|, return [=error=] with [=error code=] [=no such web extension=].
 
-1. Perform any implementation-defined steps to remove the web extension from the browser. If this fails, return [=error=] with [=error code=] [=unknown error=].
+1. Perform any implementation-defined steps to remove the web extension from the [=local end=]. If this fails, return [=error=] with [=error code=] [=unknown error=].
 
 </div>
```

whimboo:
> This is nearly ready; I tried to make some fixups, but apparently you didn't grant write access to the branch so I can't. But I suggest applying the following patch:
> [..]
> -  <dd>Tried to interact with not installed web extension.
> +  <dd>Tried to reference and unknown web extension.

`<dd>Tried to reference an unknown web extension.`

> -1. If the browser has no web extension with an id equal to |extension|, return [=error=] with [=error code=] [=no such web extension=].
> +1. If the [=local end=] has no web extension with id equal to |extension|, return [=error=] with [=error code=] [=no such web extension=].

Should this really be `local end` here? The client doesn't have extensions installed but the remote end.

jgraham:
You're quite right, it should always be `[=remote end=]`.

jgraham:
Also I just noticed in the base64 bit the diff should be
```
       1. Let |bytes| be [=forgiving-base64 decode=] |extension data spec|["<code>path</code>"].
+      1. If |bytes| is failure, return null.
       1. Let |entry| be the result of [=trying=] to [=extract a zip archive=] given |bytes|.
```

index.bs
```diff
@@ -11346,6 +11354,195 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+   <code>extension</code> field set to |extension id|.
+
+1. Return [=success=] with data |result|.
+
+</div>
+
+
+#### The webExtension.uninstall Command #### {#command-webExtension-uninstall}
+
+The <dfn export for=commands>webExtension.uninstall</dfn> command uninstalls a web extension for the browser.
```
chrmod, line 11515: @jgraham  Should this reference remote end instead of the browser?
```suggestion
The <dfn export for=commands>webExtension.uninstall</dfn> command uninstalls a web extension for the [=remote end=].
```

whimboo: Yes, it should be the same as for the other command.

chrmod:
@jgraham I've check github documentation and we should be able to grant you write access to ghostery fork https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork, but somehow I don't see that option to be available for this PR.

Instead this message is shown:
Screenshot 2024-11-28 at 13 44 28

For this PR it probably does not make sense to investigate more, but it would be best if we make it work for the feature ones.

index.bs
```diff
@@ -11346,6 +11354,195 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+      1. If |archive entry| is null, return null.
+      1. Let |bytes| be |archive entry|'s [=file entry/binary data=].
+      1. Let |entry| be the result of [=trying=] to [=extract a zip archive=] given |bytes|.
+
+    <dt>|type| is the string "<code>base64</code>"
+    <dd>
+      1. Let |bytes| be [=forgiving-base64 decode=] |extension data spec|["<code>path</code>"].
+      1. If |bytes| is failure, return null.
+      1. Let |entry| be the result of [=trying=] to [=extract a zip archive=] given |bytes|.
+      1. If |entry| is failure, return null.
```
jgraham, line 11483:
```suggestion
```

jgraham (approved):
Just one cleanup, but otherwise I think we're good to go.

chrmod:
Fantastic! Thank you all!

Issue #799, opened 2024-10-22
Fail `browsingContext.navigate` if navigation is aborted


sadym-chromium:
Currently, the [`browsingContext.navigate`](https://w3c.github.io/webdriver-bidi/#command-browsingContext-navigate) command's [await a navigation](https://w3c.github.io/webdriver-bidi/#await-a-navigation) succeeded in case of **_"navigation aborted"_**. 

Should be fail the command instead?

Repro: https://github.com/web-platform-tests/wpt/pull/48751/files

Spec:
* [`browsingContext.navigate`](https://w3c.github.io/webdriver-bidi/#command-browsingContext-navigate)
* [Aawait a navigation](https://w3c.github.io/webdriver-bidi/#await-a-navigation):
* [Navigate](https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigate) navigable...
* [navigate](https://html.spec.whatwg.org/#navigate), step 22.3: [abort a document and its descendants](https://html.spec.whatwg.org/#abort-a-document-and-its-descendants) given navigable's [active document](https://html.spec.whatwg.org/#nav-document).
* [Abort a document and its descendants](https://html.spec.whatwg.org/#abort-a-document-and-its-descendants):
* [Abort a document](https://html.spec.whatwg.org/#abort-a-document)
* If document's [during-loading navigation ID for WebDriver BiDi](https://html.spec.whatwg.org/#concept-document-navigation-id) is non-null, then:
* Invoke [WebDriver BiDi navigation aborted](https://w3c.github.io/webdriver-bidi/#webdriver-bidi-navigation-aborted) (emits `browsingContext.navigationAborted` event).
* This case does not fall under **_"If event received is "navigation failed" return [error](https://w3c.github.io/webdriver/#dfn-errors) with [error code](https://w3c.github.io/webdriver/#dfn-error-code) [unknown error](https://w3c.github.io/webdriver/#dfn-unknown-error)"_**
* **_"Return [success](https://w3c.github.io/webdriver/#dfn-success) with data body."_**

OrKoN:
Related https://github.com/w3c/webdriver-bidi/issues/112

sadym-chromium:
> jgraham: I think I need to reread this issue and come back to you

@jgraham any updates on this?

Issue #788, opened 2024-09-24
Request for a `navigationCommitted` event to inform immediately about the navigation status


whimboo:
This is a request from the Playwright folks:

> Introduce an event or method to track when a navigation is committed, beyond the current DOMContentLoaded and load events, which would be useful for some client requirements (e.g., 'should return when navigation is committed if commit is specified'). We need to know immediately if the navigation has failed or succeeded.

In Puppeteer this event is named `page.frameNavigated`.

It's not clear right now where in the HTML spec it could hook into to accurately inform about the status of the current navigation.

whimboo:
From a discussion with the Playwright folks they propose the following:

> There two events called `browsingContext.navigationFailed` and `browsingContext.navigationAborted` in the WebDriver BiDi spec, but no `browsingContext.navigationCommitted`. I'd drop `browsingContext.navigationAborted` and add `browsingContext.navigationCommitted` instead.

PR #800, opened 2024-10-23
Fail aborted navigation

sadym-chromium:
https://github.com/w3c/webdriver-bidi/issues/799

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2933,8 +2933,9 @@ To <dfn>await a navigation</dfn> given |navigable|, |request|, |wait condition|,
     "<code>download started</code>", "<code>navigation aborted</code>",
     "<code>navigation failed</code>"» and |navigation id|.
 
-1. If |event received| is "<code>navigation failed</code>"
-   return [=error=] with [=error code=] [=unknown error=].
+1. If |event received| is "<code>navigation failed</code>" or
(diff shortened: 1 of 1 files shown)


OrKoN (approved):

whimboo:
I would still like to see a debate if from a user perspective this is a acceptable solution. Comparing to our WebDriver classic implementation (where we do not fail) this would be a regression for those who want to switch to WebDriver BiDi. It's unclear how many pages will be affected and we know two major ones already.

Should the usage of `location.href` for redirecting a load really end-up as a failed navigation? Also in case of x.com I can see that they now use `<meta http-equiv="refresh" content="0; url = ...` for the redirect.

OrKoN:
I believe from the user perspective the current behavior could be considered a bug if the user specifies waiting for the load event in the navigate call because that condition is not met on the success response. The abort errors gives a clearer understanding that the current navigation was not completed and the client needs to consider additional events. Alternatively, we could specify how waiting conditions could be fulfilled tolerating an aborted navigation but then it's not clear what the command should return since it's dealing with multiple navigations. I think adding the navigation committed wait condition (as requested in other issues) would allow not throwing an abort error.

whimboo:
Just to add a reference to the issue Alex mentioned in his last comment about the navigation committed event: https://github.com/w3c/webdriver-bidi/issues/788

PR #803, opened 2024-10-25
Allow applying preload scripts to opened contexts

OrKoN:


whimboo:
This needs new tests as well, right?

index.bs
```diff
@@ -9479,6 +9490,7 @@ script=].
         functionDeclaration: text,
         ? arguments: [*script.ChannelValue],
         ? contexts: [+browsingContext.BrowsingContext],
+        ? applyToOpeningContexts: bool .default false,
```
jgraham, line 9493: My initial bikeshedding feedback is that I don't love the name `applyToOpeningContexts` because it's long and yet misleading; I don't see how auxiliary browsing contexts (in HTML spec terms) can be considered `opening`. `inherit: true` might be fine, or `addToOpenedContexts` if we want something longer and less ambiguous.

index.bs
```diff
@@ -7546,9 +7547,19 @@ To <dfn export>run WebDriver BiDi preload scripts</dfn> given |environment setti
...
 
-       1. If |preload script|'s <code>contexts</code> does not [=list/contains|contain=] |navigable id|, continue.
+       1. While |navigable| is not null:
+
+          1. Let |navigable id| be |navigable|’s [=navigable/top-level traversable=]'s id.
+
+          1. If |preload script|'s <code>contexts</code> [=list/contains=] |navigable id|,
+            set |is script run allowed| to true and break.
+
+          1. If |preload script|'s <code>applyToOpeningContexts</code> is true,
```
jgraham, line 7559: As a small detail, I'd prefer this to be `1. If |inherit| is false then break. 1. Set |navigable to |navigable|'s original opener`.

But I also feel like this doesn't quite capture the invariant that each preload script is defined exactly once so has a single setting for whether it's inherited. So conceptually if we're not inheriting we're compaing against a set containing a single value, whereas if we are inheriting we're comparing against a set that contains all the values from openers. I think making that structure explicit would make the spec easier to read.

As a different thought: if this applies to preload scripts it should also apply to event subscriptions in the same way, right? Even if we don't do it in the same PR, it could be worth checking if the same logic works in that case, or if we want to hook in to the context creation lifecycle and update the underlying datastructures to explicitly add the new context, so the getters are still doing the same lookup.

jgraham (changes requested):

index.bs
OrKoN: I am fine with `inherit`although I imagine at some point there could multiple ways to inherit configurations. The `opening` idea comes from the fact that contexts (in BiDi) have an opener ID so something is "opening" them at the time we apply this configuration. "Opened" makes it sound to me like the configuration will apply after the context is "opened" but I think this is not important.

index.bs
OrKoN: We would need to apply it to the event subscriptions and, based on the current feedback from the clients, to the viewport configuration.

index.bs
OrKoN: Updated to inherit for now.

index.bs
jgraham: Yes, basically anything that's understood as browser configuration rather than interacting with a specific document.

OrKoN:
As we discussed in the WG meeting, we likely want to have both features but the user context feature has more priority.

PR #805, opened 2024-10-26
Fix incorrect return type for 'setClientWindowState'"

harsha509:


whimboo (approved):
Good catch, and thank you for the patch!

whimboo:
@harsha509 to be able to contribute to our WebDriver BiDi spec you would have to [join the Browser Testing and Tools group](https://www.w3.org/groups/wg/browser-tools-testing/join/). Once done the IPR check would pass as well, and I can get it landed. Thanks.

PR #807, opened 2024-11-06
Include stacktrace into log messages

OrKoN:
This would align behavior in Chrome and the spec and support the following use case:

As a user I want to filter out console messages from third party libraries. For this, I could use the URLs in the stack trace.

Related discussions:

- https://github.com/puppeteer/puppeteer/issues/13264
- https://github.com/whatwg/console/issues/203

Low-priority and we can discuss later.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/807.html" title="Last updated on Jan 13, 2025, 10:51 AM UTC (a86f3e6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/807/a9e13a5...a86f3e6.html" title="Last updated on Jan 13, 2025, 10:51 AM UTC (a86f3e6)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -10854,7 +10854,7 @@ Define the following [=console steps=] with |method|, |args|, and
 
   1. Let |source| be the result of [=get the source=] given [=current Realm Record=].
 
-  1. If |method| is "<code>assert</code>", "<code>error</code>",
+  1. If |method| is "<code>assert</code>", "<code>error</code>", "<code>log</code>",
      "<code>trace</code>", or "<code>warn</code>", let |stack| be the [=current
      stack trace=]. Otherwise let |stack| be null.
 


sadym-chromium (approved):
I guess we are fine with the extra traffic, as long as the user is opting-in for the log messages.

index.bs
```diff
@@ -10854,7 +10854,7 @@ Define the following [=console steps=] with |method|, |args|, and
 
   1. Let |source| be the result of [=get the source=] given [=current Realm Record=].
 
-  1. If |method| is "<code>assert</code>", "<code>error</code>",
+  1. If |method| is "<code>assert</code>", "<code>error</code>", "<code>log</code>",
```
OrKoN, line 10857: should we perhaps do it for all methods?

OrKoN:
So I think this spec change does not depend on the behavior of the main Console spec because we are defining WebDriver BiDi-specific steps here. It should be possible to only compute stacktraces if there is a WebDriver BiDi connection and report them as WebDriver BiDi events without necessarily affecting the behavior in other instances of console use.

OrKoN:
So is everyone okay with extending the change to all console methods? or should we start with the log only as initially proposed?

OrKoN:
Closing in favor of https://github.com/w3c/webdriver-bidi/pull/854

Issue #698, opened 2024-04-10
Clarify the logic to compute the initiator type of a request


juliandescottes:
From the current spec https://w3c.github.io/webdriver-bidi/#get-the-initiator

> TODO: Get the type. It’s not quite clear how this ought to work; the CDP data depends on whether the navigation was kicked off by the parser or by script (so e.g. inserting an image from script causes the initiator to be "script"), but that doesn’t correspond to anything in Fetch.

The initiator's type can be one of `"parser" / "script" / "preflight" / "other"`. "other" is the default, "preflight" is dedicated to CORS preflight. We should clarify in which cases the type should be "parser" or "script". 

When using `fetch` or `xmlhttprequest` from a script, the type should be "script". When loading resources based on the parsing of the initial document, the type should be "parser". So we should probably focus on scenarios where elements are added via script but still rely on the parser. @OrKoN can you or someone from Google share more details on how CDP determines the initiator type? 

For comparison, Firefox DevTools has a handful of additional initiator types such as "img", "font", "stylesheet". For instance if a request is done in order to retrieve the image of an `img` element, the type is always "img" regardless of how the element was created.

OrKoN:
The initiator type in the WebDriver BiDi is probably closed to https://fetch.spec.whatwg.org/#concept-request-initiator 

Things like  "img", "font", "stylesheet" match resourceType in Puppeteer and are currently not in WebDriver BiDi spec. This probably matches https://fetch.spec.whatwg.org/#concept-request-destination or https://fetch.spec.whatwg.org/#request-initiator-type

To be honest, I am not sure Fetch defines these well (perhaps we could use the table mentioned in the fetch spec). 

I can try to trace how it is implemented in CDP but I am not sure it is very well defined there.

OrKoN:
Related https://github.com/whatwg/fetch/issues/1563

hbenl:
The initiator type could also be used to make image preload requests discoverable: https://bugzilla.mozilla.org/show_bug.cgi?id=2032946

PR #809, opened 2024-11-06
Expose initiatorType and destination on network.Request

jgraham:
Currently we have network.Initiator, which exposes some detailed information about the stack trace that generated the request on network.beforeRequestSent, plus a `type` field that's not well defined.

Fetch exposes `initiatorType` and `destination` and these are already used by other standards e.g. performanceTiming. Having data that directly corresponds to those other specs makes a good deal of sense so that e.g. performance measurements using WebDriver have the same fields with the same information as the DOM measurement APIs.

With this change we could remove `Network.Initiator.type` since that's redundant with this data, but keep the field for extra information about the stack that generated a request.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5829,6 +5829,8 @@ network.RequestData = {
     cookies: [*network.Cookie],
     headersSize: js-uint,
     bodySize: js-uint / null,
+    destination: text,
+    initiatorType: text,
     timings: network.FetchTimingInfo,
 };
 </pre>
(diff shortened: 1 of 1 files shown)


jgraham:
It's still a little bit unclear to me how this corresponds to `Network.ResourceType` in CSP, but it seems like there ought to be enough information to compute most of those values from these fields (`Ping` doesn't have an obvious mapping, but the others seem likely to). That would require someone from Chromium to verify.

jgraham:
This, plus removing `Network.Inititator.type` is a proposal for solving #698 and an alternative to #714.

whimboo:
Overall, this looks good to me. However, I’d like to refer to @OrKoN first to clarify the question about `Network.resourceType` in CDP. It seems that `destination` might align closely with what `resourceType` covers, but there could still be some differences.

Would it be possible to adopt the values defined by the Fetch specification?

index.bs
```diff
@@ -5829,6 +5829,8 @@ network.RequestData = {
     cookies: [*network.Cookie],
     headersSize: js-uint,
     bodySize: js-uint / null,
+    destination: text,
+    initiatorType: text,
```
OrKoN, line 5833:
```suggestion
    initiatorType: text / null,
```

OrKoN:
I think this change does not fully clarify https://github.com/w3c/webdriver-bidi/issues/698 (at least not in the same way Puppeteer currently expects the initiator to work, for example, by providing the referrer and the parser positions for non-script initiator types). The CDP version tries to cover the following use case: how do I track where does this particular request got initiated from providing the exact URL to a position in a resource that initiated the request if it is known. While this is difficult to spec, this spec change might not fully address that use case. 

How do you propose to deal with the removal of `Network.Inititator.type` in terms of backward compatibility?

OrKoN (approved):

whimboo:
@juliandescottes would you mind doing a final review from our side given that you mostly worked on the network features? Thanks.

juliandescottes (approved):
This looks good, thanks. I assume we will handle updating the `initiator` field of beforeRequestSent in another issue/PR?

jgraham:
> I assume we will handle updating the initiator field of beforeRequestSent in another issue/PR?

Yes.

whimboo:
> > I assume we will handle updating the initiator field of beforeRequestSent in another issue/PR?
> 
> Yes.

There is https://github.com/w3c/webdriver-bidi/pull/714. Can this still be used or do we need something new?

PR #811, opened 2024-11-11
Extend locateNodes with a locator to get navigable's container element

### LINKED ISSUE #794: Support finding iframe elements by browsing context id
Puppeteer offers a way to get the frame owner element using the browsing context/navigable ID via the `Frame.frameElement()` API.  Currently, the implementation in Puppeteer relies on finding all iframe/frame elements, serializing their contentWindow and comparing the ID of the serialized window to the id of the target frame. That does not work for Shadow roots as it is not possible to locate iframe/frame elements within closed shadow roots.

The following solutions are possible:

1) Add browsingContext.getFrameOwner(browsingContextId) -> sharedReference
2) Modify locateNodes to allow piercing shadow roots including DOM nodes
3) Extend locateNodes with a new locator type (e.g., FrameOwnerLocator)

Note that having a dedicated locator or method would allow for a more efficient implementation as generally implementations do not need to look over the entire dom tree to find a frame owner element for a given navigable. In the HTML spec terms this seems to correspond to the concept of a navigable container: https://html.spec.whatwg.org/multipage/document-sequences.html#navigable-container

Related Playwright feedback:

> Implement DOM.getFrameOwner and DOM.describeNode to enable converting between DOM node handles and Frame handles. This would improve the implementation of Locator.contentFrame() and FrameLocator.owner(), as the current workaround using contentWindow does not work with shadow DOM, as seen in the page/frame-frame-element.spec.ts tests.

OrKoN:
This change adds a new locator called `context`: when called for a parent navigable, given a child navigable as a locator, it locates the container element for the child navigable that is located within the parent navigable. While providing the parent navigable is redundant and it can be computed based on the child navigable, the interface matches existing locators which all select a node within the parent navigable that matches the locator criteria. In this case, the criterion is the child navigable identifier. 

Providing start nodes to this locator would currently trigger an invalid argument error but eventually it could be supported if there is a use case.
 
Closes #794

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/811.html" title="Last updated on Dec 19, 2024, 11:28 AM UTC (0729b55)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/811/4a85452...0729b55.html" title="Last updated on Dec 19, 2024, 11:28 AM UTC (0729b55)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2967,7 +2967,7 @@ To <dfn>await a navigation</dfn> given |navigable|, |request|, |wait condition|,
 browsingContext.Locator = (
    browsingContext.AccessibilityLocator /
    browsingContext.CssLocator /
-   browsingContext.ContainerLocator /
+   browsingContext.ContextLocator /
    browsingContext.InnerTextLocator /
    browsingContext.XPathLocator
 )
@@ -2985,8 +2985,8 @@ browsingContext.CssLocator = {
    value: text
 }
 
-browsingContext.ContainerLocator = {
-  type: "container",
+browsingContext.ContextLocator = {
+  type: "context",
   value: {
     context: text,
   }
@@ -3792,7 +3792,7 @@ To <dfn>locate nodes using CSS</dfn> with given |navigable|, |context nodes|,
 </div>
 
 <div algorithm="locate the container element">
-To <dfn>locate the container element</dfn> with given |navigable|:
+To <dfn>locate the container element</dfn> given |navigable|:
 
 1. Let |returned nodes| be an empty [=/list=].
 
@@ -4070,7 +4070,7 @@ The [=remote end steps=] with |session| and |command parameters| are:
          1. Let |result nodes| be [=locate nodes using accessibility attributes=]
             given |context nodes|, |selector|, and |maximum returned node count|.
 
-      <dt>|type| is the string "<code>container</code>"
+      <dt>|type| is the string "<code>context</code>"
       <dd>
 
          1. If |start nodes parameter| is not null,
@@ -4078,9 +4078,16 @@ The [=remote end steps=] with |session| and |command parameters| are:
 
          1. Let |selector| be |locator|["<code>value</code>"].
 
-         1. Let |child navigable| be |selector|["<code>context</code>"].
+         1. Let |context id| be |selector|["<code>context</code>"].
+
+         1. Let |child navigable| be the result of [=trying=] to [=get a navigable=] with |context id|.
+
+         1. If |child navigable|'s [=navigable/parent=] is not |navigable|,
+            return [=error=] with [=error code=] "<code>invalid argument</code>".
 
          1. Let |result nodes| be [=locate the container element=] given |child navigable|.
+
+         1. Assert: For each |node| in |result nodes|, |node|'s [=/node navigable=] is |navigable|.
 
 1. Assert: |maximum returned node count| is null or [=list/size=] of |result nodes| is less
    than or equal to |maximum returned node count|.
--- a/index.bs
+++ b/index.bs
@@ -2988,7 +2988,7 @@ browsingContext.CssLocator = {
 browsingContext.ContextLocator = {
   type: "context",
   value: {
-    context: text,
+    context: browsingContext.BrowsingContext,
   }
 }
 
--- a/index.bs
+++ b/index.bs
@@ -2987,6 +2987,9 @@ browsingContext.CssLocator = {
 
 browsingContext.ContainerLocator = {
   type: "container",
+  value: {
+    context: text,
+  }
 }
 
 browsingContext.InnerTextLocator = {
@@ -4070,9 +4073,14 @@ The [=remote end steps=] with |session| and |command parameters| are:
       <dt>|type| is the string "<code>container</code>"
       <dd>
 
-         1. Issue: throw errors if extra unused optional params are set?
+         1. If |start nodes parameter| is not null,
+            return [=error=] with [=error code=] "<code>invalid argument</code>".
 
-         1. Let |result nodes| be [=locate the container element=] given |navigable|.
+         1. Let |selector| be |locator|["<code>value</code>"].
+
+         1. Let |child navigable| be |selector|["<code>context</code>"].
+
+         1. Let |result nodes| be [=locate the container element=] given |child navigable|.
 
 1. Assert: |maximum returned node count| is null or [=list/size=] of |result nodes| is less
    than or equal to |maximum returned node count|.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2960,6 +2960,7 @@ To <dfn>await a navigation</dfn> given |navigable|, |request|, |wait condition|,
 browsingContext.Locator = (
    browsingContext.AccessibilityLocator /
    browsingContext.CssLocator /
+   browsingContext.ContainerLocator /
```
whimboo, line 2963: Is there a reason why we cannot have the container as part of the return value of the `browsingContext.getTree` command? It would be somewhat similar to the `clientWindow` field.

OrKoN: Yes, that API currently does not include any shared references from script and would not allow expressing the ownership of the returned DOM element.

OrKoN: To clarify: we could probably plumb it via the getTree along with the other options we might need to define how to return the element but to me it looks like locateNodes is a better place compared to the getTree.

jgraham: I think there's a slight misconception here: `sharedId` for nodes doesn't depend on ownership. `locateNodes` also doesn't provide a way to set the ownership. I broadly think that's fine for this API; it doesn't seem more useful to hold a reference to the container element than the Window itself.

OrKoN: Hm, when I was reviewing the locateNodes PR I recall there was the ownership parameter but indeed I do not see it after checking now. If we were to use the locateNodes for things other than a11y in Puppeteer, we would definitely need the ownership param to match the existing behavior.

OrKoN: Alright, I recall now: so in CDP we retain the resulting collection and not nodes individually so that is still different to the API here.

jgraham: The WebDriver model to this point is that the sharedId for a node is just an intrinsic property of the node, and returning it doesn't affect the lifetime of the node in any way. `executeScript` and friends also provide a way to take (shared) ownership of the (currently only root) return value.

We already have lots of APIs that return objects without providing any way to take ownership of them (notably anything returning a context). I don't immediately see why you'd need to take ownership of an iframe element with this kind of API; if something removes the iframe from the DOM you're probably in an unexpected state even if you still have a reference to the js object.

Anyway, I don't have a really strong opinion here, but if clients ~always want to know which iframe owns a nested browsing context, I think that `getTree` would work. If it's a rarer requirement, or there really is a strong use case for taking ownership of the iframe itself, then using either a separate command or a locator could work.

OrKoN: To clarify on the ownership: all element handles in Puppeteer (and maybe other clients) are retained until the client releases them. It allows getting the a handle to something and, for example, checking that it was disconnected for DOM or operating on temporary objects without polluting the global scope (e.g., creating some elements, retaining them via Puppeteer, using them in scripts later). The getter to obtain the container for a browsing context is the same kind of handle and should follow the same lifecycle. 

Currently, we can evaluate using a sharedId to retain but it would be great to avoid the round trip.

OrKoN: so we currently only have requirements to provide this data in a separate command because this data is not always needed. Therefore, I think we should not bring it into getTree to avoid the overhead of finding out the sharedId for each browsing context. I still think the locateNodes is probably the best place given that it returns nodes and any changes made to locateNodes (e.g., bringing the ownership parameter back) should also affect this method.

index.bs
```diff
@@ -4043,6 +4063,18 @@ The [=remote end steps=] with |session| and |command parameters| are:
          1. Let |result nodes| be [=locate nodes using accessibility attributes=]
             given |context nodes|, |selector|, and |maximum returned node count|.
 
+      <dt>|type| is the string "<code>container</code>"
```
jgraham, line 4066: We're not locating by container, we're locating a container by context id. So I think this would make sense if we had the type here as "context" or "containedContext" or something.

index.bs
```diff
@@ -4043,6 +4063,18 @@ The [=remote end steps=] with |session| and |command parameters| are:
...
+      <dd>
+
+         1. If |start nodes parameter| is not null,
```
jgraham, line 4069: I think this, and the confusion over whether the navigable needs to be a child of the target navigable, points slightly in the direction of a new command being preferable here, but I won't block on that or anything :)

index.bs
```diff
@@ -4043,6 +4063,18 @@ The [=remote end steps=] with |session| and |command parameters| are:
...
+            return [=error=] with [=error code=] "<code>invalid argument</code>".
+
+         1. Let |selector| be |locator|["<code>value</code>"].
+
+         1. Let |child navigable| be |selector|["<code>context</code>"].
```
jgraham, line 4074: This isn't a navigable, it's a context id.

index.bs
```diff
@@ -3776,6 +3784,18 @@ To <dfn>locate nodes using CSS</dfn> with given |navigable|, |context nodes|,
 
 </div>
 
+<div algorithm="locate the container element">
+To <dfn>locate the container element</dfn> with given |navigable|:
+
+1. Let |returned nodes| be an empty [=/list=].
+
+1. If |navigable|'s [=navigable/container=] is not null,
```
jgraham, line 3792: What if the navigable here isn't actually a child of the navigable the command was sent to? That should return an error I assume.

jgraham (changes requested):

index.bs
```diff
@@ -2995,6 +2996,13 @@ browsingContext.CssLocator = {
    value: text
 }
 
+browsingContext.ContextLocator = {
+  type: "context",
+  value: {
+    context: text,
```
sadym-chromium, line 3002:
```suggestion
    context: browsingContext.BrowsingContext,
```

OrKoN:
@jgraham @whimboo do you have any remaining concerns about this PR?

index.bs
```diff
@@ -2984,6 +2985,13 @@ browsingContext.CssLocator = {
    value: text
 }
 
+browsingContext.ContextLocator = {
```
whimboo, line 2988: I still find it puzzling when naming it `ContextLocator` but then returning a `Node` which is the container element of a context. Why was it changed from `ContainerLocator`?

OrKoN: Changed on request by @jgraham because the locator name reflects the criteria for search and not the return value. I.e., browsingContext.ContextLocator locates by a context selector,  browsingContext.CssLocator locates by a CSS selector.

whimboo: I’m still not entirely sold on this API, but if @jgraham is okay with it, we can move forward.

jgraham (approved):
I am to be honest still not convinced that this API doesn't have so many special cases compared to other locators it wouldn't make more sense to just have `browsingContext.getContainer`, but I guess if this is the direction more people are happy with then I think the implementation is basically correct.

OrKoN:
Thanks, started working on some WPT tests https://github.com/web-platform-tests/wpt/pull/50023

PR #812, opened 2024-11-12
Make sure --upgrade results in a pip --upgrade

OrKoN:
I noticed that `--upgrade` was no upgrading Bikeshed for me locally resulting in errors like for bikeshed v3

```
FATAL ERROR: Something's gone wrong with the remote data; I can't read its timestamp. Please report this!
 ✘  Did not generate, due to fatal errors
 ```

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -1,9 +1,12 @@
 #!/bin/bash
 set -ex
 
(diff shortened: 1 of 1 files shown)


whimboo (approved):
Good find!

PR #813, opened 2024-11-14
Deprecate the type field on Network.Initiator

jgraham:
This marks the field as optional and no longer sets it, so the CDDL continues to match existing implementations, but it's no longer required to set it.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5743,31 +5743,27 @@ To <dfn>deserialize header</dfn> given |protocol header|:
 
 <pre class="cddl local-cddl">
 network.Initiator = {
-    type: "parser" / "script" / "preflight" / "other",
     ? columnNumber: js-uint,
     ? lineNumber: js-uint,
     ? stackTrace: script.StackTrace,
-    ? request: network.Request
+    ? request: network.Request,
+    ? type: "parser" / "script" / "preflight" / "other"
 };
 </pre>
 
 The <code>network.Initiator</code> type represents the source of a network
 request.
 
+Note: The <code>type</code> field is included in the definition for backwards
+compatibility, but is no longer set by the [=get the initiator=] steps, and will
+be removed in a future revision of this specification. Its use is expected to be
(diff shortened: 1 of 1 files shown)


jgraham:
Another options would be for the spec to specify that the value is always "other" for now.

OrKoN (approved):

index.bs
```diff
@@ -5743,32 +5743,28 @@ To <dfn>deserialize header</dfn> given |protocol header|:
 
 <pre class="cddl local-cddl">
 network.Initiator = {
-    type: "parser" / "script" / "preflight" / "other",
     ? columnNumber: js-uint,
```
OrKoN, line 5746: it's a bit awkward that everything is optional but the initiator field itself on network.BeforeRequestSentParameters is not.

jgraham: Yeah, maybe it should be now if there's no stack.

jgraham: Updated with that change.

index.bs
```diff
@@ -5784,13 +5780,12 @@ To <dfn>get the initiator</dfn> given |request|:
...
-   <code>lineNumber</code> field set to |line number| if it's not null, or
-   omitted otherwise, the <code>stackTrace</code> field set to |stack trace| if
-   it's not null, or omitted otherwise, and the <code>request</code> field set
-   to |request id|.
+1. Return a [=/map=] matching the <code>network.Initiator</code> production, the
+   <code>columnNumber</code> field set to |column number| if it's not null, or
+   omitted otherwise, the <code>lineNumber</code> field set to |line number| if
+   it's not null, or omitted otherwise, the <code>stackTrace</code> field set to
+   |stack trace| if it's not null, or omitted otherwise, and the
+   <code>request</code> field set to |request id|.
```
juliandescottes, line 5788: I know this was merged already but based on the definition here, `request` is not actually optional and can never be missing, which means Initiator can never be empty and should not be optional in the BeforeRequestSentParameters ? Or should we remove request from the map?

PR #815, opened 2024-11-15
Remove request field from network.Initiator

juliandescottes:
Just a follow up to my comment on https://github.com/w3c/webdriver-bidi/pull/813/files#r1843522432 

By definition at the moment the initiator cannot be empty, because request id is always defined. I'm proposing to remove it, it doesn't sound useful. We also did not emit it in Firefox FWIW. 

cc @jgraham

***
<a href="https://pr-preview.s3.amazonaws.com/juliandescottes/webdriver-bidi/pull/815.html" title="Last updated on Nov 22, 2024, 11:15 AM UTC (0cfac08)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/815/46306e8...juliandescottes:0cfac08.html" title="Last updated on Nov 22, 2024, 11:15 AM UTC (0cfac08)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5765,7 +5765,7 @@ compatibility, but is no longer set by the [=get the initiator=] steps, and will
 be removed in a future revision of this specification. The
 <code>network.Initiator</code> is included in the
 <code>network.BeforeRequestSentParameters</code> which also contain the same
-request id, which makes the information redundant. See
+request id, making this information redundant. See
 [[#type-network-BaseParameters]].
 
 <div algorithm>
--- a/index.bs
+++ b/index.bs
@@ -5745,8 +5745,8 @@ To <dfn>deserialize header</dfn> given |protocol header|:
 network.Initiator = {
     ? columnNumber: js-uint,
     ? lineNumber: js-uint,
-    ? stackTrace: script.StackTrace,
     ? request: network.Request,
+    ? stackTrace: script.StackTrace,
     ? type: "parser" / "script" / "preflight" / "other"
 };
 </pre>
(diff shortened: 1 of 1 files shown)


OrKoN (approved):

index.bs
```diff
@@ -5745,8 +5745,8 @@ To <dfn>deserialize header</dfn> given |protocol header|:
 network.Initiator = {
     ? columnNumber: js-uint,
     ? lineNumber: js-uint,
-    ? stackTrace: script.StackTrace,
     ? request: network.Request,
```
whimboo, line 5748: As the PR's title suggests the `request` field should be removed. But it's still present in the CDDL. Maybe you tried to do it similarly to what @jgraham did for the `type` field. But `request` is already optional so we potentially could remove it? @OrKoN, or does Chrome return this field?

juliandescottes: Oh right I thought we should keep it for backward, but if it's fine to remove happy to do that instead (I just reordered it alphabetically otherwise)

OrKoN: I thought we keep it in CDDL for now but unset in the spec text.

juliandescottes:
I have the same build issue as other PRs, and which also affects the latest builds in the main branch 

> LINE 2049: No 'dfn' refs found for 'viewport' with spec 'css2'.
> `<a bs-line-number="2049" data-link-spec="css2" data-link-type="dfn" data-lt="viewport">viewport</a>`

Not sure how bikeshed the available definitions from external specs, I couldn't see any recent change for the css2 spec which would explain the regression.

whimboo:
The problem was fixed. So please rebase and push again. Thanks.

juliandescottes:
rebased. cc @whimboo do you want to review or approve?

index.bs
```diff
@@ -5760,11 +5760,15 @@ be removed in a future revision of this specification. Its use is expected to be
 replaced by <code>initiatorType</code> and <code>destination</code> on
 <code>network.RequestData</code>.
 
+Note: The <code>request</code> field is included in the definition for backwards
+compatibility, but is no longer set by the [=get the initiator=] steps, and will
+be removed in a future revision of this specification. The network.Initiator is
+included in the BeforeRequestSentParameters which also contain the same request
+id, which makes the information redundant.
```
whimboo, line 5767: Should we maybe add the reference to `BaseParameters` here to make it easier to find?

jgraham (approved):

PR #818, opened 2024-11-18
Fix broken fragment references in spec

tidoust:
This fixes broken and obsolete fragment references in the spec.

Most updates are straightforward. A few notes:
- `tools.ietf.org` now redirects to `datatracker.ietf.org`
- The update drops any link to "handled" in HTML. Not sure where the term was defined but it no longer seems to exist.
- "default classic script fetch options" is "default script fetch options" in practice. If "classic" matters, then that term no longer exists!

Not done as part of this update, but some of the references can likely be handled directly by Bikeshed without having to define more brittle anchors.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/818.html" title="Last updated on Nov 28, 2024, 8:05 PM UTC (49bdc4a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/818/46306e8...49bdc4a.html" title="Last updated on Nov 28, 2024, 8:05 PM UTC (49bdc4a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -40,7 +40,7 @@ spec: RFC8610; urlPrefix: https://datatracker.ietf.org/doc/html/rfc8610
 spec: RFC6265
   type: dfn
     text: Cookie; url: https://httpwg.org/specs/rfc6265.html
-    text: Cookie store; url: https://httpwg.org/specs/rfc6265.html
+    text: Cookie store; url: https://httpwg.org/specs/rfc6265.html#storage-model
 spec: RFC6265bis; urlPrefix: https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-05
   type: dfn
     text: Lax; url: section-4.1.2.7
--- a/index.bs
+++ b/index.bs
@@ -19,7 +19,7 @@ Test Suite: https://github.com/web-platform-tests/wpt/tree/master/webdriver/test
 </pre>
 
 <pre class=anchors>
-spec: RFC6455; urlPrefix: https://tools.ietf.org/html/rfc6455
+spec: RFC6455; urlPrefix: https://datatracker.ietf.org/doc/html/rfc6455
   type: dfn
     text: WebSocket URI; url: section-3
     text: Establishes a WebSocket Connection; url: section-4.1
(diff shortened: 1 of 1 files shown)


whimboo:
>     * `tools.ietf.org` now redirects to `datatracker.ietf.org`

Interesting, for me it actually forwards to https://authors.ietf.org/.

tidoust:
> > ```
> > * `tools.ietf.org` now redirects to `datatracker.ietf.org`
> > ```
> 
> Interesting, for me it actually forwards to https://authors.ietf.org/.

Ah, indeed, I hadn't noticed that the root URL forwarded somewhere else! I should have been more complete, as in "`https://tools.ietf.org/html/rfc*` now redirects to the RFC under `datatracker.ietf.org`". WebDriver Bidi links to RFCs, not to https://tools.ietf.org directly, for example: https://tools.ietf.org/html/rfc4648 forwards to https://datatracker.ietf.org/doc/html/rfc4648.

index.bs
```diff
@@ -34,23 +34,23 @@ spec: RFC6455; urlPrefix: https://tools.ietf.org/html/rfc6455
     text: Fail the WebSocket Connection; url: section-7.1.7
     text: Status Codes; url: section-7.4
     text: Handling Errors in UTF-8-Encoded Data; url: section-8.1
-spec: RFC8610; urlPrefix: https://tools.ietf.org/html/rfc8610
+spec: RFC8610; urlPrefix: https://datatracker.ietf.org/doc/html/rfc6455#section-3
```
jgraham, line 37: Having the fragment in the URL here seems unusual. Is it intentional?

index.bs
```diff
@@ -34,23 +34,23 @@ spec: RFC6455; urlPrefix: https://tools.ietf.org/html/rfc6455
...
   type: dfn
     text: match a CDDL specification; url: appendix-C
 spec: RFC6265; urlPrefix: https://httpwg.org/specs/rfc6265.html
   type: dfn
-    text: Cookie; url: section-5.3
-    text: Cookie store; url: section-5.3
-spec: RFC6265bis; urlPrefix: https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-05
+    text: Cookie; url: storage-model
```
jgraham, line 42: It feels like this was always linking to the wrong thing. Seems like maybe it should just link to `#intro` for the definition of a Cookie.

jgraham (approved):
A couple of comments, but this seems like a strict improvement, thanks!

Agreed that we should use the autorefs more often; sometimes we're poking at deep parts of specs that intentionally aren't exported, but at least with WebDriver we could export more terms to avoid needing to use so many hardcoded refs.

index.bs
tidoust: No, and even more importantly, that link targets RFC6455 whereas it should go to RFC8610. I'll fix that!

index.bs
tidoust: I updated the link to target the whole RFC (no fragment), which seems good enough?

index.bs
jgraham: I was very much missing the forest for the trees in the review then, but thanks for fixing correctly anyway :)

index.bs
jgraham: I think the "Cookie store" link should revert to targeting section 5.3 which describes how a client stores cookies.

jgraham: But a link to the whole document is fine for `Cookie`.

jgraham (approved):

Issue #817, opened 2024-11-18
Spec needs to list at least one editor for publication as W3C Technical Report


tidoust:
On the way to First Public Working Draft, I just realized that the Editor's Draft does not list any spec editor. I suppose that's intended. Now, the W3C Process requires that specs be edited by one or more editors (who must participate in the group):

> Every Technical Report published as part of the Technical Report development process is edited by **one or more editors** appointed by a Group [Chair](https://www.w3.org/policies/process/#GeneralChairs). [...] An editor must be a participant, per [§ 3.4.2 Participation in Chartered Groups](https://www.w3.org/policies/process/#group-participation) in the Group responsible for the document(s) they are editing.
https://www.w3.org/policies/process/#publication

That rule is enforced by the pubrules checker, so publication as First Public Working Draft is currently blocked on this. Who may I put as editor(s)?

jgraham:
On 11/18/24 17:09, François Daoust wrote:
> On the way to First Public Working Draft, I just realized that the 
> Editor's Draft does not list any spec editor. I suppose that's intended. 

Yes :)

But if we have to I suppose it makes sense to list me as an editor. 
There are others with significant contributions both of text and review 
who may also want to be listed.

OrKoN:
The guidance says that the editors are appointed by the group chair? I would like to nominate [myself](https://www.w3.org/users/141088/) and @sadym-chromium to be listed then.

sadym-chromium:
W3C User of @sadym-chromium: https://www.w3.org/users/128970

PR #819, opened 2024-11-19
Update Bikeshed metadata: add editors, TR link, drop level

tidoust:
This adds spec editors, as discussed in #817, adds the link to the published TR draft, and drops the notion of level so that the spec gets published as `webdriver-bidi` instead of `webdriver-bidi-1` (levels can be introduced later on if that turns out to be needed).

PR to be merged after publication as First Public Working Draft, the TR link won't exist before that.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1,12 +1,15 @@
 <pre class=metadata>
 Title: WebDriver BiDi
 Shortname: webdriver-bidi
-Level: 1
+Level: None
 Status: ED
 Group: browser-testing-tools
 URL: https://w3c.github.io/webdriver-bidi/
+TR: https://www.w3.org/TR/webdriver-bidi/
 Repository: w3c/webdriver-bidi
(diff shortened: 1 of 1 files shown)


OrKoN (approved):

jgraham (approved):

whimboo:
As @jgraham mentioned on Matrix we can merge this PR once we are on TR/.

tidoust:
> As @jgraham mentioned on Matrix we can merge this PR once we are on TR/.

Oh, we already are ;)
Spec was published as First Public Working Draft last week, see: https://www.w3.org/TR/webdriver-bidi/

(That publication is the reason why a [call for exclusions](https://lists.w3.org/Archives/Public/public-browser-tools-testing/2024OctDec/0009.html) was sent to the mailing-list)

Once this PR and #818 are merged, I'll setup automatic publication.

Issue #166, opened 2021-12-23
Does the specification leak registered event subscriptions?


whimboo:
While working on the implementation of `browsingContext.contextCreated` for Firefox and reading `browsingcontext.contextDestroyed` I cannot find anything in the specification that mentions the removal of closed browsing contexts from the [browsing context event map](https://w3c.github.io/webdriver-bidi/#event-browsing-context-event-map). That means that over time subscriptions for specific browsing contexts will remain in this map even when they aren't attached anymore.

Should we add some text to at least mention implementation specific steps to run to clean-up the map?

sadym-chromium:
I guess, as long as the browsing context has a unique ID, it's not an issue, while removing them from the map can potentially be racy with some belated events.

whimboo:
Please note that the map was recently renamed to [navigable event map](https://w3c.github.io/webdriver-bidi/#event-navigable-event-map). Also checking the HTML spec we call into `browsingContext.contextDestroyed` when closing any top-level or child navigable:

https://html.spec.whatwg.org/#destroy-a-top-level-traversable

That means that we are basically fine and should not leak any closed navigable, also because navigables have an id that will not change. I think that we can close this issue.

PR #828, opened 2024-12-04
Introduce subscription IDs

OrKoN:
This PR implements subscription IDs which can be used to unsubscribe from events. The specification is refactored
to operate on a list of subscriptions.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/828.html" title="Last updated on Dec 20, 2024, 11:23 AM UTC (8c602d0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/828/2a8cb22...8c602d0.html" title="Last updated on Dec 20, 2024, 11:23 AM UTC (8c602d0)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -747,7 +747,7 @@ Note: |navigables| is a set because a [=shared worker=] can be associated
 
 1. For each |subscription| in |session|'s [=subscriptions=]:
 
-   1. If |subscription|'s [=subscription/event names=] do not [=set/contains=] |event name|, continue.
+   1. If |subscription|'s [=subscription/event names=] do not [=set/contains=] |event name|, [=continue=].
 
    1. If |subscription| is  [=subscription/global=] return true.
 
--- a/index.bs
+++ b/index.bs
@@ -718,7 +718,7 @@ if |subscription|'s [=subscription/contexts=] is an empty set.
 
 <div algorithm>
 
-The <dfn export>set of sessions for which an event is enabled</dfn>,
+The <dfn export>set of sessions for which an event is enabled</dfn>
 given |event name| and |navigables| is:
 
 1. Let |sessions| be a new [=/set=].
@@ -740,7 +740,7 @@ To determine if an <dfn export>event is enabled</dfn> given |session|,
 Note: |navigables| is a set because a [=shared worker=] can be associated
       with multiple contexts.
 
-1. Let |top-level traversables| be the result of [=get top-level traversables=] given |navigables|.
+1. Let |top-level traversables| be [=get top-level traversables=] given |navigables|.
 
 1. For each |subscription| in |session|'s [=subscriptions=]:
 
@@ -749,9 +749,9 @@ Note: |navigables| is a set because a [=shared worker=] can be associated
    1. If |subscription|'s [=subscription/contexts=] is an empty set,
       return true.
 
-   1. Let |subscription navigables| be the result of [=get navigables=] given |subscription|'s [=subscription/contexts=].
+   1. Let |subscription navigables| be [=get navigables=] given |subscription|'s [=subscription/contexts=].
 
-   1. Let |subscription top-level traversables| be the result of [=get top-level traversables=] given |subscription navigables|.
+   1. Let |subscription top-level traversables| be [=get top-level traversables=] given |subscription navigables|.
 
    1. Return true if the [=set/intersection=] of |top-level traversables| and
       |subscription top-level traversables| is not [=list/empty=].
@@ -782,9 +782,9 @@ given |event name| and |session| is:
 
    1. Otherwise:
 
-      1. Let |navigables| to the result of [=get navigables=] given |subscription|'s [=subscription/contexts=]:
+      1. Let |navigables| to [=get navigables=] given |subscription|'s [=subscription/contexts=]:
 
-      1. Let |top-level traversables| be the result of [=get top-level traversables=] with |navigables|.
+      1. Let |top-level traversables| be [=get top-level traversables=] with |navigables|.
 
       1. [=set/Append=] each item of |top-level traversables| to |result|.
 
@@ -1110,7 +1110,7 @@ To <dfn>get navigables</dfn> given a [=/list=] of context ids |navigable ids|:
 
 1. For each |navigable id| in |navigable ids|:
 
-   1. Let |navigable| be the [=/navigable=] with id |navigable id|,
+   1. Let |navigable| be the [=/navigable=] with id |navigable id|
       if such [=/navigable=] exists, and null otherwise.
 
    1. [=set/Append=] |navigable| to |result| if |navigable| is not null.
@@ -1909,9 +1909,9 @@ The [=remote end steps=] with |session| and |command parameters| are:
 
    1. If |context ids| is not empty:
 
-      1. Let |navigables| be the result of [=get navigables=] given |context ids|.
+      1. Let |navigables| be [=get navigables=] given |context ids|.
 
-      1. Set |newly subscribed navigables| be the result of [=get top-level traversables=] given |navigables|.
+      1. Set |newly subscribed navigables| be [=get top-level traversables=] given |navigables|.
 
    1. Otherwise, |newly subscribed navigables| to a [=/set=] of all [=navigable/top-level traversables=] in the remote end.
 
--- a/index.bs
+++ b/index.bs
@@ -709,8 +709,6 @@ occurred on the [=remote end=].
 A [=BiDi session=] has a <dfn export for=event>subscriptions</dfn>
 which is a [=/list=] of [=subscriptions=].
 
-Issue: the list of subscriptions should be a map indexed by subscription
-IDs once unsubscribe by contexts is removed.
 
 A <dfn for=event>subscription</dfn> is a [=/struct=] consisting of a
 <dfn for=subscription>subscription id</dfn> (a string), <dfn for=subscription>event names</dfn> (a [=/set=] of event names)
--- a/index.bs
+++ b/index.bs
@@ -1903,7 +1903,7 @@ Issue: This needs to be generalised to work with realms too.
 <div algorithm="remote end steps for session.unsubscribe">
 The [=remote end steps=] with |session| and |command parameters| are:
 
-1. If |command parameters| matches the session.SubscriptionRequest production:
+1. If |command parameters| matches the <code>session.SubscriptionRequest</code> production:
 
    1. Let the |event names| be the value of the <code>events</code> field of
       |command parameters|.
--- a/index.bs
+++ b/index.bs
@@ -711,11 +711,12 @@ which is a [=/list=] of [=subscriptions=].
 
 
 A <dfn for=event>subscription</dfn> is a [=/struct=] consisting of a
-<dfn for=subscription>subscription id</dfn> (a string), <dfn for=subscription>event names</dfn> (a [=/set=] of event names)
-and <dfn for=subscription>contexts</dfn> (a [=/set=] of context ids).
+<dfn for=subscription>subscription id</dfn> (a string),
+<dfn for=subscription>event names</dfn> (a [=/set=] of event names)
+and <dfn for=subscription>top-level traversable ids</dfn> (a [=/set=] of IDs of [=/top-level traversables=]).
 
 A [=subscription=] |subscription| is <dfn for="subscription">global</dfn>
-if |subscription|'s [=subscription/contexts=] is an empty set.
+if |subscription|'s [=subscription/top-level traversable ids=] is an empty set.
 
 <div algorithm>
 
@@ -749,9 +750,7 @@ Note: |navigables| is a set because a [=shared worker=] can be associated
 
    1. If |subscription| is  [=subscription/global=] return true.
 
-   1. Let |subscription navigables| be [=get navigables=] given |subscription|'s [=subscription/contexts=].
-
-   1. Let |subscription top-level traversables| be [=get top-level traversables=] given |subscription navigables|.
+   1. Let |subscription top-level traversables| be [=get navigables=] given |subscription|'s [=subscription/top-level traversable ids=].
 
    1. Return true if the [=set/intersection=] of |top-level traversables| and
       |subscription top-level traversables| is not [=list/empty=].
@@ -782,9 +781,7 @@ given |event name| and |session| is:
 
    1. Otherwise:
 
-      1. Let |navigables| to [=get navigables=] given |subscription|'s [=subscription/contexts=]:
-
-      1. Let |top-level traversables| be [=get top-level traversables=] with |navigables|.
+      1. Let |top-level traversables| to [=get navigables=] given |subscription|'s [=subscription/top-level traversable ids=]:
 
       1. [=set/Append=] each item of |top-level traversables| to |result|.
 
@@ -1901,13 +1898,26 @@ The [=remote end steps=] with |session| and |command parameters| are:
 
 1. Let the |event names| be a [=/set=] containing |command parameters|["<code>events</code>"].
 
-1. Let the |context ids| be a [=/set=] containing the value of <code>contexts</code>
-   field of |command parameters| if it is present or an empty [=/set=] if it isn't.
+1. Let |newly subscribed navigables| be a [=/set=].
+
+1. Let |top-level traversable context ids| be a [=/set=].
+
+1. If |command parameters|[<code>contexts</code>] is not empty:
+
+   1. Let |navigables| be [=get navigables=] given |command parameters|[<code>contexts</code>].
+
+   1. Set |newly subscribed navigables| be [=get top-level traversables=] given |navigables|.
+
+   1. For each |navigable| in |newly subscribed navigables|:
+
+      1. [=set/Append=] |navigable|'s [=navigable id=] to |top-level traversable context ids|.
+
+1. Otherwise, |newly subscribed navigables| to a [=/set=] of all [=navigable/top-level traversables=] in the [=remote end=].
 
 1. Let |subscription| be a [=subscription=] with
    [=subscription/subscription id=] set to the string representation of a UUID,
    [=subscription/event names=] set to |event names|, and
-   [=subscription/contexts=] set to |context ids|.
+   [=subscription/top-level traversable ids=] set to |top-level traversable context ids|.
 
 1. Let |subscribe step events| be a new [=/map=].
 
@@ -1917,16 +1927,6 @@ The [=remote end steps=] with |session| and |command parameters| are:
      subscribe steps=], continue;
 
    1. Let |existing navigables| be a [=set of top-level traversables for which an event is enabled=] given |session| and |event name|.
-
-   1. Let |newly subscribed navigables| be a [=/set=].
-
-   1. If |context ids| is not empty:
-
-      1. Let |navigables| be [=get navigables=] given |context ids|.
-
-      1. Set |newly subscribed navigables| be [=get top-level traversables=] given |navigables|.
-
-   1. Otherwise, |newly subscribed navigables| to a [=/set=] of all [=navigable/top-level traversables=] in the [=remote end=].
 
    1. Set |subscribe step events|[|event name|] to [=set/difference=] of
       |newly subscribed navigables| and |existing navigables|.
@@ -1944,7 +1944,7 @@ The [=remote end steps=] with |session| and |command parameters| are:
     1. Return true if |event one|'s [=subscribe priority=] is less than |event
        two|'s subscribe priority, or false otherwise.
 
-1. If |context ids| is null, let |include global| be true, otherwise let
+1. If |subscription| is [=subscription/global=], let |include global| be true, otherwise let
    |include global| be false.
 
 1. For each |event name| → |navigables| in |subscribe step events|:
@@ -1991,17 +1991,29 @@ The [=remote end steps=] with |session| and |command parameters| are:
 
    1. Let |event names| be a [=/set=] of |command parameters|["<code>events</code>"].
 
-   1. Let |contexts| be a [=/set=] of |command parameters|["<code>contexts</code>"],
-      if |command parameters| [=map/contains=] "<code>contexts</code>", or an empty [=/set=] otherwise.
+   1. Let |top-level traversable context ids| be a [=/set=].
+
+   1. If |command parameters|[<code>contexts</code>] is not empty:
+
+      1. Let |navigables| be [=get navigables=] given |command parameters|[<code>contexts</code>].
+
+      1. Set |top-level traversables| be [=get top-level traversables=] given |navigables|.
+
+      1. For each |navigable| in |top-level traversables|:
+
+         1. [=set/Append=] |navigable|'s [=navigable id=] to |top-level traversable context ids|.
 
    1. Let |new subscriptions| to be a [=/list=].
 
    1. For each |subscription| of |session|'s [=subscriptions=]:
 
-      1. If [=set/intersection=] of |subscription|'s [=subscription/event names=] and |event names| is an empty [=/set=],
-         continue.
+      1. If [=set/intersection=] of |subscription|'s [=subscription/event names=] and |event names| is an empty [=/set=]:
 
-      1. If |contexts| is an empty [=/set=]:
+         1. [=list/append=] |subscription| to |new subscriptions|.
+
+         1. Continue.
+
+      1. If |top-level traversable context ids| is an empty [=/set=]:
 
         1. |=list/Remove=| all items [=list/contains|contained=] in |event names| from |subscription|'s [=subscription/event names=].
 
@@ -2023,7 +2035,7 @@ The [=remote end steps=] with |session| and |command parameters| are:
 
           1. For each |event name| in |subscription|'s [=subscription/event names=]:
 
-              1. Set |event map|[|event name|] to [=set/clone=] of |subscription|'s [=subscription/contexts=].
+              1. Set |event map|[|event name|] to [=set/clone=] of |subscription|'s [=subscription/top-level traversable ids=].
 
           1. For each |event name| in |event names|:
 
@@ -2031,18 +2043,20 @@ The [=remote end steps=] with |session| and |command parameters| are:
                  continue.
 
               1. Set |event map|[|event name|] to the [=set/difference=]
-                 between |event map|[|event name|] and |contexts|.
+                 between |event map|[|event name|] and |top-level traversable context ids|.
 
               1. If |event map|[|event name|] is an empty [=/set=],
 
                 1. [=map/Remove=] |event map|[|event name|].
 
-          1. For each |event name| → |remaining contexts| in |event map|:
+          1. For each |event name| → |remaining top-level traversable ids| in |event map|:
 
-            1. Let |subscription| be a [=subscription=] with
+            1. Let |partial subscription| be a [=subscription=] with
                [=subscription/subscription id=] to |subscription|'s [=subscription/subscription id=],
                [=subscription/event names=] set to a new [=/set=] containing |event name|,
-               [=subscription/contexts=] set to |remaining contexts|.
+               [=subscription/top-level traversable ids=] set to |remaining top-level traversable ids|.
+
+            1. [=list/append=] |partial subscription| to |new subscriptions|.
 
    1. Set |session|'s [=subscriptions=] to |new subscriptions|.
 
--- a/index.bs
+++ b/index.bs
@@ -2016,7 +2016,7 @@ The [=remote end steps=] with |session| and |command parameters| are:
 
       1. If |top-level traversable context ids| is an empty [=/set=]:
 
-        1. |=list/Remove=| all items [=list/contains|contained=] in |event names| from |subscription|'s [=subscription/event names=].
+        1. [=list/Remove=] all items [=list/contains|contained=] in |event names| from |subscription|'s [=subscription/event names=].
 
         1. If |subscription|'s [=subscription/event names=] is not empty:
 
--- a/index.bs
+++ b/index.bs
@@ -394,7 +394,7 @@ To <dfn export>resume</dfn> given |name|, |id| and |parameters|:
 </div>
 
 <div algorithm>
-To <dfn for=set>create</dfn> a [=/set=], from a [=/list=] |input|:
+To <dfn>create a set</dfn>, given a [=/list=] |input|:
 
 1. Let |result| be an empty [=/set=].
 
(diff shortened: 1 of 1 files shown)


OrKoN:
@jgraham I started a draft for managing subscriptions using a single list. I hope it is somewhat readable and if you think it is going in the right direction, I will work on TODOs and trying to make it landable.

index.bs
```diff
@@ -706,30 +706,15 @@ occurred on the [=remote end=].
...
-|session| and |event name|:
-
-1. Let |navigables| be an empty [=/set=].
+A [=BiDi session=] has a <dfn export for=event>subscriptions</dfn>
+which is a [=/list=] of [=subscription=].
 
-1. For each |navigable| → |events| of |session|'s [=navigable event map=]:
+A <dfn for=event>subscription</dfn> is a [=/struct=] consisting of a
+subscription id (a string), event names (a [=/list=] of event names) and
+contexts (a [=/list=] of context ids).
```
jgraham, line 714: I don't think you have to include subscribing to events for an entire user context in this patch, but it might be useful to think about how that would fit into this data structure. Infra doesn't naturally define sum types (and I guess there's a question about whether we would allow a single subscription request to include both a list of top-level traversable ids — i.e. context ids — and a list of user context ids, or require one-or-the-other), but maybe we should have something like a "subscription type" field that can either be "global" or "context" or "user context" and a "subscription data" (or something) field that's a list of contexts or a list of user contexts, or always null for global subscriptions.

index.bs
```diff
@@ -2015,14 +1902,25 @@ Issue: This needs to be generalised to work with realms too.
...
-   |command parameters|.
+1. If |command parameters| matches the session.SubscriptionRequest production:
+
+   1. Let the |list of event names| be the value of the <code>events</code> field of
+      |command parameters|.
+
+   1. Let the |list of contexts| be the value of the <code>contexts</code>
+      field of |command parameters| if it is present or an empty [=/list=] if it isn't.
+
+   1. Remove all subscriptions from |session|'s [=subscriptions=] whose event names is |list of event names|
```
jgraham, line 1913: Yeah, I think this will be the tricky part with this approach, because the contract is that at the end you're no longer subscribed to any of the events for any of the contexts, but you're still subscribed to other events for those contexts, and those events for other contexts.

So we need to handle something like
```
subscribe(contexts=[1,2], events=[A, B])
subscribe(contexts=[2, 3], events=[B, C])
unsubscribe(contexts=[1, 2], events=[B])
```

Maybe it's easier if we model subscriptions as a map from a list of events to a list of contexts (plus an id ofc, but I'll ignore that for simplicity). Then you have the above stored as:
```
[
  {
    A: Contexts([1,2]),
    B: Contexts([1,2])
  },
  {
    B: Contexts([2,3]),
    C: Contexts([2,3])
  }
}
```
and it's easier to see how unsubscribe should work: you just iterate through the list of subscriptions, and for each event that's being unsubscribed you remove the contexts you're unsubscribing from the list of subscribed contexts, so you end up with:
```
[
  {
    A: Contexts([1,2]),
  },
  {
    B: Contexts([3]),
    C: Contexts([2, 3])
  }
}
```

jgraham (commented):
In general I think this looks like the right approach, thanks!

We probably need to think carefully about the data structure for subscriptions, but I think it's going to work out :)

index.bs
OrKoN: I would probably prefer to keep subscriptions as a struct and perhaps remove the event names and contexts from the matching subscriptions (which means mutating subscriptions so we won't be able to return subscription data as is if we ever need to) and then remove "empty" subscriptions. In general, I think this unsubscribe behavior is somewhat confusing and I wonder if we want to keep it long-term or if we want to only allow ID-based unsubscribe calls eventually?

index.bs
OrKoN: I was thinking to add user contexts as another set on the subscription with empty sets indicating "all" for any fields here. 

So the subscription is type "global" iff empty(user contexts) and empty(contexts), "context" iff !empty(contexts) && empty(user contexts), "user context" iff  empty(contexts) && !empty(contexts). The structs also restrict that the attributes have to be of same type which I think means we cannot have a null. Subscriptions that have more than 1 non-empty set would be invalid (but not enforced by the types) for now. Alternatively, we can make a map out of the subscription.

index.bs
jgraham: It feels hard to me with the struct approach because each struct has a list of events E and a list of contexts C, and all of ExC are subscribed, but once you unsubscribe you are left with only some subset that are valid, so with the you need to split into i structs with the same id so that now summing over E_i x C_i gives all the valid subscriptions.

I guess the most straightforward, but inefficient, way of doing this is to model it as a totally flat list where each entry has an id, an event, and a context, and unsubscribing means removing all the entries that either match the id or the event name / context.

I agree that this behaviour is somewhat confusing, but I think we're already in a place where it's going to be very hard to deprecate the existing behaviour completely. Maybe eventually we can do it, first by making it opt-in, but I certainly think we're going to need to design in a way that makes the current model implementable.

index.bs
jgraham: That approach seems fine to me.

index.bs
OrKoN: At least I hope there will be a way to avoid introducing this behavior for user context subscriptions? 

> hard to deprecate the existing behaviour completely.

Probably, although Puppeteer, for instance, does not use unsubscribe at all at the moment so perhaps no many clients had adopted edge cases for unsubscribe. A simpler way would be to deprecate the ability to modify subscriptions via unsubscribe: could we instead make a change that unsubscribing only removes the subscriptions that exactly match the subscribe requests? 

I would probably attempt the split upon unsubscribing first with allowing duplicate subscription IDs in the list.

OrKoN: > could we instead make a change that unsubscribing only removes the subscriptions that exactly match the subscribe requests? 

> subscribe(contexts=[1,2], events=[A, B])
> subscribe(contexts=[2, 3], events=[B, C])
> unsubscribe(contexts=[1, 2], events=[B])

with such change the unsubscribe call here would do nothing (we could throw an error if nothing matched). This would allow probably most of the clients to work in almost the same way as I expect most unsubscribe requests to be the same as subscribe request.

jgraham: Requiring subscriptions to exactly match the subscribe requests feels like very breaking behaviour. For example if you have tests with a `subscribe()` command that adds the events / contexts to sets and a teardown fixture that does a single unsubscribe from all events at the end of the test.

Incidentally, for the same reason, we should allow unsubscribing from multiple ids at the same time.

I really think we should try as hard as we can to keep the existing model working for now. I don't even think it's too hard. I'd be totally happy to start emitting warnings for unsubscribe requests that don't use ids once the new model is implemented.

OrKoN: Interestingly, even our own WPT fixtures seem to unsubscribe in a symmetrical way https://github.com/web-platform-tests/wpt/blob/22b20cf0eb577a7df17f7105e47e2b1b818d07b3/webdriver/tests/support/fixtures_bidi.py#L64

I think we could still treat empty contexts set as matching "any" subscription to allow global unsubscribe (and aligned with the fields in the subscription struct). I mean I will try to keep the existing behavior but I wonder if anyone is actually using it at the moment (otherwise, it might be just extra complexity for the spec).

Selenium seems to be only doing global unsubscribes https://github.com/search?q=repo%3ASeleniumHQ%2Fselenium+session.unsubscribe+language%3AJava&type=code&l=Java

OrKoN: for the case of 

```
subscribe(events=[A])
unsubscribe(contexts=[1, 2], events=[A])
```

the current behavior seems to be to do nothing, right?

index.bs
```diff
@@ -1938,31 +1812,45 @@ Issue: This needs to be generalized to work with realms too.
...
+
+   1. Let |existing navigables| be a set of navigables for which either a global
+      subscription is defined or a subscription listing navigables's top-level navigable's id.
+
+      TODO: define the step above better.
+
+   1. Let |newly subscribed navigables| be all top-level navigables if |context ids| is empty,
+      or navigables identified by ids in |context ids|.
+
+   1. Set |subscribe step events|[|event name|] to navigables that are included in |newly subscribed navigables|
```
jgraham, line 1850: This either needs a loop or to use the set operations defined in infra.

index.bs
```diff
@@ -2015,14 +1903,67 @@ Issue: This needs to be generalised to work with realms too.
...
+
+1. Otherwise:
 
-1. Let the |list of contexts| be the value of the <code>contexts</code>
-   field of |command parameters| if it is present or null if it isn't.
+   1. Let the |subscriptions| be |command parameters|[<code>subscriptions</code>].
 
-1. [=Try=] to [=update the event map=] with |session|,
-   |list of event names|, |list of contexts| and enabled false.
+   1. Remove all subscriptions from |session|'s [=subscriptions=] whose
```
jgraham, line 1965: Rewrite this using the operations in infra.

index.bs
```diff
@@ -2015,14 +1903,67 @@ Issue: This needs to be generalised to work with realms too.
 <div algorithm="remote end steps for session.unsubscribe">
 The [=remote end steps=] with |session| and |command parameters| are:
 
-1. Let the |list of event names| be the value of the <code>events</code> field of
-   |command parameters|.
+1. If |command parameters| matches the <code>session.SubscriptionRequest</code> production:
+
+   1. Let the |event names| be the value of the <code>events</code> field of
+      |command parameters|.
```
jgraham, line 1909:
```suggestion
   1. Let |event names| be |command parameters|["<code>events</code>"].
```

index.bs
```diff
@@ -2015,14 +1903,67 @@ Issue: This needs to be generalised to work with realms too.
...
+
+   1. Let the |contexts| be the value of the <code>contexts</code>
```
jgraham, line 1911: Similar change here.

index.bs
```diff
@@ -2015,14 +1903,67 @@ Issue: This needs to be generalised to work with realms too.
...
+          1. For each |event name| in |subscription|'s event names:
+
+              1. Set |event map|[|event name|] to [=set/clone=] of |subscription|'s contexts.
+
+          1. For each |event name| in |event names|:
+
+              1. If |event map|[|event name|] does not exist,
+                 continue.
+
+              1. Remove |contexts| from |event map|[|event name|].
```
jgraham, line 1948: Infra defines [remove](https://infra.spec.whatwg.org/#list-remove), but not as a set difference operation. You could probably write `Remove all items where |contexts| [=set/contains=] item from |event map|[|event name|].` although it's pretty awkward. I think we should consider adding a `difference` operation to infra so that we could write `Set |event map|[|event name|] to the [=set/difference=] between |event map|[|event name|] and |contexts|.`

index.bs
```diff
@@ -2015,14 +1903,67 @@ Issue: This needs to be generalised to work with realms too.
...
+
+              1. If |event map|[|event name|] is an empty [=/set=],
+
+                1. [=map/Remove=] |event map|[|event name|].
+
+          1. For each |event name| → |remaining contexts| in |event map|:
```
jgraham, line 1954: So here we end up maximally splitting, so that each entry contains a single event name, and a list of contexts.

I don't think it's wrong, but I also don't think it's better than my suggestions :) (either we do this from the start and have a struct containing an id, a single event name, and a list of contexts, or we have a struct containing an id and a map from event names to contexts).

I won't block on that however.

jgraham (changes requested):
Generally I think this is correct, but some of the use of spec primitives needs to be tightened up.

Also I think this model where we start with maximally compact structs with sets of both events and contexts, but in unsubscription break them apart into one struct per event is an awkward tradeoff

index.bs
OrKoN: I still hope that we will be able to deprecate this branch. I do not think the current PR's structure is better for the use case of splitting subscriptions but if we pre-process subscriptions upfront like you suggest then we need to also duplicate context ids in the map even if we never split the subscription and take the structure into account in every algorithm. I could also add additional steps to group the events for the same context sets but I think it is also an option to leave this optimization to implementations as the number of actual subscription entries should not be observable (if we want to make it observable it is a different topic).

index.bs
```diff
@@ -2015,14 +1991,76 @@ Issue: This needs to be generalised to work with realms too.
...
+   1. Let |new subscriptions| to be a [=/list=].
+
+   1. For each |subscription| of |session|'s [=subscriptions=]:
+
+      1. If [=set/intersection=] of |subscription|'s [=subscription/event names=] and |event names| is an empty [=/set=],
+         continue.
+
+      1. If |contexts| is an empty [=/set=]:
+
+        1. |=list/Remove=| all |event names| from |subscription|'s [=subscription/event names=].
```
jgraham, line 2010: I still don't think this is quite right. Maybe something like:

```suggestion
        1. |=list/Remove=| all items [=list/contains|contained=] in |event names| from |subscription|'s [=subscription/event names=].
```

index.bs
jgraham: I feel like the data structure where we have an id and a map from event names to what the subscriptions are works pretty well across the entire spec, but you're writing the patch so it's your decision.

index.bs
```diff
@@ -2015,14 +1991,76 @@ Issue: This needs to be generalised to work with realms too.
...
+
+1. Otherwise:
+
+   1. Let |subscriptions| be a [=/set=] of |command parameters|[<code>subscriptions</code>].
+
+   1. Let |items to remove| be an empty [=/list=].
 
-1. Let the |list of contexts| be the value of the <code>contexts</code>
-   field of |command parameters| if it is present or null if it isn't.
+   1. For each |subscription| in |session|'s [=subscriptions=],
```
jgraham, line 2059: So if we try to unsubscribe with a subscription id that doesn't exist we just silently ignore it. That isn't consistent with `script.removePreloadScript` which will error in that case. I can see the case for ignoring, but given the precedent I'd prefer we error if any invalid subscription ids are passed in.

jgraham (changes requested):

index.bs
```diff
@@ -1938,31 +1891,51 @@ Issue: This needs to be generalized to work with realms too.
...
 <div algorithm="remote end steps for session.subscribe">
 The [=remote end steps=] with |session| and |command parameters| are:
 
-1. Let the |list of event names| be the value of the <code>events</code> field of
+1. Let the |event names| be a [=/set=] containing the value of <code>events</code> field of
    |command parameters|
 
-1. Let the |list of navigables| be the value of the <code>contexts</code>
-   field of |command parameters| if it is present or null if it isn't.
+1. Let the |context ids| be a [=/set=] containing the value of <code>contexts</code>
```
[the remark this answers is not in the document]
OrKoN: Good point! I think the current spec does not actually clean up navigable (neither a weak map/set is used nor I could find an explicit clean up) but we should probably do it explicitly in the "WebDriver BiDi navigable destroyed" hooks (related https://github.com/w3c/webdriver-bidi/issues/166). 

I will add upfront conversion to the top-level traversables.

index.bs
OrKoN: Let's say we would remove subscriptions if all related contexts were destroyed, should we still error in that situation?

index.bs
```diff
@@ -755,47 +741,54 @@ To determine if an <dfn export>event is enabled</dfn> given |session|,
...
+1. Let |top-level traversables| be [=get top-level traversables=] given |navigables|.
 
-1. For each |navigable| of |navigables|, append |navigable|'s [=navigable/top-level traversable=] to |top-level traversables|.
+1. For each |subscription| in |session|'s [=subscriptions=]:
 
-1. Let |event map| be the [=navigable event map=] for |session|.
+   1. If |subscription|'s [=subscription/event names=] do not [=set/contains=] |event name|, continue.
 
-1. For each |navigable| of |top-level traversables|:
+   1. If |subscription|is  [=subscription/global=] return true.
```
OrKoN, line 750:
```suggestion
   1. If |subscription| is  [=subscription/global=] return true.
```

index.bs
jgraham: No, I think the rule should be that if something was a valid subscription that hasn't been removed it's OK to remove, even if events in the meantime have rendered it a no-op. That isn't suggestive of a client bug, whereas trying to remove a subscription that was never created, or has already been removed, is.

OrKoN: I have updated the spec to throw an error. I think one edge case to consider is whether we want to throw an error if removal by ID finds no subscriptions but unsubscribe happened previously by attributes (the current spec text would throw).

index.bs
```diff
@@ -706,35 +706,23 @@ occurred on the [=remote end=].
...
-To obtain a list of <dfn>event enabled navigables</dfn> given
-|session| and |event name|:
-
-1. Let |navigables| be an empty [=/set=].
+A [=BiDi session=] has <dfn for=event>subscriptions</dfn> which is a [=/list=] of [=subscriptions=].
 
-1. For each |navigable| → |events| of |session|'s [=navigable event map=]:
+A [=BiDi session=] has <dfn for=event>internally removed subscription ids</dfn>
+which is a [=/set=] of [=subscription/subscription id=] that identifies IDs that
+were issues to the client but have been removed internally since then.
```
sadym-chromium, line 713:
```suggestion
were issued to the client but have been removed internally since then.
```

index.bs
```diff
@@ -2015,14 +1988,107 @@ Issue: This needs to be generalised to work with realms too.
 <div algorithm="remote end steps for session.unsubscribe">
 The [=remote end steps=] with |session| and |command parameters| are:
 
-1. Let the |list of event names| be the value of the <code>events</code> field of
-   |command parameters|.
+1. If |command parameters| matches the <code>session.UnsubscribeByAttributesRequest</code> production:
+
+   1. Let |event names| be a [=/set=] of |command parameters|["<code>events</code>"].
+
+   1. Let |top-level traversable context ids| be a [=/set=].
```
sadym-chromium, line 1995: The same issue with wrong navigable IDs to be interpreted as a global subscription.

index.bs
```diff
@@ -2015,14 +1988,107 @@ Issue: This needs to be generalised to work with realms too.
...
+
+         1. Continue.
 
-1. Let the |list of contexts| be the value of the <code>contexts</code>
-   field of |command parameters| if it is present or null if it isn't.
+      1. If |top-level traversable context ids| is an empty [=/set=]:
 
-1. [=Try=] to [=update the event map=] with |session|,
-   |list of event names|, |list of contexts| and enabled false.
+        1. |=list/Remove=| all items [=list/contains|contained=] in |event names| from |subscription|'s [=subscription/event names=].
```
sadym-chromium, line 2019:
```suggestion
        1. [=list/Remove=] all items [=list/contains|contained=] in |event names| from |subscription|'s [=subscription/event names=].
```

index.bs
```diff
@@ -2015,14 +1988,107 @@ Issue: This needs to be generalised to work with realms too.
...
+                 between |event map|[|event name|] and |top-level traversable context ids|.
+
+              1. If |event map|[|event name|] is an empty [=/set=],
+
+                1. [=map/Remove=] |event map|[|event name|].
+
+          1. For each |event name| → |remaining top-level traversable ids| in |event map|:
+
+            1. Let |partial subscription| be a [=subscription=] with
+               [=subscription/subscription id=] to |subscription|'s [=subscription/subscription id=],
```
sadym-chromium, line 2056:
```suggestion
               [=subscription/subscription id=] set to |subscription|'s [=subscription/subscription id=],
```

index.bs
```diff
@@ -2015,14 +1988,107 @@ Issue: This needs to be generalised to work with realms too.
...
+
+      1. If |subscriptions| [=set/contains=] |subscription|'s [=subscription/subscription id=]:
+
+         1. [=set/Append=] |subscription| to |subscriptions to remove|.
+
+         1. [=set/Append=] |subscription|'s [=subscription/subscription id=] to |found subscription ids|.
+
+   1. Let |not found subscriptions| be [=set/difference=] between |subscriptions| and |found subscription ids|.
+
+   1. Set |not found subscriptions| be [=set/difference=] between |found subscription ids| and |session|'s [=internally removed subscription ids=].
```
sadym-chromium, line 2082:
```suggestion
   1. Set |not found subscriptions| to [=set/difference=] between |found subscription ids| and |session|'s [=internally removed subscription ids=].
```

sadym-chromium, line 2082: The difference between |found subscription ids| and |session|'s [=internally removed subscription ids=] contains all the internally removed subscriptions. 

We need to iterate through the |not found subscriptions| and remove each [=internally removed subscription ids=] from there.

index.bs
```diff
@@ -1938,31 +1886,53 @@ Issue: This needs to be generalized to work with realms too.
...
-   with |session|, |list of event names| , |list of navigables| and
-   enabled true.
+   1. Set |newly subscribed navigables| be [=get top-level traversables=] given |navigables|.
+
+   1. For each |navigable| in |newly subscribed navigables|:
+
+      1. [=set/Append=] |navigable|'s [=navigable id=] to |top-level traversable context ids|.
+
+1. Otherwise, |newly subscribed navigables| to a [=/set=] of all [=navigable/top-level traversables=] in the [=remote end=].
+
```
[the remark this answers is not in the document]
OrKoN: Indeed, the current spec would throw an error here. I think |top-level traversable context ids| is empty is not sufficient but I will publish an update to this in the next patch set.

index.bs
sadym-chromium: UPD: my bad, the "difference" is not commutative.

sadym-chromium (approved):

index.bs
```diff
@@ -1938,31 +1886,59 @@ Issue: This needs to be generalized to work with realms too.
...
+   1. Let |navigables| be [=get navigables by ids=] given |input context ids|.
+
+   1. If [=list/size=] of |navigables| does not equal [=list/size=] of |input context ids|:
+
+      1. Return [=error=] with [=error code=] [=invalid argument=].
 
-1. Let the |list of navigables| be the value of the <code>contexts</code>
-   field of |command parameters| if it is present or null if it isn't.
+   1. Set |newly subscribed navigables| be [=get top-level traversables=] given |navigables|.
```
jgraham, line 1916:
```suggestion
   1. Set |newly subscribed navigables| to [=get top-level traversables=] given |navigables|.
```

jgraham: Also can we rename `newly subscribed navigables` to something else like `subscription navigables` or something. This confused me into thinking it would end up containing only the navigables that we are _newly_ subscribing to rather than all the (top-level) navigables that the subscription applies to.

index.bs
```diff
@@ -1938,31 +1886,59 @@ Issue: This needs to be generalized to work with realms too.
...
       </pre>
    </dd>
 </dl>
 
 <div algorithm="remote end steps for session.subscribe">
 The [=remote end steps=] with |session| and |command parameters| are:
 
-1. Let the |list of event names| be the value of the <code>events</code> field of
-   |command parameters|
+1. Let the |event names| be a [=/set=] containing |command parameters|["<code>events</code>"].
```
jgraham, line 1900: This could be interpreted as a set containing a single list. The tedious way to do this would be something like
```
1. Let |event names| be a [=/set=].

1. For each |event| of |command parameters|["<code>events</code>"], [=set/append=] |event| to |event names|.
```

I think we could probably say `Let |event names| be |command parameters|["<code>events</code>"] interpreted as a [=/set=]`, but that's technically ambiguous if we have a duplicate element (infra just asserts that there are not duplicates in the set).

index.bs
```diff
@@ -1938,31 +1886,59 @@ Issue: This needs to be generalized to work with realms too.
...
+
+1. Let |newly subscribed navigables| be a [=/set=].
+
+1. Let |top-level traversable context ids| be a [=/set=].
+
+1. Let |input context ids| be a [=/set=] of |command parameters|[<code>contexts</code>].
```
jgraham, line 1906: Similar comment about set construction.

index.bs
```diff
@@ -2015,14 +1994,113 @@ Issue: This needs to be generalised to work with realms too.
 <div algorithm="remote end steps for session.unsubscribe">
 The [=remote end steps=] with |session| and |command parameters| are:
 
-1. Let the |list of event names| be the value of the <code>events</code> field of
-   |command parameters|.
+1. If |command parameters| matches the <code>session.UnsubscribeByAttributesRequest</code> production:
+
+   1. Let |event names| be a [=/set=] of |command parameters|["<code>events</code>"].
```
jgraham, line 1999: Similar comment about set construction.

index.bs
```diff
@@ -2015,14 +1994,113 @@ Issue: This needs to be generalised to work with realms too.
...
+
+   1. Let |top-level traversable context ids| be a [=/set=].
+
+   1. Let |input context ids| be a [=/set=] of |command parameters|[<code>contexts</code>].
```
jgraham, line 2003: Similar comment about set construction.

index.bs
```diff
@@ -2015,14 +1994,113 @@ Issue: This needs to be generalised to work with realms too.
...
-   field of |command parameters| if it is present or null if it isn't.
+         1. Continue.
 
-1. [=Try=] to [=update the event map=] with |session|,
-   |list of event names|, |list of contexts| and enabled false.
+      1. If |top-level traversable context ids| is an empty [=/set=]:
+
+        1. [=list/Remove=] all items [=list/contains|contained=] in |event names| from |subscription|'s [=subscription/event names=].
+
+        1. If |subscription|'s [=subscription/event names=] is not empty:
```
jgraham, line 2033: Otherwise shouldn't we add the subscription id to `internally removed subscription ids`? Otherwise you can have code like:
```
// In the harness
subscribe(events=[A])
// In a test
subId = subscribe(events=[A])
// In the harness, for some reason
unsubscribe(events=[A])
// In the test teardown
unsubscribe(subscription=subId)
```
and the second unsubscribe will fail, even though you passed in a totally valid handle. I think it should be an invariant that you only get an error if you use a subscription id to subscribe more than once, even if some subsequent action made the subscription a no-op (it's problematic either way to mix unsubscribing by ids and events, but I think this is less bad because at least a test fixture can gather a list of subscriptions it created and know that passing them to unsubscribe won't break anything. Otherwise I think we should at least return a set of subscription ids that were invalidated, but that's a weird model).

index.bs
```diff
@@ -2015,14 +1994,113 @@ Issue: This needs to be generalised to work with realms too.
...
+                 continue.
+
+              1. Set |event map|[|event name|] to the [=set/difference=]
+                 between |event map|[|event name|] and |top-level traversable context ids|.
+
+              1. If |event map|[|event name|] is an empty [=/set=],
+
+                1. [=map/Remove=] |event map|[|event name|].
+
+          1. For each |event name| → |remaining top-level traversable ids| in |event map|:
```
jgraham, line 2065: Similar comment, if this is empty it shouldn't invalidate the subscription id.

index.bs
```diff
@@ -2015,14 +1994,113 @@ Issue: This needs to be generalised to work with realms too.
...
+               [=subscription/event names=] set to a new [=/set=] containing |event name|,
+               [=subscription/top-level traversable ids=] set to |remaining top-level traversable ids|.
+
+            1. [=list/append=] |partial subscription| to |new subscriptions|.
+
+   1. Set |session|'s [=subscriptions=] to |new subscriptions|.
+
+1. Otherwise:
+
+   1. Let |subscriptions| be a [=/set=] of |command parameters|[<code>subscriptions</code>].
```
jgraham, line 2078: Similar comment about converting a list to a set

index.bs
```diff
@@ -2015,14 +1994,113 @@ Issue: This needs to be generalised to work with realms too.
...
+
+      1. If |subscriptions| [=set/contains=] |subscription|'s [=subscription/subscription id=]:
+
+         1. [=set/Append=] |subscription| to |subscriptions to remove|.
+
+         1. [=set/Append=] |subscription|'s [=subscription/subscription id=] to |found subscription ids|.
+
+   1. Let |not found subscriptions| be [=set/difference=] between |subscriptions| and |found subscription ids|.
+
+   1. Set |not found subscriptions| to [=set/difference=] between |found subscription ids| and |session|'s [=internally removed subscription ids=].
```
jgraham, line 2094: This seems wrong; you're creating a set and then immediately overwriting it. I assume the second line should be `between |not found subscriptions| and |session|'s [=internally removed subscription ids=]`?

index.bs
```diff
@@ -2015,14 +1994,113 @@ Issue: This needs to be generalised to work with realms too.
...
+
+   1. If |not found subscriptions| is not empty:
+
+      1. Return [=error=] with [=error code=] [=invalid argument=].
+
+   1. Set |session|'s [=internally removed subscription ids=] to
+      [=set/difference=] between |session|'s [=internally removed subscription ids=] |subscriptions|.
```
jgraham, line 2101:
```suggestion
   1. Set |session|'s [=internally removed subscription ids=] to
      [=set/difference=] between |session|'s [=internally removed subscription ids=] and |subscriptions|.
```

index.bs
```diff
@@ -706,35 +706,23 @@ occurred on the [=remote end=].
...
-<div algorithm>
-
-To obtain a list of <dfn>event enabled navigables</dfn> given
-|session| and |event name|:
-
-1. Let |navigables| be an empty [=/set=].
+A [=BiDi session=] has <dfn for=event>subscriptions</dfn> which is a [=/list=] of [=subscriptions=].
 
-1. For each |navigable| → |events| of |session|'s [=navigable event map=]:
+A [=BiDi session=] has <dfn for=event>internally removed subscription ids</dfn>
```
jgraham, line 711: Given the later comments, I suggest replacing this with `A [=BiDi session=] has a <dfn for=event>known subscription ids=] which is a [=/set=] of all [=subscription/subscription ids=] that have been issued to the
[=local end=] but which have not yet been unsubscribed`. Then you unconditionally append to this when creating a subscription, and remove from it when unsubscribing by id, and get to remove a lot of logic in other places.

jgraham (changes requested):

index.bs
OrKoN: My idea was that we would still want to error if unsubscribe happened via commands indicating some client bugs/redundant calls but only suppress errors if we removed a subscription due to navigables being disposed and other internal reasons.

So in this scenario, I would indeed expect an error in the last unsubscribe as the subscriptions has already been removed by the client which indicates redundant client calls. I think it could be viewed as sending `unsubscribe(subscription=subId)` twice.

index.bs
OrKoN: Indeed!

index.bs
OrKoN: Added a helper, will propose it to the infra spec.

index.bs
OrKoN: Anyway, I think either way works and tracking known subscription ids makes the spec easier, so I adopted your proposal into the text.

jgraham (approved):
A couple of style nits, but I think this is ready to land without more review.

Thanks for all the work, and for upstreaming the set operations into infra :)

OrKoN:
I filed https://github.com/GoogleChromeLabs/chromium-bidi/issues/2925 to track the work on implementation and WPT

index.bs
```diff
@@ -757,48 +742,25 @@ Note: |navigables| is a set because a [=shared worker=] can be associated
...
 
-1. For each |navigable| of |navigables|, append |navigable|'s [=navigable/top-level traversable=] to |top-level traversables|.
+1. For each |navigable| of |navigables|,
+   append |navigable|'s [=navigable/top-level traversable=] to |top-level traversables|.
 
-1. Let |event map| be the [=navigable event map=] for |session|.
+1. For each |subscription| in |session|'s [=subscriptions=]:
 
-1. For each |navigable| of |top-level traversables|:
+   1. If |subscription|'s event names do not [=set/contains=] |event name|, continue.
```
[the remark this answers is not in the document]
juniormalicdem: All proposetory delete

PR #831, opened 2024-12-12
Use `<code>` for references to a CDDL type

tidoust:
The spec wrapped the return type of commands in  `<pre class="cddl">`, even when the return type is actually a reference to CDDL type defined elsewhere.

The convention with IDL blocks is that any `<pre class="idl">` block creates a new IDL definition. I'd like to follow the same convention for CDDL.

On top of it, from a semantic perspective, such a reference is more "a fragment of computer code" (`<code>`) than "a block of preformatted text" (`<pre>`).

Note: when and if Bikeshed starts supporting CDDL, this could be replaced by something like `<a cddl-type>EmptyResult</a>` (or `{^EmptyResult^}` if the shorthand gets adopted).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/831.html" title="Last updated on Dec 12, 2024, 1:18 PM UTC (d2d44c4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/831/b61fb78...d2d44c4.html" title="Last updated on Dec 12, 2024, 1:18 PM UTC (d2d44c4)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1893,9 +1893,9 @@ The <dfn export for=commands>session.end</dfn> command ends the current
    </dd>
    <dt>Result Type</dt>
    <dd>
-      <pre class="cddl">
+      <code>
       EmptyResult
-      </pre>
+      </code>
    </dd>
 </dl>
 
@@ -1938,9 +1938,9 @@ Issue: This needs to be generalized to work with realms too.
    </dd>
    <dt>Result Type</dt>
    <dd>
-      <pre class="cddl">
+      <code>
       EmptyResult
-      </pre>
+      </code>
    </dd>
 </dl>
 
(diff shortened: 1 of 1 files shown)


OrKoN (approved):

index.bs
```diff
@@ -1893,9 +1893,9 @@ The <dfn export for=commands>session.end</dfn> command ends the current
    </dd>
    <dt>Result Type</dt>
    <dd>
-      <pre class="cddl">
+      <code>
```
OrKoN, line 1896: maybe we should be defining a type for each return type:

```
session.EndResult = EmptyResult
```

so that one can easily find the matching return type event if it is empty.

tidoust: Not sure who you're asking ;) That seems good to me but I'm not sure I understand the consequences this could have on implementations.

Would you like me to update the PR to add types for each return type as suggested?

jgraham: We can if it helps something. The reason we didn't is that it seemed like extra boilerplate to have to specifically define a novel return type for every command, even if it's just an alias for an existing type (but we still want the spec to say what it returns).

OrKoN: I think we can land this CL though to unblock the CDDL extraction.

PR #836, opened 2024-12-20
Introduce user context subscriptions

OrKoN:
This feature will allow scoping events per user context to get events for all newly created browsing contexts within a user context.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/836.html" title="Last updated on Jan 15, 2025, 10:49 AM UTC (a76d539)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/836/2c01081...a76d539.html" title="Last updated on Jan 15, 2025, 10:49 AM UTC (a76d539)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -713,11 +713,13 @@ all [=subscription/subscription ids=] that have been issued to the [=local end=]
 
 A <dfn for=event>subscription</dfn> is a [=/struct=] consisting of a
 <dfn for=subscription>subscription id</dfn> (a string),
-<dfn for=subscription>event names</dfn> (a [=/set=] of event names)
-and <dfn for=subscription>top-level traversable ids</dfn> (a [=/set=] of IDs of [=/top-level traversables=]).
+<dfn for=subscription>event names</dfn> (a [=/set=] of event names),
+<dfn for=subscription>top-level traversable ids</dfn> (a [=/set=] of IDs of [=/top-level traversables=])
+and <dfn for=subscription>user context ids</dfn> (a [=/set=] of IDs of [=user contexts=]).
 
 A [=subscription=] |subscription| is <dfn for="subscription">global</dfn>
-if |subscription|'s [=subscription/top-level traversable ids=] is an empty set.
+if |subscription|'s [=subscription/top-level traversable ids=] is an empty set
+and |subscription|'s [=user context ids=] is an empty set.
 
 <div algorithm>
 
@@ -750,10 +752,19 @@ Note: |navigables| is a set because a [=shared worker=] can be associated
 
    1. If |subscription| is [=subscription/global=] return true.
 
-   1. Let |subscription top-level traversables| be [=get navigables by ids=] with |subscription|'s [=subscription/top-level traversable ids=].
+   1. If [=subscription/user context ids=] is not empty:
 
-   1. If the [=set/intersection=] of |top-level traversables| and
-      |subscription top-level traversables| is not [=list/empty=] return true.
+      1. [=list/For each=] |navigable| in |top-level traversables|:
+
+         1. If |subscription|'s [=subscription/user context ids=] [=set/contains=]
+            |navigable|'s [=associated user context=]'s [=user context id=], return true.
+
+   1. Otherwise:
+
+      1. Let |subscription top-level traversables| be [=get navigables by ids=] with |subscription|'s [=subscription/top-level traversable ids=].
+
+      1. If the [=set/intersection=] of |top-level traversables| and
+          |subscription top-level traversables| is not [=list/empty=] return true.
 
 1. Return false.
 
@@ -778,6 +789,14 @@ given |event name| and |session| is:
         1. [=set/Append=] |traversable| to |result|.
 
       1. [=Break=].
+
+   1. Otherwise, if [=subscription/user context ids=] is not empty:
+
+      1. For each |traversable| in remote end's [=/top-level traversables=]:
+
+         1. [=set/Append=] |traversable| to |result| if
+            |subscription|'s [=subscription/user context ids=] [=set/contains=]
+            |traversable|'s [=associated user context=]'s [=user context id=].
 
    1. Otherwise:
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -750,10 +752,19 @@ Note: |navigables| is a set because a [=shared worker=] can be associated
...
+
+         1. If |subscription|'s [=subscription/user context ids=] [=set/contains=]
+            |navigable|'s [=associated user context=]'s [=user context id=], return true.
+
+   1. Otherwise:
+
+      1. Let |subscription top-level traversables| be [=get navigables by ids=] with |subscription|'s [=subscription/top-level traversable ids=].
+
+      1. If the [=set/intersection=] of |top-level traversables| and
+          |subscription top-level traversables| is not [=list/empty=] return true.
```
jgraham, line 767:
```suggestion
      1. If the [=set/intersection=] of |top-level traversables| and
         |subscription top-level traversables| is not [=list/empty=] return true.
```

index.bs
```diff
@@ -1436,6 +1455,31 @@ To <dfn export>get user context</dfn> given |user context id|:
 
 </div>
 
+When a [=user context=] is [=set/remove|removed=] from the
+[=set of user contexts=], [=remove user context subscriptions=]:
```
jgraham, line 1459:
```suggestion
When a [=user context=] is [=set/remove|removed=] from the
[=set of user contexts=], [=remove user context subscriptions=].
```

But also, can we move this up into the paragraph starting "Implementations may [=set/remove=] any [=empty user context=], with exception of[…]" (I can't add that as a suggestion, sadly).

index.bs
```diff
@@ -1436,6 +1455,31 @@ To <dfn export>get user context</dfn> given |user context id|:
...
+
+<div algorithm>
+
+To <dfn export>remove user context subscriptions</dfn>:
```
jgraham, line 1463: So, this does create undefined behaviour, because it makes it observable whether a specific user context was removed or not, which is implementation defined (I guess technically that was already the case since it affects what's in storage). It isn't a blocker here, but I wonder if we should define that any user context created via `createUserContext` actually is removed if it goes from 1 to 0 browsing contexts.

jgraham (changes requested):
Just a few minor comments, otherwise this looks good, thanks!

index.bs
OrKoN: I am not sure I get why this creates undefined behavior. I added this algorithm to define a clean-up step to avoid accumulating subscriptions if the client does not unsubscribe (unsubscribe requests with the ID of removed subscriptions can still be issued). I think it is not observable to the local end. Could you please clarify?

>  It isn't a blocker here, but I wonder if we should define that any user context created via createUserContext actually is removed if it goes from 1 to 0 browsing contexts.

In Chrome, the user context stays there if all browsing contexts in it are removed.

index.bs
```diff
@@ -1918,12 +1963,26 @@ The [=remote end steps=] with |session| and |command parameters| are:
 
       1. [=set/Append=] |navigable|'s [=navigable id=] to |top-level traversable context ids|.
 
+1. Otherwise, if |input user context ids| is not empty:
```
jgraham, line 1966: Aren't we missing the definition of `|input user context ids|`?

Also I think we need to error if both |input user context ids| and |input context ids| are set (and should be clear in the definition of a subscription that even though it's a product type, those fields are mutually exclusive for any given subscription).

OrKoN: Indeed, let me check, I think I may have lost it while rebasing

index.bs
jgraham: I guess it's actually not externally visible whether or not this cleanup happened, so it's fine.

The problem I'm thinking of is actually already an issue. Say you create a user context, create a tab in that user context, then close the tab. Now an implementation is permitted to remove the user context. But whether or not you do that is visible if you then try to create another new tab in that user context.

It kind of makes sense for gecko because you can delete the user context in the browser UI. So if someone actually interacts with the browser and does that all bets are off. But maybe the spec should be clearer that it's not expected that user contexts (or their associated storage) are removed without user interaction.

index.bs
```diff
@@ -713,11 +713,13 @@ all [=subscription/subscription ids=] that have been issued to the [=local end=]
 
 A <dfn for=event>subscription</dfn> is a [=/struct=] consisting of a
 <dfn for=subscription>subscription id</dfn> (a string),
-<dfn for=subscription>event names</dfn> (a [=/set=] of event names)
-and <dfn for=subscription>top-level traversable ids</dfn> (a [=/set=] of IDs of [=/top-level traversables=]).
+<dfn for=subscription>event names</dfn> (a [=/set=] of event names),
+<dfn for=subscription>top-level traversable ids</dfn> (a [=/set=] of IDs of [=/top-level traversables=])
+and <dfn for=subscription>user context ids</dfn> (a [=/set=] of IDs of [=user contexts=]).
```
jgraham, line 718:
```suggestion
and <dfn for=subscription>user context ids</dfn> (a [=/set=] of IDs of [=user contexts=]).

Note: A [=subscription=] can have non-empty [=subscription/top-level traversable ids=],
or non-empty [=subscription/user context ids=], but not both.
```

index.bs
```diff
@@ -1898,12 +1943,17 @@ The [=remote end steps=] with |session| and |command parameters| are:
 
 1. Let the |event names| be [=set/create|create a set=] with |command parameters|["<code>events</code>"].
 
+1. Let |input user context ids| be [=set/create|create a set=] with |command parameters|[<code>contexts</code>].
```
OrKoN, line 1946:
```suggestion
1. Let |input user context ids| be [=set/create|create a set=] with |command parameters|[<code>userContexts</code>].
```

jgraham (approved):

index.bs
OrKoN: So the same situation can happen in Chrome if the user closes a user context via the UI. I think it makes sense to clarify that user contexts should not get removed randomly by the implementation, but removing them as the result of a direct user action is okay.

whimboo:
@OrKoN are you going to add some tests for this new feature?

OrKoN:
@whimboo see https://github.com/GoogleChromeLabs/chromium-bidi/issues/2983

OrKoN:
@whimboo if you would like to work on tests earlier than us, as always it would be very much appreciated. Please leave a comment on the issue if that is the case, otherwise, we will add the tests once we start working on the implementation.

whimboo:
It would be @lutien who will work on it, but right now we are not there yet.

lutien:
Hey @OrKoN, I've seen that you've added invalid tests for this argument, but I would like to add more tests now if you're not working on it yet. Let me know if that's the case.

OrKoN:
> Hey @OrKoN, I've seen that you've added invalid tests for this argument, but I would like to add more tests now if you're not working on it yet. Let me know if that's the case.

I planned to start on the second batch of tests on Monday

OrKoN:
I have started working on the tests

whimboo:
@lutien I assume we have enough tests now so that we can remove the `needs-tests` label?

lutien:
yes, I removed the label

PR #847, opened 2025-01-09
Unsubscribe by attributes should check that all events and context ids exist in subscriptions

OrKoN:
In https://github.com/w3c/webdriver-bidi/pull/828 I also overlooked that removal of subscriptions should only apply if all unsubscribe attributes match existing subscription data. To remain backward compatible, this PR adds steps to account for that and cover the cases like the following:

For events:

```
subscribe(events=[A, B])
unsubscribe(events=[A]) => works
unsubscribe(events=[A, B]) => expected to throw
```

For contexts:
```
subscribe(contexts=[A, B])
unsubscribe(contexts=[A]) => works
unsubscribe(contexts=[A, B]) => expected to throw
```

```
subscribe(events=[A])
unsubscribe(events=[A], contexts=[A]) => expected to throw
```

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/847.html" title="Last updated on Jan 10, 2025, 2:52 PM UTC (26665d2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/847/87cd2c0...26665d2.html" title="Last updated on Jan 10, 2025, 2:52 PM UTC (26665d2)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2054,21 +2054,47 @@ The [=remote end steps=] with |session| and |command parameters| are:
 
    1. Let |new subscriptions| to be a [=/list=].
 
+   1. Let |match| be false.
+
    1. For each |subscription| of |session|'s [=subscriptions=]:
+
+      1. If |match| is true:
+
+         1. [=list/append=] |subscription| to |new subscriptions|.
+
+         1. [=Continue=].
 
       1. If [=set/intersection=] of |subscription|'s [=subscription/event names=] and |event names| is an empty [=/set=]:
 
          1. [=list/append=] |subscription| to |new subscriptions|.
 
-         1. Continue.
+         1. [=Continue=].
 
       1. If |top-level traversable context ids| is an empty [=/set=]:
 
-        1. [=list/Remove=] all items [=list/contains|contained=] in |event names| from |subscription|'s [=subscription/event names=].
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2107,6 +2136,11 @@ The [=remote end steps=] with |session| and |command parameters| are:
 
             1. [=list/append=] |partial subscription| to |new subscriptions|.
 
+   1. If |matched events| does not equal |event names|, return [=error=] with [=error code=] [=invalid argument=].
+
+   1. If |top-level traversable context ids| is not empty and |matched contexts| does not equal
```
sadym-chromium, line 2141: Technically we can simply compare `|matched contexts|` and `|top-level traversable context ids|`, as for global subscriptions they are both empty, for non-global the latest is non-empty. But the current version makes the algo more readable.

sadym-chromium (approved):

sadym-chromium (approved):

OrKoN:
@jgraham PTAL (this is another spec regression from my subscription ID refactoring that I discovered while actually working on the implementation, I believe this is the last one).

jgraham (approved):

PR #854, opened 2025-01-13
Always include stacktrace for console messages

OrKoN:
As an alternative to https://github.com/w3c/webdriver-bidi/pull/807 it makes sense to always include the stacktrace so that the client can figure out where the logs are coming from.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/854.html" title="Last updated on Jan 14, 2025, 5:05 PM UTC (e6001f1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/854/e25f510...e6001f1.html" title="Last updated on Jan 14, 2025, 5:05 PM UTC (e6001f1)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -11029,17 +11029,14 @@ Define the following [=console steps=] with |method|, |args|, and
 
   1. Let |source| be the result of [=get the source=] given [=current Realm Record=].
 
-  1. If |method| is "<code>assert</code>", "<code>error</code>",
-     "<code>trace</code>", or "<code>warn</code>", let |stack| be the [=current
-     stack trace=]. Otherwise let |stack| be null.
+  1. Let |stack| be the [=current stack trace=].
 
   1. Let |entry| be a [=/map=] matching the <code>log.ConsoleLogEntry</code> production,
      with the the <code>level</code> field set to |level|, the <code>text</code>
      field set to |text|, the <code>timestamp</code> field set to |timestamp|, the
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -11029,9 +11029,7 @@ Define the following [=console steps=] with |method|, |args|, and
 
   1. Let |source| be the result of [=get the source=] given [=current Realm Record=].
 
-  1. If |method| is "<code>assert</code>", "<code>error</code>",
-     "<code>trace</code>", or "<code>warn</code>", let |stack| be the [=current
-     stack trace=]. Otherwise let |stack| be null.
+  1. Let |stack| be the [=current stack trace=].
```
juliandescottes, line 11032: With this change stack can't be null so we can simplify the next step: 

> the <code>stackTrace</code> field set to |stack| if |stack| is not null, or
> omitted otherwise,

can become

> the <code>stackTrace</code> field set to |stack|

Also I think `stackTrace` is now mandatory in `log.BaseLogEntry`?

juliandescottes (changes requested):
Thanks! Some small changes needed now that stack will never be null

index.bs
juliandescottes: Ah yes indeed. A bit weird to have an unused / undefined log type here, but that's another topic :) 

Thanks for the patch and sorry about the delay

juliandescottes (approved):

PR #855, opened 2025-01-13
Add browsingContext.navigationCommitted event and wait for it when no wait condition is provided

OrKoN:
Closes https://github.com/w3c/webdriver-bidi/issues/788
HTML spec: https://github.com/whatwg/html/pull/10910

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/855.html" title="Last updated on Jan 15, 2025, 7:45 PM UTC (5e34be5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/855/e25f510...5e34be5.html" title="Last updated on Jan 15, 2025, 7:45 PM UTC (5e34be5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2840,6 +2840,7 @@ BrowsingContextEvent = (
   browsingContext.HistoryUpdated //
   browsingContext.Load //
   browsingContext.NavigationAborted //
+  browsingContext.NavigationCommitted //
   browsingContext.NavigationFailed //
   browsingContext.NavigationStarted //
   browsingContext.UserPromptClosed //
@@ -3041,9 +3042,13 @@ To <dfn>await a navigation</dfn> given |navigable|, |request|, |wait condition|,
 
   1. Return [=success=] with data |body|.
 
-1. If |wait condition| is "<code>interactive</code>", let |event name| be
-   "<code>domContentLoaded</code>", otherwise let |event name| be
-   "<code>load</code>".
+1. If |wait condition| is "<code>committed</code>", let |event name| be
+   "<code>committed</code>".
+
+1. Otherwise, if |wait condition| is "<code>interactive</code>", let |event name| be
+   "<code>domContentLoaded</code>".
+
+1. Otherwise, let |event name| be "<code>load</code>".
 
 1. Let (|event received|, |status|) be [=await=] given «|event name|,
     "<code>download started</code>", "<code>navigation aborted</code>",
@@ -3168,7 +3173,7 @@ To <dfn>get the navigation info</dfn>, given |navigable| and |navigation status|
 #### The browsingContext.ReadinessState Type #### {#type-browsingContext-ReadinessState}
 
 <pre class="cddl remote-cddl">
-browsingContext.ReadinessState = "none" / "interactive" / "complete"
+browsingContext.ReadinessState = "none" / "committed" / "interactive" / "complete"
 </pre>
 
 The <code>browsingContext.ReadinessState</code> type represents the stage of
@@ -5111,6 +5116,44 @@ aborted</dfn> steps given |navigable| and |navigation status|:
 
 1. For each |session| in the [=set of sessions for which an event is enabled=]
    given "<code>browsingContext.navigationAborted</code>" and |related navigables|:
+
+  1. [=Emit an event=] with |session| and |body|.
+
+</div>
+
+#### The browsingContext.navigationCommitted Event #### {#event-browsingContext-navigationCommitted}
+
+<dl>
+   <dt>Event Type</dt>
+   <dd>
+      <pre class="cddl local-cddl">
+        browsingContext.NavigationCommitted = (
+         method: "browsingContext.navigationCommitted",
+         params: browsingContext.NavigationInfo
+        )
+      </pre>
+   </dd>
+</dl>
+
+<div algorithm>
+The [=remote end event trigger=] is the <dfn export>WebDriver BiDi navigation
+committed</dfn> steps given |navigable| and |navigation status|:
+
+1. Let |params| be the result of [=get the navigation info=] given |navigable|
+   and |navigation status|.
+
+1. Let |body| be a [=/map=] matching the
+    <code>browsingContext.NavigationCommitted</code> production, with the
+    <code>params</code> field set to |params|.
+
+1. Let |navigation id| be |navigation status|'s id.
+
+1. Let |related navigables| be a [=/set=] containing |navigable|.
+
+1. [=Resume=] with "<code>navigation committed</code>", |navigation id|, and |navigation status|.
+
+1. For each |session| in the [=set of sessions for which an event is enabled=]
+   given "<code>browsingContext.navigationCommitted</code>" and |related navigables|:
 
   1. [=Emit an event=] with |session| and |body|.
 
--- a/index.bs
+++ b/index.bs
@@ -5136,8 +5136,8 @@ aborted</dfn> steps given |navigable| and |navigation status|:
 </dl>
 
 <div algorithm>
-The [=remote end event trigger=] is the <dfn export>WebDriver BiDi navigation
-committed</dfn> steps given |navigable| and |navigation status|:
+The [=remote end event trigger=] is the <dfn export>WebDriver BiDi navigation committed</dfn> steps
+given |navigable| and |navigation status|:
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -5116,6 +5121,44 @@ aborted</dfn> steps given |navigable| and |navigation status|:
...
+         method: "browsingContext.navigationCommitted",
+         params: browsingContext.NavigationInfo
+        )
+      </pre>
+   </dd>
+</dl>
+
+<div algorithm>
+The [=remote end event trigger=] is the <dfn export>WebDriver BiDi navigation
+committed</dfn> steps given |navigable| and |navigation status|:
```
sadym-chromium, line 5140: nit: avoid line breaks in <dfn> allows for easier search in the source code

sadym-chromium (approved):

index.bs
OrKoN, line 5140:
```suggestion
The [=remote end event trigger=] is the <dfn export>WebDriver BiDi navigation committed</dfn> steps
given |navigable| and |navigation status|:
```

sadym-chromium (approved):

index.bs
```diff
@@ -3168,7 +3173,7 @@ To <dfn>get the navigation info</dfn>, given |navigable| and |navigation status|
 #### The browsingContext.ReadinessState Type #### {#type-browsingContext-ReadinessState}
 
 <pre class="cddl remote-cddl">
-browsingContext.ReadinessState = "none" / "interactive" / "complete"
+browsingContext.ReadinessState = "none" / "committed" / "interactive" / "complete"
```
whimboo, line 3176: So this means that we are fine for the `wait` argument of `browsingContext.navigate` to allow `committed` as well? Asking because it causes a side-effect.

OrKoN: Could you please clarify what you mean by the side-effect?

whimboo: So far we spoke about a committed event but not that we have another readiness state that could be used by other APIs to control the navigation flow like for `browsingContext.navigate`. In the current form this value for the `wait` argument wouldn't be denied but also not used in [await a navigation](https://w3c.github.io/webdriver-bidi/#await-a-navigation) because we don't have a check. So it's basically a no-op.

OrKoN: I believe the value is checked on line 3045 (part of the await navigation algorithm you linked) so it does not appear to be no-op. I think it's needed for internal consistency but we can discuss it separately.

whimboo: Ah indeed. Even through I opened the preview in a different tab I actually checked the public BiDi spec page. Sorry for that.

While investigating some more Playwright tests I saw the following:

> await page.goto(server.EMPTY_PAGE, { waitUntil: 'commit' })

Is that a condition that would help Puppeteer users as well that we should add to the navigate steps to return on the committed step? Or do it in a follow-up PR?

OrKoN: Yeah, in Puppeteer we always waited for commit by default and switched to waiting for interactive by default for tests to pass and handle errors but we would prefer to roll back to waiting for commit as it could be a little more robust in some situation. I think we do not expose it as an option on the API though but maybe we should.

> Is that a condition that would help Puppeteer users as well that we should add to the navigate steps to return on the committed step? Or do it in a follow-up PR?

I didn't get the question, sorry. I believe this PR already returns navigate steps on "committed" if it is configured so. We could extract the whole waitUntil part into a separate PR if it makes it easier.

index.bs
```diff
@@ -5116,6 +5121,44 @@ aborted</dfn> steps given |navigable| and |navigation status|:
...
+   and |navigation status|.
+
+1. Let |body| be a [=/map=] matching the
+    <code>browsingContext.NavigationCommitted</code> production, with the
+    <code>params</code> field set to |params|.
+
+1. Let |navigation id| be |navigation status|'s id.
+
+1. Let |related navigables| be a [=/set=] containing |navigable|.
+
```
whimboo, line 5152: Lets use the same order as the other navigation events:

```suggestion
1. Let |related navigables| be a [=/set=] containing |navigable|.

1. Let |navigation id| be |navigation status|'s id.

```

index.bs
whimboo: No that is fine. So lets keep it in. Thanks.

It would be great if you could actually write some WPT tests for the event and the wait argument given that it seems to be easy for you to add support for `committed` (CDP already supports it).

whimboo (commented):
What should we do with the `browsingContext.navigationAborted` event? Should we deprecate it? See https://github.com/w3c/webdriver-bidi/issues/788#issuecomment-2586998329

index.bs
OrKoN: Well, the current order is the same as the order of steps in navigationFailed.

OrKoN:
> What should we do with the `browsingContext.navigationAborted` event? Should we deprecate it? See [#788 (comment)](https://github.com/w3c/webdriver-bidi/issues/788#issuecomment-2586998329)

what is the specific reason for deprecating navigationAborted? I think we can discuss it separately.

index.bs
OrKoN: I filed https://github.com/GoogleChromeLabs/chromium-bidi/issues/2982 to write the tests as part of the implementation.

index.bs
whimboo: But different to the `domContentLoaded` and `load`. Would be nice to have it in the same order everywhere for navigation events.

whimboo:
> > What should we do with the `browsingContext.navigationAborted` event? Should we deprecate it? See [#788 (comment)](https://github.com/w3c/webdriver-bidi/issues/788#issuecomment-2586998329)
> 
> what is the specific reason for deprecating navigationAborted? I think we can discuss it separately.

See the details in the referenced comment. That's all what I got. Happy to move it to a follow-up.

index.bs
OrKoN: So I am not sure the order matters for anything but I'd suggest to open a follow-up editorial PR to update those if it is important for anything. I applied your suggestion nonetheless.

index.bs
jgraham: I tend to think we want this to continue to match the `document.readyState` property, but for the purposes of awaiting a navigation make "none" work like "committed", unless we have a use case where you really don't even care if the navigation started or not, but still need the return value of the command for some reason.

OrKoN: I guess if we are not concerned about breaking changes, I am in favor of replacing `none` to mean `committed`  (cc @sadym-chromium)

sadym-chromium: > I am in favor of replacing `none` to mean `committed`

YES!

OrKoN: Updated the PR, please take a look.

jgraham (approved):
The API change seem good. It would be nice to make sure we have clarity at least amongst the WebDriver folks on the HTML side of the change before landing this.

index.bs
whimboo: That's actually a great idea. I support that as well because it will reduce failures due to process switches (not in all cases like redirects but most of them). We potentially might break some users but lets see how severe this will be. Benefit for sure will be that we will know the target URL at that point and clearly can return a proper navigation info or error response.

whimboo (approved):
Thanks!

whimboo:
@OrKoN the HTML spec PR got merged. I assume this one is ready as well?

OrKoN:
Implementation tracking on our side https://github.com/GoogleChromeLabs/chromium-bidi/issues/2982

OrKoN:
If someone plans to look into WPT before we get to it during the implementation, please post on the issue.

whimboo:
Tests landed via https://github.com/web-platform-tests/wpt/pull/50413. 

@sadym-chromium are those complete? (I didn't check them yet). Thanks!

sadym-chromium:
> Tests landed via [web-platform-tests/wpt#50413](https://github.com/web-platform-tests/wpt/pull/50413).
> 
> @sadym-chromium are those complete? (I didn't check them yet). Thanks!

After the PR https://github.com/web-platform-tests/wpt/pull/50419 is merged, I believe the tests are complete. But of course they can be extended with more exotic cases.

sadym-chromium:
> @sadym-chromium are those complete? (I didn't check them yet). Thanks!

UPD: the tests in https://github.com/web-platform-tests/wpt/pull/50413 covers only the event, not the wait condition. I'm working on the navigation tests.

PR #857, opened 2025-01-14
Fix syntax error in browser.SetClientWindowStateParameters

### LINKED ISSUE #853: Syntax error in `browser.SetClientWindowStateParameters` definition?
If I'm understanding the CDDL spec correctly, and inferring the intended definition of `browser.SetClientWindowStateParameters` correctly, then [`browser.SetClientWindowStateParameters` is defined incorrectly](https://w3c.github.io/webdriver-bidi/#command-browser-setClientWindowState); it uses `//` when it meant to use `/`.

That is, `//` is just a weakly-binding `/`, allowing you to define groups inline without having to abstract them to types. The current definition of `browser.SetClientWindowStateParameters` is:

```
browser.SetClientWindowStateParameters = {
  clientWindow: browser.ClientWindow,
  browser.ClientWindowNamedState // browser.ClientWindowRectState
}
```

This is, afaict, equivalent to:

```
browser.SetClientWindowStateParameters = {
  (clientWindow: browser.ClientWindow, browser.ClientWindowNamedState) // 
  browser.ClientWindowRectState
}
```

That is, the first two items form a group, and the third item forms an alternative group. But I suspect what was intended was for `clientWindow` to always be present, and then have a choice between `browser.ClientWindowNamedState` and `browser.ClientWindowRectState`. To do that, you would need to have written:

```
browser.SetClientWindowStateParameters = {
  clientWindow: browser.ClientWindow,
  browser.ClientWindowNamedState / browser.ClientWindowRectState
}
```

tabatkins:
Fixes #853.

***
<a href="https://pr-preview.s3.amazonaws.com/tabatkins/webdriver-bidi/pull/857.html" title="Last updated on Jan 15, 2025, 7:22 AM UTC (9cb04e1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/857/e25f510...tabatkins:9cb04e1.html" title="Last updated on Jan 15, 2025, 7:22 AM UTC (9cb04e1)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2698,7 +2698,7 @@ dimensions of a [=client window=].
 
       browser.SetClientWindowStateParameters = {
         clientWindow: browser.ClientWindow,
-        browser.ClientWindowNamedState // browser.ClientWindowRectState
+        browser.ClientWindowNamedState / browser.ClientWindowRectState
       }
 
       browser.ClientWindowNamedState = (
--- a/index.bs
+++ b/index.bs
@@ -2698,7 +2698,7 @@ dimensions of a [=client window=].
 
       browser.SetClientWindowStateParameters = {
         clientWindow: browser.ClientWindow,
-        browser.ClientWindowNamedState / browser.ClientWindowRectState
+        (browser.ClientWindowNamedState // browser.ClientWindowRectState)
       }
 
       browser.ClientWindowNamedState = (


index.bs
```diff
@@ -2698,7 +2698,7 @@ dimensions of a [=client window=].
 
       browser.SetClientWindowStateParameters = {
         clientWindow: browser.ClientWindow,
-        browser.ClientWindowNamedState // browser.ClientWindowRectState
+        browser.ClientWindowNamedState / browser.ClientWindowRectState
```
OrKoN, line 2701:
```suggestion
        (browser.ClientWindowNamedState // browser.ClientWindowRectState)
```

(seeing the failing tests) should not it be the following? since `/` for choosing between types and `//` for choosing between groups?

tabatkins: Hm. I could be misreading the RFC, but as far as I can tell the only difference between `/` and `//` is the early- vs late-binding behavior; otherwise groups *are* types, right? Like, you'd only *use* `//` specifically to deal with groups, to avoid having to use `/` with parens around the groups. 

I see the parse failure, but I'm not certain that's not a bug in the parser you're using. (But I might be wrong.)

whimboo: Isn't it actually the following that we need?

```
browser.SetClientWindowStateParameters = {
  clientWindow: browser.ClientWindow,
  ( browser.ClientWindowNamedState // browser.ClientWindowRectState )
}

browser.ClientWindowNamedState = {
  state: "fullscreen" / "maximized" / "minimized"
}

browser.ClientWindowRectState = {
  state: "normal",
  ? width: js-uint,
  ? height: js-uint,
  ? x: js-int,
  ? y: js-int
}
```

Both `ClientWindowNamedState` and `ClientWindowRectState` are basically maps which would require curly braces (not rounded ones). And when listing them within `SetClientWindowStateParameters` we actually need rounded braces to correctly express the alternatives. Or do I see it wrongly?

tabatkins: Yeah, just adding parens definitely works as well, regardless of how we read the spec about `/` vs `//`.

whimboo: Would you mind fixing the `ClientWindowNamedState` and `ClientWindowRectState` definitions as well as mentioned in my last comment?

whimboo: Whereby it looks like that other commands have a similar issue and that we are not consistent in the usage of `(` and `{`. So lets keep it out of this PR.

whimboo (approved):

Issue #849, opened 2025-01-10
Add support for "userContexts" argument to "script.addPreloadScript" command


lutien:
Similar to other commands (e.g. session.subscribe) we should add a possibility to add preload scripts to all existing and future browsing contexts belonging to certain `userContexts`.

PR #862, opened 2025-01-16
Support preload scripts per user context

OrKoN:
Closes https://github.com/w3c/webdriver-bidi/issues/849

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/862.html" title="Last updated on Jan 17, 2025, 10:34 AM UTC (f8311a9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/862/374e7ba...f8311a9.html" title="Last updated on Jan 17, 2025, 10:34 AM UTC (f8311a9)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -9784,6 +9784,10 @@ script=].
 <div algorithm="remote end steps for script.addPreloadScript">
 The [=remote end steps=] given |session| and |command parameters| are:
 
+1. If |command parameters| [=map/contains=] "<code>userContexts</code>"
+   and |command parameters| [=map/contains=] "<code>contexts</code>",
+   return [=error=] with [=error code=] [=invalid argument=].
+
 1. Let |function declaration| be the <code>functionDeclaration</code> field of |command
    parameters|.
 
--- a/index.bs
+++ b/index.bs
@@ -9790,7 +9790,7 @@ The [=remote end steps=] given |session| and |command parameters| are:
 1. Let |arguments| be the <code>arguments</code> field of |command
    parameters| if present, or an empty [=/list=] otherwise.
 
-1. Let |user contexts| be [=set/create|create a set=] with |command parameters|[<code>userContexts</code>]
+1. Let |user contexts| be [=set/create|create a set=] with |command parameters|["<code>userContexts</code>"]
    if |command parameters| [=map/contains=] <code>userContexts</code> and an empty [=/set=] otherwise.
 
 1. For each |user context id| of |user contexts|:
--- a/index.bs
+++ b/index.bs
@@ -7832,7 +7832,9 @@ To <dfn export>run WebDriver BiDi preload scripts</dfn> given |environment setti
 
     1. If |preload script|'s <code>contexts</code> is not null:
 
-       1. If |preload script|'s <code>contexts</code> does not [=list/contain=] |user context id|,
+       1. Let |navigable id| be |navigable|’s [=navigable/top-level traversable=]'s id.
+
+       1. If |preload script|'s <code>contexts</code> does not [=list/contain=] |navigable id|,
           [=continue=].
 
     1. If |preload script|'s <code>sandbox</code> is not null, let |realm| be [=get
--- a/index.bs
+++ b/index.bs
@@ -7799,8 +7799,9 @@ TODO: Extend this to scripts in other kinds of realms.
 A [=BiDi session=] has a <dfn>preload script map</dfn> which is a [=/map=] in
 which the keys are [[!RFC9562|UUID]]s, and the values are [=structs=] with an <a
 for=struct>item</a> named <code>function declaration</code>, which is a string,
-<code>arguments</code>, <code>contexts</code>, which is a list or null, and an item named <code>sandbox</code> which is a string
-or null.
+<code>arguments</code>, <code>contexts</code>, which is a list or null,
+an item named <code>sandbox</code>, which is a string or null,
+and an item named <code>applyToOpeningContexts</code>, which is a boolean.
 
 Note: If executing a [=preload script=] fails, either due to a syntax error, or
 a runtime exception, an [[ECMAScript]] exception is reported in the realm in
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -9777,6 +9788,15 @@ The [=remote end steps=] given |session| and |command parameters| are:
 1. Let |arguments| be the <code>arguments</code> field of |command
    parameters| if present, or an empty [=/list=] otherwise.
 
+1. Let |user contexts| be [=set/create=] with |command parameters|[<code>userContexts</code>]
```
juliandescottes, line 9791: nitpick: the rest of the spec seems to use `[=set/create|create a set=]` for set creation
```suggestion
1. Let |user contexts| be [=set/create|create a set=] with |command parameters|[<code>userContexts</code>]
```

juliandescottes, line 9791:
```suggestion
1. Let |user contexts| be [=set/create=] with |command parameters|["<code>userContexts</code>"]
```

Missing quotes here, although I see we missed it in a few other spots in the spec

index.bs
```diff
@@ -9777,6 +9788,15 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+   if |command parameters| [=map/contains=] [<code>userContexts</code>] and an empty [=/set=] otherwise.
```
juliandescottes, line 9792: Square brackets not necessary. Again, sometimes we seem to use quotes, sometimes not. Feel free to add or not
```suggestion
   if |command parameters| [=map/contains=] <code>userContexts</code> and an empty [=/set=] otherwise.
```

index.bs
```diff
@@ -7797,10 +7797,12 @@ before any author-defined script have run.
...
-which the keys are [[!RFC9562|UUID]]s, and the values are [=structs=] with an <a
-for=struct>item</a> named <code>function declaration</code>, which is a string,
-<code>arguments</code>, <code>contexts</code>, which is a list or null, and an item named <code>sandbox</code> which is a string
-or null.
+which the keys are [[!RFC9562|UUID]]s, and the values are [=structs=] with
+an [=struct/item=] named <code>function declaration</code>, which is a string,
+an [=struct/item=] named <code>arguments</code>, which is a list,
+an [=struct/item=] named <code>contexts</code>, which is a list or null,
+an [=struct/item=] named <code>sandbox</code>, which is a string or null,
+and an [=struct/item=] named <code>user contexts</code>, which is a [=/set=], initially empty.
```
juliandescottes, line 7805: Not sure it makes sense to state that it's initially empty? For each preload script, the `user contexts` set will be assigned the value provided in the command. It's not like it starts empty and can be filled afterwards? We could say it's potentially empty, but not sure that's worth calling out?

index.bs
```diff
@@ -7814,16 +7816,24 @@ To <dfn export>run WebDriver BiDi preload scripts</dfn> given |environment setti
...
+
+       1. If |preload script|'s <code>user contexts</code> does not [=list/contain=] |user context id|,
+          [=continue=].
 
-       1. Let |navigable id| be |navigable|’s [=navigable/top-level traversable=]'s id.
+    1. If |preload script|'s <code>contexts</code> is not null:
 
-       1. If |preload script|'s <code>contexts</code> does not [=list/contains|contain=] |navigable id|, continue.
+       1. If |preload script|'s <code>contexts</code> does not [=list/contain=] |user context id|,
+          [=continue=].
```
juliandescottes, line 7836: This looks wrong. We still need to get the `navigable id` and check if it is in contexts.

juliandescottes (changes requested):
Thanks, looks great, some nits and minor issues to fix.

index.bs
OrKoN: Indeed!

index.bs
OrKoN, line 7805:
```suggestion
and an [=struct/item=] named <code>user contexts</code>, which is a [=/set=].
```

OrKoN: No preferences here, updated.

index.bs
```diff
@@ -7814,16 +7816,26 @@ To <dfn export>run WebDriver BiDi preload scripts</dfn> given |environment setti
...
+1. Let |user context id| be |user context|'s [=user context id=].
+
 1. For each |session| in [=active BiDi sessions=]:
 
   1. For each |preload script| in |session|'s [=preload script map=]'s
      [=values=]:
 
+    1. If |preload script|'s <code>user contexts</code>'s [=set/size=] is not zero:
+
+       1. If |preload script|'s <code>user contexts</code> does not [=list/contain=] |user context id|,
```
jgraham, line 7830: So this is giving us "and" semantics for preload scripts i.e. you have to match both a user context id and a context id, if both are provided. For subscriptions, I think we have "or" semantics i.e. matching one or the other is fine. Generally I think I'd prefer either "or" semantics or we only allow the commands to provide either user contexts or contexts, but not both.

OrKoN: Indeed, changed to the "or" semantics, I think the "and" logic does not make sense here, same as in events.

sadym-chromium: Should we make `contexts` and `userContexts` mutually exclusive?

sadym-chromium: This would simplify understanding API edge cases.

OrKoN: They are mutually exclusive in the latest patch.

sadym-chromium: UPD: oh, they already are, so nevermind.

index.bs
```diff
@@ -9791,17 +9810,29 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
 
+1. Otherwise, if |command parameters| [=map/contains=] <code>userContexts</code>:
+
+  1. Set |user contexts| to [=set/create|create a set=] with |command parameters|["<code>userContexts</code>"].
+
+  1. For each |user context id| of |user contexts|:
+
+    1. Set |user context| to [=get user context=] with |user context id|.
+
+    1. If |user context| is null, return [=error=] with [=error code=] [=no such user context=].
```
sadym-chromium, line 9821: nit: maybe iterate through the `|command parameters|["<code>userContexts</code>"]`, assert valid values and append them to existing map `|user contexts|` to increase readability and align with `contexts` logic one step above?

sadym-chromium (approved):

jgraham (approved):

lutien:
@OrKoN, I'm planning to work on the tests for this feature. Let me know if you have any objections.

OrKoN:
I already started working on some tests https://github.com/web-platform-tests/wpt/pull/50290

OrKoN:
oh right, sorry, I had user context events on my mind.

PR #863, opened 2025-01-21
Prevent navigable being overridden by parent navigable

whimboo:
Within [`get the navigable info`](https://w3c.github.io/webdriver-bidi/#get-the-navigable-info) we are overriding the passed in `navigable` when trying to get the navigable id of the parent navigable. Further steps in this method will then base on the parent instead of the current navigable.


whimboo:
Interesting that the build and deploy tasks fail now. It is caused by code that landed long long ago. I'm going to open another PR with those fixes.

PR #867, opened 2025-01-22
Always handle subscription IDs in session.unsubscribe if provided

OrKoN:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/867.html" title="Last updated on Jan 24, 2025, 8:01 AM UTC (b541c46)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/867/d11969b...b541c46.html" title="Last updated on Jan 24, 2025, 8:01 AM UTC (b541c46)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2104,6 +2104,8 @@ The [=remote end steps=] with |session| and |command parameters| are:
 
 1. If |command parameters| does not [=map/contain=] "<code>subscriptions</code>":
 
+   Note: The condition implies that |command parameters| is matching the session.UnsubscribeByAttributesRequest production.
+
    1. Let |event names| be an empty [=/set=].
 
    1. For each entry |name| in |command parameters|["<code>events</code>"],
--- a/index.bs
+++ b/index.bs
@@ -2102,7 +2102,7 @@ and will be removed in the future versions.
 <div algorithm="remote end steps for session.unsubscribe">
 The [=remote end steps=] with |session| and |command parameters| are:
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2102,7 +2102,7 @@ and will be removed in the future versions.
 <div algorithm="remote end steps for session.unsubscribe">
 The [=remote end steps=] with |session| and |command parameters| are:
 
-1. If |command parameters| matches the <code>session.UnsubscribeByAttributesRequest</code> production:
+1. If |command parameters| does not [=map/contain=] "<code>subscriptions</code>":
```
juliandescottes, line 2105: Quick question, here it's implied that `command parameters` already matches either `session.UnsubscribeByAttributesRequest` or `session.UnsubscribeByIDRequest`, and if `subscriptions` is not in `parameters`, then it has to match `session.UnsubscribeByAttributesRequest`. I think it makes sense overall but it's a bit hard to follow and might be worth adding a NOTE if we don't want to explicitly call out `and matches session.UnsubscribeByAttributesRequest` (which I agree is redundant)

juliandescottes (commented):
Overall looks good to me, but it makes the algorithm a bit harder to follow (see comment). Would still like to make sure that @jgraham is fine with this as well.

jgraham (approved):

index.bs
OrKoN: Added a note.

PR #729, opened 2024-06-13
Change the spec to use HTML's navigables instead of browsing contexts

OrKoN:
This PR updates the specification to define interactions with browsing contexts in terms of HTML spec's navigables without changing the API  (which keeps using "browsingContext" and "context" to refer to navigables).

Closes https://github.com/w3c/webdriver-bidi/issues/91

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/729.html" title="Last updated on Jul 17, 2024, 10:25 AM UTC (a8145a6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/729/0fa2d99...a8145a6.html" title="Last updated on Jul 17, 2024, 10:25 AM UTC (a8145a6)">Diff</a>

PR #876, opened 2025-02-11
Add support for "userContexts" argument to "browsingContext.setViewport" command.

### LINKED ISSUE #851: Add support for "userContexts" argument to "browsingContext.setViewport" command
Similar to other commands (e.g. session.subscribe) we should add a possibility to set viewport settings to all existing and future browsing contexts belonging to certain `userContexts`.

lutien:
A proposal to add "userContexts" argument to "browsingContext.setViewport" command.

Closes #851

***
<a href="https://pr-preview.s3.amazonaws.com/lutien/webdriver-bidi/pull/876.html" title="Last updated on Mar 4, 2025, 11:15 AM UTC (2fd7fdd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/876/198ddae...lutien:2fd7fdd.html" title="Last updated on Mar 4, 2025, 11:15 AM UTC (2fd7fdd)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4712,7 +4712,7 @@ before the [=run WebDriver BiDi preload scripts=] algorithm is invoked:
 
 1. If [=viewport overrides map=] [=map/contains=] |user context|:
 
-   1. If the <code>viewport</code> field of |[=viewport overrides map=][|user context|] is present:
+   1. If the <code>viewport</code> field of [=viewport overrides map=][|user context|] is not null:
 
       1. [=Set viewport=] with |navigable| and [=viewport overrides map=][|user context|]["<code>viewport</code>"].
 
--- a/index.bs
+++ b/index.bs
@@ -4667,7 +4667,7 @@ To <dfn>set device pixel ratio override</dfn> given |navigable| and |device pixe
 
 1. If |device pixel ratio| is not null:
 
-   1. When the [=select an image source from a source set=] are run, act as if
+   1. When the [=select an image source from a source set=] steps are run, act as if
       the implementation's pixel density was set to |device pixel ratio| when selecting an image.
 
    1. For the purposes of the [=resolution media feature=], act as if
@@ -4785,6 +4785,8 @@ The [=remote end steps=] with |command parameters| are:
 
          1. [=list/Append=] |top-level traversable| to |navigables|.
 
+1. Otherwise, return [=error=] with [=error code=] [=invalid argument=].
+
 1. If |command parameters| [=map/contains=] the <code>viewport</code> field:
 
    1. Let |viewport| be the |command parameters|["<code>viewport</code>"].
@@ -4793,7 +4795,7 @@ The [=remote end steps=] with |command parameters| are:
 
       1. [=Set viewport=] with |navigable| and |viewport|.
 
-1. Run the [[cssom-view-1#resizing-viewports]] steps.
+      1. Run the [[cssom-view-1#resizing-viewports]] steps with |navigable|'s [=active document=].
 
 1. If |command parameters| [=map/contains=] the <code>devicePixelRatio</code> field:
 
--- a/index.bs
+++ b/index.bs
@@ -2966,6 +2966,13 @@ between [=navigables=] and device pixel ratio overrides. It is initially empty.
 Note: this map is not cleared when the final session ends i.e. device pixel
 ratio overrides outlive any WebDriver session.
 
+A [=remote end=] has a <dfn>viewport overrides map</dfn> which is a weak map
+between [=user contexts=] and [=structs=], with an [=struct/item=] named <code>viewport</code>
+and an [=struct/item=] named <code>devicePixelRatio</code>, which is a float or null.
+An [=struct/item=] <code>viewport</code> is a [=struct=] or null, with an [=struct/item=]
+named <code>height</code>, which is an integer, and an [=struct/item=] named <code>width</code>,
+which is an integer.
+
 ### Types ### {#module-browsingcontext-types}
 
 #### The browsingContext.BrowsingContext Type #### {#type-browsingContext-Browsingcontext}
@@ -4629,9 +4636,10 @@ The <dfn export for=commands>browsingContext.setViewport</dfn> command modifies 
       )
 
       browsingContext.SetViewportParameters = {
-        context: browsingContext.BrowsingContext,
+        ? context: browsingContext.BrowsingContext,
         ? viewport: browsingContext.Viewport / null,
         ? devicePixelRatio: (float .gt 0.0) / null,
+        ? userContexts: [+browser.UserContext],
       }
 
       browsingContext.Viewport = {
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -4648,71 +4656,154 @@ The <dfn export for=commands>browsingContext.setViewport</dfn> command modifies
...
+   <code>height</code> of |viewport| in CSS pixels.
+
+1. Otherwise, set the |navigable|'s [=layout viewport=] to the
+   implementation-defined default.
+
+</div>
+
+<div algorithm="set viewport when a top-level traversable is created">
+
+When the [=set up a window environment settings object=]
```
lutien, line 4708: It's probably a wrong place to hook on, but since `WebDriver BiDi navigable created` should happen when everything is set up, and likely we have to set up viewport even before preload scripts run we need something else here, and will have to add something to html spec. So this is more of placeholder for now.

index.bs
```diff
@@ -4648,71 +4656,154 @@ The <dfn export for=commands>browsingContext.setViewport</dfn> command modifies
    </dd>
 </dl>
 
+<div algorithm>
+To <dfn>set device pixel ratio override</dfn> given |navigable| and |device pixel ratio|:
+
+1. Let |navigable| be the [=/navigable=] whose [=navigable/active document=] is
```
jgraham, line 4662: Is there a case where this isn't a no-op?

index.bs
jgraham: Do we actually need to do this once per document, or only really once per navigable? It seems like ideally we run this once per top-level traversable, since we seem to skip over everything else, and then you can assert that the input navigable is a top-level traversable, and skip a bunch of checks below.

I think for now it would be better to have a sketch of under what conditions we want to run these steps (maybe "after creating a document in a new top-level traversable, before running the 'Run WebDriver BiDi preload scripts' steps") rather than being more specific in a way that's misleading.

index.bs
```diff
@@ -4648,71 +4656,154 @@ The <dfn export for=commands>browsingContext.setViewport</dfn> command modifies
...
 
-   1. Let |viewport| be the |command parameters|["<code>viewport</code>"].
+   1. If |navigable| is not a [=/top-level traversable=], return [=error=] with
+      [=error code=] [=invalid argument=].
 
-   1. If |viewport| is not null, set the width of |navigable|'s [=layout
-      viewport=] to be the <code>width</code> of |viewport| in CSS pixels and
-      set the height of the |navigable|'s [=layout viewport=] to be the
-      <code>height</code> of |viewport| in CSS pixels.
+   1. If the implementation is unable to adjust the |navigable|'s [=layout viewport=]
```
jgraham, line 4756: We have this here; what happens if it fails when we try to adjust it in the user context case?

index.bs
```diff
@@ -4648,71 +4656,154 @@ The <dfn export for=commands>browsingContext.setViewport</dfn> command modifies
...
 
-1. If |command parameters| [=map/contains=] the <code>devicePixelRatio</code> field:
+   1. For each |user context id| of |command parameters|["<code>userContexts</code>"]:
 
-   1. Let |device pixel ratio| be the |command
-      parameters|["<code>devicePixelRatio</code>"].
+      1. Set |user context| to [=get user context=] with |user context id|.
 
-   1. For the |navigable| and all [=descendant navigables=]:
+      1. If |user context| is null, return [=error=] with [=error code=] [=no such user context=].
```
jgraham, line 4768: This failure inside the loop means that we'd partially apply the command to any user contexts before the one that doesn't exist; we should probably fail if any don't exist.

jgraham (changes requested):

lutien:
The build fails, but it doesn't look that it comes from my changes. Locally, I can reproduce it on main branch as well.

index.bs
```diff
@@ -4648,71 +4659,152 @@ The <dfn export for=commands>browsingContext.setViewport</dfn> command modifies
...
+1. Let |navigable| be the [=/navigable=] whose [=navigable/active document=] is
+   |navigable|'s [=navigable/active document=].
 
-1. Let |navigable id| be the value of the <code>context</code> field of |command
-   parameters|.
+1. If |device pixel ratio| is not null:
 
-1. Let |navigable| be the result of [=trying=] to [=get a navigable=] with
-   |navigable id|.
+   1. When the [=select an image source from a source set=] are run, act as if
```
juliandescottes, line 4670:
```suggestion
   1. When the [=select an image source from a source set=] steps are run, act as if
```

index.bs
```diff
@@ -4648,71 +4659,152 @@ The <dfn export for=commands>browsingContext.setViewport</dfn> command modifies
...
+The [=remote end steps=] with |command parameters| are:
+
+1. If the implementation is unable to adjust the [=layout viewport=]
    parameters with the given |command parameters| for any reason, return
    [=error=] with [=error code=] [=unsupported operation=].
 
-1. If |command parameters| [=map/contains=] the <code>viewport</code> field:
+1. If |command parameters| [=map/contains=] "<code>userContexts</code>"
+   and |command parameters| [=map/contains=] "<code>context</code>",
+   return [=error=] with [=error code=] [=invalid argument=].
```
juliandescottes, line 4740: We should probably also raise an error if both parameters are missing? It will effectively be a noop and won't crash or anything, but the user should know about it. Otherwise they might assume it just applied setViewport to all navigables?

index.bs
```diff
@@ -4648,71 +4659,152 @@ The <dfn export for=commands>browsingContext.setViewport</dfn> command modifies
...
+
+1. If |command parameters| [=map/contains=] the <code>viewport</code> field:
+
+   1. Let |viewport| be the |command parameters|["<code>viewport</code>"].
+
+   1. For each |navigable| of |navigables|:
+
+      1. [=Set viewport=] with |navigable| and |viewport|.
+
+1. Run the [[cssom-view-1#resizing-viewports]] steps.
```
juliandescottes, line 4796: (not related to your PR, but those steps seem to require a document to run against?)

index.bs
lutien: Yeah, that's true, added ✅

index.bs
```diff
@@ -4648,71 +4659,154 @@ The <dfn export for=commands>browsingContext.setViewport</dfn> command modifies
    </dd>
 </dl>
 
-<div algorithm="remote end steps for browsingContext.setViewport">
+<div algorithm>
```
OrKoN, line 4662: I think this helper algorithms could be moved outside of the command section so that remote end steps follow the CDDL type definitions for the command.

index.bs
```diff
@@ -4648,71 +4659,154 @@ The <dfn export for=commands>browsingContext.setViewport</dfn> command modifies
...
+   return [=error=] with [=error code=] [=invalid argument=].
 
-   1. Let |viewport| be the |command parameters|["<code>viewport</code>"].
+1. Let |navigables| be a [=/set=].
 
-   1. If |viewport| is not null, set the width of |navigable|'s [=layout
-      viewport=] to be the <code>width</code> of |viewport| in CSS pixels and
-      set the height of the |navigable|'s [=layout viewport=] to be the
-      <code>height</code> of |viewport| in CSS pixels.
+1. If the <code>context</code> field of |command parameters| is present:
```
OrKoN, line 4744: I think we could use `[=map/contains=]` instead of "is present"

index.bs
```diff
@@ -4648,71 +4659,154 @@ The <dfn export for=commands>browsingContext.setViewport</dfn> command modifies
...
+      1. If |user context| is null, return [=error=] with [=error code=] [=no such user context=].
 
-         1. [=map/Set=] [=device pixel ratio overrides=][|navigable|] to |device pixel ratio|.
+      1. [=set/Append=] |user context| to |user contexts|.
 
-            Note: This will take an effect because of the patch of [[#patchs-determine-the-device-pixel-ratio]].
+   1. For each |user context| of |user contexts|:
 
-      1. Otherwise:
+      1. [=map/Set=] [=viewport overrides map=][|user context|] to a new [=/map=].
```
OrKoN, line 4771: isn't `[=viewport overrides map=][|user context|]` defined as a struct?

index.bs
```diff
@@ -4648,71 +4659,154 @@ The <dfn export for=commands>browsingContext.setViewport</dfn> command modifies
...
+
+         1. Set |viewport overrides map|[|user context|]["<code>viewport</code>"]
+            to |command parameters|["<code>viewport</code>"].
+
+      1. If |command parameters| [=map/contains=] the <code>devicePixelRatio</code> field:
+
+         1. Set |viewport overrides map|[|user context|]["<code>devicePixelRatio</code>"]
+            to |command parameters|["<code>devicePixelRatio</code>"].
+
+      1. [=list/For each=] |top-level traversable| in the list of all [=/top-level traversables=]
```
OrKoN, line 4783:
```suggestion
      1. [=list/For each=] |top-level traversable| of the list of all [=/top-level traversables=]
```

I think the infra standard recommends `for each |item| of |list|`.

index.bs
```diff
@@ -4648,71 +4659,154 @@ The <dfn export for=commands>browsingContext.setViewport</dfn> command modifies
...
+<div algorithm="set viewport when a top-level traversable is created">
+
+After creating a document in a new [=navigable/top-level traversable=] |navigable| and
+before the [=run WebDriver BiDi preload scripts=] algorithm is invoked:
+
+1. Let |user context| be |navigable|'s [=associated user context=].
+
+1. If [=viewport overrides map=] [=map/contains=] |user context|:
+
+   1. If |viewport overrides map|[|user context|] [=map/contains=] the <code>viewport</code> field:
```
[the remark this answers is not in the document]
lutien: Yeah, I guess that's probably true. I've updated to just check if a field is present.

index.bs
lutien: Fixed ✅

index.bs
lutien: Interestingly, I just copied and pasted what we have already to extract it in the new algorithm. And the way it's written now is the result of [refactoring to use navigable everywhere](https://github.com/w3c/webdriver-bidi/pull/729). Before, the idea was to use a navigable whose active document matches the browsing context's active document. I think we probably don't need this anymore since we use the navigable directly. So I've removed this step. Let me know if I'm missing something.

index.bs
lutien: Updated ✅

index.bs
lutien: Yeah, I guess, but we also have the same with "browsingContext.locateNodes" or "script.callFunction"

index.bs
OrKoN: Thanks for updating! the refactoring was semi-mechanical and we might have missed things like this.

index.bs
```diff
@@ -2966,6 +2966,16 @@ between [=navigables=] and device pixel ratio overrides. It is initially empty.
 Note: this map is not cleared when the final session ends i.e. device pixel
 ratio overrides outlive any WebDriver session.
 
+A <dfn for="viewport-configuration">viewport dimensions</dfn> is a [=struct=] with an [=struct/item=] named
+<dfn attribute for="viewport-dimensions">height</dfn> which is an integer and
+a [=struct/item=] named <dfn attribute for="viewport-dimensions">width</dfn> which is an integer.
+
+A <dfn>viewport configuration</dfn> is a [=struct=] with an [=struct/item=] named
```
OrKoN, line 2973: So I am still not sure if a struct is suitable here and I am not sure if we can check if a field is present in a struct. 

https://infra.spec.whatwg.org/#structs says that each item needs to be of a defined type which I interpret as it cannot be null and needs to have a value of a given type. WDYT? cc @jgraham

jgraham: Nullable values are definitely allowed in structs (HTML uses them, for example). I think infra is just saying you're supposed to specify what the type is when you define the struct, but the equivalent of `Option<T>` (or `T | null` if that's your preferred syntax) is fine :)

OrKoN: So to check if there is a value we should assume the struct item is always there and check if it is null?  That SGTM.

index.bs
```diff
@@ -4648,71 +4659,151 @@ The <dfn export for=commands>browsingContext.setViewport</dfn> command modifies
...
+<div algorithm="set viewport when a top-level traversable is created">
+
+After creating a document in a new [=navigable/top-level traversable=] |navigable| and
+before the [=run WebDriver BiDi preload scripts=] algorithm is invoked:
+
+1. Let |user context| be |navigable|'s [=associated user context=].
+
+1. If [=viewport overrides map=] [=map/contains=] |user context|:
+
+   1. If the <code>viewport</code> field of |[=viewport overrides map=][|user context|] is present:
```
OrKoN, line 4715:
```suggestion
   1. If the <code>viewport</code> field of [=viewport overrides map=][|user context|] is not null:
```

index.bs
```diff
@@ -4648,71 +4659,151 @@ The <dfn export for=commands>browsingContext.setViewport</dfn> command modifies
...
 
-1. If the implementation is unable to adjust the |navigable|'s [=layout viewport=]
+      1. [=Set viewport=] with |navigable| and [=viewport overrides map=][|user context|]["<code>viewport</code>"].
+
+   1. If the <code>devicePixelRatio</code> field of |[=viewport overrides map=][|user context|] is present:
```
OrKoN, line 4719:
```suggestion
   1. If the <code>devicePixelRatio</code> field of [=viewport overrides map=][|user context|] is not null:
```

OrKoN (approved):
LGTM (cc @sadym-chromium)

index.bs
```diff
@@ -4648,71 +4659,151 @@ The <dfn export for=commands>browsingContext.setViewport</dfn> command modifies
...
-<div algorithm="remote end steps for browsingContext.setViewport">
+<div algorithm>
+To <dfn>set device pixel ratio override</dfn> given |navigable| and |device pixel ratio|:
 
-The [=remote end steps=] with |command parameters| are:
+1. If |device pixel ratio| is not null:
 
-1. Let |navigable id| be the value of the <code>context</code> field of |command
-   parameters|.
+   1. When the [=select an image source from a source set=] steps are run, act as if
```
jgraham, line 4667: Should this say something about what it applies to e.g. "for [=documents=] loaded in |navigable|"?

index.bs
```diff
@@ -4648,71 +4659,151 @@ The <dfn export for=commands>browsingContext.setViewport</dfn> command modifies
...
 
-            Note: This will take an effect because of the patch of [[#patchs-determine-the-device-pixel-ratio]].
+   1. For each |user context| of |user contexts|:
 
-      1. Otherwise:
+      1. [=map/Set=] [=viewport overrides map=][|user context|] to a struct.
 
-         1. When the [=select an image source from a source set=] steps are run, use the implementation's default behavior,
-            without any changes made by previous invocations of these steps.
+      1. If |command parameters| [=map/contains=] the <code>viewport</code> field:
```
jgraham, line 4770:
```suggestion
      1. If |command parameters| [=map/contains=] "<code>viewport</code>":
```

and similarly for the others

index.bs
```diff
@@ -4648,71 +4659,151 @@ The <dfn export for=commands>browsingContext.setViewport</dfn> command modifies
...
+1. If [=viewport overrides map=] [=map/contains=] |user context|:
+
+   1. If the <code>viewport</code> field of [=viewport overrides map=][|user context|] is not null:
 
-1. If the implementation is unable to adjust the |navigable|'s [=layout viewport=]
+      1. [=Set viewport=] with |navigable| and [=viewport overrides map=][|user context|]["<code>viewport</code>"].
+
+   1. If the <code>devicePixelRatio</code> field of [=viewport overrides map=][|user context|] is not null:
+
+      1. For the |navigable| and all [=descendant navigables=]:
```
sadym-chromium, line 4721: I guess the new descendant navigables created after the top level navigable is created will not have the "devicePixelRatio" updated.

index.bs
lutien: yeah, I guess it applies to the second step is well, so I've added before to work for both

index.bs
lutien: Yeah, James also pointed it out, I've updated the algorithm to apply to all kind of navigables, so it should hopefully cover it

jgraham (approved):
I think there might still be some lingering issues with bfcache (since in that case we want the updated setting to apply even to the existing document), but since we need to move this to HTML anyway, I think this is OK for now.

index.bs
```diff
@@ -4648,71 +4659,156 @@ The <dfn export for=commands>browsingContext.setViewport</dfn> command modifies
...
+   viewport=] to be the <code>width</code> of |viewport| in CSS pixels and
+   set the height of the |navigable|'s [=layout viewport=] to be the
+   <code>height</code> of |viewport| in CSS pixels.
+
+1. Otherwise, set the |navigable|'s [=layout viewport=] to the
+   implementation-defined default.
+
+</div>
+
+<div algorithm="set viewport when a top-level traversable is created">
```
sadym-chromium, line 4710:
```suggestion
<div algorithm="set viewport when a navigable is created">
```

index.bs
```diff
@@ -4648,71 +4659,156 @@ The <dfn export for=commands>browsingContext.setViewport</dfn> command modifies
...
+
+After creating a document in a new [=/navigable=]|navigable| and
```
sadym-chromium, line 4712:
```suggestion
After creating a document in a new [=/navigable=] |navigable| and
```

index.bs
```diff
@@ -4648,71 +4659,156 @@ The <dfn export for=commands>browsingContext.setViewport</dfn> command modifies
...
+   implementation-defined default.
+
+</div>
+
+<div algorithm="set viewport when a navigable is created">
+
+After creating a document in a new [=/navigable=] |navigable| and
+before the [=run WebDriver BiDi preload scripts=] algorithm is invoked:
+
+TODO: Move it as a hook in the html spec instead.
```
sadym-chromium, line 4715: We already have a "WebDriver BiDi navigable created" hook. I believe it's the right moment to set the viewports, and we don't need an additional hook.

lutien: But at that moment we don't have (guaranteed) a document created, which we need to apply viewport settings to

sadym-chromium: indeed

sadym-chromium (approved):

lutien:
Thanks everybody for the feedback! 
I'm going to merge it now

PR #888, opened 2025-03-12
Add "emulation.setGeolocationOverride" command

### LINKED ISSUE #343: Emulation: Support geolocation override
One request often heard from users is that they want to simulate a browser at a specific geolocation for their sites. WebDriver should provide a mechanism to facilitate this testing.

lutien:
Closes #343

PR in the geolocation spec: https://github.com/w3c/geolocation/pull/170

***
<a href="https://pr-preview.s3.amazonaws.com/lutien/webdriver-bidi/pull/888.html" title="Last updated on Mar 20, 2025, 4:24 PM UTC (3629b26)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/888/3a04347...lutien:3629b26.html" title="Last updated on Mar 20, 2025, 4:24 PM UTC (3629b26)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5629,6 +5629,10 @@ The <dfn export for=commands>emulation.setGeolocationOverride</dfn> command modi
 
 The [=remote end steps=] with |command parameters| are:
 
+1. If |command parameters| [=map/contains=] "<code>altitudeAccuracy</code>"
+   and |command parameters| doesn't [=map/contain=] "<code>altitude</code>",
+   return [=error=] with [=error code=] [=invalid argument=].
+
 1. If |command parameters| [=map/contains=] "<code>userContexts</code>"
    and |command parameters| [=map/contains=] "<code>context</code>",
    return [=error=] with [=error code=] [=invalid argument=].
--- a/index.bs
+++ b/index.bs
@@ -5575,12 +5575,12 @@ EmulationCommand = (
 
 A <dfn>geolocation override</dfn> is a [=struct=] with an [=struct/item=] named
 <dfn attribute for="geolocation override">latitude</dfn> which is a float,
-a [=struct/item=] named <dfn attribute for="viewport-dimensions">longitude</dfn> which is a float,
-a [=struct/item=] named <dfn attribute for="viewport-dimensions">accuracy</dfn> which is a float,
-a [=struct/item=] named <dfn attribute for="viewport-dimensions">altitude</dfn> which is a float or null,
-a [=struct/item=] named <dfn attribute for="viewport-dimensions">altitudeAccuracy</dfn> which is a float or null,
-a [=struct/item=] named <dfn attribute for="viewport-dimensions">heading</dfn> which is a float or null,
-a [=struct/item=] named <dfn attribute for="viewport-dimensions">speed</dfn> which is a float or null.
+a [=struct/item=] named <dfn attribute for="geolocation override">longitude</dfn> which is a float,
+a [=struct/item=] named <dfn attribute for="geolocation override">accuracy</dfn> which is a float,
+a [=struct/item=] named <dfn attribute for="geolocation override">altitude</dfn> which is a float or null,
+a [=struct/item=] named <dfn attribute for="geolocation override">altitudeAccuracy</dfn> which is a float or null,
+a [=struct/item=] named <dfn attribute for="geolocation override">heading</dfn> which is a float or null,
+a [=struct/item=] named <dfn attribute for="geolocation override">speed</dfn> which is a float or null.
 
 A [=remote end=] has a <dfn>geolocation overrides map</dfn> which is a weak map between [=user contexts=] and [=geolocation override=].
 
--- a/index.bs
+++ b/index.bs
@@ -5574,8 +5574,7 @@ EmulationCommand = (
 </pre>
 
 A <dfn>geolocation override</dfn> is a [=struct=] with:
-* [=struct/item=] named
-<dfn attribute for="geolocation override">latitude</dfn> which is a float;
+* [=struct/item=] named <dfn attribute for="geolocation override">latitude</dfn> which is a float;
 * [=struct/item=] named <dfn attribute for="geolocation override">longitude</dfn> which is a float;
 * [=struct/item=] named <dfn attribute for="geolocation override">accuracy</dfn> which is a float;
 * [=struct/item=] named <dfn attribute for="geolocation override">altitude</dfn> which is a float or null;
--- a/index.bs
+++ b/index.bs
@@ -180,6 +180,9 @@ spec: GEOMETRY; urlPrefix: https://drafts.fxtf.org/geometry/
     text: y coordinate; url: rectangle-y-coordinate
     text: width dimension; url: rectangle-width-dimension
     text: height dimension; url: rectangle-height-dimension
+spec: GEOLOCATION; urlPrefix: https://www.w3.org/TR/geolocation/
+  type: dfn
+    text: set emulated position data; url: #dfn-set-emulated-position-data
 spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
   type: dfn
     text: 2D context creation algorithm; url: canvas.html#2d-context-creation-algorithm
@@ -12270,10 +12273,6 @@ Insert the following steps at the start of the [=determine the device pixel rati
 
 1. If [=device pixel ratio overrides=] [=map/contains=] <var ignore>window</var>'s [=window/navigable=], return [=device pixel ratio overrides=][<var ignore>window</var>'s [=window/navigable=]].
 
-
-## Geolocation ## {#patchs-geolocation}
-
-<dfn>Set emulated position data</dfn> algorithm is going to be added in the geolocation specification.
 
 # Appendices # {#appendices}
 
--- a/index.bs
+++ b/index.bs
@@ -1191,6 +1191,25 @@ To <dfn>get valid navigables by ids</dfn> given a [=/list=] of context ids |navi
 
 </div>
 
+<div algorithm>
+
+To <dfn>get valid top-level traversables by ids</dfn> given a [=/list=] of context ids |navigable ids|:
+
+1. Let |result| be an empty [=/set=].
+
+1. For each |navigable id| in |navigable ids|:
+
+   1. Let |navigable| be the result of [=trying=] to [=get a navigable=] with |navigable id|.
+
+   1. If |navigable| is not a [=/top-level traversable=], return [=error=]
+      with [=error code=] [=invalid argument=].
+
+   1. [=set/Append=] |navigable| to |result|.
+
+1. Return [=success=] with data |result|.
+
+</div>
+
 <div algorithm> To <dfn export>emit an event</dfn> given |session|, and |body|:
 
 1. [=Assert=]: |body| has [=map/size=] 2 and [=map/contains=] "<code>method</code>"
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -5523,6 +5529,126 @@ opened</dfn> steps given |window|, |type|, |message|, and optional |default valu
 
 </div>
 
+## The emulation Module ## {#module-emulation}
+
+The <dfn export for=modules>emulation</dfn> module contains commands and events
```
shs96c, line 5534: Won't this become quite a large module over time? Would it be better to keep (eg) the geolocation emulation in a `geolocation` module? This would also aid discoverability, as all the commands to do with geolocation would all be in the same place.

jgraham: Are we expecting a lot of geolocation-specific functionality? To me it seems more likely that pattern would lead to a lot of modules for each thing we want to override (geolocation / timezone / locale / etc.) all containing one function to override that thing. Of course it doesn't really matter either way, but considering them all kinds of environment emulation seems reasonable to me.

index.bs
```diff
@@ -5523,6 +5529,126 @@ opened</dfn> steps given |window|, |type|, |message|, and optional |default valu
...
+
+The [=remote end steps=] with |command parameters| are:
+
+1. If |command parameters| [=map/contains=] "<code>userContexts</code>"
+   and |command parameters| [=map/contains=] "<code>context</code>",
+   return [=error=] with [=error code=] [=invalid argument=].
+
+1. Let |navigables| be a [=/set=].
+
+1. If the <code>contexts</code> field of |command parameters| is present:
```
jgraham, line 5609: We now have some helpers for parts of this function, like https://w3c.github.io/webdriver-bidi/#get-valid-navigables-by-ids. Given that a lot of these commands end up doing basically the same setup of "get a list of top-level-traversables, fail if any arguments aren't that" (and similar things for user contexts), maybe we can try to centralise more of the logic in those (or new, related) helpers.

index.bs
```diff
@@ -5523,6 +5529,126 @@ opened</dfn> steps given |window|, |type|, |message|, and optional |default valu
...
+   1. For each |user context| of |user contexts|:
+
+      1. [=map/Set=] [=geolocation overrides map=][|user context|] to a |command parameters|["<code>coordinates</code>"].
+
+      1. [=list/For each=] |top-level traversable| of the list of all [=/top-level traversables=]
+         whose [=associated user context=] is |user context|:
+
+         1. [=list/Append=] |top-level traversable| to |navigables|.
+
+1. Otherwise, return [=error=] with [=error code=] [=invalid argument=].
```
jgraham, line 5642: I don't like the multiple-otherwise formulation ("otherwise" is already confusing because it's not clear if it should be a separate step or not). I'd rather have this failure at the top.

lutien: yeah, I agree, updated

index.bs
lutien: Sure, make sense, added two new helpers for top-level traversables and user contexts.

index.bs
```diff
@@ -5523,6 +5558,112 @@ opened</dfn> steps given |window|, |type|, |message|, and optional |default valu
...
+</pre>
+
+A <dfn>geolocation override</dfn> is a [=struct=] with an [=struct/item=] named
+<dfn attribute for="geolocation override">latitude</dfn> which is a float,
+a [=struct/item=] named <dfn attribute for="viewport-dimensions">longitude</dfn> which is a float,
+a [=struct/item=] named <dfn attribute for="viewport-dimensions">accuracy</dfn> which is a float,
+a [=struct/item=] named <dfn attribute for="viewport-dimensions">altitude</dfn> which is a float or null,
+a [=struct/item=] named <dfn attribute for="viewport-dimensions">altitudeAccuracy</dfn> which is a float or null,
+a [=struct/item=] named <dfn attribute for="viewport-dimensions">heading</dfn> which is a float or null,
+a [=struct/item=] named <dfn attribute for="viewport-dimensions">speed</dfn> which is a float or null.
```
reillyeon, line 5583:
```suggestion
a [=struct/item=] named <dfn attribute for="geolocation override">longitude</dfn> which is a float,
a [=struct/item=] named <dfn attribute for="geolocation override">accuracy</dfn> which is a float,
a [=struct/item=] named <dfn attribute for="geolocation override">altitude</dfn> which is a float or null,
a [=struct/item=] named <dfn attribute for="geolocation override">altitudeAccuracy</dfn> which is a float or null,
a [=struct/item=] named <dfn attribute for="geolocation override">heading</dfn> which is a float or null,
a [=struct/item=] named <dfn attribute for="geolocation override">speed</dfn> which is a float or null.
```

index.bs
```diff
@@ -5523,6 +5558,112 @@ opened</dfn> steps given |window|, |type|, |message|, and optional |default valu
...
+        ? contexts: [+browsingContext.BrowsingContext],
+        ? userContexts: [+browser.UserContext],
+      }
+
+      emulation.GeolocationCoordinates = {
+         latitude: float,
+         longitude: float,
+         ? accuracy: float .default 1.0,
+         ? altitude: float / null .default null,
+         ? altitudeAccuracy: float / null .default null,
```
reillyeon, line 5614: Please add validation that `altitude` must be set if `altitudeAccuracy` is set.

reillyeon (commented):
This makes sense from my perspective as Geolocation API editor.

index.bs
```diff
@@ -5523,6 +5558,116 @@ opened</dfn> steps given |window|, |type|, |message|, and optional |default valu
...
+The <dfn export for=modules>emulation</dfn> module contains commands and events
+relating to emulation of browser APIs.
+
+### Definition ### {#module-emulation-definition}
+
+[=remote end definition=]
+
+<pre class="cddl remote-cddl">
+EmulationCommand = (
+  emulation.setGeolocationOverride
```
whimboo, line 5572: In CDP there is also `clearGeolocationOverride`. I wonder if we should separate it as well.

index.bs
```diff
@@ -5523,6 +5558,116 @@ opened</dfn> steps given |window|, |type|, |message|, and optional |default valu
...
+)
+</pre>
+
+A <dfn>geolocation override</dfn> is a [=struct=] with an [=struct/item=] named
+<dfn attribute for="geolocation override">latitude</dfn> which is a float,
+a [=struct/item=] named <dfn attribute for="geolocation override">longitude</dfn> which is a float,
+a [=struct/item=] named <dfn attribute for="geolocation override">accuracy</dfn> which is a float,
```
whimboo, line 5579: Should this be `longitudeAccuracy` or what is the difference to `altitudeAccuracy` below?

lutien: My understanding is that [accuracy field defines accuracy to both `latitude` and `longitude`](https://w3c.github.io/geolocation/#latitude-longitude-and-accuracy-attributes). [altitudeAccuracy defines accuracy for `altitude`](https://w3c.github.io/geolocation/#altitude-and-altitudeaccuracy-attributes). That's the naming which is used in geolocation spec.

index.bs
lutien: We could, but we also use `null` to reset the overrides in `browsingContext.setViewport` command.

index.bs
whimboo: Ah I see. So I wonder if we need a similar check if both `latitude` and `logitude` are present if the `accuracy` is set.

index.bs
whimboo: Good point. I'm fine with a single API then. But lets keep the comment open in case others have some feedback as well.

index.bs
lutien: `latitude` and `logitude` are mandatory fields :)

index.bs
```diff
@@ -4713,7 +4751,7 @@ To <dfn>set viewport</dfn> given |navigable| and |viewport|:
 
 </div>
 
-<div algorithm="set viewport when a navigable is created">
+<div algorithm="set viewport and geolocation override when a navigable is created">
```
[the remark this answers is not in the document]
lutien: It's not luckily (in my opinion), but it might happen that we have to apply other emulations at another point of navigable creation.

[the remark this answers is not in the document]
sadym-chromium: It will be timezone, locale, user-agent, network conditions. So maybe it makes sense to rename it to "configure navigable".

index.bs
sadym-chromium: If I'm reading Geolocation specification PR correctly, having `coordinates: null` makes the geolocation not available, instead of removing the emulation https://pr-preview.s3.amazonaws.com/w3c/geolocation/pull/170.html#dfn-set-the-system-coordinates:

> If coordinates is not given, clear the system location and thus make the position unavailable.

index.bs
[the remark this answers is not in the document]
lutien: Ok, I've renamed the algorithm

sadym-chromium:
I drafted a hook in geolocation: https://github.com/w3c/geolocation/pull/183. @lutien I think it matches the proposed in this PR API. WDYT?

index.bs
```diff
@@ -5523,6 +5558,116 @@ opened</dfn> steps given |window|, |type|, |message|, and optional |default valu
...
+1. If |command parameters| [=map/contains=] "<code>altitudeAccuracy</code>"
+   and |command parameters| doesn't [=map/contain=] "<code>altitude</code>",
+   return [=error=] with [=error code=] [=invalid argument=].
+
+1. If |command parameters| [=map/contains=] "<code>userContexts</code>"
+   and |command parameters| [=map/contains=] "<code>context</code>",
+   return [=error=] with [=error code=] [=invalid argument=].
+
+1. If |command parameters| doesn't [=map/contain=] "<code>userContexts</code>"
+   and |command parameters| doesn't [=map/contains=] "<code>context</code>",
```
jgraham, line 5641:
```suggestion
   and |command parameters| doesn't [=map/contain=] "<code>context</code>",
```

jgraham (approved):

lutien:
> I drafted a hook in geolocation: [w3c/geolocation#183](https://github.com/w3c/geolocation/pull/183). @lutien I think it matches the proposed in this PR API. WDYT?

Thanks! Yes, I think overall it looks as I expected

index.bs
```diff
@@ -4722,6 +4760,11 @@ TODO: Move it as a hook in the html spec instead.
 
 1. Let |user context| be |navigable|'s [=associated user context=].
 
+1. If [=geolocation overrides map=] [=map/contains=] |user context| and
```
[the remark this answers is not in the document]
lutien: hmm... But we already have a hook extracted which accepts a navigable and `coordinates`. We could of course have one more algorithm with the list of navigables and put here this one navigable in the list first, but I'm not sure if it makes a big difference. Do you also want for existing navigables to use the data from the`geolocation overrides ma`? If yes, then we would also need to have a step to get an associated user context from the navigable first. So again, not sure if it makes it simpler. Did I miss anything?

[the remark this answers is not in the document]
sadym-chromium: hm. I thought about preparation for the future emulations. But maybe it's too early. Feel free to ignore the comment above.

index.bs
sadym-chromium: In the [new geolocation PR](https://github.com/w3c/geolocation/pull/183/files#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051R813), `null` is considered as "disable navigation and provide the real data".

sadym-chromium (commented):
Please update the reference to the geolocation spec

index.bs
```diff
@@ -12126,6 +12271,10 @@ Insert the following steps at the start of the [=determine the device pixel rati
 1. If [=device pixel ratio overrides=] [=map/contains=] <var ignore>window</var>'s [=window/navigable=], return [=device pixel ratio overrides=][<var ignore>window</var>'s [=window/navigable=]].
 
 
+## Geolocation ## {#patchs-geolocation}
+
+<dfn>Set the system coordinates</dfn> algorithm is going to be added in the geolocation specification.
```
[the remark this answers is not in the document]
lutien: Sure, updated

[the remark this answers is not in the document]
sadym-chromium: What I mean is I believe you can make this PR pretending the https://github.com/w3c/geolocation/pull/183 is merged, so that we don't need a follow-up change in this spec after the geolocation's one is landed.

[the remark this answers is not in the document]
lutien: Right, I was not sure if the bikeshed would be happy about links, which don't exist yet, but looks like it's fine. So updated ✅

whimboo:
FYI tests will be added via https://bugzilla.mozilla.org/show_bug.cgi?id=1954992.

Issue #789, opened 2024-09-24
Allow browser configuration by user context and not only globally (capabilities)


whimboo:
This is a request coming from Playwright:

> Allow to configure options (aka capabilities) such as `proxy` and `acceptInsecureCerts` per user context rather than per session, so that same browser process can be reused for testing with different options.

We could use the capabilities as default settings, and allow to override those when creating a new user context by specifying new values via optional arguments.

Note that it's not clear if all those settings can be actually handled per user context for all the browsers.

sadym-chromium:
I assume extending [`browser.createUserContext`](https://w3c.github.io/webdriver-bidi/#command-browser-createUserContext) with the following parameters is closing the original requirements:
* acceptInsecureCerts
* proxy
* unhandledPromptBehavior

sadym-chromium:
WIP on `acceptInsecureCerts`

sadym-chromium:
For `unhandledPromptBehavior`, what is the expected behavior for the following scenario:
1. User creates a session with capabilities "accept".
2. In BiDi, user creates a user context with capability "ignore".
3. User prompt is shown.
4. User sends a classic command.

Questions:
* *Should the user prompt be accepted or not?* I'm inclined to answer "no" prioritizing user context override over capability.
* *Should the classic command fail due to the alert?* I'm inclined to answer "yes".

However, to reduce the chance of confusion, we can deny using "createUserContext" with "unhandledPromptBehavior" if the capability "unhandledPromptBehavior" is set.

sadym-chromium:
WIP: [Specify `browsingContext.setUnhandledUserPromptBehavior` by sadym-chromium · Pull Request #896 · w3c/webdriver-bidi](https://github.com/w3c/webdriver-bidi/pull/896)

sadym-chromium:
WIP: [Extend `browser.createUserContext` with `proxy` parameter. by sadym-chromium · Pull Request #897 · w3c/webdriver-bidi](https://github.com/w3c/webdriver-bidi/pull/897)

whimboo:
@sadym-chromium would you mind to add a new entry to the roadmap covering the browser configuration per user context?

sadym-chromium:
I believe the original feature request is addressed, feel free to either reopen or file a new feature request with specific configuration.

whimboo:
Interesting that we had another issue for that. I completely missed it. We can dupe to issue #775.

PR #895, opened 2025-03-31
Extend `browser.createUserContext` with `acceptInsecureCerts` parameter

sadym-chromium:
Addressing #789. Allow for setting `acceptInsecureCerts` for per user context.

Open questions: 
* [ ] How to tight service worker to a user context?
* [ ] Do we need even better integration with Classic? In order to properly specify behavior in case of capability set to "acceptInsecureCerts: true", and user context to "acceptInsecureCerts: false", the classic command on the user context's browsing context will still ignore certificate errors due to this:
> If the [remote end](https://w3c.github.io/webdriver/#dfn-remote-ends)'s [accept insecure TLS](https://w3c.github.io/webdriver/#dfn-accept-insecure-tls) flag is true, no certificate errors that would normally cause the user agent to abort and show a security warning are to hinder navigation to the requested address.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/895.html" title="Last updated on May 7, 2025, 2:44 PM UTC (0fa3d18)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/895/c1c706c...0fa3d18.html" title="Last updated on May 7, 2025, 2:44 PM UTC (0fa3d18)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1615,7 +1615,7 @@ To <dfn>cleanup the session</dfn> given |session|:
 1. For each |user context| in the |session|'s
    [=user contexts with insecure certificates overrides=]:
 
-   1. [=Override insecure certificates behavior=] with |user context| and null.
+   1. [=Restore insecure certificates behavior=] with |user context|.
 
    1. [=set/Remove=] |user context| from |session|'s
       [=user contexts with insecure certificates overrides=].
@@ -2687,20 +2687,19 @@ The <dfn export for=commands>browser.createUserContext</dfn> command creates a
 <div algorithm="remote end steps for browser.createUserContext">
 
 <div algorithm>
+To <dfn>restore insecure certificates behavior</dfn> given |user context|:
+
+1. If [=endpoint node=]'s [=accept insecure TLS=] flag is true,
+   [=override insecure certificates behavior=] with |user context| and true.
+
+1. Otherwise, [=override insecure certificates behavior=] with |user context| and
+   false.
+
+</div>
+
+<div algorithm>
 To <dfn>override insecure certificates behavior</dfn> given |user context| and
 |acceptInsecureCerts|:
-
-1. If |acceptInsecureCerts| is null:
-
-  Note: restore default behavior.
-
-  1. If [=endpoint node=]'s [=accept insecure TLS=] flag is true,
-     [=override insecure certificates behavior=] with |user context| and true.
-
-  1. Otherwise, [=override insecure certificates behavior=] with |user context| and
-     false.
-
-  1. Return.
 
 1. If |acceptInsecureCerts| is true:
 
--- a/index.bs
+++ b/index.bs
@@ -1505,6 +1505,10 @@ time. However they are not required to remove such [=user contexts=]. [=User
 contexts=] that are not [=empty user contexts=] must not be removed from the
 [=set of user contexts=].
 
+A [=BiDi session=] has a
+<dfn>user contexts with insecure certificates overrides</dfn>, which is a [=/set=] of
+[=user contexts=].
+
 When a [=user context=] is [=set/remove|removed=] from the
 [=set of user contexts=], [=remove user context subscriptions=].
 
@@ -1607,6 +1611,14 @@ To <dfn>end the session</dfn> given |session|:
 To <dfn>cleanup the session</dfn> given |session|:
 
 1. [=Close the WebSocket connections=] with |session|.
+
+1. For each |user context| in the |session|'s
+   [=user contexts with insecure certificates overrides=]:
+
+   1. [=Override insecure certificates behavior=] with |user context| and null.
+
+   1. [=set/Remove=] |user context| from |session|'s
+      [=user contexts with insecure certificates overrides=].
 
 1. If [=active sessions=] is [=list/empty=], [=cleanup remote end state=].
 
@@ -2674,7 +2686,46 @@ The <dfn export for=commands>browser.createUserContext</dfn> command creates a
 
 <div algorithm="remote end steps for browser.createUserContext">
 
-The [=remote end steps=] with |command parameters| are:
+<div algorithm>
+To <dfn>override insecure certificates behavior</dfn> given |user context| and
+|acceptInsecureCerts|:
+
+1. If |acceptInsecureCerts| is null:
+
+  Note: restore default behavior.
+
+  1. If [=remote end=]'s [=accept insecure TLS=] flag is true,
+     [=override insecure certificates behavior=] with |user context| and true.
+
+  1. Otherwise, [=override insecure certificates behavior=] with |user context| and
+     false.
+
+  1. Return.
+
+1. If |acceptInsecureCerts| is true:
+
+   1. If [=remote end=] doesn't support accepting insecure TLS connections,
+      return [=error=] with [=error code=] [=unsupported operation=].
+
+   1. Perform implementation-defined steps to disable any certificate errors in the
+      scope of |user context|. This should affect all the network requests associated
+      with the |user context|, e.g. initiated by [=/navigable=] which
+      [=associated user context=] is |user context|.
+
+   1. Return.
+
+1. Assert |acceptInsecureCerts| is false.
+
+1. Perform implementation-defined steps to enable all the required TLS certificates
+   checks for all associated with |user context| network requests. This should affect
+   all the network requests associated with the |user context|, e.g. initiated by
+   [=/navigable=] which [=associated user context=] is |user context|.
+
+TODO: mention service workers.
+
+</div>
+
+The [=remote end steps=] with |session| and |command parameters| are:
 
 1. Let |user context| be a new [=user context=].
 
@@ -2683,17 +2734,11 @@ The [=remote end steps=] with |command parameters| are:
   Note: If "<code>acceptInsecureCerts</code>" is set, it overrides the
   [=accept insecure TLS=] flag's behavior.
 
-  1. If |command parameters|["<code>acceptInsecureCerts</code>"] is true, then
-     perform implementation-defined steps to disable any certificate errors in the
-     scope of |user context|.
+  1. [=Override insecure certificates behavior=] with |user context| and
+     |command parameters|["<code>acceptInsecureCerts</code>"].
 
-  1. If |command parameters|["<code>acceptInsecureCerts</code>"] is false, then
-     perform implementation-defined steps to force certificate checks in the scope of
-     |user context|.
-
-  1. If |command parameters|["<code>acceptInsecureCerts</code>"] is null, then
-     perform implementation-defined steps to restore the default handling of
-     certificate errors in the scope of |user context|.
+  1. [=set/Append=] |user context| to |session|'s
+     [=user contexts with insecure certificates overrides=].
 
 1. [=set/Append=] |user context| to the [=set of user contexts=].
 
--- a/index.bs
+++ b/index.bs
@@ -2690,6 +2690,22 @@ The <dfn export for=commands>browser.createUserContext</dfn> command creates a
 <div algorithm="remote end steps for browser.createUserContext">
 
 <div algorithm>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2655,8 +2663,12 @@ The <dfn export for=commands>browser.createUserContext</dfn> command creates a
       <pre class="cddl remote-cddl">
       browser.CreateUserContext = (
         method: "browser.createUserContext",
-        params: EmptyParams,
+        params: browser.CreateUserContextParameters,
       )
+
+      browser.CreateUserContextParameters = {
+        ? acceptInsecureCerts: bool
```
whimboo, line 2670: For now just a question to get a discussion started... For other features we decided against a parameter object for the `createUserContext` command and ended-up in having a `userContext` argument instead. This applies to `script.addPreloadScript` and `browsingContext.setViewport` so far. 

Adding parameters now would open up a different path in configuring user contexts. I find this a bit confusing and wonder if we should have some command for the network module to configure network behavior (similar like throttling, offline mode etc in the future).

whimboo (commented):
Thank you @sadym-chromium for getting started on this PR! I know it's still in Draft mode but let me add one comment given that this would affect the way of implementation a lot.

index.bs
sadym-chromium: While `acceptInsecureCerts` is implementable via a dedicated command, the `proxy` setting is shared in the scope of the user context, and should be set at the moment the user context is created. So I believe we will need to extend it anyways. However, I can move the `acceptInsecureCerts` into `network` namespace in a dedicated command.

sadym-chromium: the PR is updated

whimboo: Hi @sadym-chromium. We discussed it yesterday internally and agreed on that the former solution was better. Also it would match the same logic as setting the proxy settings which are per user context only. Would you mind reverting the last commit? Thanks!

sadym-chromium: > Hi @sadym-chromium. We discussed it yesterday internally and agreed on that the former solution was better. Also it would match the same logic as setting the proxy settings which are per user context only. Would you mind reverting the last commit? Thanks!

Thanks git, no sweat ) done.

index.bs
```diff
@@ -2669,10 +2674,27 @@ The <dfn export for=commands>browser.createUserContext</dfn> command creates a
 
 <div algorithm="remote end steps for browser.createUserContext">
 
-The [=remote end steps=] are:
+The [=remote end steps=] with |command parameters| are:
 
 1. Let |user context| be a new [=user context=].
 
+1. If |command parameters| [=map/contain=] "<code>acceptInsecureCerts</code>":
```
jgraham, line 2681: I think you need to integrate this better with the WebDriver classic spec (which admittedly is a mess).

For example: what happens if the WebDriver session ends? If you end a classic session then the flag is unset, and cert checking resumes. But BiDi supports multiple sessions, and this is global state, so do we need to revert it only when the last session ends? Or when the session that set the flag ends (I guess the latter). Currently it seems like it would persist indefinitely.  Also, in theory, the capability can be unsupported. It seems reasonable that we handle the unsupported case here too (e.g. by returning an error).

(A really good patch here would have a check in fetch at the point at which a request is started. That would allow answering questions like "what about service workers?". But I understand that's a lot of work, so I don't consider it a blocker)

jgraham (changes requested):

jgraham:
>>Currently it seems like it would persist indefinitely.
> This is intentional.

I think we should try rather hard to avoid persisting this beyond the end of WebDriver sessions. I don't think we want to allow a scenario where https cert checking is disabled for some browsing contexts even when automation is disconnected; that would seem to weaken some security properties that WebDriver classic has.

index.bs
sadym-chromium: Done:
* [x] `acceptInsecureCerts` is tight to the session created the user context. When the session is disconnected, the default behavior taken from WebDriverClassic is restored.
* [x] If the `acceptInsecureCerts` is not supported, error will be returned.

Open question:
* [ ] WRT the service workers, IDK how to tight one to the user context. Do you have any ideas of that, or do you think we can leave it for now?

sadym-chromium: FWIW: I believe the TLS check happens in [fetch: create a connection](https://fetch.spec.whatwg.org/#create-a-connection), step 2, _"... create a connection ..."_.

> Let connection be a new [connection](https://fetch.spec.whatwg.org/#concept-connection) whose [key](https://fetch.spec.whatwg.org/#connection-key) is key, [origin](https://fetch.spec.whatwg.org/#connection-origin) is origin, [credentials](https://fetch.spec.whatwg.org/#connection-credentials) is credentials, and [timing info](https://fetch.spec.whatwg.org/#concept-connection-timing-info) is timingInfo. [Record connection timing info](https://fetch.spec.whatwg.org/#record-connection-timing-info) given connection and use connection to establish an HTTP connection to host, taking proxy and origin into account, with the following caveats: [[HTTP]](https://fetch.spec.whatwg.org/#biblio-http) [[HTTP1]](https://fetch.spec.whatwg.org/#biblio-http1) [[TLS]](https://fetch.spec.whatwg.org/#biblio-tls)

index.bs
```diff
@@ -2669,10 +2686,59 @@ The <dfn export for=commands>browser.createUserContext</dfn> command creates a
...
+<div algorithm>
+To <dfn>override insecure certificates behavior</dfn> given |user context| and
+|acceptInsecureCerts|:
+
+1. If |acceptInsecureCerts| is true:
+
+   1. If [=endpoint node=] doesn't support accepting insecure TLS connections,
+      return [=error=] with [=error code=] [=unsupported operation=].
+
+   1. Perform implementation-defined steps to disable all the TLS certificates checks
```
OrKoN, line 2709: nit: I am not sure we can and would be disabling all possible checks (i.e., we still fail if certificate is malformed). I think what we want is to allow self-signed and expired certificates so perhaps we could be more precise here.

index.bs
```diff
@@ -2669,10 +2686,59 @@ The <dfn export for=commands>browser.createUserContext</dfn> command creates a
...
+      for all associated with |user context| network requests. This should affect all
+      the network requests associated with the |user context|, e.g. initiated by
+      [=/navigable=] which [=associated user context=] is |user context|.
```
OrKoN, line 2712:
```suggestion
      [=/navigable=] whose [=associated user context=]'s [=user context id=] is |user context|.
```

OrKoN, line 2712:
```suggestion
      [=/navigable=] whose [=associated user context=] is |user context|.
```

index.bs
```diff
@@ -2669,10 +2686,59 @@ The <dfn export for=commands>browser.createUserContext</dfn> command creates a
...
+To <dfn>override insecure certificates behavior</dfn> given |user context| and
+|acceptInsecureCerts|:
+
+1. If |acceptInsecureCerts| is true:
+
+   1. If [=endpoint node=] doesn't support accepting insecure TLS connections,
+      return [=error=] with [=error code=] [=unsupported operation=].
+
+   1. Perform implementation-defined steps to disable all the TLS certificates checks
+      for all associated with |user context| network requests. This should affect all
```
OrKoN, line 2710:
```suggestion
      for all network requests associated with |user context|. This should affect all
```

index.bs
```diff
@@ -2669,10 +2686,59 @@ The <dfn export for=commands>browser.createUserContext</dfn> command creates a
...
+
+1. If |acceptInsecureCerts| is true:
+
+   1. If [=endpoint node=] doesn't support accepting insecure TLS connections,
+      return [=error=] with [=error code=] [=unsupported operation=].
+
+   1. Perform implementation-defined steps to disable all the TLS certificates checks
+      for all network requests associated with the [=user context=] with [=user context id=] |user context| . This should affect all
+      the network requests associated with the |user context|, e.g. initiated by
+      [=/navigable=] whose [=associated user context=]'s [=user context id=] is |user context|.
```
OrKoN, line 2712:
```suggestion
      [=/navigable=] whose [=associated user context=] is |user context|.
```

index.bs
```diff
@@ -2669,10 +2686,59 @@ The <dfn export for=commands>browser.createUserContext</dfn> command creates a
...
+
+   1. Return.
+
+1. Assert |acceptInsecureCerts| is false.
+
+1. Perform implementation-defined steps to enable all the required TLS certificates
+   checks for all associated with |user context| network requests. This should affect
```
OrKoN, line 2719:
```suggestion
   checks for all network requests associated with |user context|. This should affect
```

index.bs
```diff
@@ -2669,10 +2686,59 @@ The <dfn export for=commands>browser.createUserContext</dfn> command creates a
...
+   all the network requests associated with the |user context|, e.g. initiated by
+   [=/navigable=] which [=associated user context=] is |user context|.
```
OrKoN, line 2721:
```suggestion
   [=/navigables=] whose [=associated user context=] is |user context|.
```

index.bs
```diff
@@ -2669,10 +2686,59 @@ The <dfn export for=commands>browser.createUserContext</dfn> command creates a
...
+
+1. If |acceptInsecureCerts| is true:
+
+   1. If [=endpoint node=] doesn't support accepting insecure TLS connections,
+      return [=error=] with [=error code=] [=unsupported operation=].
+
+   1. Perform implementation-defined steps to disable all the TLS certificates checks
+      for all network requests associated with the [=user context=] |user context| . This should affect all
+      the network requests associated with the |user context|, e.g. initiated by
+      [=/navigable=] whose [=associated user context=] is |user context|.
```
OrKoN, line 2712:
```suggestion
      [=/navigables=] whose [=associated user context=] is |user context|.
```

index.bs
sadym-chromium: Good point. I added the specific step of certificate validation to skip.

sadym-chromium:
@jgraham WDYT?

index.bs
```diff
@@ -2669,10 +2689,61 @@ The <dfn export for=commands>browser.createUserContext</dfn> command creates a
...
+<div algorithm>
+To <dfn>override insecure certificates behavior</dfn> given |user context| and
+|acceptInsecureCerts|:
+
+1. If |acceptInsecureCerts| is true:
+
+   1. If [=endpoint node=] doesn't support accepting insecure TLS connections,
+      return [=error=] with [=error code=] [=unsupported operation=].
+
+   1. Perform implementation-defined steps to skip steps 6.1.3.a and any other
```
jgraham, line 2712: It would be nicer if the steps linked to the right part of the spec. Also 6.1.3 is a section and a is the step (I think). So maybe something like:

```
For any network request |request|for which the [=request originates in context=] steps with |request| and |context| return true, when running the [=Basic Certificate Processing Steps=] for that request, skip step a, along with any other implementation-specific certificate validation steps.
```
and 

```
A <dfn>request originates in context</a> given |request| and |context| are:
1. Let |settings| be |request|'s  [=request/client=].
1. Let |related navigables| be [=get related navigables=] with |settings|.
1. For |navigable| in |navigables|:
  1. If |navigable|'s [=associated user context=] is |context| return true.
1. Return false.
 ```

**Review by jgraham — COMMENTED** [review]
This basically looks fine, made some editorial suggestions.

File: index.bs (hunk 10 above)
Comment (sadym-chromium) [reply index.bs]: Thanks! Done.

File: index.bs [hunk 11]
```diff
@@ -2669,10 +2689,61 @@ The <dfn export for=commands>browser.createUserContext</dfn> command creates a
 
 <div algorithm="remote end steps for browser.createUserContext">
 
-The [=remote end steps=] are:
+<div algorithm>
+To <dfn>restore insecure certificates behavior</dfn> given |user context|:
+
+1. If [=endpoint node=]'s [=accept insecure TLS=] flag is true,
```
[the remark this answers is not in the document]

jgraham (approved):

PR #896, opened 2025-04-01
Extend `browser.createUserContext` with `unhandledPromptBehavior` parameter

sadym-chromium:
Addressing #789

Analogue of [`acceptInsecureCerts`](https://github.com/w3c/webdriver-bidi/pull/895), add `unhandledPromptBehavior` parameter to `browser.createUserContext`.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/896.html" title="Last updated on Jul 4, 2025, 3:07 PM UTC (3ec0ea6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/896/f89d8e9...3ec0ea6.html" title="Last updated on Jul 4, 2025, 3:07 PM UTC (3ec0ea6)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3120,7 +3120,7 @@ A <dfn>viewport configuration</dfn> is a [=struct=] with an [=struct/item=] name
 <dfn attribute for="viewport-configuration">viewport</dfn> which is a [=viewport-configuration/viewport dimensions=]
 or null and an [=struct/item=] named <dfn attribute for="viewport-configuration">devicePixelRatio</dfn> which is a float or null.
 
-A <dfn>unhandled prompt behavior</dfn> is a [=struct=] with:
+An <dfn>unhandled prompt behavior</dfn> is a [=struct=] with:
 * [=struct/Item=] named <dfn attribute for="unhandled-prompt-behavior-alert">alert</dfn> which is a string or null (default: null);
 * [=struct/Item=] named <dfn attribute for="unhandled-prompt-behavior-beforeUnload">beforeUnload</dfn> which is a string or null (default: null);
 * [=struct/Item=] named <dfn attribute for="unhandled-prompt-behavior-confirm">confirm</dfn> which is a string or null (default: null);
--- a/index.bs
+++ b/index.bs
@@ -3120,13 +3120,13 @@ A <dfn>viewport configuration</dfn> is a [=struct=] with an [=struct/item=] name
 <dfn attribute for="viewport-configuration">viewport</dfn> which is a [=viewport-configuration/viewport dimensions=]
 or null and an [=struct/item=] named <dfn attribute for="viewport-configuration">devicePixelRatio</dfn> which is a float or null.
 
-An <dfn>unhandled prompt behavior</dfn> is a [=struct=] with:
-* [=struct/Item=] named <dfn attribute for="unhandled-prompt-behavior-alert">alert</dfn> which is a string or null (default: null);
-* [=struct/Item=] named <dfn attribute for="unhandled-prompt-behavior-beforeUnload">beforeUnload</dfn> which is a string or null (default: null);
-* [=struct/Item=] named <dfn attribute for="unhandled-prompt-behavior-confirm">confirm</dfn> which is a string or null (default: null);
-* [=struct/Item=] named <dfn attribute for="unhandled-prompt-behavior-default">default</dfn> which is a string or null (default: null);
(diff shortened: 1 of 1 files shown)


burg:
I prefer the separate command that configures the session behavior, optionally for a user context. At least compared to a capability, it would make it easier to share implementations in WebKit. The current `unhandledPromptBehavior` capability is already quite complicated and I'm not sure why it has to be a capability anyway.

sadym-chromium:
> I prefer the separate command that configures the session behavior, optionally for a user context. At least compared to a capability, it would make it easier to share implementations in WebKit. The current `unhandledPromptBehavior` capability is already quite complicated and I'm not sure why it has to be a capability anyway.

I'm not sure how having a separate command would ease the implementation, as we need to specify the behavior per user context.

OrKoN (approved):

sadym-chromium:
@jgraham @gsnedders friendly ping

sadym-chromium:
@whimboo @juliandescottes please take a look

index.bs
```diff
@@ -3114,11 +3119,22 @@ A <dfn>viewport configuration</dfn> is a [=struct=] with an [=struct/item=] name
 <dfn attribute for="viewport-configuration">viewport</dfn> which is a [=viewport-configuration/viewport dimensions=]
 or null and an [=struct/item=] named <dfn attribute for="viewport-configuration">devicePixelRatio</dfn> which is a float or null.
 
+An <dfn>unhandled prompt behavior</dfn> is a [=struct=] with:
+* [=struct/Item=] named <dfn attribute for="unhandled-prompt-behavior-alert">alert</dfn> which is a string or null (default: null);
```
whimboo: Doesn't `or null` already indicate that it is the default. At least that is what we are using for the other override maps.

Also I wonder if it should actually have values of kind struct with `handler` and `notify`. I assume we do not want to have to deserialize the behavior each time when we access this map.

index.bs
```diff
@@ -5720,6 +5736,26 @@ closed</dfn> steps given |window|, |type|, |accepted| and optional |user text|
...
 </dl>
 
+<div algorithm>
+To <dfn>get navigable's user prompt handler</dfn> given |type| and |navigable|:
+
+1. Let |user context| be |navigable|'s [=associated user context=].
+
+1. If [=unhandled prompt behavior overrides map=] contains |user context|:
+
+  1. Let |behavior map| be [=unhandled prompt behavior overrides map=][|user context|].
```
whimboo: Entries for this overrides map are not a map but a string or null at the moment.

sadym-chromium: The `|behavior map|` at this point is an `unhandled prompt behavior` struct, not a string.

sadym-chromium: I renamed it a bit to avoid confusion

index.bs
sadym-chromium: The struct represents the [`session.UserPromptHandler`](https://www.w3.org/TR/webdriver-bidi/#cddl-type-sessionuserprompthandler) with keys "alert" / "beforeUnload" / "confirm" / "default" / "file", "prompt" and values "accept" / "dismiss" / "ignore".

sadym-chromium: This struct is used to store the overrides and does not allow for classics string values like  "dismiss and notify".

sadym-chromium:
@whimboo I believe your concerns are addressed. Please take a look

index.bs
```diff
@@ -5721,6 +5737,27 @@ closed</dfn> steps given |window|, |type|, |accepted| and optional |user text|
    </dd>
 </dl>
 
+<div algorithm>
+To <dfn>get navigable's user prompt handler</dfn> given |type| and |navigable|:
+
+1. Let |user context| be |navigable|'s [=associated user context=].
+
+1. If [=unhandled prompt behavior overrides map=] contains |user context|:
```
juliandescottes: nit: map/contains
```suggestion
1. If [=unhandled prompt behavior overrides map=] [=map/contains=] |user context|:
```

sadym-chromium:
> FWIW, this looks good to me, but maybe wait for @whimboo to give a final sign off.

thanks for the review!

sadym-chromium:
@whimboo friendly ping

sadym-chromium:
I hope @whimboo does not mind me dropping their "request change" review, as I believe the concerns were addressed, and no feedback was received since last week.

index.bs
whimboo: Ok, yes. That makes sense. Not sure why I thought about notify here.

PR #897, opened 2025-04-01
Extend `browser.createUserContext` with `proxy` parameter

sadym-chromium:
Extend `browser.createUserContext` with `proxy` configuration.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/897.html" title="Last updated on May 20, 2025, 4:06 PM UTC (50e0c62)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/897/18006eb...50e0c62.html" title="Last updated on May 20, 2025, 4:06 PM UTC (50e0c62)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2712,10 +2712,10 @@ The [=remote end steps=] with |session| and |command parameters| are:
 
 1. If |command parameters| [=map/contains=] "<code>proxy</code>":
 
-   1. Let |proxy| be [=deserialize as a proxy=] with |command parameters|["<code>proxy</code>"].
+   1. Let |proxy configuration| be |command parameters|["<code>proxy</code>"].
 
    1. Take implementation-defined steps to ensure that requests originating from the
-      |user context| use the proxy defined by the |proxy| configuration. If the
+      |user context| use the proxy defined by the |proxy configuration|. If the
       defined proxy cannot be configured return [=error=] with [=error code=]
       [=invalid argument=].
 
--- a/index.bs
+++ b/index.bs
@@ -7960,6 +7960,20 @@ request sent</dfn> steps given |request|:
             1. Take implementation-defined steps to ensure that |request| uses
                the proxy defined by the |proxy configuration|.
 
+         Note: |user context| can be in not more then one
+         [=user context to proxy configuration map=].
+
+         1. If |session|'s [=user context to proxy configuration map=]
+            [=set/contains|contain=] |user context|:
+
+            1. Let |proxy configuration| be |session|'s
+               [=user context to proxy configuration map=][|user context|].
+
+            1. Assert |proxy configuration| can be configured.
+
+            1. Take implementation-defined steps to ensure that |request| uses
+               the proxy defined by the |proxy configuration|.
+
 1. If [=before request sent map=] does not contain |request|, set [=before
    request sent map=][|request|] to a new set.
 
--- a/index.bs
+++ b/index.bs
@@ -2720,8 +2720,9 @@ The [=remote end steps=] with |session| and |command parameters| are:
 
    1. Let |proxy configuration| be |command parameters|["<code>proxy</code>"].
 
-   1. If the |proxy configuration| cannot be configured return [=error=] with
-      [=error code=] [=invalid argument=].
+   1. If the [=remote end=] is unable to configure proxy settings per [=user context=], 
+       or is unable to configure the proxy with |proxy configuration|, return  [=error=] with
+       [=error code=] [=unsupported operation=].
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2669,10 +2674,18 @@ The <dfn export for=commands>browser.createUserContext</dfn> command creates a
...
 <div algorithm="remote end steps for browser.createUserContext">
 
-The [=remote end steps=] are:
+The [=remote end steps=] with |command parameters| are:
 
 1. Let |user context| be a new [=user context=].
 
+1. If |command parameters| [=map/contains=] "<code>proxy</code>":
+
+   1. Let |proxy| be [=deserialize as a proxy=] with |command parameters|["<code>proxy</code>"].
```
OrKoN, line 2683: I think it should be `[=trying=] to [=deserialize as a proxy=]`

OrKoN: generally, it seems to be used for validation only in the classic spec and validation is covered by CDDL?

index.bs
```diff
@@ -2669,10 +2674,18 @@ The <dfn export for=commands>browser.createUserContext</dfn> command creates a
...
+
+   1. Take implementation-defined steps to set the |user context|'s proxy using the
```
OrKoN, line 2685: Let's rephrase to that we take implementation-defined steps to ensure that requests originating from the user context use the proxy defined by the proxy configuration? `set the |user context|'s proxy using` does not sound precise.

index.bs
sadym-chromium: fair point. Done.

index.bs

index.bs
```diff
@@ -60,6 +60,7 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
     text: close the session; url: dfn-close-the-session
     text: cookie domain; url: dfn-cookie-domain
     text: cookie expiry time; url: dfn-cookie-expiry-time
+    text: deserialize as a proxy; url: dfn-deserialize-as-a-proxy
```
OrKoN, line 63: not needed any longer?

OrKoN (approved):
LGTM

index.bs
```diff
@@ -2708,6 +2716,17 @@ The [=remote end steps=] with |session| and |command parameters| are:
       [=user context to accept insecure certificates override map=][|user context|]
       to |acceptInsecureCerts|.
 
+1. If |command parameters| [=map/contains=] "<code>proxy</code>":
+
+   1. Let |proxy configuration| be |command parameters|["<code>proxy</code>"].
+
+   1. If the |proxy configuration| cannot be configured return [=error=] with
+      [=error code=] [=invalid argument=].
```
jgraham, line 2724: I think "if |configuration" cannot be configured" is a bit confusing, even though it's close to the language in WebDriver. Maybe:

```suggestion
   1. If the [=remote end=] is unable to configure proxy settings per [=user context=], 
       or is unable to configure the proxy with |proxy configuration|, return  [=error=] with
       [=error code=] [=unsupported operation=].
```

index.bs
```diff
@@ -7928,6 +7947,20 @@ request sent</dfn> steps given |request|:
...
+         Note: |user context| can be in not more then one
+         [=user context to proxy configuration map=].
+
+         1. If |session|'s [=user context to proxy configuration map=]
+            [=set/contains=] |user context|:
+
+            1. Let |proxy configuration| be |session|'s
+               [=user context to proxy configuration map=][|user context|].
+
+            1. Assert |proxy configuration| can be configured.
```
jgraham, line 7959: I would remove this one.

index.bs
```diff
@@ -7928,6 +7947,20 @@ request sent</dfn> steps given |request|:
...
+
+            1. Take implementation-defined steps to ensure that |request| uses
+               the proxy defined by the |proxy configuration|.
```
jgraham, line 7962:
```suggestion
            1. Take implementation-defined steps to ensure that |request| uses
               the proxy settings defined by the |proxy configuration|.
               
               Note: the settings are validated when the user context is created and so
               are assumed to be valid at this stage; any error accessing the proxy will
               be reported as a network error when handling the request.
```

jgraham (approved):
I think this is basically fine. I have some suggestions, but they don't block. I also wonder if we should try to link to https://w3c.github.io/webdriver/#dfn-proxy-configuration somewhere to define what the proxy configuration object means.

index.bs
sadym-chromium: Agree. The note you added is enough

PR #902, opened 2025-04-10
Add more detailed validation of geolocation coordinates.

### LINKED ISSUE #900: Add more detailed validation to provided coordinates to "emulation.setGeolocationOverride" command
CDDL in `emulation.setGeolocationOverride` command at the moment covers only basic validation for `coordinates` argument. At the same time, the geolocation spec describes in more details what are possible values for properties of the `coordinates` argument. I think we should either extend the CDDL or add validation steps to `emulation.setGeolocationOverride` command.

lutien:
Closes #900

The validation data comes from [the geolocation spec](https://www.w3.org/TR/geolocation/#acquire-a-position).

***
<a href="https://pr-preview.s3.amazonaws.com/lutien/webdriver-bidi/pull/902.html" title="Last updated on Apr 10, 2025, 2:21 PM UTC (274f0ef)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/902/aaec7b6...lutien:274f0ef.html" title="Last updated on Apr 10, 2025, 2:21 PM UTC (274f0ef)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5632,11 +5632,11 @@ The <dfn export for=commands>emulation.setGeolocationOverride</dfn> command modi
       emulation.GeolocationCoordinates = {
          latitude: float,
          longitude: float,
-         ? accuracy: float .default 1.0,
+         ? accuracy: (float .ge 0.0) .default 1.0,
          ? altitude: float / null .default null,
-         ? altitudeAccuracy: float / null .default null,
-         ? heading: float / null .default null,
-         ? speed: float / null .default null,
+         ? altitudeAccuracy: (float .ge 0.0) / null .default null,
+         ? heading: (0.0...360.0) / null .default null,
+         ? speed: (float .ge 0.0) / null .default null,
       }
     </pre>
    </dd>
--- a/index.bs
+++ b/index.bs
@@ -5630,8 +5630,8 @@ The <dfn export for=commands>emulation.setGeolocationOverride</dfn> command modi
       }
 
       emulation.GeolocationCoordinates = {
-         latitude: float,
-         longitude: float,
+         latitude: -90.0..90.0,
+         longitude: -180.0..180.0,
          ? accuracy: (float .ge 0.0) .default 1.0,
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -5632,11 +5632,11 @@ The <dfn export for=commands>emulation.setGeolocationOverride</dfn> command modi
       emulation.GeolocationCoordinates = {
          latitude: float,
          longitude: float,
```
whimboo, line 5634: For both fields the geolocation spec says that their value is given in degrees. Shall we as well limit to `0.0...360.0)`?

lutien: Yep, updated (but to valid ranges for these values)

index.bs
```diff
@@ -5630,13 +5630,13 @@ The <dfn export for=commands>emulation.setGeolocationOverride</dfn> command modi
       }
 
       emulation.GeolocationCoordinates = {
-         latitude: float,
-         longitude: float,
-         ? accuracy: float .default 1.0,
+         latitude: -90.0..90.0,
```
whimboo, line 5633: I can see that we have such definitions already in the spec but it's a bit sad that we loose the actual type definition here by using dots. Would the following work as well?

```suggestion
         latitude: float .ge -90.0 .le 90.0
```

lutien: Probably we could, but I'm not sure what exactly is the problem. We have numbers with decimal places in the range, this implies that the value should be of type float.

OrKoN (approved):

index.bs
whimboo: So we have both variants in use in the specification. It's not a blocker for this PR but maybe we should find a solution to use one or the other.

whimboo (approved):

index.bs
whimboo: I filed https://github.com/w3c/webdriver-bidi/issues/907 for it.

PR #484, opened 2023-07-10
Add `type` for message types

jrandolf-google:
Fixed: https://github.com/w3c/webdriver-bidi/issues/480

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/484.html" title="Last updated on Jul 10, 2023, 12:23 PM UTC (770693d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/484/cbc17e9...770693d.html" title="Last updated on Jul 10, 2023, 12:23 PM UTC (770693d)">Diff</a>

PR #903, opened 2025-04-11
Use Event production in 'emit an event' assertion

lauromoura:
The `emit an event` algorithm previously asserted that the event body only contains `method` and `params`, which is no longer the case after #484 added the `type` field.

Instead of such low-level assertion, this commit changes the assertion to match against the Event type production.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1215,8 +1215,7 @@ To <dfn>get valid top-level traversables by ids</dfn> given a [=/list=] of conte
 
 <div algorithm> To <dfn export>emit an event</dfn> given |session|, and |body|:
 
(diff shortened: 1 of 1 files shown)


OrKoN (approved):
Thanks!

whimboo (approved):
Thanks a lot for fixing that. There is no update needed for the BiDi client in the web-platform tests because we [already added the `event` type check](https://github.com/web-platform-tests/wpt/blob/c805c049458288caf3d47c10ba05e5c85c46e302/tools/webdriver/webdriver/bidi/client.py#L203) around the time when the patch got landed.

PR #905, opened 2025-04-14
Forced colors mode WebDriver emulation command

alisonmaher:
Forced Colors Mode is difficult to test via WPT because it is dependent on OS state (Window High Contrast) to trigger the feature to light up.

The CSSWG resolved to add a WebDriver extension to aid in testing this feature: https://github.com/w3c/csswg-drafts/issues/11824#issuecomment-2769343959

I originally drafted the emulation based on the original WebDriver spec (https://github.com/w3c/csswg-drafts/pull/11823) but in review, it was suggested to use the WebDriver Bidi spec.

The `Emulation Module` seemed to be the best place for this to live, and it is recommended in the spec to extend existing modules directly in the WebDriver Bidi spec rather than in the spec the related feature is defined.

This change adds a new `ForcedColorsModeThemeOverride` WebDriver `EmulationCommand` (https://alisonmaher.github.io/alisonmaher/webdriver-bidi/index.html#module-emulation-definition) with a corresponding `setForcedColorsModeThemeOverride` definition (https://alisonmaher.github.io/alisonmaher/webdriver-bidi/index.html#module-emulation-commands).

This emulation will require also a change to the color adjust spec to define "emulated theme data" in step 6.1 of the defined remote ends steps for the new emulation command (similar to what is done for the geolocation emulation).

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5591,9 +5591,16 @@ relating to emulation of browser APIs.
 
 <pre class="cddl remote-cddl">
 EmulationCommand = (
+  emulation.SetForcedColorsModeThemeOverride //
   emulation.SetGeolocationOverride
 )
 </pre>
+
+The <dfn>forced colors mode theme override</dfn> is a [=struct=] with:
+* [=struct/item=] named <dfn attribute for="forced colors mode theme override">theme</dfn> which is a string.
+
+A [=remote end=] has a <dfn>forced colors mode theme overrides map</dfn> which is a weak map
+between [=user contexts=] and [=forced colors mode theme override=].
 
 A <dfn>geolocation override</dfn> is a [=struct=] with:
 * [=struct/item=] named <dfn attribute for="geolocation override">latitude</dfn> which is a float;
@@ -5606,8 +5613,77 @@ A <dfn>geolocation override</dfn> is a [=struct=] with:
 
 A [=remote end=] has a <dfn>geolocation overrides map</dfn> which is a weak map between [=user contexts=] and [=geolocation override=].
 
-
 ### Commands ### {#module-emulation-commands}
+
+#### The emulation.setForcedColorsModeThemeOverride Command ####  {#command-emulation-setForcedColorsModeThemeOverride}
+
+The <dfn export for=commands>emulation.setForcedColorsModeThemeOverride</dfn> command modifies
+[=forced colors mode=] theming characteristics on the given top-level traversables or user contexts.
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl remote-cddl">
+      emulation.SetForcedColorsModeThemeOverride = (
+        method: "emulation.setForcedColorsModeThemeOverride",
+        params: emulation.SetForcedColorsModeThemeOverrideParameters
+      )
+
+      emulation.SetForcedColorsModeThemeOverrideParameters = {
+        theme: [+emulation.ForcedColorsModeAutomationTheme],
+        ? contexts: [+browsingContext.BrowsingContext],
+        ? userContexts: [+browser.UserContext],
+      }
+
+      emulation.ForcedColorsModeAutomationTheme = "none" / "light" / "dark"
+    </pre>
+   </dd>
+   <dt>Result Type</dt>
+   <dd>
+    <code>
+     EmptyResult
+    </code>
+   </dd>
+</dl>
+
+<div algorithm="remote end steps for emulation.setForcedColorsModeThemeOverride">
+
+The [=remote end steps=] with |command parameters| are:
+
+1. If |command parameters| [=map/contains=] "<code>userContexts</code>"
+   and |command parameters| [=map/contains=] "<code>context</code>",
+   return [=error=] with [=error code=] [=invalid argument=].
+
+1. If |command parameters| doesn't [=map/contain=] "<code>userContexts</code>"
+   and |command parameters| doesn't [=map/contain=] "<code>context</code>",
+   return [=error=] with [=error code=] [=invalid argument=].
+
+1. Let |navigables| be a [=/set=].
+
+1. If the <code>contexts</code> field of |command parameters| is present:
+
+   1. Let |navigables| be the result of [=trying=] to [=get valid top-level traversables by ids=] with |command parameters|["<code>contexts</code>"].
+
+1. Otherwise, if the <code>userContexts</code> field of |command parameters| is present:
+
+   1. Let |user contexts| be the result of [=trying=] to [=get valid user contexts=] with |command parameters|["<code>userContexts</code>"].
+
+   1. For each |user context| of |user contexts|:
+
+      1. [=map/Set=] [=forced colors mode theme overrides map=][|user context|] to |command parameters|["<code>theme</code>"].
+
+      1. [=list/For each=] |top-level traversable| of the list of all [=/top-level traversables=]
+         whose [=associated user context=] is |user context|:
+
+         1. [=list/Append=] |top-level traversable| to |navigables|.
+
+1. For each |navigable| of |navigables|:
+
+   1. [=Set emulated theme data=] with |navigable| and |command parameters|["<code>theme</code>"].
+
+1. Return [=success=] with data null.
+
+</div>
 
 #### The emulation.setGeolocationOverride Command ####  {#command-emulation-setGeolocationOverride}
 


index.bs
```diff
@@ -5606,9 +5613,84 @@ A <dfn>geolocation override</dfn> is a [=struct=] with:
...
+      1. [=map/Set=] [=forced colors mode theme overrides map=][|user context|] to |command parameters|["<code>theme</code>"].
+
+      1. [=list/For each=] |top-level traversable| of the list of all [=/top-level traversables=]
+         whose [=associated user context=] is |user context|:
+
+         1. [=list/Append=] |top-level traversable| to |navigables|.
+
+1. For each |navigable| of |navigables|:
+
+   1. [=Set emulated forced colors theme data=] with |navigable| and |command parameters|["<code>theme</code>"].
```
alisonmaher, line 5688: As a note for reviewers, this is dependent on https://github.com/w3c/csswg-drafts/pull/11823

alisonmaher:
@OrKoN great point. This was something I considered, as well, but I ended up deciding against it to keep these concepts conceptually separate. I personally think the CDP design is slightly confusing that you have to set both media features to change the Forced Colors theme for emulation purposes. And keeping them separate would allow us to diverge what theme types we allow for each feature emulation in the future, if required.

But I'm open to splitting that out or making the theme input type a more generic name to allow us to add a similar mechanism for the preferred color scheme in the future if that ends up being preferrable.

Behind the scenes, I think we can use the same pipeline as the existing Chrome DevTools emulation for forced colorsw mode (although I added some additional system color mappings to the color adjust spec PR draft for this that we'll need to double check are correctly mapped in the existing emulation). Otherwise, this should result in equivalent behavior to the Chrome Devtools emulation.

OrKoN:
I can start a review but I would love input from @jgraham @sadym-chromium @whimboo @gsnedders on whether we want to have a single extendable command for media feature emulation or if we prefer to have a command per feature. I am leaning towards having a single command that can be extended with various media features.

index.bs
```diff
@@ -5606,9 +5613,84 @@ A <dfn>geolocation override</dfn> is a [=struct=] with:
...
+
+<div algorithm="remote end steps for emulation.setForcedColorsModeThemeOverride">
+
+The [=remote end steps=] with |command parameters| are:
+
+1. If |command parameters| doesn't [=map/contain=] "<code>theme</code>",
+   return [=error=] with [=error code=] [=invalid argument=].
+
+1. If |command parameters| [=map/contains=] "<code>theme</code>",
+   and "<code>theme</code>" is not "<code>none</code>", "<code>light</code>", or "<code>dark</code>",
```
OrKoN, line 5657: these two steps are already covered by the CDDL validation so they can be removed

alisonmaher:
@OrKoN great, thank you! And I would like to point out that Forced Colors Mode, as compared to other media features, has a lot more involved when it comes to the emulation than other media features.

System colors start overriding styles on the page, along with other interactions.

Granted that we do emulate forced colors mode in Chromium devtools using the media feature as the input, but generally speaking, the effects of that are more than just the media feature evaluating to some value, which I think is what makes this fairly unique amongst the other media features.

Open to some other design here, though, if it ends up making the most sense.

index.bs
OrKoN, line 5688: I think `set emulated forced colors theme data` should be called for each new top-level traversable if the user context has an override? The geolocation override solved via a TODO to include a call into the HTML spec so we can either add something there or (better) if the CSS spec could define a what point the override should start applying for the new traversable.

OrKoN (changes requested):

index.bs
alisonmaher: Sounds good, I can make that update.

Regarding the CSS spec, I defined what `set emulated forced colors theme data` means at https://drafts.csswg.org/css-color-adjust-1/#color-adjust-emulation

And then I define how this is integrated into the feature at https://drafts.csswg.org/css-color-adjust-1/#forced (see the last paragraph of that section).

Is there anything else that would be required by the CSS spec for this? I'm also happy to make any updates there that would be better. Was mainly trying to follow similarly to what geolocation was doing

alisonmaher: @OrKoN Just getting back around to making these edits, but I wanted to clarify this comment in particular.

From my understanding of the current pipeline as it is currently drafted, the `emulation.setForcedColorsModeThemeOverride` command would be used in a WPT to enable the emulation.

We end up calling `set emulated forced colors theme data` on each navigable, which include the top-level traversable.

When this is set, we end up going to https://drafts.csswg.org/css-color-adjust-1/#color-adjust-emulation, which requires a style recalculation, which ends up taking us to https://drafts.csswg.org/css-color-adjust-1/#forced, which is where we take the emulation theme into account and apply the necessary emulation of forced colors mode.

When you suggest that we should `set emulated forced colors theme data` should be called for each new top-level traversable if the user context has an override, is this in addition to the above, or in replacement of the above?

index.bs
alisonmaher: Removed the first two steps in the latest commit. I had a clarification on the second point below, so I'll follow up on that one once I have a better understanding on what you're looking for on that one.

index.bs
alisonmaher: This has now landed.

alisonmaher: @OrKoN Now that I think about it, perhaps the text at https://drafts.csswg.org/css-color-adjust-1/#forced that states "If [get emulated forced colors theme data](https://drafts.csswg.org/css-color-adjust-1/#get-emulated-forced-colors-theme-data) is not "[none](https://drafts.csswg.org/css-color-adjust-1/#dom-forcedcolorsmodeautomationtheme-none)", the user agent should bypass the above operating system color themes, and instead do the following based on the resulting value of [emulated forced colors theme data](https://drafts.csswg.org/css-color-adjust-1/#emulated-forced-colors-theme-data):..." would cover what you were looking for?

OrKoN: Sorry for a slow reply. I meant the following scenario:

1) the client calls `emulation.setForcedColorsModeThemeOverride(userContexts=[A])` (a user context is like an incognito window in Chrome, holds multiple tabs). As specified, `Set emulated forced colors theme data` is not called because no navigables exist in the user context.
2) the client then calls `browsingContext.create(userContext=A)` that creates a new top-level traversable in the user context.

Now the client expects that the override automatically applies to the newly created traversable but I believe no spec text covers it. I think https://drafts.csswg.org/css-color-adjust-1/#forced does not help because https://drafts.csswg.org/css-color-adjust-1/#emulated-forced-colors-theme-data would be none.  

As I mentioned, the geolocation solved it by adding a TODO for future integration with the HTML spec https://w3c.github.io/webdriver-bidi/#command-browsingContext-create:~:text=TODO%3A%20Move%20it%20as%20a%20hook%20in%20the%20html%20spec%20instead and updating the state according to the override. But if we can integrate it into the CSS color adjust spec by having the spec invoke steps in the WebDriver BiDi spec to determine whether a given navigable has an override or not.

https://drafts.csswg.org/css-color-adjust-1/#emulated-forced-colors-theme-data also does not seem to accept a navigable as a parameter? It's given an instance of ForcedColorsModeAutomationTheme but as far as I see that is not associated with any navigable because it's an enum?

alisonmaher: Thanks for the detailed explanation and example.

> https://drafts.csswg.org/css-color-adjust-1/#emulated-forced-colors-theme-data also does not seem to accept a navigable as a parameter? It's given an instance of ForcedColorsModeAutomationTheme but as far as I see that is not associated with any navigable because it's an enum?

The setter does take one in, just separately from ForcedColorsModeAutomation theme via text "To set emulated forced colors theme data, given [navigable](https://html.spec.whatwg.org/multipage/nav-history-apis.html#window-navigable) navigable and an emulatedThemeData". And in this PR we pass in the navigable and theme when setting.

However, you're right that the getter doesn't, so it sounds like perhaps the two things that need to happen are:
1. Pass in a navigable to the getter.
2. Figure out some way for https://drafts.csswg.org/css-color-adjust-1/#forced to access the current navigable to pass into the getter. I think this would solve the case you raised since a new tab or window will walk this path, and as long as we have the right info to pass into the setter, that should fix this.

Although to double check, we do note in the getter that "Let navigable be theme’s [relevant global object](https://html.spec.whatwg.org/multipage/webappapis.html#concept-relevant-global)’s [associated Document](https://html.spec.whatwg.org/multipage/nav-history-apis.html#concept-document-window)’s [node navigable](https://html.spec.whatwg.org/multipage/document-sequences.html#node-navigable)." Would this cover your case, or not quite? If not, what's the best way to get the right navigable - are there other specs that would be good to reference for this?

OrKoN: From https://drafts.csswg.org/css-color-adjust-1/#forced I am not sure when the browsers should call the getter and how they provide the required theme parameter? IIUC the theme cannot have a relevant global object because theme is not a platform object? 

I am not sure it addresses the original problem: navigables might not exist when the WebDriver override is configured so there is no navigable to call the setter on.

alisonmaher: The getter in https://drafts.csswg.org/css-color-adjust-1/#forced will be called anytime there is a style recalculation, or on a refresh. From my read of `browsingContext.create()` this will create a new window or tab, so this code will be exercised in that case.

However, if `emulation.setForcedColorsModeThemeOverride()` is called and no navigable exists yet, should that be a no-op? I think the way this algorithm is set up, we only call the setter on navigables? Unless we should be updating this PR in some way to also go through this path whenever `browsingContext.create()` is called (although I'm not sure how we'd know what the theme set to is, as you already mentioned)?

Should we be tying the algorithm in this PR to something other than a navigable?

alisonmaher: @OrKoN friendly ping on this. And also open to setting up a quick call later this week or next in case that would be easier (if so, I can reach out on slack to set something up). Want to make sure I fully understand what updates you are suggesting for this one

gsnedders (commented):
I would argue that this is notably different than most other emulation commands; c.f. https://github.com/w3c/csswg-drafts/issues/11824#issuecomment-2770404993

I still believe that it needs to be conforming to not actually support forced colour palettes.

For example, macOS provides no means to set a forced colour palette or enable any forced colour mode, and nor does Safari. Under the current Color Adjustment ED, it would be conforming for Safari to support the `forced-color-adjust` property and its three values by doing nothing, because the UA would never be in forced colours mode.

However, if WebDriver BiDi requires support for the `emulation.setForcedColorsModeThemeOverride` command, and by reference requires support for the emulated forced colours theme data, then the UA must have an implementation of forced colour modes and palettes that it otherwise would not be required to.

I don't know exactly how we want to deal with this, but my first suggestion would be that we want to behave as if `emulation.ForcedColorsModeAutomationTheme` has a smaller set of values than it actually does (and thus matching the message would cause an "invalid argument" error), and support only "none". Alternatively, we could just make support of  `emulation.setForcedColorsModeThemeOverride` optional.

jgraham:
I think it's totally reasonable to allow implementations to return Unsupported Operation if they don't implement the feature; we already do this in many cases (e.g. window state, `setViewport`, taking screenshots, etc.)

index.bs
OrKoN: sorry for the slow response, let me leave a specific suggestion for this PR and see if it clarifies the changes! If not, we can discuss it in the next WG meeting https://www.w3.org/wiki/WebDriver

OrKoN: I realized I cannot actually add a comment on the part not included in the diff. In the following algorithm on the spec that starts with:

```
After creating a document in a new [=/navigable=] |navigable| and
before the [=run WebDriver BiDi preload scripts=] algorithm is invoked:

TODO: Move it as a hook in the html spec instead.
```

We should add a step within `If navigable is a top-level traversable:`:

```
1. If [=forced colors mode theme overrides map=] [=map/contains=] |user context|,
      [=set emulated forced colors theme data=] with |navigable| and [=forced colors mode theme overrides map=][|user context|].
```

As I mentioned, it is also a possibility to refactor the css-color-adjust-spec-1 to call WebDriver BiDi algorithms to check if there is an override for a given navigable. But I think the change above should be sufficient.

alisonmaher: This helps a lot, thanks for the clarification. I completely missed there was a note about geolocation already there, and the TODO was to move it instead of to add something about it. Apologies for the confusion, what you were mentioning earlier makes a lot more sense now. Updated per your suggestion.

OrKoN (approved):
Looks good to me, @sadym-chromium could you please review as well?

alisonmaher:
Friendly ping on this review. @sadym-chromium @jgraham any additional feedback from your side?

sadym-chromium:
Sorry, it fell out of my attention. I'll check it out next week due to vacation. Please let us know if you need the feedback earlier.

index.bs
```diff
@@ -5606,9 +5616,78 @@ A <dfn>geolocation override</dfn> is a [=struct=] with:
...
+        params: emulation.SetForcedColorsModeThemeOverrideParameters
+      )
+
+      emulation.SetForcedColorsModeThemeOverrideParameters = {
+        theme: [+emulation.ForcedColorsModeAutomationTheme],
+        ? contexts: [+browsingContext.BrowsingContext],
+        ? userContexts: [+browser.UserContext],
+      }
+
+      emulation.ForcedColorsModeAutomationTheme = "none" / "light" / "dark"
```
sadym-chromium: 1. CSS Color Adjustment spec [uses a string literal "none"](https://drafts.csswg.org/css-color-adjust-1/#dom-forcedcolorsmodeautomationtheme-none) as an indicator of no overrides. In WebDriver BiDi, e.g. [`emulation.setGeolocationOverride`](https://www.w3.org/TR/webdriver-bidi/#cddl-key-emulationsetgeolocationoverrideparameters-coordinates) the value `null` is used to flag it. Please follow the BiDi style. Please note that the call to `[=Set emulated forced colors theme data=]` should be done with string literal "none".
2. `theme` cannot have multiple values.

```suggestion
      emulation.SetForcedColorsModeThemeOverrideParameters = {
        theme: emulation.ForcedColorsModeAutomationTheme / null,
        ? contexts: [+browsingContext.BrowsingContext],
        ? userContexts: [+browser.UserContext],
      }

      emulation.ForcedColorsModeAutomationTheme = "light" / "dark"
```

sadym-chromium (changes requested):
Great work! Please update the API to align with other method params (in comment).

index.bs
alisonmaher: Thanks @sadym-chromium for the feedback! I've updated according to your suggestion in the latest commit: https://github.com/w3c/webdriver-bidi/pull/905/commits/9966851a2cfd3814f8325b20ea001944ec31fb05.

> Please note that the call to [=Set emulated forced colors theme data=] should be done with string literal "none".

I added a note to both places where we call [=Set emulated forced colors theme data=] that null should be converted to "none", but please feel free to let me know if you think that should live in a more central location (or if you think that is implied).

index.bs
```diff
@@ -4774,6 +4774,10 @@ TODO: Move it as a hook in the html spec instead.
    1. If [=geolocation overrides map=] [=map/contains=] |user context|,
       [=set emulated position data=] with |navigable| and [=geolocation overrides map=][|user context|].
 
+   1. If [=forced colors mode theme overrides map=] [=map/contains=] |user context|,
+      [=set emulated forced colors theme data=] with |navigable| and [=forced colors mode theme overrides map=][|user context|],
+      where a [=forced colors mode theme overrides map=][|user context|] value of <code>null</code> is converted to <code>"none"</code>.
+
```
```suggestion
   1. If [=forced colors mode theme overrides map=] [=map/contains=] |user context|:
      1. Let |theme| be [=forced colors mode theme overrides map=][|user context|].
      1. If |theme| is null, set |theme| to "<code>none</code>".
      1. [=Set emulated forced colors theme data=] with |navigable| and |theme|.
```

index.bs
```diff
@@ -5606,9 +5617,79 @@ A <dfn>geolocation override</dfn> is a [=struct=] with:
...
+      1. [=list/For each=] |top-level traversable| of the list of all [=/top-level traversables=]
+         whose [=associated user context=] is |user context|:
+
+         1. [=list/Append=] |top-level traversable| to |navigables|.
+
+1. For each |navigable| of |navigables|:
+
+   1. [=Set emulated forced colors theme data=] with |navigable| and |command parameters|["<code>theme</code>"],
+      where a |command parameters|["<code>theme</code>"] value of <code>null</code> is converted to <code>"none"</code>.
```
```suggestion
   1. Let |theme| be |command parameters|["<code>theme</code>"].
   1. If |theme| is null, set |theme| to "<code>none</code>".
   1. [=Set emulated forced colors theme data=] with |navigable| and |theme|.
```

sadym-chromium (approved):
LGTM with the patches in the comments

index.bs
alisonmaher: Great idea, that is much cleaner. Updated!

index.bs
alisonmaher: Done

alisonmaher:
@sadym-chromium @OrKoN thanks again for the detailed reviews! If all looks good, I'll need help in landing the PR since I don't have write permissions.

There is a [check that failed](https://github.com/w3c/webdriver-bidi/actions/runs/15736770275/job/44351497386?pr=905) on the latest commit, but it looks unrelated to me, although it doesn't seem like I have permissions to re-run the check myself.

OrKoN:
We still need input from other vendors before we can land this. The error needs to be fixed by changing the way cddl blocks are formatted (changed with https://github.com/w3c/webdriver-bidi/pull/927)

alisonmaher:
> We still need input from other vendors before we can land this. The error needs to be fixed by changing the way cddl blocks are formatted (changed with https://github.com/w3c/webdriver-bidi/pull/927)

@OrKoN thanks, I rebased the branch and updated the cddl usage that I added, and that seems to have fixed it.

Is that last review we're waiting on from @jgraham or are there others you are hoping to have take a look, as well?

index.bs
```diff
@@ -5747,9 +5762,82 @@ A <dfn>geolocation override</dfn> is a [=struct=] with:
...
+
+1. For each |navigable| of |navigables|:
+
+   1. Let |theme| be |command parameters|["<code>theme</code>"].
+
+   1. If |theme| is null, set |theme| to "<code>none</code>".
+
+   1. [=Set emulated forced colors theme data=] with |navigable| and |theme|.
+
+1. Return [=success=] with data null.
```
sadym-chromium: Oh, one more small thing I didn't recognize before: 
```suggestion
1. Let |theme| be |command parameters|["<code>theme</code>"].

1. If |theme| is null, set |theme| to "<code>none</code>".

1. For each |navigable| of |navigables|:

   1. [=Set emulated forced colors theme data=] with |navigable| and |theme|.

1. Return [=success=] with data null.
```

alisonmaher: Ah good point, updated per your suggestion

alisonmaher:
@jgraham friendly ping to see if you have any feedback on this one? Thanks in advance!

alisonmaher:
@OrKoN @sadym-chromium - should I continue to wait for other reviews on this one, or has enough time passed to consider this ok to land?

sadym-chromium:
> @OrKoN @sadym-chromium - should I continue to wait for other reviews on this one, or has enough time passed to consider this ok to land?

@jgraham @whimboo @juliandescottes @lutien could you please take a look?

sadym-chromium:
> @OrKoN @sadym-chromium - should I continue to wait for other reviews on this one, or has enough time passed to consider this ok to land?

We should wait for at least one other vendor's approval.

index.bs
```diff
@@ -4892,6 +4892,14 @@ TODO: Move it as a hook in the html spec instead.
    1. If [=geolocation overrides map=] [=map/contains=] |user context|,
       [=set emulated position data=] with |navigable| and [=geolocation overrides map=][|user context|].
 
+   1. If [=forced colors mode theme overrides map=] [=map/contains=] |user context|:
+
+      1. Let |theme| be [=forced colors mode theme overrides map=][|user context|].
+
+      1. If |theme| is null, set |theme| to "<code>none</code>".
+
+      1. [=Set emulated forced colors theme data=] with |navigable| and |theme|.
```
lutien, line 4899: @alisonmaher, as far as I understood [this is the hook](https://drafts.csswg.org/css-color-adjust-1/#set-emulated-forced-colors-theme-data) which you're using here, could you link in the document header?

alisonmaher: @lutien that's correct, and the link seems to work if you go to my hosted version of the spec updates: https://alisonmaher.github.io/alisonmaher/webdriver-bidi/index.html#module-emulation-commands.

Do you mean link it in the PR description? Or is there some documentation for these sorts of external links I should add in the doc itself?

lutien: > @lutien that's correct, and the link seems to work if you go to my hosted version of the spec updates: https://alisonmaher.github.io/alisonmaher/webdriver-bidi/index.html#module-emulation-commands.
> 
hmm... it also works in the PR preview, so I guess it's somehow found automatically.

I meant adding it in the document like e.g. for geolocation override: https://github.com/w3c/webdriver-bidi/blob/main/index.bs#L201. But I guess if it works anyway, it's probably fine 🙂

alisonmaher: Sounds good, yeah, I think bikeshed is usually good at finding these unless it is ambiguous in some way. In this case, there should only be one definition of this phrase, so I think it should work as-is from what I can tell

index.bs
```diff
@@ -5813,13 +5821,20 @@ relating to emulation of browser APIs.
...
   emulation.SetTimezoneOverride
 )
 </pre>
 
+The <dfn>forced colors mode theme override</dfn> is a [=struct=] with:
+* [=struct/item=] named <dfn attribute for="forced colors mode theme override">theme</dfn> which is a string.
+
+A [=remote end=] has a <dfn>forced colors mode theme overrides map</dfn> which is a weak map
+between [=user contexts=] and [=forced colors mode theme override=].
+
```
juliandescottes: We declare `forced colors mode theme overrides map` as a map between `user context` and a struct (`forced colors mode theme override`), but then we directly store `emulation.ForcedColorsModeAutomationTheme / null` in it (so string / null).

Maybe remove the `<dfn>forced colors mode theme override</dfn>` definition and make it a map from user contexts to string?

index.bs
```diff
@@ -5840,6 +5855,79 @@ A [=remote end=] has a <dfn>screen orientation overrides map</dfn> which is a we
...
+      1. [=map/Set=] [=forced colors mode theme overrides map=][|user context|] to |command parameters|["<code>theme</code>"].
+
+      1. [=list/For each=] |top-level traversable| of the list of all [=/top-level traversables=]
+         whose [=associated user context=] is |user context|:
+
+         1. [=list/Append=] |top-level traversable| to |navigables|.
+
+1. Let |theme| be |command parameters|["<code>theme</code>"].
+
+1. If |theme| is null, set |theme| to "<code>none</code>".
```
juliandescottes: suggestion: I would move this step around L5900 and set `|theme|` directly in `forced colors mode theme overrides map`. This way we don't have to handle null -> "none" again in `configure new navigable`

index.bs
```diff
@@ -5840,6 +5855,79 @@ A [=remote end=] has a <dfn>screen orientation overrides map</dfn> which is a we
...
+        params: emulation.SetForcedColorsModeThemeOverrideParameters
+      )
+
+      emulation.SetForcedColorsModeThemeOverrideParameters = {
+        theme: emulation.ForcedColorsModeAutomationTheme / null,
+        ? contexts: [+browsingContext.BrowsingContext],
+        ? userContexts: [+browser.UserContext],
+      }
+
+      emulation.ForcedColorsModeAutomationTheme = "light" / "dark"
```
juliandescottes: nit: I would drop `Automation` from the name

index.bs
```diff
@@ -4892,6 +4892,14 @@ TODO: Move it as a hook in the html spec instead.
    1. If [=geolocation overrides map=] [=map/contains=] |user context|,
       [=set emulated position data=] with |navigable| and [=geolocation overrides map=][|user context|].
 
+   1. If [=forced colors mode theme overrides map=] [=map/contains=] |user context|:
+
+      1. Let |theme| be [=forced colors mode theme overrides map=][|user context|].
+
+      1. If |theme| is null, set |theme| to "<code>none</code>".
```
juliandescottes: See other comment, it would be nice to handle conversion from null to "none" only once in the command.

index.bs
```diff
@@ -5840,6 +5855,79 @@ A [=remote end=] has a <dfn>screen orientation overrides map</dfn> which is a we
...
+   and |command parameters| doesn't [=map/contain=] "<code>context</code>",
+   return [=error=] with [=error code=] [=invalid argument=].
+
+1. Let |navigables| be a [=/set=].
+
+1. If the <code>contexts</code> field of |command parameters| is present:
+
+   1. Let |navigables| be the result of [=trying=] to [=get valid top-level traversables by ids=] with |command parameters|["<code>contexts</code>"].
+
+1. Otherwise, if the <code>userContexts</code> field of |command parameters| is present:
```
```suggestion
1. Otherwise:
```

Based on the earlier checks, if `contexts` is not present, then `userContexts` has to be present at this point.

alisonmaher: That's a good point, although this matches what other emulation commands are doing (example: https://www.w3.org/TR/webdriver-bidi/#command-emulation-setLocaleOverride). Do you think it makes sense to update all to just "Otherwise:", diverge on this one, or leave as-is for consistency?

index.bs
alisonmaher: This matches the naming in the CSS color adjust spec: https://drafts.csswg.org/css-color-adjust-1/#emualte-forced-colors-mode

I think clarifying this is automation related is at least helpful there to avoid confusion with the preferred color scheme of light or dark that gets set when forced colors is enabled (and makes it clear wherever used that it is only related to automation, not the theme set at the OS level, for example).

If you feel strongly, lmk and I suspect we'd want to update that in the CSS spec, too, in that case

index.bs
alisonmaher: I'm guessing you are suggesting to remove null and just accept string. And just do `emulation.ForcedColorsModeAutomationTheme` instead of `emulation.ForcedColorsModeAutomationTheme / null`. That way we don't need to convert none to null"?

That was originally how this was spec'd but it was suggested that using null was more consistent with how this spec is done: https://github.com/w3c/webdriver-bidi/pull/905#discussion_r2152869458. Would you prefer to remove null, or were you suggesting a different approach?

juliandescottes: My main issue is that right now the spec expects the values of the map to be structs, but we only store strings or null. So at least we should remove `forced colors mode theme override`, and make the map between `user contexts` and `string or null`.

> That was originally how this was spec'd but it was suggested that using null was more consistent with how this spec is done: https://github.com/w3c/webdriver-bidi/pull/905#discussion_r2152869458.

Ah I had not read the initial discussion, sorry :) But I think the comment is mostly about the command parameter.
We can accept `emulation.ForcedColorsModeAutomationTheme / null` in the command, but still store the translated string "none" in the map. That avoids duplicating the logic in two places. So my suggestion is: keep `emulation.ForcedColorsModeAutomationTheme / null` in the command parameter but only store a string in the map.

Follow up question: Since the map only contains entries for user contexts, could we just remove entries from the map when `null` is provided? Is there a scenario where it's useful to apply the "none" theme to a newly created navigable ?

index.bs
juliandescottes: Let's leave that to a follow up issue. Sounds good to have "Automation" in the css-color-adjust spec but not sure we need to match it on our side. Maybe that could use a NOTE with a link to the CSS spec type.

index.bs
juliandescottes: I think`setLocaleOverride` also just uses `Otherwise:` (step 7)?

Although it adds a `Assert the userContexts field of command parameters is present.` step right below (which really feels unnecessary here).  `setTimezoneOverride` override also follows the same pattern.

Only `setScreenOrientationOverride` is inconsistent here and uses `Otherwise, if the <code>userContexts</code> field of |command parameters| is present:`. 

I will file an issue to fix `setScreenOrientationOverride`, here let's use the pattern from `setLocaleOverride` and `setTimezoneOverride`.  You can also add the `Assert ...` step for consistency, but I'll mention it in the issue to see if we can get rid of it.

index.bs
SebastianZ: For reference, I just merged https://github.com/w3c/csswg-drafts/pull/12520. So, if this spec. wants to reference it, please make sure the link is correct!

Sebastian

index.bs
alisonmaher: I see, I've updated to remove that definition and map to string or null instead. I'm not sure that it allows us to remove those conversions. Will comment on why in the other related discussion

index.bs
alisonmaher: Gotcha, yeah I guess we could have those use different names since here, it is implied to be related to automation. I've updated the name and added a note. Lmk if you have any edits based on the updates.

index.bs
alisonmaher: This conversion is needed because "Set emulated forced colors theme data" is on the CSS side and expects none, light or dark. So I think we need this conversion here and above, but lmk if I'm misunderstanding the suggestion

index.bs
alisonmaher: I see, yeah I guess they aren't all consistent already. Update to just "Otherwise:", thanks!

index.bs
alisonmaher: > Follow up question: Since the map only contains entries for user contexts, could we just remove entries from the map when null is provided? Is there a scenario where it's useful to apply the "none" theme to a newly created navigable ?

Whoops, sorry, forgot to comment on this piece. The forced colors spec in CSS will check the type and if not none, emulate. So I think it is helpful to set this no matter what so that we do the correct calculations later on in that spec, but open to other suggestions there

index.bs
juliandescottes: My suggestion is not to remove the conversion, but to do it earlier in the command and to store the converted value in the map. With or without my suggestion we will use the same values when call  "Set emulated forced colors theme data". 

If I call the command with theme = null and userContexts = ["foo"]:
- current PR: 
  - store `null` in the map
  - convert `null` to "none"
  - call CSS endpoint with "none" for all existing and relevant navigables
  - when navigable is created in "foo": 
    - get theme from map -> null
    - convert `null` to "none"
    - call CSS endpoint with "none"
- suggestion: 
  - convert `null` to "none"
  - set "none" in the map
  - call CSS endpoint with "none" for all existing and relevant navigables
  - when navigable is created in "foo": 
    - get theme from map ->"none"
    - call CSS endpoint with "none"

From an implementation point of view, we can implement that more efficiently so that's not a concern. But from a spec standpoint, having the conversion duplicated in 2 different algorithms is unnecessary and means we need to keep them in sync.

But that's not a blocker, the current solution also works. I'll let you have a look and decide what to do with the suggestion :)

index.bs
```diff
@@ -5840,6 +5852,81 @@ A [=remote end=] has a <dfn>screen orientation overrides map</dfn> which is a we
...
+
+The [=remote end steps=] with |command parameters| are:
+
+1. If |command parameters| [=map/contains=] "<code>userContexts</code>"
+   and |command parameters| [=map/contains=] "<code>context</code>",
+   return [=error=] with [=error code=] [=invalid argument=].
+
+1. If |command parameters| doesn't [=map/contain=] "<code>userContexts</code>"
+   and |command parameters| doesn't [=map/contain=] "<code>context</code>",
+   return [=error=] with [=error code=] [=invalid argument=].
```
```suggestion
1. If |command parameters| [=map/contains=] "<code>userContexts</code>"
   and |command parameters| [=map/contains=] "<code>contexts</code>",
   return [=error=] with [=error code=] [=invalid argument=].

1. If |command parameters| doesn't [=map/contain=] "<code>userContexts</code>"
   and |command parameters| doesn't [=map/contain=] "<code>contexts</code>",
   return [=error=] with [=error code=] [=invalid argument=].
```

juliandescottes (approved):
thanks! One last nit about context -> contexts in the parameter validation, and as mentioned the suggestion about the null->none conversion doesn't have to block merging this.

index.bs
alisonmaher: Hmm this may be a stupid question, but quick clarification. If we are always setting the value to "none" in the map, when is it ever going to be null? In which case, would we need to convert (or support null to being with). Or is there some place where things are added to the map by default with null that I may not be familiar with?

index.bs
alisonmaher: Great catch, thanks! Update both to `contexts`

index.bs
juliandescottes: We want to support `null` as the command parameter, for consistency with the other commands which already support `null` as a means to "reset" the emulation. That's the part which is exposed to clients and libraries and we should strive to have a consistent API. Past that point, it's all just internal implementation, I don't think there is any reason to set `null` in the map.

index.bs
alisonmaher: Removed this instance

index.bs
alisonmaher: Thanks for the clarification, that makes more sense now. So essentially it is about the input type as opposed to the stored type. Updated as suggested so now we covert before storing the value in the map. Thanks for the suggestion

alisonmaher:
[juliandescottes](https://github.com/juliandescottes) Thanks for the review, all comments should be addressed with the latest commit. Please let me know if you have any other edits. If all looks good, if you, or one of the other reviewers is able to help merge, that would be great since I don't have permissions for that. Thanks again for all the great feedback!

index.bs
```diff
@@ -5840,6 +5850,81 @@ A [=remote end=] has a <dfn>screen orientation overrides map</dfn> which is a we
...
+
+   1. Let |navigables| be the result of [=trying=] to [=get valid top-level traversables by ids=] with |command parameters|["<code>contexts</code>"].
+
+1. Otherwise:
+
+   1. Let |user contexts| be the result of [=trying=] to [=get valid user contexts=] with |command parameters|["<code>userContexts</code>"].
+
+   1. Let |theme| be |command parameters|["<code>theme</code>"].
+
+   1. If |theme| is null, set |theme| to "<code>none</code>".
```
```suggestion
1. If |command parameters| [=map/contains=] "<code>userContexts</code>"
   and |command parameters| [=map/contains=] "<code>contexts</code>",
   return [=error=] with [=error code=] [=invalid argument=].

1. If |command parameters| doesn't [=map/contain=] "<code>userContexts</code>"
   and |command parameters| doesn't [=map/contain=] "<code>contexts</code>",
   return [=error=] with [=error code=] [=invalid argument=].

1. Let |theme| be |command parameters|["<code>theme</code>"].

1. If |theme| is null, set |theme| to "<code>none</code>".

1. Let |navigables| be a [=/set=].

1. If the <code>contexts</code> field of |command parameters| is present:

   1. Let |navigables| be the result of [=trying=] to [=get valid top-level traversables by ids=] with |command parameters|["<code>contexts</code>"].

1. Otherwise:

   1. Let |user contexts| be the result of [=trying=] to [=get valid user contexts=] with |command parameters|["<code>userContexts</code>"].
```

`theme` needs to be defined before the if/otherwise, because it's also used in the `For each |navigable| of |navigables|`

juliandescottes (approved):
Thanks for addressing my comments 🙏 LGTM with one last thing to update: make sure to define `theme` before the `if` otherwise it will only have a value when users provide `userContexts`.

index.bs
alisonmaher: Ah whoops, you're right, updated!

juliandescottes (approved):
thanks!

juliandescottes:
@alisonmaher looks like the PR needs to be rebased before merging, let us know when it's done.

alisonmaher:
> @alisonmaher looks like the PR needs to be rebased before merging, let us know when it's done.

@juliandescottes  on my end it was showing up as green with no merge conflicts, but just pulled in the latest updates from main just to be safe

juliandescottes:
> > @alisonmaher looks like the PR needs to be rebased before merging, let us know when it's done.
> 
> @juliandescottes on my end it was showing up as green with no merge conflicts, but just pulled in the latest updates from main just to be safe

I still can't rebase and merge the PR 
image

I just pulled your PR and did `git rebase main` and I have conflicts. Can you make sure your `webdriver-bidi` main branch is up to date, and rebase your PR branch on top of it ?

sadym-chromium:
> > > @alisonmaher looks like the PR needs to be rebased before merging, let us know when it's done.
> > 
> > 
> > @juliandescottes on my end it was showing up as green with no merge conflicts, but just pulled in the latest updates from main just to be safe
> 
> I still can't rebase and merge the PR image
> 
> I just pulled your PR and did `git rebase main` and I have conflicts. Can you make sure your `webdriver-bidi` main branch is up to date, and rebase your PR branch on top of it ?

Screenshot 2025-07-29 at 09 21 26

juliandescottes:
@sadym-chromium very odd, I still can't rebase and merge on my end, and as I said I was getting conflicts with the branch. Feel free to land if that works on your side I guess?

juliandescottes:
(ok I didn't think I could select squash in the UI, the dropdown seemed disabled to me...)

alisonmaher:
Thanks for helping land and for all the reviews!

PR #911, opened 2025-04-29
emulate geolocation POSITION_UNAVAILABLE error

sadym-chromium:
WPT tests: https://github.com/web-platform-tests/wpt/pull/52250

I don't think we need to emulate other errors (PERMISSION_DENIED, TIMEOUT), as it can be tested without emulation.

The geolocation emulation spec was written to support any kind of error emulations, which is an overkill, and is tricky to be implemented, so I propose to limit the error emulation only to "POSITION_UNAVAILABLE" on the WebDriver BiDi side.

From geolocation spec:
> If `emulatedPositionData` is a [GeolocationPositionError](https://www.w3.org/TR/geolocation/#dom-geolocationpositionerror):
[Call back with error](https://www.w3.org/TR/geolocation/#dfn-call-back-with-error) passing errorCallback and emulatedPositionData.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5624,8 +5624,10 @@ The <dfn export for=commands>emulation.setGeolocationOverride</dfn> command modi
       )
 
       emulation.SetGeolocationOverrideParameters = {
-        ? coordinates: emulation.GeolocationCoordinates / null,
-        ? error: emulation.GeolocationPositionError,
+        (
+          (coordinates: emulation.GeolocationCoordinates / null) //
+          (error: emulation.GeolocationPositionError)
+        ),
         ? contexts: [+browsingContext.BrowsingContext],
         ? userContexts: [+browser.UserContext],
       }
--- a/index.bs
+++ b/index.bs
@@ -182,6 +182,7 @@ spec: GEOMETRY; urlPrefix: https://drafts.fxtf.org/geometry/
     text: height dimension; url: rectangle-height-dimension
 spec: GEOLOCATION; urlPrefix: https://www.w3.org/TR/geolocation/
   type: dfn
+    text: POSITION_UNAVAILABLE; url: #dom-geolocationpositionerror-position_unavailable
     text: set emulated position data; url: #dfn-set-emulated-position-data
 spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
   type: dfn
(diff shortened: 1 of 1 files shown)


OrKoN (approved):

index.bs
```diff
@@ -5651,9 +5657,31 @@ The <dfn export for=commands>emulation.setGeolocationOverride</dfn> command modi
...
+   "<code>altitude</code>", return [=error=] with [=error code=]
+   [=invalid argument=].
+
+1. If |command parameters| [=map/contains=] "<code>error</code>":
+
+   1. Assert |command parameters|["<code>error</code>"]["<code>type</code>"] equals
+      "<code>positionUnavailable</code>".
+
+   1. Let |emulated position data| be a [=/map=] with value of
+      «"code" → [=POSITION_UNAVAILABLE=], "message" → ""».
```
sadym-chromium, line 5681: Empty message here needed for check
> If `emulatedPositionData` is a [GeolocationPositionError](https://www.w3.org/TR/geolocation/#dom-geolocationpositionerror)

index.bs
```diff
@@ -5623,7 +5624,8 @@ The <dfn export for=commands>emulation.setGeolocationOverride</dfn> command modi
       )
 
       emulation.SetGeolocationOverrideParameters = {
-        coordinates: emulation.GeolocationCoordinates / null,
+        ? coordinates: emulation.GeolocationCoordinates / null,
+        ? error: emulation.GeolocationPositionError,
```
whimboo, line 5628: To make it mutual exclusive and to enforce that at least one entry needs to be present we probably should use a group choice:

```suggestion
    (
      (coordinates: emulation.GeolocationCoordinates / null) //
      (error: emulation.GeolocationPositionError)
    ),
```

It could also be moved out to a separate definition for a better readability.

OrKoN (approved):

sadym-chromium:
@juliandescottes any other concerns?

index.bs
```diff
@@ -5651,9 +5660,34 @@ The <dfn export for=commands>emulation.setGeolocationOverride</dfn> command modi
...
-1. If |command parameters| [=map/contains=] "<code>altitudeAccuracy</code>"
-   and |command parameters| doesn't [=map/contain=] "<code>altitude</code>",
-   return [=error=] with [=error code=] [=invalid argument=].
+1. If |command parameters| [=map/contains=] "<code>coordinates</code>"
+   and |command parameters| [=map/contains=] "<code>error</code>", return [=error=]
+   with [=error code=] [=invalid argument=].
+
+1. If |command parameters| doesn't [=map/contain=] "<code>coordinates</code>"
+   and |command parameters| doesn't [=map/contain=] "<code>error</code>", return
+   [=error=] with [=error code=] [=invalid argument=].
```
whimboo, line 5669: This should be covered by the new group so it's not needed anymore.

sadym-chromium: indeed, thanks

juliandescottes (approved):
LGTM, thanks!

PR #914, opened 2025-05-06
fix(cddl): fix CDDL syntax for emulation.SetGeolocationOverrideParameters

christian-bromann:
I find the current definition of `emulation.SetGeolocationOverrideParameters` somewhat unusual and inconsistent with the style used throughout the rest of the specification.

```cddl
   (
   ),
}
```

Aside from concerns about whether this is valid CDDL syntax—and the fact that some parsers may or may not support it—I recommend refactoring this definition to better align with the established structure used elsewhere in the spec. Specifically, we typically define separate types for success and error cases, then unify them with a top-level choice. For example:

```cddl
script.EvaluateResult = (
  script.EvaluateResultSuccess /
  script.EvaluateResultException
)

script.EvaluateResultSuccess = {
  type: "success",
  result: script.RemoteValue,
  realm: script.Realm
}

script.EvaluateResultException = {
  type: "exception",
  exceptionDetails: script.ExceptionDetails,
  realm: script.Realm
}
```

Applying a similar pattern to the emulation namespace would improve readability, tooling compatibility, and consistency across the specification.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5624,14 +5624,23 @@ The <dfn export for=commands>emulation.setGeolocationOverride</dfn> command modi
         params: emulation.SetGeolocationOverrideParameters
       )
 
-      emulation.SetGeolocationOverrideParameters = {
-        (
-          (coordinates: emulation.GeolocationCoordinates / null) //
-          (error: emulation.GeolocationPositionError)
-        ),
+      emulation.SetGeolocationOverrideParameters = (
+        emulation.SetGeolocationOverrideSuccessParameters /
+        emulation.SetGeolocationOverrideErrorParameters
+      )
+
+      emulation.SetGeolocationOverrideBaseParameters = (
         ? contexts: [+browsingContext.BrowsingContext],
         ? userContexts: [+browser.UserContext],
-      }
+      )
+
+      emulation.SetGeolocationOverrideSuccessParameters = {
+        coordinates: emulation.GeolocationCoordinates / null
+      } & emulation.SetGeolocationOverrideBaseParameters
+
+      emulation.SetGeolocationOverrideErrorParameters = {
+        error: emulation.GeolocationPositionError
+      } & emulation.SetGeolocationOverrideBaseParameters
 
       emulation.GeolocationCoordinates = {
          latitude: -90.0..90.0,


index.bs
```diff
@@ -5624,13 +5624,24 @@ The <dfn export for=commands>emulation.setGeolocationOverride</dfn> command modi
...
+        emulation.SetGeolocationOverrideErrorParameters
+      )
+
+      emulation.SetGeolocationOverrideBaseParameters = (
         ? contexts: [+browsingContext.BrowsingContext],
         ? userContexts: [+browser.UserContext],
+      )
+
+      emulation.SetGeolocationOverrideSuccessParameters = {
+        emulation.SetGeolocationOverrideBaseParameters,
```
whimboo, line 5638: I wonder if we would need a `type` field here as well similar to the evaluation types.

whimboo (commented):
Thanks @christian-bromann. Personally I definitely like that better. I've just a question that you can find inline.

CC @sadym-chromium and @jgraham

index.bs
sadym-chromium: I'm in favor of adding "type: success/error" field.

christian-bromann: Personally IMHO the existence of an error keyword only when the command fails is indication enough for me to understand that something went wrong. I am happy to add a type property to these entities. The current places where we do this are the following:

```cddl
Message = (
  CommandResponse /
  ErrorResponse /
  Event
)

CommandResponse = {
  type: "success",
  id: js-uint,
  result: ResultData,
  Extensible
}

ErrorResponse = {
  type: "error",
  id: js-uint / null,
  error: ErrorCode,
  message: text,
  ? stacktrace: text,
  Extensible
}
```
as well as:
```cddl
script.EvaluateResult = (
  script.EvaluateResultSuccess /
  script.EvaluateResultException
)

script.EvaluateResultSuccess = {
  type: "success",
  result: script.RemoteValue,
  realm: script.Realm
}

script.EvaluateResultException = {
  type: "exception",
  exceptionDetails: script.ExceptionDetails
  realm: script.Realm
}
```
Thoughts?

jgraham: These are command parameters, not responses. However I really think we should name types where possible rather than depending on introspection of the fields (consider e.g. what happens if you have both a `coordinates` _and_ and `error` field in the parameters), so I'm in favour of adding the `type` field.

nvborisenko: The same situation:

```
browser.SetClientWindowStateParameters = {
  clientWindow: browser.ClientWindow,
  (browser.ClientWindowNamedState // browser.ClientWindowRectState)
}

browser.ClientWindowNamedState = (
  state: "fullscreen" / "maximized" / "minimized"
)

browser.ClientWindowRectState = (
  state: "normal",
  ? width: js-uint,
  ? height: js-uint,
  ? x: js-int,
  ? y: js-int,
)
```

And this variant is very appreciated from my point of view who deals with strong typing programming language.

christian-bromann: Thanks for all the feedback 🙏  I've amended the changeset accordingly, please let me know your thoughts!

index.bs
```diff
@@ -5624,13 +5624,20 @@ The <dfn export for=commands>emulation.setGeolocationOverride</dfn> command modi
...
 
-      emulation.SetGeolocationOverrideParameters = {
-        (
-          (coordinates: emulation.GeolocationCoordinates / null) //
-          (error: emulation.GeolocationPositionError)
-        ),
+      emulation.SetGeolocationOverrideParameters = (
         ? contexts: [+browsingContext.BrowsingContext],
         ? userContexts: [+browser.UserContext],
+        (emulation.SetGeolocationOverrideSuccessParameters // emulation.SetGeolocationOverrideErrorParameters)
```
OrKoN, line 5630: I believe SetGeolocationOverrideSuccessParameters and SetGeolocationOverrideErrorParameters have to be groups (`()`), not types (map `{}`).

index.bs
```diff
@@ -5624,13 +5624,20 @@ The <dfn export for=commands>emulation.setGeolocationOverride</dfn> command modi
         params: emulation.SetGeolocationOverrideParameters
       )
 
-      emulation.SetGeolocationOverrideParameters = {
-        (
-          (coordinates: emulation.GeolocationCoordinates / null) //
-          (error: emulation.GeolocationPositionError)
-        ),
+      emulation.SetGeolocationOverrideParameters = (
```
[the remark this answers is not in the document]
OrKoN: Right, I think this should stay a map

index.bs
christian-bromann: That is correct, updated!

index.bs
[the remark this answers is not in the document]
christian-bromann: While it is possible to insert a group into a map, e.g.:
```
foobar = {
  someGroup,
  key: value
}
```

It is not possible to include group choices, because the map expects a single group, not a group choice expression*.

RFC 8610 (Section 3.5) says:
> When a group is used as a member of a map, the entries in the group are included at that location in the map.

But that refers to a group, not a group choice. Group choices (`//`) are only valid inside a group.

index.bs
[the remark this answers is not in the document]
christian-bromann: I am happy to raise another PR with some guidelines.

index.bs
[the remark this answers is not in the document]
OrKoN: `(emulation.SetGeolocationPosition // emulation.GeolocationPositionError)` defines a new group so it is valid to put it into a map. We have other places where this works. The current PR version does provide any type for the params key.

index.bs
```diff
@@ -5624,14 +5624,21 @@ The <dfn export for=commands>emulation.setGeolocationOverride</dfn> command modi
...
+      )
+
+      emulation.SetGeolocationPosition = (
+        type: "position",
+        coordinates: emulation.GeolocationCoordinates / null
+      )
+
+      emulation.GeolocationPositionError = (
+         type: "error"
+         error: emulation.GeolocationPositionUnavailableError
```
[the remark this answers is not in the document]
sadym-chromium: that was my motivation too

index.bs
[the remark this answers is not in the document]
christian-bromann: @OrKoN it is not allowed to have a group choice (`//`) between two maps (`{ ... }`). This aligns with other places in the spec, e.g.

- `(browser.ClientWindowNamedState // browser.ClientWindowRectState)` ([ref](https://github.com/w3c/webdriver-bidi/blob/82b7005a80b946cb7734368cb6feb10cc047b5c9/index.bs#L2861)) where both `browser.ClientWindowNamedState` and `browser.ClientWindowRectState` are groups (`( ... )`)
- `(network.ContinueWithAuthCredentials // network.ContinueWithAuthNoCredentials)` ([ref](https://github.com/w3c/webdriver-bidi/blob/82b7005a80b946cb7734368cb6feb10cc047b5c9/index.bs#L7404)) where also both `network.ContinueWithAuthCredentials` and `network.ContinueWithAuthNoCredentials` are groups

The CDDL specification indirectly defines this in [RFC 8610 §3.9 — Group-to-Group Composition](https://datatracker.ietf.org/doc/html/rfc8610#section-3.9):

> The operator `//` is used to express a choice between groups. The right-hand side of the operator must be a group.

This means:

- `A // B` is only valid if both A and B are groups.
- A map (`{ ... }`) is a type, not a group, and therefore cannot be used directly in `//`.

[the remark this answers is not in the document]
OrKoN: Sorry, I am not following: both SetGeolocationPosition and GeolocationPositionError are groups, not maps. SetGeolocationOverrideParameters has to remain a map:

```
     (emulation.SetGeolocationPosition // emulation.GeolocationPositionError)
}
```

PR #915, opened 2025-05-06
fix(cddl): fix usage of .and operator

christian-bromann:
Section 3.8.5 of [RFC 8610](https://datatracker.ietf.org/doc/html/rfc8610#section-3.8.5) does define two control operators: `.within` and `.and`. These are used in type controls — not as top-level structural merging operators like &.

`.and` is a valid CDDL control operator — but its use is limited. It is a control operator, not a general-purpose group or map merging operator like `&`. `.and` is used to constrain a value to conform to both a structural schema and a semantic type.
It takes two types, and a data item matches the type `X .and Y` if it matches both X and Y.

So this:

```
mytype = text .and tstr
```

Means: the data must match both `text` (the CDDL type) and `tstr` (presumably a semantic constraint/type definition).

It is not valid to write:

```
{ a: int } .and SomeOtherMap
```

IMHO this is not a valid application of `.and`, because you’re not applying it to a value, you’re trying to merge map structures — which is outside `.and`'s purpose.

A valid way to write:

```
color = text .and "red" / "green" / "blue"
```

This would mean: a string, constrained to only be "red", "green", or "blue".

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/915.html" title="Last updated on May 6, 2025, 11:10 PM UTC (4a7735e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/915/5e3b177...4a7735e.html" title="Last updated on May 6, 2025, 11:10 PM UTC (4a7735e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -8412,10 +8412,10 @@ script.LocalValue = (
   script.PrimitiveProtocolValue /
   script.ChannelValue /
   script.ArrayLocalValue /
-  script.DateLocalValue /
+  { script.DateLocalValue } /
   script.MapLocalValue /
   script.ObjectLocalValue /
-  script.RegExpLocalValue /
+  { script.RegExpLocalValue } /
   script.SetLocalValue
 )
 
--- a/index.bs
+++ b/index.bs
@@ -8426,10 +8426,10 @@ script.ArrayLocalValue = {
   value: script.ListLocalValue,
 }
 
-script.DateLocalValue = {
+script.DateLocalValue = (
   type: "date",
   value: text
-}
+)
 
 script.MappingLocalValue = [*[(script.LocalValue / text), script.LocalValue]];
 
@@ -8448,10 +8448,10 @@ script.RegExpValue = {
   ? flags: text,
 }
 
-script.RegExpLocalValue = {
+script.RegExpLocalValue = (
   type: "regexp",
   value: script.RegExpValue,
-}
+)
 
 script.SetLocalValue = {
   type: "set",
@@ -9204,15 +9204,17 @@ script.FunctionRemoteValue = {
   ? internalId: script.InternalId,
 }
 
-script.RegExpRemoteValue = {
+script.RegExpRemoteValue = &(
+  script.RegExpLocalValue,
   ? handle: script.Handle,
   ? internalId: script.InternalId,
-} .and script.RegExpLocalValue
+)
 
-script.DateRemoteValue = {
+script.DateRemoteValue = &(
+  script.DateLocalValue,
   ? handle: script.Handle,
   ? internalId: script.InternalId,
-} .and script.DateLocalValue
+)
 
 script.MapRemoteValue = {
   type: "map",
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -8426,10 +8426,10 @@ script.ArrayLocalValue = {
   value: script.ListLocalValue,
 }
 
-script.DateLocalValue = {
+script.DateLocalValue = (
```
whimboo, line 8429: Mind explaining why these changes are needed?

jimevans: @whimboo IIUC, they're needed because the `.and` operator is being used incorrectly. It is not intended to be a "generic merge operator," which is how we're using it here.

whimboo: I read through the CDDL spec, and it seems that using rounded brackets is indeed necessary when defining a named subgroup that will be embedded into another group.

OrKoN, line 8429: I think DateLocalValue is used in a type choice (/) in LocalValue and it looks like making it a group would not be quite correct?

whimboo (approved):

index.bs
whimboo: Alternatively we could use `{ script.DateLocalValue }` in the definition of `script.LocalValue`.

OrKoN: By reading the CDDL RFC I am not sure it is even allowed to use type choice operator on groups. Our current version of the parser fails to parse this PR.

christian-bromann: That is correct. I am not aware of the limitation that groups can't be used in a type choice.

OrKoN: Testing with something like: 

```
A = (
  a: int
)

B = {
  b: int
}

Root = A / B
```

does not work while 

```
A = {
  a: int
}

B = {
  b: int
}

Root = A / B
``` 

validates correctly.

whimboo: I think what Alex is referring to is the mix of groups and maps, which is not allowed.

OrKoN: Yes, that's what I meant.

> Type operators (such as "/" for type choice)
>   operate on types, while group operators (such as "//" for group
>  choice) operate on groups.  Types can simply be used in groups, but
> groups need to be bracketed (as arrays or maps) to become types

https://datatracker.ietf.org/doc/html/rfc8610#section-3.11

I think we need to fix this before we land this PR (can it be as easy as wrapping groups into `{}` in the LocalValue definition like @whimboo suggested). I think removal of the `.and` struct is correct.

jimevans: > I think what Alex is referring to is the mix of groups and maps, which is not allowed.

@whimboo Yeah, I realized that after the fact, and just didn't delete my comment. Sorry for the wasted cognitive cycles.

whimboo (changes requested):
As discussed we need to wrap the groups into a map for the type choice.

whimboo (approved):
Looks good to me. @OrKoN if you are fine with the changes as well you can merge it.

Thank you @christian-bromann for fixing it!

OrKoN (approved):

Issue #774, opened 2024-09-17
Emulate browser locale (including Accept-Language header, Intl APIs, and Navigator.language)


whimboo:
It would be useful if BiDi clients could fully emulate the browser’s locale, including key functionalities such as:

1. **`Accept-Language` HTTP Headers**:
   - Allow the browser to send the appropriate `Accept-Language` header according to the emulated locale settings.
   
2. **JavaScript APIs** (`Date`, `Intl`):
   - Ensure that JavaScript Date and Internationalization APIs (`Intl`) reflect the correct locale-based formats (e.g., date, currency, numbers).

3. **`Navigator.language`**:
   - Properly simulate the `navigator.language` value in the emulated environment.

#### Expected Behavior:
The emulation should consistently apply the locale across:
- **HTTP Headers**: Emulating `Accept-Language` based on the locale.
- **JS APIs**: Reflecting the locale in Date formatting and `Intl` APIs for things like number and currency formatting.
- **Navigator Properties**: Ensuring that `navigator.language` matches the emulated locale.

#### Use Case:
Many modern web applications use locale information for rendering content, formatting dates/numbers, and providing localized experiences to users. Having full locale emulation is critical for testing applications in different regions and languages to ensure proper functionality.

#### Potential Implementation:
1. **`Accept-Language` Header**:
   - Similar to the ability to set custom HTTP headers via `page.setExtraHTTPHeaders()`, but tied to the locale setting of the browser.

2. **`Intl` and `Date` APIs**:
   - Ensure that JavaScript's `Date` and `Intl` APIs automatically reflect the locale and language settings passed during browser initialization (e.g., `page.setLocale()`).

3. **`Navigator.language`**:
   - Override the `navigator.language` property so that it correctly reflects the locale setting (e.g., `en-US`, `fr-FR`).

#### Impact:
Implementing this feature would significantly improve localization and internationalization (i18n) testing for web applications, as it would allow a more realistic and complete emulation of how the browser behaves with respect to different locales.

sadym-chromium:
I believe the issue is addressed. Feel free to re-open or open a new one if something is not covered.

whimboo:
[...]

PR #924, opened 2025-05-19
`emulation.setLocaleOverride`

sadym-chromium:
Addressing https://github.com/w3c/webdriver-bidi/issues/774. 

Emulate locales in [`Intl`](https://tc39.es/ecma402/#intl-object). Other i18n emulation is out of scope of this PR. 

Emulate locale for browsing contexts or for user contexts.

The ecmascript specification of [Intl](https://tc39.es/ecma402/#intl-object) is tight to [DefaultLocale()](https://tc39.es/ecma402/#sec-defaultlocale), which is an [implementation-defined](https://tc39.es/ecma262/#implementation-defined) abstract operation. This PR defines additional steps in case of the locale override is set.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/924.html" title="Last updated on Jun 26, 2025, 12:07 PM UTC (b71a1d5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/924/f8595a0...b71a1d5.html" title="Last updated on Jun 26, 2025, 12:07 PM UTC (b71a1d5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -127,6 +127,10 @@ spec: CONSOLE; urlPrefix: https://console.spec.whatwg.org
     text: formatter; url: formatter
     text: formatting specifier; url: formatting-specifiers
     text: printer; url: printer
+spec: ECMASCRIPT-I18N; urlPrefix: https://tc39.es/ecma402/
+  type: dfn
+    text: DefaultLocale; url: #sec-defaultlocale
+    text: IsStructurallyValidLanguageTag; url: #sec-isstructurallyvalidlanguagetag
 spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
   type: dfn
     text: Array; url: sec-array-objects
@@ -3116,6 +3120,12 @@ or null and an [=struct/item=] named <dfn attribute for="viewport-configuration"
 
 A [=remote end=] has a <dfn>viewport overrides map</dfn> which is a weak map between [=user contexts=] and [=viewport configuration=].
 
+A [=remote end=] has a <dfn>navigable to locale overrides map</dfn> which is a weak
+map between [=navigables=] and strings.
+
+A [=remote end=] has a <dfn>user context to locale overrides map</dfn> which is a
+weak map between [=user contexts=] and strings.
+
 ### Types ### {#module-browsingcontext-types}
 
 #### The browsingContext.BrowsingContext Type #### {#type-browsingContext-Browsingcontext}
@@ -4869,6 +4879,10 @@ TODO: Move it as a hook in the html spec instead.
       [=set emulated screen orientation=] with |navigable| and
       [=screen orientation overrides map=][|user context|].
 
+   1. If [=user context to locale overrides map=] [=map/contains=] |user context|,
+      [=set emulated locale=] with |navigable| and
+      [=user context to locale overrides map=][|user context|].
+
 1. If [=viewport overrides map=] [=map/contains=] |user context|:
 
    1. If |navigable| is a [=/top-level traversable=] and the <code>viewport</code> field of
@@ -5768,6 +5782,7 @@ relating to emulation of browser APIs.
 <pre class="cddl" data-cddl-module="remote-cddl">
 EmulationCommand = (
   emulation.SetGeolocationOverride //
+  emulation.SetLocaleOverride //
   emulation.SetScreenOrientationOverride
 )
 </pre>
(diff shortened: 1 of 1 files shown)


OrKoN (approved):
LGTM

index.bs
```diff
@@ -3061,6 +3065,12 @@ or null and an [=struct/item=] named <dfn attribute for="viewport-configuration"
 
 A [=remote end=] has a <dfn>viewport overrides map</dfn> which is a weak map between [=user contexts=] and [=viewport configuration=].
 
+A [=remote end=] has a <dfn>navigable to locale overrides map</dfn> which is a weak
+map between [=navigables=] and strings.
+
+A [=remote end=] has a <dfn>user context to locale overrides map</dfn> which is a
+weak map between [=user contexts=] and strings.
```
whimboo, line 3072: Lets make sure that the naming of the map is done similarly to other override maps (see above):

```suggestion
A [=remote end=] has a <dfn>locale overrides map</dfn> which is a
weak map between [=user contexts=] and strings.
```

index.bs
```diff
@@ -3061,6 +3065,12 @@ or null and an [=struct/item=] named <dfn attribute for="viewport-configuration"
 
 A [=remote end=] has a <dfn>viewport overrides map</dfn> which is a weak map between [=user contexts=] and [=viewport configuration=].
 
+A [=remote end=] has a <dfn>navigable to locale overrides map</dfn> which is a weak
+map between [=navigables=] and strings.
```
whimboo, line 3069: Why is that needed? We don't have it eg. for the viewport overrides map.

index.bs
```diff
@@ -5750,6 +5765,104 @@ The [=remote end steps=] with |command parameters| are:
...
+    <code>
+     EmptyResult
+    </code>
+   </dd>
+</dl>
+
+<div>
+To <dfn>set emulated locale</dfn> given |navigable| and |locale|:
+
+1. If |locale| is null, [=map/remove=] |navigable| from
```
whimboo, line 5800: For other emulation commands we do not have a global map from the navigable to the feature. Why do we need it here?

whimboo (commented):
Thank you @sadym-chromium to get started on this API! For the beginning some initial questions and comments.

index.bs
sadym-chromium: There are 2 maps: for user contexts and for navigable, and it is required to distinguish between them

index.bs
```diff
@@ -8217,6 +8330,33 @@ ScriptEvent = (
 )
 </pre>
 
+<div algorithm="updated DefaultLocale steps">
+Whenever the [=DefaultLocale=] algorithm is invoked, the [=remote end=] must run the
```
jgraham: I think this makes it sound worse that it is; the ECMAScript-I18N spec just says that it's implementation defined, so we're just defining part of that algorithm for implementations of WebDriver-BiDi. So I'd say it more like:

```
The [=DefaultLocale=] algorithm is implementation defined. A WebDriver-BiDi [=remote end=] must have an implementation that runs the following steps:
[…]
1. Run implementation-defined steps in accordance with the requirements of the [=DefaultLocale=] specification.
```

index.bs
```diff
@@ -3061,6 +3065,12 @@ or null and an [=struct/item=] named <dfn attribute for="viewport-configuration"
 
 A [=remote end=] has a <dfn>viewport overrides map</dfn> which is a weak map between [=user contexts=] and [=viewport configuration=].
 
+A [=remote end=] has a <dfn>navigable to locale overrides map</dfn> which is a weak
```
jgraham: Is all this state expected to persist after all BiDi sessions are closed? For `accept insecure certs` and `proxy configuration` we store the state on the session, and unset it once that session ends. We aren't doing that for viewports or geolocation override, but I think we probably should be (certainly for geolocation override). Otherwise there isn't any way for a user to reset the state other than starting a new webdriver session and clearing it (well, there could be browser UI, but that's a big lift compared to not leaking state past the end of the session).

index.bs
```diff
@@ -4810,6 +4820,10 @@ TODO: Move it as a hook in the html spec instead.
    1. If [=geolocation overrides map=] [=map/contains=] |user context|,
       [=set emulated position data=] with |navigable| and [=geolocation overrides map=][|user context|].
 
+   1. If [=user context to locale overrides map=] [=map/contains=] |user context|,
```
jgraham: I don't recall if there's any reason to prefer this vs just at point of use checking the navigable first and the navigable's user context second (this might be a better implementation strategy since it could be a perf concern, but from the point of view of the spec, checking at time of use seems easier).

index.bs
sadym-chromium: As long as the emulation is a global state, the user can remove emulations form another session by passing `locale: null`, but they have to be aware of such an override. However, I filed an issue to clarify the behavior in such a case: https://github.com/w3c/webdriver-bidi/issues/938

index.bs
sadym-chromium: thanks, done

index.bs
sadym-chromium: I believe having a single `[=set emulated locale=]` is more flexible, specifically if at some point it will be moved to i18n spec.

index.bs
sadym-chromium: Optionally we can make the `locale overrides map` to have keys of both user contexts and browsing contexts, but it could be harder to read.

sadym-chromium:
@jgraham @whimboo I'm about to start specifying `emulation.setUserAgentOverride`, and it will be quite close to this one, so I'd prefer to have a consensus on this approach first.

index.bs
jgraham: I'm not sure I understand? It would still be a single `[=set emulated locale=]`, but it would look like:

```
[…]
    1. If [=navigable to locale overrides map=] [=map/contains=]
       |top-level traversable|:
       
        1. Return [=navigable to locale overrides map=][|top-level traversable|].

    1. Let |user context| be |top-level-traversable|'s [=associated user context=].

    1. If [=user context to locale overrides map=] [=map/contains=] |user context|:

          1. Return [=user context to locale overrides map=][|user context|]
 […]
```

In practice I think the chance of this moving to the ECMA spec is pretty much zero.

sadym-chromium: I updated the PR so that the `locale overrides map` is checked each use.

sadym-chromium:
> I'm not sure I understand? It would still be a single [=set emulated locale=], but it would look like:

@jgraham done.

sadym-chromium:
Also after some considerations I merged the maps into a single `[=locale overrides map=]` which can store both navigables and user contexts.

PR #925, opened 2025-05-23
Sadym/emulate orientation

sadym-chromium:


index.bs
```diff
@@ -5770,6 +5784,127 @@ The [=remote end steps=] with |command parameters| are:
...
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl remote-cddl">
+      emulation.SetOrientationOverride = (
+        method: "emulation.setOrientationOverride",
+        params: emulation.SetOrientationOverrideParameters
+      )
+
+      emulation.OrientationAngel = (0, 90, 180, 270)
```
OrKoN, line 5801:
```suggestion
      emulation.OrientationAngle = (0, 90, 180, 270)
```

index.bs
```diff
@@ -5770,6 +5784,127 @@ The [=remote end steps=] with |command parameters| are:
...
+
+      emulation.OrientationType = (
+          "portrait-primary",
+          "portrait-secondary",
+          "landscape-primary",
+          "landscape-secondary"
+      )
+
+      emulation.Orientation = {
+        angel: emulation.OrientationAngel,
```
OrKoN, line 5811:
```suggestion
        angle: emulation.OrientationAngle,
```

?

index.bs
```diff
@@ -5770,6 +5784,127 @@ The [=remote end steps=] with |command parameters| are:
...
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl remote-cddl">
+      emulation.SetOrientationOverride = (
+        method: "emulation.setOrientationOverride",
+        params: emulation.SetOrientationOverrideParameters
+      )
+
+      emulation.OrientationAngle = (0, 90, 180, 270)
```
OrKoN, line 5801:
```suggestion
      emulation.OrientationAngle = 0 / 90 / 180 / 270
```

is it equivalent to this? we usually use `/`

index.bs
```diff
@@ -5770,6 +5784,127 @@ The [=remote end steps=] with |command parameters| are:
...
+
+      emulation.OrientationType = (
+          "portrait-primary",
+          "portrait-secondary",
+          "landscape-primary",
+          "landscape-secondary"
+      )
```
OrKoN, line 5808:
```suggestion
      emulation.OrientationType = "portrait-primary" / "portrait-secondary" / "landscape-primary" / "landscape-secondary"
```

OrKoN (approved):
LGTM Thanks!

index.bs
```diff
@@ -5770,6 +5784,122 @@ The [=remote end steps=] with |command parameters| are:
...
+
+<div algorithm="remote end steps for emulation.setOrientationOverride">
+
+The [=remote end steps=] with |command parameters| are:
+
+1. If the implementation is unable to adjust the [=screen orientations=]
+   parameters with the given |command parameters| for any reason, return
+   [=error=] with [=error code=] [=unsupported operation=].
+
+1. If |command parameters| [=map/contains=] "<code>userContexts</code>"
```
OrKoN, line 5860: As discussed, I think from here we could define a re-usable algorithm for a command applicable to userContexts with contexts, starting from here, everything seems exactly the same as for other similar commands and the variable bits are:

1) per-user-context steps like `[=map/Set=] [=orientation overrides map=][|user context|] to |emulated orientation|.`
2) per-navigable steps like `[=Set emulated orientation=] with |navigable| and |emulated orientation|`.

It would be great to try to refactor the spec and de-duplicate these common steps.

whimboo (changes requested):
Generally looks fine for me as well. But there is one thing which doesn't look correct to me. Please see the inline comment.

index.bs
```diff
@@ -5770,6 +5784,124 @@ The [=remote end steps=] with |command parameters| are:
...
+        params: emulation.SetScreenOrientationOverrideParameters
+      )
+
+      emulation.ScreenOrientationAngle = 0 / 90 / 180 / 270
+
+      emulation.ScreenOrientationType = "portrait-primary" / "portrait-secondary" / "landscape-primary" / "landscape-secondary"
+
+      emulation.ScreenOrientation = {
+        angle: emulation.ScreenOrientationAngle,
+        type: emulation.ScreenOrientationType
```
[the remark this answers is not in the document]
whimboo: Ah ok. So what if we would hide the angles and instead have a field that just allows to set `natural portrait` or `natural landscape`? Then we wouldn't have to expose the angles at all, and it should be clear what is actually set. Right now we would have to do validity checks and raise an error in case of invalid combinations like [landscape-secondary](https://w3c.github.io/screen-orientation/#dfn-secondary) and `0` degrees. This makes it complicated to use and to remember which values are actually valid.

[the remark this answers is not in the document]
whimboo: Yes, that is what I was thinking of. And even through we cannot rotate the device with emulating the `natural orientation` it allows to more easily set the target screen orientation by consumers because it's better to imagine?

So if there is uncertainty lets get some feedback from @jgraham as well.

index.bs
```diff
@@ -5662,6 +5673,9 @@ A <dfn>geolocation override</dfn> is a [=struct=] with:
 
 A [=remote end=] has a <dfn>geolocation overrides map</dfn> which is a weak map between [=user contexts=] and [=geolocation override=].
 
+A [=remote end=] has a <dfn>screen orientation overrides map</dfn> which is a weak
+map between [=user contexts=] and [=/map=] matching the
```
jgraham: Internally we should avoid using protocol values directly, but convert them into primitives or structs.

index.bs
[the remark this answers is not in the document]
sadym-chromium: > Although personally I don't really like exposing natural orientation here, because this command does not update the viewport to change the natural orientation...

Screen orientation refers to the physical orientation of a device's display, while viewport orientation is the current orientation of the browser's visible area. This makes screen and viewport's orientations independent.

index.bs
jgraham: I think there's something we can do, but I'm not a huge fan of specs that define algorithms that take callbacks; I think it makes the logic harder to follow.

index.bs
```diff
@@ -5770,6 +5784,153 @@ The [=remote end steps=] with |command parameters| are:
...
+
+   1. Set |top-level navigable|'s [=current orientation type=] to
+      implementation-defined default.
+
+1. Otherwise:
+
+   1. Let |emulated orientation type| be
+      |emulated screen orientation|["<code>type</code>"].
+
+   1. If |emulated screen orientation|["<code>natural</code>"] equals
```
jgraham: Can we look this up in https://www.w3.org/TR/screen-orientation/#dfn-screen-orientation-values-lists instead of repeating the values here? Not a blocker, but might be nicer.

index.bs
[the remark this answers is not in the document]
jgraham: I think what we have makes sense (but I'm not an expert in this area).

index.bs

index.bs

sadym-chromium:
@whimboo is it still blocking?

index.bs
```diff
@@ -5855,6 +5874,129 @@ The [=remote end steps=] with |command parameters| are:
...
+   <dd>
+    <code>
+     EmptyResult
+    </code>
+   </dd>
+</dl>
+
+<div algorithm>
+
+To <dfn>set emulated screen orientation</dfn> given |emulated screen orientation|:
```
whimboo: This method also needs the navigable as argument.

PR #927, opened 2025-05-28
Adapt CDDL declarations to new version of Bikeshed

tidoust:
Some time ago, I mentioned that I had been working on adding support for CDDL highlighting and linking in Bikeshed. That's now done in [version 5.2.0 of Bikeshed](https://github.com/speced/bikeshed/issues/1773#issuecomment-2914475958), released yesterday, see [Bikeshed documentation](https://speced.github.io/bikeshed/#cddl) for details.

This update makes a minimal set of changes to adjust the CDDL declarations to define the CDDL modules and blocks in a way that Bikeshed understands. In particular, Bikeshed needs the blocks to reference the CDDL modules through a `data-cddl-module` attribute (as opposed to them appearing in the `class` attribute).

Bikeshed now also generates a CDDL index at the end of the spec. This CDDL index is divided into a remote end definition part and a local end definition part. Each part contains all definitions defined as applying to the underlying module. Generation of the index can be disabled if needed.

The definitions of the CDDL modules themselves require using a new `cddl-module` definition type. Changing the definition type means that the resulting IDs and the way to reference the definitions from other specs changes. Specs that reference these definitions in WebDriver BiDi need a small update accordingly.

Further updates are doable to start linking the spec prose with the CDDL definition where that makes sense. The auto-linking part already allows to follow CDDL structures that appear in the blocks, which I personally prefer to the raw blocks that the spec had otherwise.

I also updated the script that extracts the CDDL definitions. Now, [CDDL extracts](https://github.com/w3c/webref/tree/main/ed/cddl) are now automatically published in Webref every 6 hours (and validated using the CDDL parser I developed for integration in Bikeshed). I don't know to what extent that script is still used and useful.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/927.html" title="Last updated on Jun 11, 2025, 8:06 AM UTC (c1f3e3e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/927/bd16afc...c1f3e3e.html" title="Last updated on Jun 11, 2025, 8:06 AM UTC (c1f3e3e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5424,7 +5424,7 @@ started</dfn> steps given |navigable| and |navigation status|:
 <dl>
    <dt>Event Type</dt>
    <dd>
-      <pre class="cddl local-cddl">
+      <pre class="cddl" data-cddl-module="local-cddl">
         browsingContext.DownloadEnd = (
           method: "browsingContext.downloadEnd",
           params: browsingContext.DownloadEndParams
--- a/index.bs
+++ b/index.bs
@@ -413,8 +413,8 @@ protocol. These terms are distinct from their representation at the
 
 The protocol is defined using a [[!RFC8610|CDDL]] definition. For the
 convenience of implementers two separate CDDL definitions are defined; the
-<dfn export>remote end definition</dfn> which defines the format of messages produced
-on the [=local end=] and consumed on the [=remote end=], and the <dfn export>local end
+<dfn cddl-module export lt="remote end definition|remote-cddl">remote end definition</dfn> which defines the format of messages produced
+on the [=local end=] and consumed on the [=remote end=], and the <dfn cddl-module export lt="local end definition|local-cddl">local end
 definition</dfn> which defines the format of messages produced on the [=remote end=]
 and consumed on the [=local end=]
 
@@ -428,7 +428,7 @@ the remainder of the specification.
 
 [=Remote end definition=]
 
-<pre class="cddl remote-cddl">
+<pre class="cddl" data-cddl-module="remote-cddl">
 Command = {
   id: js-uint,
   CommandData,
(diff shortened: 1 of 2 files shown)


whimboo:
@tidoust Thank you for creating this PR! I had a quick look and as well checked the preview. Could you explain why the published page has CDDL highlighting and the preview doesn't have?

tidoust:
> @tidoust Thank you for creating this PR! I had a quick look and as well checked the preview. Could you explain why the published page has CDDL highlighting and the preview doesn't have?

Oh, I hadn't noticed that the PR preview was wrong. PR Preview uses https://api.csswg.org/bikeshed/ to generate the spec. That service still had the previous version of Bikeshed from March when I created the PR, so no CDDL support. The service updates itself automatically and now runs the latest version. I don't know how to re-generate the PR preview though.

whimboo:
Ah great! You probably only have to push a small change that you can revert right after again. It should then re-create the preview.

OrKoN:
We were using these scripts to also process [external specifications](https://w3c.github.io/webdriver-bidi/#external-specifications) with CDDL declarations such as https://webbluetoothcg.github.io/web-bluetooth/ and https://www.w3.org/TR/permissions/ We would need to update CDDL declarations in there too.

OrKoN (approved):

tidoust:
> We would need to update CDDL declarations in there too.

Yes, I'm planning to propose updates to these specifications as well once this PR is merged.

OrKoN:
Is this PR ready to be merged?

tidoust:
It should be!

PR #932, opened 2025-06-11
Remove FTP proxy which is no longer supported by browsers

whimboo:
See https://github.com/w3c/webdriver/issues/1900 for the discussion.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1740,7 +1740,6 @@ session.DirectProxyConfiguration = (
 
 session.ManualProxyConfiguration = (
    proxyType: "manual",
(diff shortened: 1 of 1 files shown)


whimboo:
@christian-bromann, @titusfortner, and @diemol do you have any objections about removing FTP support for proxy? If not do you want to get new releases out for your Selenium bindings first or will you just follow up?

Note that the same also applies for WebDriver classic (https://github.com/w3c/webdriver/pull/1901).

diemol:
All good from our side at Selenium. Thank you for the ping, @whimboo!

PR #937, opened 2025-06-19
touch

juliandescottes:
Just a sanity check to verify the build

***
<a href="https://pr-preview.s3.amazonaws.com/juliandescottes/webdriver-bidi/pull/937.html" title="Last updated on Jun 19, 2025, 9:29 AM UTC (ceb7f1d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/937/7b70258...juliandescottes:ceb7f1d.html" title="Last updated on Jun 19, 2025, 9:29 AM UTC (ceb7f1d)">Diff</a>

PR #939, opened 2025-06-19
Fix references to CDDL module definitions

tidoust:
The CDDL modules are now defined using the "cddl-module" dfn type. This means they can no longer be referenced using the `[= =]` shorthand but now need the `{^ ^}` shorthand.

Internal references were missed in previous update (and did not trigger any error during the build because the xref database still contained the previous definition of type `dfn`).

Via
https://github.com/w3c/webdriver-bidi/pull/937#issuecomment-2987413818

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/939.html" title="Last updated on Jun 19, 2025, 10:32 AM UTC (2fcce96)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/939/7b70258...2fcce96.html" title="Last updated on Jun 19, 2025, 10:32 AM UTC (2fcce96)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -413,8 +413,8 @@ protocol. These terms are distinct from their representation at the
 
 The protocol is defined using a [[!RFC8610|CDDL]] definition. For the
 convenience of implementers two separate CDDL definitions are defined; the
-<dfn cddl-module export lt="remote end definition|remote-cddl">remote end definition</dfn> which defines the format of messages produced
-on the [=local end=] and consumed on the [=remote end=], and the <dfn cddl-module export lt="local end definition|local-cddl">local end
+<dfn cddl-module export lt="Remote end definition|remote end definition|remote-cddl">remote end definition</dfn> which defines the format of messages produced
+on the [=local end=] and consumed on the [=remote end=], and the <dfn cddl-module export lt="Local end definition|local end definition|local-cddl">local end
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -413,20 +413,20 @@ protocol. These terms are distinct from their representation at the
 
 The protocol is defined using a [[!RFC8610|CDDL]] definition. For the
 convenience of implementers two separate CDDL definitions are defined; the
-<dfn cddl-module export lt="remote end definition|remote-cddl">remote end definition</dfn> which defines the format of messages produced
-on the [=local end=] and consumed on the [=remote end=], and the <dfn cddl-module export lt="local end definition|local-cddl">local end
+<dfn cddl-module export lt="Remote end definition|remote end definition|remote-cddl">remote end definition</dfn> which defines the format of messages produced
```
tidoust, line 416: Note: Bikeshed is case sensitive for anything that is not a generic dfn, so uppercase variant needs to be specified as well.

jgraham: That's quite annoying :)

jgraham (approved):

sadym-chromium (approved):
LGTM as long as the build is fixed

PR #940, opened 2025-06-24
Resume blocked requests if no active session can resume anymore

### LINKED ISSUE #936: Cleanup intercepted requests on session end
On Firefox side, we currently have a leak for some network.addIntercept wdspec tests (https://bugzilla.mozilla.org/show_bug.cgi?id=1963548)
In many addIntercept tests, we simply add intercepts and block requests, but we never resume them from the tests. 
And we haven't implemented any logic to unblock the requests when our session ends either.

It seems that unblocking requests at the end of the session addresses the leak on our side. 
Would it make sense to update the spec to unblock requests (either by resuming or canceling) when a session ends if no other session has this request in its blocked request map?

cc @sadym-chromium  @OrKoN does Chrome implement any cleanup for blocked requests when session ends at the moment?

juliandescottes:
Fixes #936 

On session cleanup (which happens after the session was removed from the active sessions in session end), resume requests which can no longer be resumed by any other active session.

***
<a href="https://pr-preview.s3.amazonaws.com/juliandescottes/webdriver-bidi/pull/940.html" title="Last updated on Jun 25, 2025, 1:21 PM UTC (a616e87)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/940/6120680...juliandescottes:a616e87.html" title="Last updated on Jun 25, 2025, 1:21 PM UTC (a616e87)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1646,6 +1646,11 @@ To <dfn>cleanup the session</dfn> given |session|:
    1. [=map/Remove=] |session|'s
       [=user context to proxy configuration map=][|user context|].
 
+1. For each |request id| in |session|'s [=blocked request map=]:
+
+   1. [=Resume=] with "<code>continue request</code>", |request id|
+      and (null, "<code>incomplete</code>").
+
 1. For each |collector| in |session|'s [=network collectors=]:
 
    1. Let |collector id| be |collector|'s <code>collector</code>.
--- a/index.bs
+++ b/index.bs
@@ -1646,10 +1646,11 @@ To <dfn>cleanup the session</dfn> given |session|:
    1. [=map/Remove=] |session|'s
       [=user context to proxy configuration map=][|user context|].
 
-1. For each |request id| in |session|'s [=blocked request map=]:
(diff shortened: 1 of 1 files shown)


juliandescottes:
As discussed with @whimboo on Matrix, by design a given request can probably only be in one blocked request map. So we probably don't even have to check other sessions?

index.bs
```diff
@@ -1646,6 +1646,11 @@ To <dfn>cleanup the session</dfn> given |session|:
    1. [=map/Remove=] |session|'s
       [=user context to proxy configuration map=][|user context|].
 
+1. For each |request id| in |session|'s [=blocked request map=]:
+
+   1. [=Resume=] with "<code>continue request</code>", |request id|
+      and (null, "<code>incomplete</code>").
```
jgraham: Is `null` here correct? Don't we need to continue with request's existing response, like in the `continueResponse` command?

juliandescottes: You're right! I thought that since we don't want to update the response we could just return null, but this is wrong. Will update.

PR #942, opened 2025-06-25
Allow "sameSite" cookie property to have value "default".

lutien:
In case, setting the cookie with `document.cookie` Firefox doesn't set `sameSite` property. At the moment for BiDi `storage.getCookies` API [we fall back to `sameSite` property to `None`](https://bugzilla.mozilla.org/show_bug.cgi?id=1971488) but we would like to have a separate value "default" in this case to better reflect the actual browser behavior. 
This PR allows `sameSite` property to have the value `default` in this case.

***
<a href="https://pr-preview.s3.amazonaws.com/lutien/webdriver-bidi/pull/942.html" title="Last updated on Jun 30, 2025, 10:45 AM UTC (df3155c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/942/7831d08...lutien:df3155c.html" title="Last updated on Jun 30, 2025, 10:45 AM UTC (df3155c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -325,17 +325,6 @@ spec: STREAMS; urlPrefix: https://streams.spec.whatwg.org/
     text: ReadableStream; url: #readablestream
 </pre>
 
-<pre class="biblio">sa
-{
-    "SAME-SITE-COOKIES": {
-        "authors": ["Steven Bingler", "Mike West", "John Wilander"],
-        "href": "https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-20",
-        "publisher": "IETF",
-        "title": "Cookies: HTTP State Management Mechanism"
-    }
-}
-</pre>
-
 <style>
 var {
   color: #cd5c5c
@@ -6870,9 +6859,6 @@ The <code>network.Cookie</code> type represents a cookie.
 
 <div algorithm>
 To <dfn>serialize cookie</dfn> given |stored cookie|:
-
-Note: The definitions of |stored cookie|'s fields are from [[COOKIES]], except
-same-site-flag, which is from [[SAME-SITE-COOKIES]].
 
 1. Let |name| be the result of [=UTF-8 decode=] with |stored cookie|'s name field.
 


index.bs
```diff
@@ -6278,7 +6278,7 @@ network.Cookie = {
     size: js-uint,
     httpOnly: bool,
     secure: bool,
-    sameSite: network.SameSite,
+    ? sameSite: network.SameSite,
     ? expiry: js-uint,
```
```suggestion
    ? expiry: js-uint,
    ? sameSite: network.SameSite,
```

index.bs
```diff
@@ -6312,7 +6312,7 @@ samesite-flag, which is from [[SAME-SITE-COOKIES]].
 1. Let |secure| be true if |stored cookie|'s secure-only-flag is true, or false
    otherwise.
 
-1. Let |same site| be "<code>none</code>" if |stored cookie|'s samesite-flag is
+1. Let |same site| be null if not set, otherwise "<code>none</code>" if |stored cookie|'s samesite-flag is
```
whimboo: We should only include it into the production map when it is not `null`.

whimboo (changes requested):

index.bs
```diff
@@ -6323,7 +6323,7 @@ samesite-flag, which is from [[SAME-SITE-COOKIES]].
    |expiry| if it's not null, or omitted otherwise, the <code>size</code> field
    set to |size|, the <code>httpOnly</code> field set to |http only|, the
    <code>secure</code> field set to |secure|, and the <code>sameSite</code>
-   field set to |same site|.
+   field set to |same site| if it's not null.
```
```suggestion
   field set to |same site| if it's not null, or omitted otherwise.
```

index.bs
juliandescottes: I don't understand the `if not set`? Does it mean if `|stored cookie|'s samesite-flag` is not set?

As I mentioned on bugzilla, at the moment I feel like the samesite cookie spec does not really allow the samesite flag to not be set (https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-cookie-same-site#section-4.2) ? But maybe I'm not reading this correctly.

lutien: > I don't understand the `if not set`? Does it mean if `|stored cookie|'s samesite-flag` is not set?

I've updated the sentence, does it sound better now?

> As I mentioned on bugzilla, at the moment I feel like the samesite cookie spec does not really allow the samesite flag to not be set (https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-cookie-same-site#section-4.2) ? But maybe I'm not reading this correctly.

I think it's probably right that computed value of `SameSite` should be equal to `none`/`lax`/`strict`. And that we expose now an internal value (similar to [web-extensions where it's called `unspecified`](https://developer.chrome.com/docs/extensions/reference/api/cookies#type-SameSiteStatus)). I'm not sure how critical it's, if we maybe should add something to make it clear that we use a different kind of `SameSite` (and if it's ok in general). So summoning @jgraham to take a look as well.

jgraham: OK, so it looks like that draft has now been folded into https://www.ietf.org/archive/id/draft-ietf-httpbis-rfc6265bis-20.html so we should update the reference. That seems to store the values None/Lax/Strict/Default (see step 17 in https://www.ietf.org/archive/id/draft-ietf-httpbis-rfc6265bis-20.html#name-storage-model), so I think Default corresponds to Gecko's "Unset" state, and is probably what we want here.

index.bs
```diff
@@ -6871,7 +6872,7 @@ The <code>network.Cookie</code> type represents a cookie.
 To <dfn>serialize cookie</dfn> given |stored cookie|:
 
 Note: The definitions of |stored cookie|'s fields are from [[COOKIES]], except
-samesite-flag, which is from [[SAME-SITE-COOKIES]].
+same-site-flag, which is from [[SAME-SITE-COOKIES]].
```
jgraham: This is wrong now, I think; we're just following RFC6265bis everywhere?

lutien: Ah, alright, I removed it now. I was just not sure how all these overrides work

lutien:
@sadym-chromium, do you want to take another look? We changed the approach, so would like to confirm that you're still fine with these changes

sadym-chromium (approved):

index.bs
```diff
@@ -6859,8 +6849,8 @@ network.Cookie = {
     size: js-uint,
     httpOnly: bool,
     secure: bool,
-    sameSite: network.SameSite,
     ? expiry: js-uint,
+    sameSite: network.SameSite,
```
whimboo, line 6853: @lutien nit, this entry should have been moved back given that it stays as required and should be listed before optional ones. Mind a quick follow-up patch?

PR #944, opened 2025-07-01
Added note about possibility of capping the cookie expiry value

whimboo:
Equivalent to https://github.com/w3c/webdriver/pull/1907 which was for WebDriver classic.

***
<a href="https://pr-preview.s3.amazonaws.com/whimboo/webdriver-bidi/pull/944.html" title="Last updated on Jul 1, 2025, 2:13 PM UTC (02f49cc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/944/0b2239c...whimboo:02f49cc.html" title="Last updated on Jul 1, 2025, 2:13 PM UTC (02f49cc)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -49,6 +49,7 @@ spec: RFC6265
     text: Cookie store; url: https://httpwg.org/specs/rfc6265.html#storage-model
 spec: RFC6265bis; urlPrefix: https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-20.html
   type: dfn
+    text: Cookie Lifetime Limits; url: #cookie-lifetime-limits
     text: Lax; url: section-4.1.2.7
     text: Strict; url: section-4.1.2.7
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -12334,6 +12335,8 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
    <dd><p>|cookie spec|["<code>expiry</code>"] if it exists, otherwise leave unset to
     indicate that this is a session cookie.
 
+    Note: The cookie&apos;s expiry value might be limited by the remote end in accordance with the [=Cookie Lifetime Limits=].
```
juliandescottes: Just curious, I don't think we have `&apos;`  in other parts of the spec. Is this better/intentional?

whimboo: We have it at least in WebDriver classic and I only copied the content from over there. And indeed in this spec we use the `'` character directly.

PR #946, opened 2025-07-07
`emulation.setScriptingEnabled`

sadym-chromium:
`emulation. setScriptingEnabled` command. Related [HTML PR](https://github.com/whatwg/html/pull/11441) should be landed after this one.

* Store the state in `scripting enabled overrides map`.
* Provide a hook "WebDriver BiDi scripting is enabled" to be called by HTML's ["Scripting is enabled"](https://html.spec.whatwg.org/multipage/webappapis.html#concept-environment-script) method.
* Force scripting enabled for scripts initiated by BiDi protocol user via "bypassDisabledScripting" argument.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/946.html" title="Last updated on Aug 11, 2025, 9:42 PM UTC (a651416)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/946/0a6dc6c...a651416.html" title="Last updated on Aug 11, 2025, 9:42 PM UTC (a651416)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -243,7 +243,6 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
     text: report an error; url: webappapis.html#report-the-error
     text: run the animation frame callbacks; url: imagebitmap-and-animations.html#run-the-animation-frame-callbacks
     text: same origin domain; url: browsers.html#same-origin-domain
-    text: scripting is disabled; url: webappapis.html#concept-environment-noscript
     text: select an image source from a source set; url: images.html#select-an-image-source-from-a-source-set
     text: selected files; url: input.html#concept-input-type-file-selected
     text: session history entry; url: browsing-the-web.html#session-history-entry
@@ -6068,67 +6067,37 @@ disabling JavaScript on web pages.
 Note: only emulation of disabled Javascript is supported.
 
 <div algorithm>
+The <dfn export>WebDriver BiDi scripting is enabled</dfn> steps given
+[=environment settings object=] |settings| are:
 
-To <dfn>emulate javascript enabled status</dfn> given |environment settings| and
-|emulated javascript enabled status|:
+1. Let |navigable| be |settings|'s [=relevant global object=]'s
+   <a>associated <code>Document</code></a>'s [=/node navigable=].
 
-1. Assert |emulated javascript enabled status| is null of false.
+1. Let |top-level traversable| be |navigable|’s [=navigable/top-level traversable=].
 
-1. If |emulated javascript enabled status| is false, perform implementation defined
-   steps to emulate the user disabled scripting for |environment settings|, so that
-   |environment settings|'s [=scripting is disabled=].
+1. If [=javascript enabled overrides map=] contains |top-level traversable|:
 
-1. Otherwise if |emulated javascript enabled status| is null, perform implementation
-   defined steps to restore the default scripting enabled status for
-   |environment settings|.
+   1. If [=javascript enabled overrides map=][|top-level traversable|] is false,
+      return false.
 
-</div>
+   1. Return true.
 
-<div algorithm>
+1. Let |user context| be  |top-level traversable|'s [=associated user context=].
 
-To <dfn>get emulated javascript enabled status</dfn> given |navigables|:
+1. If [=javascript enabled overrides map=] contains |user context|, return
+   [=javascript enabled overrides map=][|user context|].
 
-1. For each |navigable| or |navigables|:
+   1. If [=javascript enabled overrides map=][|user context|] is false, return false.
 
-   1. Let |top-level traversable| be |navigable|’s
-      [=navigable/top-level traversable=].
+   1. Return true.
 
-   1. If [=javascript enabled overrides map=] contains |top-level traversable|,
-      return [=javascript enabled overrides map=][|top-level traversable|].
-
-   1. Let |user context| be  |top-level traversable|'s [=associated user context=].
-
-   1. If [=javascript enabled overrides map=] contains |user context|, return
-      [=javascript enabled overrides map=][|user context|].
-
-   1. Return null.
+1. Return true.
 
 </div>
 
-<div algorithm>
-
-To <dfn>update javascript enabled status</dfn>:
-
-1. For each |environment settings| of [=environment settings objects=]:
-
-   1. Let |related navigables| be the result of [=get related navigables=] given
-      |environment settings|.
-
-   1. Let |emulated javascript enabled status| be the result of
-      [=get emulated javascript enabled status=] given |related navigables|.
-
-   1. [=Emulate javascript enabled status=] given |environment settings| and
-      |emulated javascript enabled status|.
-
-</div>
-
-<div algorithm="remote end steps for emulation.setEnableScriptingOverride">
+<div algorithm="remote end steps for emulation.setJavascriptEnabled">
 
 The [=remote end steps=] with |command parameters| are:
-
-1. If the implementation is unable to adjust the [=scripting is disabled=]
-   state for any reason, return [=error=] with [=error code=]
-   [=unsupported operation=].
 
 1. If |command parameters| [=map/contains=] "<code>userContexts</code>"
    and |command parameters| [=map/contains=] "<code>context</code>",
@@ -6140,12 +6109,10 @@ The [=remote end steps=] with |command parameters| are:
 
 1. Let |emulated javascript enabled status| be null.
 
-1. If |command parameters| [=map/contains=] "<code>enable</code>":
+1. If |command parameters| [=map/contains=] "<code>enabled</code>":
 
    1. Set |emulated javascript enabled status| to
-      |command parameters|["<code>enable</code>"].
-
-1. Let |navigables| be a [=/set=].
+      |command parameters|["<code>enabled</code>"].
 
 1. If the <code>contexts</code> field of |command parameters| is present:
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -5798,6 +5803,109 @@ map between [=user contexts=] and [=screen orientation override=].
...
+   </dd>
+</dl>
+
+<div algorithm>
+
+To <dfn>get emulated script disabled status</dfn> given |navigables|:
+
+1. For each |navigable| or |navigables|:
+
+   1. If [=disable scripting overrides map=] contains |navigable|, return
```
sadym-chromium: TODO: get  top-level navigable.

index.bs
```diff
@@ -5798,6 +5803,109 @@ map between [=user contexts=] and [=screen orientation override=].
...
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl" data-cddl-module="remote-cddl">
+      emulation.SetDisableScriptingOverride = (
+        method: "emulation.setDisableScriptingOverride",
+        params: emulation.SetDisableScriptingOverrideParameters
+      )
+
+      emulation.SetDisableScriptingOverrideParameters = {
+        disable: true / null,
```
jgraham: Only taking `true` or `null` is pretty odd. I think `true` / `false` / `null`, where `null` means "revert to browser default settings" makes more sense.

index.bs
```diff
@@ -5798,6 +5803,109 @@ map between [=user contexts=] and [=screen orientation override=].
 
 ### Commands ### {#module-emulation-commands}
 
+#### The emulation.setDisableScriptingOverride Command ####  {#command-emulation-setDisableScriptingOverride}
+
+The <dfn export for=commands>emulation.setDisableScriptingOverride</dfn> command
```
jgraham: At risk of bikeshedding names, what about `emulation.setJavascriptEnabled`? I don't think we need to worry about non-JS languages, and unlike the other commands in this module, this seems more like _configuration_ than providing mock ("override") data (tbh I'd also be OK with it being in the `browsingContext` module, even though you can apply it to user contexts).

index.bs
jgraham: It makes sense to me that this would work on a per-navigable basis, not just for top-level traversables.

sadym-chromium: what about inheritance? What should it do in case of the emulation is enabled for a child navigable, but then explicitly disabled for the parent? I believe we had more-or-less the same issue with subscriptions and ended up to stick to the top-level traversables for the same reason.

index.bs
sadym-chromium: So my concern is if we want to force-enable JS in case of the user disabled it.

index.bs
sadym-chromium: Agree, but only in case if we agree on `disable: bool / null` instead of `disable: true / null`.

OrKoN:
> Requires extending HTML "create a classic script" algorithm with a flag "bypass disabled scripting".

would expressions or functions evaluated via WebDriver BiDi still work with this feature?

Could you please link the issue this PR addreses in the description?

OrKoN:
> > Requires extending HTML "create a classic script" algorithm with a flag "bypass disabled scripting".
> 
> would expressions or functions evaluated via WebDriver BiDi still work with this feature?
> 
> Could you please link the issue this PR addreses in the description?

Is there a draft for the HTML spec?

index.bs
whimboo: All the existing APIs (maybe I miss one?) require a top-level browsing context to be specified. So it would be the first API that would behave differently. Do we really want such a mix?

sadym-chromium:
> would expressions or functions evaluated via WebDriver BiDi still work with this feature?

Yes, it should.

> Is there a draft for the HTML spec?

Added to the description: https://github.com/whatwg/html/pull/11441

index.bs
OrKoN: wdyt about `enabled: bool / null` and `emulation.setJavascriptEnabled` or `setScriptingEnabled`?

sadym-chromium: I added an open question to the description. As mentioned, I don't mind the name, but rather the behavior.

index.bs
sadym-chromium: > At risk of bikeshedding names

I would kindly ask you to propose the better names if you think of any for any of my proposals. It would allow for save some mental effort for the protocol users in the future )

index.bs
jgraham: Scripting is specifically tied to a navigable (each navigable has its own global, etc.) so in this case I'm not sure what the problem would be. Arguably there aren't clear use cases for per-navigable overrides but I don't think there's the same level of implementation complexity as with other features.

index.bs
jgraham: I think if you have an untrusted webdriver session connected to your browser there are more significant problems than "maybe it will re-enable scripting".

index.bs
```diff
@@ -5798,6 +5803,109 @@ map between [=user contexts=] and [=screen orientation override=].
...
+Note: TODO check if `disable: false` make sense.
+Note: TODO check if the operation is supported.
+
+The [=remote end steps=] with |command parameters| are:
+
+1. If |command parameters| [=map/contains=] "<code>userContexts</code>"
+   and |command parameters| [=map/contains=] "<code>context</code>",
+   return [=error=] with [=error code=] [=invalid argument=].
+
+1. If |command parameters| doesn't [=map/contain=] "<code>userContexts</code>"
```
jgraham: It seems like it makes sense for this switch to work globally as well (maybe this is also the case for some other emulation features).

OrKoN:
> Should protocol ENABLE JavaScript if the user disabled it before?

I would say yes, do we have a precedent in CDP for it?

sadym-chromium:
> I would say yes, do we have a precedent in CDP for it?

CDP does not allow for enabling the script execution if it was disabled by the user:
1. User setting is reflected in [embedder_script_enabled_](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/inspector/dev_tools_emulator.h;drc=17252ebab8c8f8977b19616706fa98dcf4d9d7ae;l=156).
2. Which should be `true` in order to enable script execution: [`SetScriptEnabled(!script_execution_disabled_ && embedder_script_enabled_)`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/inspector/dev_tools_emulator.cc;drc=17252ebab8c8f8977b19616706fa98dcf4d9d7ae;l=558).

index.bs
sadym-chromium: I filed an issue for that: https://github.com/w3c/webdriver-bidi/issues/956

index.bs
jgraham: I'm not quite sure what's wrong with `setJavascriptEnabled` with a boolean `enabled` that's either `false` or `null`.

sadym-chromium: > I'm not quite sure what's wrong with `setJavascriptEnabled` with a boolean `enabled` that's either `false` or `null`.

nothing, it should work fine

index.bs
```diff
@@ -5903,6 +5908,152 @@ The [=remote end steps=] with |command parameters| are:
...
+
+The <dfn export for=commands>emulation.setJavascriptEnabled</dfn> command emulates
+disabling JavaScript on web pages.
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl" data-cddl-module="remote-cddl">
+      emulation.SetJavascriptEnabled = (
+        method: "emulation.setJavascriptEnabled",
```
whimboo: Really bikeshedding but it's called `JavaScript`. So we maybe want to use a capital `S` as well?

```suggestion
      emulation.SetJavaScriptEnabled = (
        method: "emulation.setJavaScriptEnabled",
```

sadym-chromium: We have precedents of using both [`JavaScript`](https://www.w3.org/TR/webdriver-bidi/#cddl-type-logjavascriptlogentry:~:text=type%20represents%20a-,JavaScript,-exception.) and [`Javascript`](https://www.w3.org/TR/webdriver-bidi/#cddl-type-logjavascriptlogentry), so I'd stick to  `Javascript` unless you insist on `JavaScript`.

whimboo: Ah, renaming would make it complicated for the https://www.w3.org/TR/webdriver-bidi/#cddl-key-logjavascriptlogentry-type - and cause backward incompatible changes. So lets leave as it is.

index.bs
sadym-chromium: I believe this discussed on WG with a decision: make it for top-level now with possible extension later on.

index.bs
```diff
@@ -5946,6 +5950,110 @@ The [=remote end steps=] with |command parameters| are:
...
+
+The <dfn export for=commands>emulation.setScriptingEnabled</dfn> command emulates
+disabling JavaScript on web pages.
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl" data-cddl-module="remote-cddl">
+      emulation.SetScriptingEnabled = (
+        method: "emulation.setScriptingEnabled",
```
sadym-chromium: Method name aligns with HTML's ["Scripting is enabled"](https://html.spec.whatwg.org/multipage/webappapis.html#concept-environment-script).

sadym-chromium:
@sadym-chromium @jgraham @whimboo please take another look at re-worked version.

index.bs
```diff
@@ -11542,7 +11650,8 @@ Note: the |function declaration| is parenthesized and evaluated.
    «"<code>(</code>", |function declaration|, "<code>)</code>"»
 
 1. Let |function script| be the result of [=create a classic script=] with
-   |parenthesized function declaration|, |environment settings|, |base URL|, and |options|.
+   |parenthesized function declaration|, |environment settings|, |base URL|,
+   |options| and <code>bypass disabled scripting</code> of true.
```
juliandescottes: `bypass disabled scripting of true` feels weird when reading it, not sure how to pass those optional arguments. Maybe `let |bypassDisabledScripting| be false` and then pass `|bypassDisabledScripting|` ?

whimboo:
@sadym-chromium I assume you are going to provide wpt tests for this API?

sadym-chromium:
> @sadym-chromium I assume you are going to provide wpt tests for this API?

yes, I'm on it

sadym-chromium:
> @sadym-chromium I assume you are going to provide wpt tests for this API?

https://github.com/web-platform-tests/wpt/pull/54289

PR #947, opened 2025-07-07
`emulation.setUserAgentOverride`

sadym-chromium:
Initial draft of `emulation.setUserAgentOverride` command. Requires defining a list of related navigables for the "default User-Agent value" caller.


index.bs
```diff
@@ -6146,6 +6153,125 @@ The [=remote end steps=] with |command parameters| are:
...
+    </code>
+   </dd>
+</dl>
+
+<div algorithm="updated `User-Agent` getter steps">
+The [=default User-Agent value=] getter steps is implementation defined. A
+WebDriver-BiDi [=remote end=] must have an implementation that runs the following
+steps:
+
+Issue: get |related navigables|.
```
sadym-chromium: I found 2 usages of the ["default `User-Agent` value"](https://fetch.spec.whatwg.org/#default-user-agent-value):
1. [Set a header](https://fetch.spec.whatwg.org/#ref-for-default-user-agent-value) for network request.
2. [NavigatorId](https://html.spec.whatwg.org/#navigatorid)'s ["appVersion"](https://html.spec.whatwg.org/#dom-navigator-appversion) and ["userAgent"](https://html.spec.whatwg.org/#dom-navigator-useragent).

We need to fins a way to specify what "related navigables" means in those cases.

OrKoN: I guess we could propose an update to the fetch and html specs? I think there is a way how find related navigable for an HTTP request (like in interception) and for a document, does it work here?

OrKoN: Yeah, I think we should attempt creating patches for the HTML and Fetch spec. WDYT?

juliandescottes: For reference the algorithm we use to map request to navigable is https://www.w3.org/TR/webdriver-bidi/#get-navigable-for-request

To hook from the fetch spec, you can look at the fetch PR at https://whatpr.org/fetch/1540.html#concept-http-fetch (we still haven't landed this so it makes a bit hard to follow the BiDi/fetch integration)

index.bs
```diff
@@ -6146,6 +6156,142 @@ The [=remote end steps=] with |command parameters| are:
...
+
+1. If [=default User-Agent value=] was accessed from [=NavigatorID=] interface:
+
+   1. Let |realm| be [=current Realm Record=].
+
+   1. Let |environment settings| be the [=environment settings object=] whose
+      [=realm execution context=]'s Realm component is |realm|.
+
+   1. Let |related navigables| be the result of [=get related navigables=] given
+      |environment settings|.
```
sadym-chromium: WDYT about this approach? @OrKoN @juliandescottes @jgraham ?

index.bs
sadym-chromium: I tried another approach

index.bs
sadym-chromium: @juliandescottes @jgraham WDYT about this approach?

juliandescottes: From a maintenance perspective, I prefer adding hooks in the relevant specs because otherwise the other specs might change and break the relationship with BiDi unintentionally.

From the point of view of someone trying to read and implement the specification, you still need to jump between the two specifications to understand the behavior. Possibly even harder to follow because you need to identify all the spots in the other spec which access `[=default User-Agent value=]`, which sounds painful.

The main added value of this approach is that we don't have to update other specs, but apart from that hooks seem preferable to me. Is there something which makes it more difficult for the NavigatorID case? For fetch we already have pending hooks in https://github.com/whatwg/fetch/pull/1540 so for consistency it also seems better to stick to hooks.

index.bs
sadym-chromium: I added a hook "WebDriver BiDi emulated User-Agent" and used it from HTML and Fetch specs. Please take another look.

index.bs

sadym-chromium:
@OrKoN @juliandescottes please take a look

sadym-chromium:
@juliandescottes please take a look

sadym-chromium:
@juliandescottes @whimboo PTAL

index.bs
```diff
@@ -6277,6 +6293,124 @@ The [=remote end steps=] with |command parameters| are:
...
+1. Return null.
+
+</div>
+
+<div algorithm="remote end steps for emulation.setUserAgentOverride">
+
+The [=remote end steps=] with |command parameters| are:
+
+1. If |command parameters| [=map/contains=] "<code>userContexts</code>"
+   and |command parameters| [=map/contains=] "<code>context</code>",
```
```suggestion
   and |command parameters| [=map/contains=] "<code>contexts</code>",
```

index.bs
```diff
@@ -6277,6 +6293,124 @@ The [=remote end steps=] with |command parameters| are:
...
+   return [=error=] with [=error code=] [=invalid argument=].
+
+1. Let |emulated user agent| be null.
+
+1. If |command parameters| [=map/contains=] "<code>userAgent</code>":
+
+   1. Set |emulated user agent| to |command parameters|["<code>userAgent</code>"].
```
juliandescottes: userAgent is mandatory so I think we can skip the check (but I can see we used the same pattern in many other emulation commands, where we can probably do the same simplification)
```suggestion
1. Set |emulated user agent| to |command parameters|["<code>userAgent</code>"].
```

index.bs
```diff
@@ -6277,6 +6293,124 @@ The [=remote end steps=] with |command parameters| are:
...
+
+1. If |command parameters| [=map/contains=] "<code>context</code>":
```
```suggestion
1. If |command parameters| [=map/contains=] "<code>contexts</code>":
```

index.bs
```diff
@@ -6277,6 +6293,124 @@ The [=remote end steps=] with |command parameters| are:
...
+   1. If |default emulated user agent| is not null, return
+      |default emulated user agent|.
+
+1. Return null.
+
+</div>
+
+<div algorithm="remote end steps for emulation.setUserAgentOverride">
+
+The [=remote end steps=] with |command parameters| are:
```
juliandescottes: we need the |session| here as well 
```suggestion
The [=remote end steps=] given |session|  and |command parameters| are:
```

index.bs
```diff
@@ -5827,10 +5833,20 @@ EmulationCommand = (
...
   emulation.SetScreenOrientationOverride //
   emulation.SetScriptingEnabled //
-  emulation.SetTimezoneOverride
+  emulation.SetTimezoneOverride //
+  emulation.SetUserAgentOverride
 )
 </pre>
 
+A [=BiDi session=] has an <dfn for=session>emulated user agent</dfn> which is a
+[=struct=] with an [=struct/item=] named
```
juliandescottes, line 5838: Just out of curiosity, why are we attaching this to the session rather than the remote end? Is it to make cleanup easier on session end?

index.bs
```diff
@@ -235,6 +240,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
     text: navigable; for:window; url: nav-history-apis.html#window-navigable
     text: navigables; url: document-sequences.html#navigables
     text: navigation id; url: browsing-the-web.html#navigation-id
+    text: NavigatorID; url: system-state.html#navigatorid
```
juliandescottes: question: are we using this?

juliandescottes (changes requested):
Thanks @sadym-chromium ! Overall that looks fine to me. Just some nits and suggestions. Before approving I just want to check the reason why we attach this to the session rather than to the remote end. Back to you

index.bs
sadym-chromium: removed the artifact

index.bs
sadym-chromium: Yes, attach it to session so that the emulation will be gone after the session is ended.

index.bs
sadym-chromium: fixed here and in other places

index.bs

index.bs
sadym-chromium: done here and in other places

sadym-chromium:
@juliandescottes PTAL I believe I addressed your comments, thanks!

juliandescottes (approved):
Oh I didn't notice that the context/contexts error was also in the other commands. Thanks for fixing them as well. Looks good to me now!

index.bs
whimboo: But what makes it different to the other emulation commands which have it all attached to the remote end?

Also other emulation commands have a single weak map for user contexts and navigables. Here we now have two of them. Maybe you can explain why that is needed?

index.bs
```diff
@@ -6279,6 +6294,120 @@ The [=remote end steps=] with |command parameters| are:
...
+
+</div>
+
+<div algorithm="remote end steps for emulation.setUserAgentOverride">
+
+The [=remote end steps=] given |session| and |command parameters| are:
+
+1. If |command parameters| [=map/contains=] "<code>userContexts</code>"
+   and |command parameters| [=map/contains=] "<code>contexts</code>",
+   return [=error=] with [=error code=] [=invalid argument=].
```
whimboo, line 6364: We also need to fail if whether `userContexts` nor `contexts` are given.

index.bs
```diff
@@ -6279,6 +6294,120 @@ The [=remote end steps=] with |command parameters| are:
...
+
+1. Let |emulated user agent| be |command parameters|["<code>userAgent</code>"].
+
+1. If |command parameters| [=map/contains=] "<code>contexts</code>":
```
whimboo, line 6368: Mind using the same phrasing as for all the other emulation commands?

index.bs
juliandescottes: (not in this case, if none are given we set it globally)

whimboo: Oh wait. So emulating the user agent will be the first emulation command that will allow us to set some overrides globally? Is that something that we discussed / agreed on - maybe I missed it because I have to leave early from the W3C meetings?

sadym-chromium: We already can set global [`network.setExtraHeaders`](https://www.w3.org/TR/webdriver-bidi/#extra-headers-default-headers), so it's not the first time we use this pattern. Also we have [this issue](https://github.com/w3c/webdriver-bidi/issues/956) filed after [feedback for `emulation.setScriptingEnabled` command](https://github.com/w3c/webdriver-bidi/pull/946#discussion_r2192043506).

index.bs
sadym-chromium: > But what makes it different to the other emulation commands which have it all attached to the remote end?

I don't have a strict rule for that. My rule of thumb is "can it expose some unwilling side-effects", and I find this emulation to make sense to stop with the session. We can change it later if there are strict opinions or scenarios for that.

> Also other emulation commands have a single weak map for user contexts and navigables. Here we now have two of them. Maybe you can explain why that is needed?

I used the same pattern @jgraham used in [extra headers](https://www.w3.org/TR/webdriver-bidi/#session-extra-headers): I store a struct with "default" (global), "per user context' and "per navigable" settings. The other commands didn't support "global" settings, so the single map was used there, but this approach is not extendable for "global".

index.bs
sadym-chromium: Other emulation commands don't support global settings, so the pattern don't quite work for this one.

sadym-chromium: Also refer to this PR: https://github.com/w3c/webdriver-bidi/pull/1001

Issue #776, opened 2024-09-18
Support emulation of network conditions


whimboo:
Currently, network condition emulation applies globally to the browser (if possible at all), but it would be very useful to emulate network disconnection or throttling for specific browsing contexts (such as tabs or windows) independently.

This feature would provide more control over testing scenarios where individual contexts simulate network loss or restricted connectivity, while other contexts remain unaffected.

Certainly! Here's a suggestion for a GitHub issue to request **network offline emulation per browsing context** in WebDriver BiDi or a similar automation context:

#### **Use Cases**
1. **Simulating Network Outages for Specific Tabs**: Being able to emulate a network outage for a single tab or window while keeping other tabs connected. This is useful for testing progressive web apps (PWAs), handling network errors in single-page applications, or multi-tab synchronization.
   
2. **Multi-Context Testing**: Testing how different contexts react to network issues independently. This is especially useful in applications where different parts of the app rely on separate network calls (e.g., one window working while another goes offline).
   
3. **Offline-First Applications**: Testing apps that are designed to work offline in a more realistic scenario where only certain windows/tabs experience network issues, rather than the entire browser being set to offline mode.

#### **Proposal**
Introduce an API to WebDriver BiDi that allows specifying custom network conditions (e.g., offline, latency, download/upload throttling) for specific browsing contexts, rather than affecting the entire browser.

The API could look like:

- **Command to set network offline for a specific context**:
    ```json
    {
      "method": "network.setNetworkConditions",
      "params": {
        "contexts": [ids],  // The IDs of the browsing contexts or undefined for global settings
        "offline": true,           // Set network to offline
        "latency": 0,              // Optionally set latency
        "downloadThroughput": -1,  // Optional: Disable downloads
        "uploadThroughput": -1     // Optional: Disable uploads
      }
    }
    ```

#### **Benefits**
- **Granular Control**: This feature allows testing different network conditions in a more targeted way, improving the ability to simulate real-world scenarios for developers building resilient, network-dependent applications.
- **Better Testing for PWAs**: Many progressive web apps require granular control over network state to properly test offline functionality.
- **Non-disruptive Testing**: Since this feature would apply to specific browsing contexts, it allows for more controlled testing without affecting the entire browser or other ongoing operations.

#### **Questions**
- Are there any known limitations that might prevent isolating network conditions for individual browsing contexts?

#### **References**
[...]

PR #948, opened 2025-07-08
`emulation.setNetworkConditions:offline`

sadym-chromium:
First step addressing https://github.com/w3c/webdriver-bidi/issues/776.
* Fail all the intercepted requests, WebSocket and WebTransport.
* Html spec: https://github.com/whatwg/html/pull/11575
* Fetch spec: https://github.com/whatwg/fetch/pull/1852

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/948.html" title="Last updated on Oct 7, 2025, 7:46 AM UTC (86db360)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/948/20868f9...86db360.html" title="Last updated on Oct 7, 2025, 7:46 AM UTC (86db360)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -331,7 +331,6 @@ spec: STREAMS; urlPrefix: https://streams.spec.whatwg.org/
 spec: WebSockets; urlPrefix: https://websockets.spec.whatwg.org/
   type: dfn
     text: WebSocket; url: #websocket
-    text: Make disappear; url: #make-disappear
 spec: WebTransport; urlPrefix: https://w3c.github.io/webtransport/
   type: dfn
     text: Cleanup WebTransport; url: #webtransport-cleanup
@@ -6347,7 +6346,7 @@ To <dfn>apply network conditions</dfn>:
 
   1. If the result of [=WebDriver BiDi network is offline=] with |environment settings| is true:
 
-     1. [=Make disappear=] |webSocket|.
+     1. [=Fail the WebSocket connection=] |webSocket|.
 
 1. For each [=WebTransport=] object |webTransport|:
 
--- a/index.bs
+++ b/index.bs
@@ -9797,6 +9797,23 @@ are:
 </div>
 
 <div algorithm>
+The <dfn export>WebDriver BiDi network is offline</dfn> steps given
+[=environment settings object=] |settings| are:
+
+1. Let |navigable| be |settings|'s [=relevant global object=]'s
+   <a>associated <code>Document</code></a>'s [=/node navigable=].
+
+1. Let |emulated network conditions| be the result of
+   [=get emulated network conditions=] with [|navigable|].
+
+1. If |emulated network conditions| is not null and |emulated network conditions|'s
+   [=emulated-network-conditions-struct/offline=] is true, return true.
+
+1. Return false.
+
+</div>
+
+<div algorithm>
 The [=remote end event trigger=] is the <dfn export>WebDriver BiDi before request sent</dfn> steps
 given [=/request=] |request|:
 
--- a/index.bs
+++ b/index.bs
@@ -1543,6 +1543,14 @@ A [=BiDi session=] has a
 A [=BiDi session=] has a <dfn>user context to proxy configuration map</dfn>, which is
 a [=/map=] between [=user contexts=] and [=proxy configuration=].
 
+An <dfn>emulated network conditions struct</dfn> is a [=struct=] with:
+* [=struct/item=] named <dfn id="emulated-network-conditions-struct-offline"
+  for="emulated-network-conditions-struct">offline</dfn> which is a boolean or null.
+
+A [=BiDi session=] has an <dfn>emulated network conditions map</dfn>, which is a weak
+map between [=navigables=] or [=user contexts=] and
+[=emulated network conditions struct=] or null.
+
 When a [=user context=] is [=set/remove|removed=] from the
 [=set of user contexts=], [=remove user context subscriptions=].
 
@@ -5930,6 +5938,7 @@ EmulationCommand = (
   emulation.SetForcedColorsModeThemeOverride //
   emulation.SetGeolocationOverride //
   emulation.SetLocaleOverride //
+  emulation.SetNetworkConditions //
   emulation.SetScreenOrientationOverride //
   emulation.SetScriptingEnabled //
   emulation.SetTimezoneOverride //
@@ -6271,6 +6280,86 @@ The [=remote end steps=] with |command parameters| are:
    1. If |emulated locale| is null, [=map/remove=] |navigable| from [=locale overrides map=].
 
    1. Otherwise, [=map/set=] [=locale overrides map=][|navigable|] to |emulated locale|.
+
+1. Return [=success=] with data null.
+
+</div>
+
+#### The emulation.setNetworkConditions Command ####  {#command-emulation-setNetworkConditions}
+
+The <dfn export for=commands>emulation.setNetworkConditions</dfn> command
+emulates specific network conditions for the given browsing context or for a user
+context.
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl" data-cddl-module="remote-cddl">
+      emulation.SetNetworkConditions = (
+        method: "emulation.setNetworkConditions",
+        params: emulation.setNetworkConditionsParameters
+      )
+
+      emulation.setNetworkConditionsParameters = {
+        networkConditions: emulation.NetworkConditions / null,
+        ? contexts: [+browsingContext.BrowsingContext],
+        ? userContexts: [+browser.UserContext],
+      }
+
+      emulation.NetworkConditions = emulation.NetworkConditionsOffline
+
+      emulation.NetworkConditionsOffline = {
+        type: "offline"
+      }
+    </pre>
+   </dd>
+   <dt>Result Type</dt>
+   <dd>
+    <code>
+     EmptyResult
+    </code>
+   </dd>
+</dl>
+
+<div algorithm="remote end steps for emulation.setNetworkConditions">
+
+The [=remote end steps=] with |command parameters| are:
+
+1. If |command parameters| [=map/contains=] "<code>userContexts</code>"
+   and |command parameters| [=map/contains=] "<code>context</code>",
+   return [=error=] with [=error code=] [=invalid argument=].
+
+1. If |command parameters| doesn't [=map/contain=] "<code>userContexts</code>"
+   and |command parameters| doesn't [=map/contain=] "<code>context</code>",
+   return [=error=] with [=error code=] [=invalid argument=].
+
+1. Let |emulated network conditions| be null.
+
+1. If |command parameters|["<code>networkConditions</code>"] is not null and
+      |command parameters|["<code>networkConditions</code>"]["<code>type</code>"]
+      equals "<code>offline</code>", set |emulated network conditions| to a new
+      [=emulated network conditions struct=] with
+      [=emulated-network-conditions-struct/offline=] set to true.
+
+1. If the <code>contexts</code> field of |command parameters| is present:
+
+   1. Let |navigables| be the result of [=trying=] to
+      [=get valid top-level traversables by ids=] with
+      |command parameters|["<code>contexts</code>"].
+
+   1. For each |navigable| of |navigables|:
+
+      1. [=map/Set=] [=emulated network conditions map=][|navigable|] to
+         |emulated network conditions|.
+
+1. If the <code>userContexts</code> field of |command parameters| is present:
+
+   1. Let |user contexts| be the result of [=trying=] to [=get valid user contexts=]
+      with |command parameters|["<code>userContexts</code>"].
+
+   1. For each |user context| of |user contexts|:
+
+      1. [=map/Set=] [=emulated network conditions map=][|user context|] to |emulated network conditions|.
 
 1. Return [=success=] with data null.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -9129,8 +9241,17 @@ request sent</dfn> steps given |request|:
 
     Note: While waiting, no further processing of the request occurs.
 
+1. Let |emulated network conditions| be the result of
+   [=get emulated network conditions=] with |related navigables|.
+
+1. If |emulated network conditions| is not null and
+   |emulated network conditions|'s [=emulated-network-conditions-struct/offline=] is
+   true, return ([=network error=], "<code>complete</code>").
+
```
juliandescottes, line 9999: What is the added value of an offline emulation which only rejects on `beforeRequestSent`. It seems you can easily implement this with addIntercept + failRequest? Even if we ignore navigator.online, an offline mode should impact ongoing requests, websockets, allow you to test service workers etc... I feel like a limited feature like this would be more confusing than anything else?

OrKoN: I was under impression that this was discussed in the WG? I do not think the offline mode should apply to ongoing requests and with this it should apply to web sockets and service worker requests as they have to start via fetch? The advantage is that you do no pay the cost of interception and have a dedicated command that can be extended with further aspects specific to the network condition emulation. The question I would have if this allows successfully testing the offline mode behavior of the browser which is probably a bit more than failing the network requests?

juliandescottes: > I was under impression that this was discussed in the WG?

I thought so as well, but I wanted to check the details and I can't find any summary for this discussion. So I'm just giving my feedback based on what is currently proposed. I couldn't find a matching topic in the last 6 meetings, but maybe this was discussed informally?

For service workers, I would need to check in details. I was thinking about a fetch request which gets captured by an already registered service worker, returning a response from the cache. This should still work in offline mode, but not sure that's the case here?

For websocket, is fetch used for anything but establishing the connection at https://websockets.spec.whatwg.org/#concept-websocket-establish ? I was thinking about ws messages on an already open websocket.

> The question I would have if this allows successfully testing the offline mode behavior of the browser which is probably a bit more than failing the network requests?

I think the answer to this is no, but I don't know if the offline mode is completely specified.

But if the WG agreed to have something that only blocks new fetch requests as an offline mode, I don't mean to block the PR (would be great if I could find the discussion though :) )

OrKoN: I was not attending unfortunately so I do not know for sure :)

jgraham: I don't remember such a discussion and broadly agree with Julian's concerns here; I'd certainly expect that in offline mode `navigatior.onLine` should be false, and for the `offline` event to fire when the state changes. I also agree it would be extremely surprising if e.g. open websockets (or server sent events eventsources) continue to work in "offline" mode.

juliandescottes:
Notes from the Working Group meeting:
> 18:33 <jdescottes> sadym: feature requested by users is setting network conditions. Started with offline mode. Technically can be done with Network Interception. Will be difficult to specify bandwidth etc.
18:34 <jdescottes> sadym: talking about offline: several aspects : network requests and navigator online. With network interception we cannot influence navigator online, bur we can handle requests
18:35 <jdescottes> sadym: currently navigator online is defined in a way which should return false if the user can't reach the network. Unclear behavior
18:36 <jdescottes> sadym: proposal is to add a hook `is offline` so that the html spec can call this to know what to return
18:36 <jdescottes> q+
18:36 * Zakim sees jdescottes on the speaker queue
18:36 <jdescottes> ack jdescottes
18:36 * Zakim sees no one on the speaker queue
18:37 <jdescottes> jdescottes: hook for html spec sounds good to me, especially since we can't block all network for now
18:38 <jimevans> q+
18:38 * Zakim sees jimevans on the speaker queue
18:38 <jdescottes> sadym: should we split or land everything in one PR
18:40 <jdescottes> jimevans: can be more work, but I would prefer to see the full implementation landed at once. I would expect everything to behave as offline/online when setting this emulation. 
18:40 <jdescottes> sadym: sounds good

index.bs
sadym-chromium: I updated the PR and created an HTML PR addressing `navigatior.onLine`. @juliandescottes @jgraham please take another look.

sadym-chromium: The existing connections are still out of scope of the PR. Other concerns I believe are addressed.

sadym-chromium:
@juliandescottes @OrKoN could you please take another look?

index.bs
```diff
@@ -1541,6 +1541,20 @@ A [=BiDi session=] has a
 A [=BiDi session=] has a <dfn>user context to proxy configuration map</dfn>, which is
 a [=/map=] between [=user contexts=] and [=proxy configuration=].
 
+An <dfn>emulated network conditions struct</dfn> is a [=struct=] with:
+* [=struct/item=] named <dfn id="emulated-network-conditions-struct-offline"
+  for="emulated-network-conditions-struct">offline</dfn> which is a boolean or null.
+
+A [=BiDi session=] has a <dfn for=session>emulated network conditions</dfn> which is
```
OrKoN, line 1556: the generated preview seems to be out of sync with the changes here?

sadym-chromium: now it's up to date

index.bs
```diff
@@ -9490,8 +9640,17 @@ request sent</dfn> steps given |request|:
 
     Note: While waiting, no further processing of the request occurs.
 
+1. Let |emulated network conditions| be the result of
+   [=get emulated network conditions=] with |related navigables|.
+
+1. If |emulated network conditions| is not null and
+   |emulated network conditions|'s [=emulated-network-conditions-struct/offline=] is
+   true, return ([=network error=], "<code>complete</code>").
```
juliandescottes, line 9998: If I understand correctly, we have the `WebDriver BiDi network is offline` algorithm, which could be called by fetch before trying to issue any network activity, and that's what we plan to use in the future in order to avoid websocket messages for instance?

So I'm wondering if we should keep explicitly blocking requests on our end in beforeRequestSent? Shouldn't this be also delegated to the fetch spec?

My issue here is that offline mode should not block things such as serviceworker handled requests. Those trigger a beforeRequestSent event, so they would be canceled here. Same thing for resources cached locally (eg JS/CSS) which don't perform a network access. They should be loaded even in offline mode, but this will also prevent to do so.

I would wait for James here to get some more feedback, but my feeling is that we should have the `WebDriver BiDi network is offline` algorithm, but then it should be up to `fetch` to decide when to call it, because that's the only place where we really know if/when we are hitting the network?

WDYT?

sadym-chromium: Added fetch PR. Also I added the logic to drop all the WebSocket and WebTransport connections.

sadym-chromium:
@juliandescottes PTAL

index.bs
```diff
@@ -9383,6 +9520,56 @@ and |user context| are:
...
+The steps to <dfn>get emulated network conditions</dfn> given |related navigables|
+are:
+
+1. For each |navigable| of |related navigables|:
+
+   1. Let |top-level traversable| be |navigable|’s [=navigable/top-level traversable=].
+
+   1. Let |user context| be |top-level traversable|'s [=associated user context=].
+
+   1. For each |session| in [=active BiDI sessions=]:
```
```suggestion
   1. For each |session| in [=active BiDi sessions=]:
```

index.bs
```diff
@@ -6153,6 +6175,121 @@ The [=remote end steps=] with |command parameters| are:
...
+1. For each [=WebSocket=] object |webSocket|:
+
+  1. Let |realm| be |webSocket|'s [=relevant Realm=].
+
+  1. Let |environment settings| be the [=environment settings object=] whose
+     [=realm execution context=]'s Realm component is |realm|.
+
+  1. If the result of [=WebDriver BiDi network is offline=] with |environment settings| is true:
+
+     1. [=Make disappear=] |webSocket|.
```
jgraham: Is this right? It's going to send a close frame to the server, which wouldn't happen if the browser was actually offline. Do browsers' existing offline modes actually send this frame when activated on a page with an open websocket?

index.bs
```diff
@@ -6153,6 +6175,121 @@ The [=remote end steps=] with |command parameters| are:
...
+
+The [=remote end steps=] with |command parameters| and |session| are:
+
+1. If |command parameters| [=map/contains=] "<code>userContexts</code>"
+   and |command parameters| [=map/contains=] "<code>context</code>",
+   return [=error=] with [=error code=] [=invalid argument=].
+
+1. Let |emulated network conditions| be null.
+
+1. If |command parameters|["<code>networkConditions</code>"] is not null and
```
jgraham, line 6374: For now the check for type can just be an assert. I see you're trying to define this for extensibility although it does make the API rather odd in the short term, and the obvious extensions into network conditions seem like they're going to be rather challenging to standardise. I wonder if we'd be better with just a `network.setOffline({contexts, userContexts, offline: <bool>})` command for now.

juliandescottes (commented):
Thanks for the update, and sorry for the delay. Overall I think the update for BiDi looks good, but I'm not sure how we should treat this PR since we have a dependency on HTML and Fetch. 

I prefer to let James or other people more used to the spec world comment on this, my feeling is that we might need at least some agreement on the other PRs that they are ok with the direction. We can also discuss it next wednesday during the meeting.

jgraham (commented):
Apparently I made some comments a while ago and didn't submit them…

index.bs
sadym-chromium: If we don't plan to extend for other network conditions (e.g. high-latency or low-bandwidth), I'd propose to:
1. Move the method to `emulation.setNetworkOffline`, as it influences not only the network, but the navigator as well.
2. Allow for only `offline: true / null`, as we cannot emulate being online when offline, we can only restore the default behavior.

index.bs
sadym-chromium: Switched to [Fail the WebSocket connection](https://datatracker.ietf.org/doc/html/rfc6455#section-7.1.7)

sadym-chromium:
@jgraham could you please take another look?

sadym-chromium:
@juliandescottes @jgraham friendly ping

index.bs
sadym-chromium: Actually, I believe that it still makes sense to keep it extendable and eventually introduce other conditions.

index.bs
```diff
@@ -9631,6 +9767,56 @@ and |user context| are:
...
+         [=emulated network conditions/navigable network conditions=]
+         [=map/contains=] |top-level traversable|, return |session|'s
+         [=emulated network conditions=]'s
+         [=emulated network conditions/navigable network conditions=][|top-level traversable|].
+
+      1. If |session|'s [=emulated network conditions=]'s
+         [=emulated network conditions/user context network conditions=]
+         [=map/contains=] |user context|, return |session|'s
+         [=emulated network conditions=]'s
+         [=emulated network conditions/user context network conditions=][|user context|].
```
juliandescottes, line 9854: In the steps for `emulation.setNetworkConditions`, I think we can set `null` as the `emulated network conditions` in `navigable network conditions` or `user context network conditions`. Should we null check here as well?

index.bs
```diff
@@ -9631,6 +9767,56 @@ and |user context| are:
...
+The steps to <dfn>get emulated network conditions</dfn> given |related navigables|
+are:
+
+1. For each |navigable| of |related navigables|:
+
+   1. Let |top-level traversable| be |navigable|’s [=navigable/top-level traversable=].
+
+   1. Let |user context| be |top-level traversable|'s [=associated user context=].
+
+   1. For each |session| in [=active BiDi sessions=]:
```
juliandescottes, line 9840: Should we store the conditions on the session or on the remote end?
With the per session approach, the first session to be checked always wins. Whereas if it's per remote end, the last condition set wins, which feels easier to manage for users than (arbitrary) session order?

index.bs
```diff
@@ -9631,6 +9767,56 @@ and |user context| are:
 
 </div>
 
+<div algorithm>
+The steps to <dfn>get emulated network conditions</dfn> given |related navigables|
+are:
+
+1. For each |navigable| of |related navigables|:
```
juliandescottes, line 9834: note: probably not important as long as we can only disable, but I'm wondering about scenarios where later we could support other network conditions eg throttling.

If navigable 1 is throttled and navigable 2 is offline, what should we do? We should probably defer that discussion to whenever we implement more conditions, but still curious if you have any thoughts about this.

juliandescottes (commented):
Thanks Maksim, looks fine to me, just a couple of questions.

index.bs
sadym-chromium: > note: probably not important as long as we can only disable, but I'm wondering about scenarios where later we could support other network conditions eg throttling.
> 
> If navigable 1 is throttled and navigable 2 is offline, what should we do? We should probably defer that discussion to whenever we implement more conditions, but still curious if you have any thoughts about this.

This spec version defines the behavior as "the first created navigable wins". This scenario happens for shared workers. However IDK what exactly Chromium does in this case, and I don't have a strong opinion what it should do.

index.bs
sadym-chromium: I don't have a strong opinion here, as currently we don't support multiple sessions. For now I changed the spec, so that we check all sessions' navigables, then  user contexts, and then default settings.

index.bs
sadym-chromium: Updated. We should not set null to `user context`'s or `navigable`'s network conditions.

juliandescottes (approved):

index.bs
jgraham: Personally, I still feel like a separate command would have been the better design for now. The only advantage I see of a single command is that it will make it easier in the future to switch atomically from a "network offline" state to an online state with some other configuration. But that would be possible either way (e.g. by allowing the configuration to be set when offline, and applying it when online).

This would have the advantage of not making the existing protocol more complex in service of uncertain future requirements.

However since there is already general agreement about the approach here I don't consider it a blocker.

sadym-chromium: Ack, thanks for the clarification! I imagine we can extend the emulated conditions with some vendor-specific types in the nearest future until the proper specification is landed.

whimboo:
@sadym-chromium Will you also work on the wpt tests whenever you are going to work on supporting this feature?

sadym-chromium:
> @sadym-chromium Will you also work on the wpt tests whenever you are going to work on supporting this feature?

The wpr tests are landed

whimboo:
Ah, there was no link to https://github.com/web-platform-tests/wpt/pull/55300 in this PR. Thanks!

PR #949, opened 2025-07-09
`emulation.setTimezoneOverride`

sadym-chromium:
Specify `emulation.setTimezoneOverride`

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/949.html" title="Last updated on Jul 17, 2025, 1:06 PM UTC (96f45d7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/949/26516ad...96f45d7.html" title="Last updated on Jul 17, 2025, 1:06 PM UTC (96f45d7)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5982,14 +5982,13 @@ The [=DefaultLocale=] algorithm is implementation defined. A WebDriver-BiDi
 
 1. Let |realm| be [=current Realm Record=].
 
-1. Let |global object| be |realm|'s [=realm/global object=].
+1. Let |environment settings| be the [=environment settings object=] whose
+   [=realm execution context=]'s Realm component is |realm|.
 
-1. If |global object| implements {{WindowProxy}}:
+1. Let |related navigables| be the result of [=get related navigables=] given
+   |environment settings|.
 
-    1. Let |window| be the value of |global object|'s \[[WindowProxy]]
-       [=internal slot=].
-
-    1. Let |navigable| be |window|'s [=window/navigable=].
+1. For each |navigable| of |related navigables|:
 
     1. Let |top-level traversable| be |navigable|’s
        [=navigable/top-level traversable=].
@@ -6226,14 +6225,13 @@ steps:
 
 1. Let |realm| be [=current Realm Record=].
 
-1. Let |global object| be |realm|'s [=realm/global object=].
+1. Let |environment settings| be the [=environment settings object=] whose
+   [=realm execution context=]'s Realm component is |realm|.
 
-1. If |global object| implements {{WindowProxy}}:
+1. Let |related navigables| be the result of [=get related navigables=] given
+   |environment settings|.
 
-    1. Let |window| be the value of |global object|'s \[[WindowProxy]]
-       [=internal slot=].
-
-    1. Let |navigable| be |window|'s [=window/navigable=].
+1. For each |navigable| of |related navigables|:
 
     1. Let |top-level traversable| be |navigable|’s
        [=navigable/top-level traversable=].
--- a/index.bs
+++ b/index.bs
@@ -136,6 +136,7 @@ spec: ECMASCRIPT-I18N; urlPrefix: https://tc39.es/ecma402/
 spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
   type: dfn
     text: Array; url: sec-array-objects
+    text: AvailableNamedTimeZoneIdentifiers; url: sec-availablenamedtimezoneidentifiers
     text: Await; url: await
     text: BigInt; url: sec-bigint-constructor
     text: Call; url: sec-call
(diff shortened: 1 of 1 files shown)


OrKoN (approved):

index.bs
```diff
@@ -6181,6 +6187,127 @@ The [=remote end steps=] with |command parameters| are:
...
+
+1. Let |emulated timezone| be null.
+
+1. If |command parameters| [=map/contains=] "<code>timezone</code>":
+
+   1. Set |emulated timezone| to |command parameters|["<code>timezone</code>"].
+
+   1. If [=IsTimeZoneOffsetString=](|emulated timezone|) returns false and
+      [=AvailableNamedTimeZoneIdentifiers=] does not [=list/contain=]
+      |emulated timezone|, return [=error=] with [=error code=] [=invalid argument=].
```
lutien, line 6275: Just to make sure I understood it right, does it mean that we want to support here named timezones (e.g., `Europe/Berlin`) and timezone offsets (e.g., `-120`).

sadym-chromium: > Just to make sure I understood it right, does it mean that we want to support here named timezones (e.g., `Europe/Berlin`) and timezone offsets (e.g., `-120`).

The timezone is either an [IANA time zone name](https://www.iana.org/time-zones) (e.g. "Europe/Berlin", "America/New_York") or a specific [UTC offset](https://tc39.es/ecma262/#prod-UTCOffset) (e.g. "+02:00" or "-09:30").

sadym-chromium:
@whimboo @lutien @juliandescottes @jgraham any other concerns?

lutien (approved):
Looks fine to me, but maybe @juliandescottes, @whimboo or @jgraham still want to take a look

index.bs
```diff
@@ -6181,6 +6187,127 @@ The [=remote end steps=] with |command parameters| are:
...
+    1. Let |user context| be |top-level traversable|'s [=associated user context=].
+
+    1. If [=timezone overrides map=] [=map/contains=] |top-level traversable|, set
+       |emulated timezone| to [=timezone overrides map=][|top-level traversable|].
+
+    1. Otherwise, if [=timezone overrides map=] [=map/contains=]
+       |user context|, set |emulated timezone| to
+       [=timezone overrides map=][|user context|].
+
+1. If |emulated timezone| is not null, return |emulated timezone|.
```
whimboo, line 6248: AFAICS this is a no-op. Step 1 sets it to `null` but it doesn't get modified until this line. So we can remove it?

index.bs
```diff
@@ -6181,6 +6187,127 @@ The [=remote end steps=] with |command parameters| are:
...
+
+      1. [=list/For each=] |top-level traversable| of the list of all
+         [=/top-level traversables=] whose [=associated user context=] is
+         |user context|:
+
+         1. [=list/Append=] |top-level traversable| to |navigables|.
+
+1. For each |navigable| of |navigables|:
+
+   1. [=map/Set=] [=timezone overrides map=][|navigable|] to |emulated timezone|.
```
whimboo, line 6305: And where do we actually set the timezone settings? This only updates our map but doesn't make a modification to the actual navigable.

index.bs
```diff
@@ -6181,6 +6187,127 @@ The [=remote end steps=] with |command parameters| are:
...
+   </dd>
+   <dt>Result Type</dt>
+   <dd>
+    <code>
+     EmptyResult
+    </code>
+   </dd>
+</dl>
+
+<div algorithm="updated SystemTimeZoneIdentifier steps">
```
whimboo, line 6219: Where is this being called from?

sadym-chromium: > The [SystemTimeZoneIdentifier](https://tc39.es/ecma262/#sec-systemtimezoneidentifier) algorithm is implementation defined. A WebDriver-BiDi [remote end](https://w3c.github.io/webdriver/#dfn-remote-ends) must have an implementation that runs the following steps...

index.bs
sadym-chromium: `|emulated timezone|` is set in line [6242](https://github.com/w3c/webdriver-bidi/pull/949/files#diff-5e793325cd2bfc452e268a4aa2f02b4024dd9584bd1db3c2595f61f1ecf7b985R6242) and [6245](https://github.com/w3c/webdriver-bidi/pull/949/files#diff-5e793325cd2bfc452e268a4aa2f02b4024dd9584bd1db3c2595f61f1ecf7b985R6245), if the `[=timezone overrides map=]` has `|top-level traversable|` or `|user context|`.

index.bs
sadym-chromium: We access it in the `updated SystemTimeZoneIdentifier steps` algorithm.

index.bs
whimboo: Right, but it means that the timezone override only takes effect when the first call to any date and time related API occurs by the page. So maybe that's fine?

index.bs
whimboo: Oh, I completely missed that.

index.bs
sadym-chromium: I read the spec as it invokes "SystemTimeZoneIdentifier()" each time it needs a timezone and doesn't cache it anywhere.

PR #950, opened 2025-07-09
fix (network.fetchError): define errorText as implementation defined string

### LINKED ISSUE #631: network.fetchError: define how errorText should be retrieved / computed
At the moment the spec does not say how the `errorText` of the `network.fetchError` event is retrieved:

> TODO: Set the errorText field of params.

Do we want to align with how CDP behaves here?

juliandescottes:
Fixes #631

The fetch spec does not define the error message which leads to abort a request, so the fetchError parameter will most likely be an implementation defined string.

Ideally it should not be used for assertions, but can be still be useful for clients.

***
<a href="https://pr-preview.s3.amazonaws.com/juliandescottes/webdriver-bidi/pull/950.html" title="Last updated on Jul 9, 2025, 8:01 PM UTC (186a655)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/950/a8b68a1...juliandescottes:186a655.html" title="Last updated on Jul 9, 2025, 8:01 PM UTC (186a655)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -9052,7 +9052,8 @@ error</dfn> steps given |request|:
   1. Let |params| be the result of [=process a network event=] with |session|
      "<code>network.fetchError</code>", and |request|.
 
-  1. TODO: Set the <code>errorText</code> field of |params|.
+  1. Set the <code>errorText</code> field of |params| to an implementation-defined string
+     describing the error which caused the request to be aborted.
 
   1. Assert: |params| matches the <code>network.FetchErrorParameters</code>
      production.


index.bs
```diff
@@ -9052,7 +9052,8 @@ error</dfn> steps given |request|:
   1. Let |params| be the result of [=process a network event=] with |session|
      "<code>network.fetchError</code>", and |request|.
 
-  1. TODO: Set the <code>errorText</code> field of |params|.
+  1. Set the <code>errorText</code> field of |params| to an implementation-defined string
+     describing the error which led to abord the request.
```
whimboo: Maybe the following?

```suggestion
     describing the error which caused the request to be aborted.
```

juliandescottes: I fixed the typo abord -> abort. Regarding the actual sentence, I'll wait for more feedback to update but no strong opinion I can change it if it seems clearer.

jgraham: I think @whimboo's suggestion is a bit clearer, so I'd suggest taking that.

whimboo (approved):

PR #960, opened 2025-07-11
Add network.setExtraHeaders command

jgraham:
This allows appending, or overriding, HTTP headers for all requests, or for all requests originating from a specific user context or a specific navigable.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/960.html" title="Last updated on Jul 11, 2025, 10:25 AM UTC (4533b20)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/960/a8b68a1...4533b20.html" title="Last updated on Jul 11, 2025, 10:25 AM UTC (4533b20)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6204,7 +6204,8 @@ NetworkCommand = (
   network.ProvideResponse //
   network.RemoveDataCollector //
   network.RemoveIntercept //
-  network.SetCacheBehavior
+  network.SetCacheBehavior //
+  network.SetExtraHeaders
 )
 
 </pre>
@@ -6237,6 +6238,15 @@ initially "<code>default</code>".
 A [=remote end=] has a <dfn>navigable cache behavior map</dfn> which is a weak
 map between [=/top-level traversables=] and strings representing cache
 behavior. It is initially empty.
+
+A [=BiDi session=] has a <dfn for=session>extra headers</dfn> which is a
+[=struct=] with a [=struct/item=] named <dfn for="extra headers">default
+headers</dfn>, which is a [=/header list=]. (initially set to an empty
+[=/header list=]), a [=struct/item=] named
+<dfn for="extra headers">user context headers</dfn>, which is a weak map
+between [=user contexts=] and [/=header lists=], and  a [=struct/item=] named
+<dfn for="extra headers">navigable headers</dfn>, which is a weak map
+between [=navigables=] and [=/header lists=].
 
 ### Network Data Collection ### {#network-data-collection}
 
@@ -6587,21 +6597,8 @@ To <dfn>update the response</dfn> given |session|, |command| and |command parame
 
 1. If |command parameters| [=map/contains=] "<code>headers</code>":
 
-  1. Let |headers| be an empty [=/header list=].
-
-  1. For |header| in |command parameters|["<code>headers</code>"]:
-
-    1. Let |deserialized header| be [=deserialize header=] with |header|.
-
-    1. If |deserialized header|'s name does not match the [=field-name token=]
-       production, return [=error=] with [=error code=]
-      "<code>invalid argument</code>".
-
-    1. If |deserialized header|'s value does not match the [=header value=]
-       production, return [=error=] with [=error code=]
-      "<code>invalid argument</code>".
-
-    1. Append |deserialized header| to |headers|.
+  1. Let |headers| be the result of [=trying=] to [=create a headers list=] with
+     |command parameters|["<code>headers</code>"].
 
   1. Set |response|'s [=response/header list=] to |headers|.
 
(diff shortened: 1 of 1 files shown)


jgraham:
Note that the sematics here are to always overwrite. We could support more options e.g. appending or extending additional headers, but there is ofc a complexity tradeoff.

index.bs
```diff
@@ -6238,6 +6239,15 @@ A [=remote end=] has a <dfn>navigable cache behavior map</dfn> which is a weak
 map between [=/top-level traversables=] and strings representing cache
 behavior. It is initially empty.
 
+A [=BiDi session=] has a <dfn for=session>extra headers</dfn> which is a
+[=struct=] with a [=struct/item=] named <dfn for="extra headers">default
```
```suggestion
[=struct=] with an [=struct/item=] named <dfn for="extra headers">default
```

index.bs
```diff
@@ -6238,6 +6239,15 @@ A [=remote end=] has a <dfn>navigable cache behavior map</dfn> which is a weak
...
+headers</dfn>, which is a [=/header list=]. (initially set to an empty
```
```suggestion
headers</dfn>, which is a [=/header list=] (initially set to an empty
```

index.bs
```diff
@@ -6238,6 +6239,15 @@ A [=remote end=] has a <dfn>navigable cache behavior map</dfn> which is a weak
...
+[=/header list=]), a [=struct/item=] named
```
```suggestion
[=/header list=]), an [=struct/item=] named
```

whimboo:
@jgraham please rebase your PR on main to get the node.js upgrade, which will fix the test failure.

index.bs
```diff
@@ -8962,14 +9109,17 @@ request sent</dfn> steps given |request|:
 
 1. Let |response status| be "<code>incomplete</code>".
 
+1. For each |session| in [=active BiDI sessions=]:
```
```suggestion
1. For each |session| in [=active BiDi sessions=]:
```

juliandescottes:
> Note that the sematics here are to always overwrite. We could support more options e.g. appending or extending additional headers, but there is ofc a complexity tradeoff.

FWIW, that seems to match what the current CDP API is doing.

index.bs
```diff
@@ -6238,6 +6239,15 @@ A [=remote end=] has a <dfn>navigable cache behavior map</dfn> which is a weak
 map between [=/top-level traversables=] and strings representing cache
 behavior. It is initially empty.
 
+A [=BiDi session=] has a <dfn for=session>extra headers</dfn> which is a
+[=struct=] with a [=struct/item=] named <dfn for="extra headers">default
+headers</dfn>, which is a [=/header list=]. (initially set to an empty
+[=/header list=]), a [=struct/item=] named
+<dfn for="extra headers">user context headers</dfn>, which is a weak map
+between [=user contexts=] and [/=header lists=], and  a [=struct/item=] named
```
```suggestion
between [=user contexts=] and [=/header lists=], and  a [=struct/item=] named
```

index.bs
```diff
@@ -8906,7 +9053,7 @@ request sent</dfn> steps given |request|:
    1. If the [=request originates in user context=] steps with |request| and
       |user context| return true:
 
-      1. For each |session| in [=active BiDI sessions=]
+      1. For each |session| in [=active BiDI sessions=]:
```
```suggestion
      1. For each |session| in [=active BiDi sessions=]:
```
while we are touching the line

index.bs
```diff
@@ -8962,14 +9109,17 @@ request sent</dfn> steps given |request|:
 
 1. Let |response status| be "<code>incomplete</code>".
 
+1. For each |session| in [=active BiDI sessions=]:
+
+  1. [=Update request headers=] with |session|, |request| and
+     |related navigables|.
```
juliandescottes, line 9119: For multisession, this means that the order in which we will apply headers (and potentially overwrite them) is based on the order of sessions in the `active BiDi sessions`, not based on the order in which setExtraHeaders commands have been called. I'm just noting it here - especially since multi session support is not implemented anywhere.

jgraham:
> Should we restrict the API to only accept top level contexts, and in update request headers resolve the top level context for the related navigable?

I'm not sure. I think there are use cases where you really only want a request in a specific iframe to contain extra headers (consider e.g. a cross-origin iframe that's being used for auth). On the other hand one could implement that using request interception, and not inheriting does mean that you don't get any extra headers when an iframe first loads, unless you override at the user context level (which is even less specific).

jgraham:
Heh, I think you convinced me that the right starting point is probably inheriting into subframes (and that's easier if we only allow setting it on top-level traversables), since harder cases can always be handled by request interception. Do we know what clients expect here?

juliandescottes:
> Heh, I think you convinced me that the right starting point is probably inheriting into subframes (and that's easier if we only allow setting it on top-level traversables), since harder cases can always be handled by request interception. Do we know what clients expect here?

Ah 😅 I haven't found puppeteer or playwright tests covering the iframe scenarios. Will try to build something, in the meantime cc @OrKoN , do you know if in CDP `setExtraHTTPHeaders` also inherits to iframes loaded in the page? Looking at the documentation and implementation in puppeteer, it feels like setExtraHTTPHeaders is always global. It's a command on the Network domain which doesn't take any argument to be scoped down. But maybe I'm misreading the code.

OrKoN:
CDP sessions are always implicitly scoped to a CDP target (main frame including local sub-frames, worker, out-of-process frames including their local sub-frames), you can send the same command to a different session and it does not change the command params. Puppeteer exposes it one the page level https://pptr.dev/api/puppeteer.page.setextrahttpheaders and that implies all the frames (local and cross-process) as well dedicated workers. I cannot remember off the top of my head if all kinds of workers actually support it but ideally they should.

index.bs
```diff
@@ -8778,6 +8797,138 @@ The [=remote end steps=] given <var ignore>session</var> and |command parameters
...
+     with |command parameters|["<code>contexts</code>"].
+
+  1. Let |target| be |session|'s [=session/extra headers=]'
+     [=extra headers/navigable headers=]
+
+  1. For |navigable| in |navigables|:
+
+    1. Set |target|[|navigable|] to |headers|.
+
+    1. Return [=success=] with data null.
```
```suggestion
  1. Return [=success=] with data null.
```

(otherwise we return after the first navigable update)

PR #961, opened 2025-07-11
Convert scripts/cddl into an ESM

jgraham:
This avoids a CI failure that's started happening with CommonJS imports


scripts/cddl/generate.js
```diff
@@ -8,13 +8,15 @@
  * - pathToSpec: a path to the specification to extract CDDL definitions from. Default: _dirname/../../index.bs.
```
whimboo, line 8: When we have those scripts now as ESM modules we probably should as well update the filename to have the `.mjs` extension. See https://nodejs.org/api/esm.html#enabling for more details.

PR #965, opened 2025-07-17
fix: Always resume pending collections in maybe collect body

juliandescottes:
Two issues spotted while implementing for Firefox:
- should always resume in case a collected data is found
- size and bytes should only be set when enough space could be allocated.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6429,11 +6429,11 @@ To <dfn>maybe collect network response body</dfn> given |sessions|, |request| an
 
       1. [=Allocate size to record data=] given |size|.
 
-1. Set |collected data|'s <code>bytes</code> to |bytes|.
+      1. Set |collected data|'s <code>bytes</code> to |bytes|.
+
+      1. Set |collected data|'s <code>size</code> to |size|.
 
 1. Set |collected data|'s <code>pending</code> to false.
-
-1. Set |collected data|'s <code>size</code> to |size|.
 
 1. [=Resume=] with "<code>network data collected</code>" and (|request|'s [=request id=], "response").
 
--- a/index.bs
+++ b/index.bs
@@ -6366,9 +6366,22 @@ To <dfn>maybe collect network response body</dfn> given |sessions|, |request| an
 
    Note: For redirects, only the final response body is stored.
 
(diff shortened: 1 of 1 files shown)


juliandescottes:
@jgraham @OrKoN gentle ping. I'm moving forward with the implementation on Firefox side assuming those changes are ok.

sadym-chromium:
> size and bytes should only be set when enough space could be allocated.

I don't think this issue is addressed.

index.bs
```diff
@@ -6429,11 +6439,9 @@ To <dfn>maybe collect network response body</dfn> given |sessions|, |request| an
 
       1. [=Allocate size to record data=] given |size|.
 
-1. Set |collected data|'s <code>bytes</code> to |bytes|.
-
-1. Set |collected data|'s <code>pending</code> to false.
+      1. Set |collected data|'s <code>bytes</code> to |bytes|.
```
juliandescottes, line 6442: cc @sadym-chromium 

Increasing the indentation here and for the following line ensures we only set the `size` and `bytes` if we allocated size for this network data. If we reach this point we already know that the size of the data is less than the total max size and that it matches a collector.

juliandescottes:
> > size and bytes should only be set when enough space could be allocated.
> 
> I don't think this issue is addressed.

Thanks for the review, see my inline comment pointing at the fix.

index.bs
sadym-chromium: Correct, but `[=Allocate size to record data=]` does not return errors.

OrKoN (approved):

index.bs
juliandescottes: `[=Allocate size to record data=]` is always expected to succeed on the spec side at the moment, it doesn't perform any check. If implementations have to handle errors there, it's up to the implementation to deal with this. Not sure I understand the question.

sadym-chromium: Oh, I missed this condition:
> If size is less than or equal to collector’s max encoded data size, [append](https://infra.spec.whatwg.org/#list-append) collector’s collector to collected data’s collectors.

juliandescottes:
Thanks for the reviews, merging!

PR #970, opened 2025-07-22
`browser.setDownloadBehavior`

sadym-chromium:
In puppeteer, [there is a way](https://pptr.dev/api/puppeteer.browsercontextoptions) to [deny all the downloads](https://pptr.dev/api/puppeteer.downloadpolicy), or allow for [downloading to a custom folder](https://pptr.dev/api/puppeteer.downloadbehavior#downloadpath). This PR is intended to provide a way to customize the download behavior per user context or globally.

1. Should be followed up with [HTML PR](https://github.com/whatwg/html/pull/11474) invoking new hooks.
1. Add a new hook "WebDriver BiDi download will begin" which is intended to replace "WebDriver BiDi download started".
1. "WebDriver BiDi download will begin" returns "download behavior struct", which instructs the user agent how to handle downloads.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/970.html" title="Last updated on Sep 16, 2025, 5:43 PM UTC (5bad586)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/970/f1d6555...5bad586.html" title="Last updated on Sep 16, 2025, 5:43 PM UTC (5bad586)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5611,7 +5611,7 @@ The [=remote end event trigger=] is the
 The [=remote end event trigger=] is the <dfn export>WebDriver BiDi download
 started</dfn> steps given |navigable| and |navigation status|:
 
-Issue: Removed after HTML spec switched to [=WebDriver BiDi download will begin=]
+Issue: Remove after HTML spec switched to [=WebDriver BiDi download will begin=]
 (https://github.com/whatwg/html/pull/11474).
 
 1. Let |navigation info| be the result of [=get the navigation info=] given |navigable| and


index.bs
```diff
@@ -3022,6 +3023,114 @@ those after the window has reached its new state.
...
 </div>
 
+#### The browser.setDownloadBehavior Command #### {#command-browser-setDownloadBehavior}
+
+A <dfn>download behavior struct</dfn> is a [=struct=] with:
+* [=struct/item=] named <dfn id="download-behavior-struct-allowed" for="download-behavior-struct">allowed</dfn> which is a boolean;
+* [=struct/item=] named <dfn id="download-behavior-struct-destination-folder" for="download-behavior-struct">destinationFolder</dfn> which is a string or null.
+
+A [=remote end=] has a <dfn>download behavior map</dfn> which is a weak map between
+[=user contexts=] or "<code>global</code>" and [=download behavior struct=] or null.
```
juliandescottes: nit: not sure what's the definition of weak map in spec world, but a "global" string doesn't sound like a typical weak map key?

sadym-chromium:
> I can see most automation libraries provide ways to set a download path, so that sounds like a useful feature. Libraries could already monitor downloads and move files manually to achieve the same effect, but still having it built in sounds more practical. And there are already preferences to drive this in most browsers anyway...

Testing libraries are not guaranteed to have access to the filesystem, so I would leave this logic to the browser.

> Regarding the API itself, I can see CDP has this on the Browser domain, but it also has the download events there, whereas for us they are on BrowsingContext. I would have expected this on BrowsingContext, but interested to know what others think.

The `Browser` module was chosen as implementation-wise in CDP we cannot set the download behavior per browsing context, only per user context. So having this method in BrowsingContext module and not being able to configure per browsing context seems a bit odd.

index.bs

sadym-chromium:
@juliandescottes PTAL

index.bs
```diff
@@ -5457,11 +5572,52 @@ complete</dfn> steps given |navigable| and |navigation status|:
...
+   |navigable| and |navigation status|.
+
+1. Let |params| be a [=/map=] matching the
+   <code>browsingContext.DownloadWillBeginParams</code> production, with the
+   <code>context</code> field set to |navigation info|["<code>context</code>"], the
+   <code>navigation</code> field set to |navigation info|["<code>navigation</code>"],
+   the <code>timestamp</code> field set to
+   |navigation info|["<code>timestamp</code>"], the <code>url</code> field set to
+   |navigation info|["<code>url</code>"] and <code>suggestedFilename</code> field set
+   to |navigation status|'s <code>suggestedFilename</code>.
```
juliandescottes, line 5586: I had some test updates planned for this at https://phabricator.services.mozilla.com/D264048 (before I knew your PR was touching this).

I think they should still be aligned with the proposed changes but Chrome is failing them in 2 spots:
- there is a navigation id set for the download created by a download attribute
- the navigation id for downloadWillBegin is different from the navigation id of navigationStarted for the navigation download. 

First item sounds ok, I guess you will update the implementation. But for the second item, do you agree the navigation ids in the 2 events should be equal?

juliandescottes:
@sadym-chromium it seems the current HTML PR generates a random UUID for the navigation in case of download attribute links, but in the meeting we said:

> orkon: We discussed with sadym. I think we should not emit a navigationStarted event. Instead, the navigation ID should be optional.

Am I misunderstanding the PR or did we minute this incorrectly?

sadym-chromium:
> @sadym-chromium it seems the current HTML PR generates a random UUID for the navigation in case of download attribute links, but in the meeting we said:
> 
> > orkon: We discussed with sadym. I think we should not emit a navigationStarted event. Instead, the navigation ID should be optional.
> 
> Am I misunderstanding the PR or did we minute this incorrectly?

I forgot to update the spec. Now it is null for [downloading hyperlinks](https://whatpr.org/html/11474/links.html#downloading-hyperlinks).

index.bs
```diff
@@ -5457,11 +5572,52 @@ complete</dfn> steps given |navigable| and |navigation status|:
...
+1. Return |download behavior|.
+
+</div>
+
 <div algorithm>
 The [=remote end event trigger=] is the <dfn export>WebDriver BiDi download
 started</dfn> steps given |navigable| and |navigation status|:
 
- 1. Let |navigation info| be the result of [=get the navigation info=] given |navigable| and
+Issue: TODO: Removed after HTML spec switched to [=WebDriver BiDi download will begin=].
```
juliandescottes: Maybe mention https://github.com/whatwg/html/pull/11474 explicitly?

juliandescottes (approved):
Works for me, thanks. 
If there are issues with the tests I mentioned in my other comment, can you add your feedback on Phabricator?

index.bs
sadym-chromium: > there is a navigation id set for the download created by a download attribute

It should be `null`.

> the navigation id for downloadWillBegin is different from the navigation id of navigationStarted for the navigation download.

Agree, the navigation IDs for navigation lead to the download and in the download should be equal. IDK if we can implement it tough.

index.bs

index.bs
```diff
@@ -3022,6 +3023,120 @@ those after the window has reached its new state.
...
+* [=struct/item=] named <dfn id="download-behavior-struct-destination-folder" for="download-behavior-struct">destinationFolder</dfn> which is a string or null.
+
+A [=remote end=] has a <dfn>download behavior</dfn> which is a [=struct=] with an [=struct/item=]
+named <dfn for="download behavior">default download behavior</dfn>, which is a
+[=download behavior struct=] or null, and an [=struct/item=] named <dfn
+for="download behavior">user context download behavior</dfn>, which is a weak map between
+[=user contexts=] and [=download behavior struct=].
+
+
+The <dfn export for=commands>browser.setDownloadBehavior</dfn> command returns a
```
OrKoN: The sentence does not look correct?

sadym-chromium: removed artifact

index.bs
```diff
@@ -5457,11 +5568,53 @@ complete</dfn> steps given |navigable| and |navigation status|:
...
+1. Return |download behavior|.
+
+</div>
+
 <div algorithm>
 The [=remote end event trigger=] is the <dfn export>WebDriver BiDi download
 started</dfn> steps given |navigable| and |navigation status|:
 
- 1. Let |navigation info| be the result of [=get the navigation info=] given |navigable| and
+Issue: Removed after HTML spec switched to [=WebDriver BiDi download will begin=]
```
OrKoN:
```suggestion
Issue: Remove after HTML spec switched to [=WebDriver BiDi download will begin=]
```
?

PR #972, opened 2025-08-01
fix: Do not link data collection to network event subscription

juliandescottes:
With the current spec, you could essentially get deadlock if you subscribe to responseStarted but not to responseCompleted and try to use network.getData.

This is because we only trigger the data collector algorithms for sessions which subscribe to those events. The algorithms should be executed regardless of the state of network event subscription.

***
<a href="https://pr-preview.s3.amazonaws.com/juliandescottes/webdriver-bidi/pull/972.html" title="Last updated on Aug 1, 2025, 4:07 PM UTC (68fa7f3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/972/5c647ae...juliandescottes:68fa7f3.html" title="Last updated on Aug 1, 2025, 4:07 PM UTC (68fa7f3)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6534,11 +6534,11 @@ To <dfn>match collector for navigable</dfn> given |collector| and |navigable|:
 </div>
 
 <div algorithm>
-To <dfn export>clone network response body</dfn> given |sessions|, |request| and |response body|:
+To <dfn export>clone network response body</dfn> given |request| and |response body|:
 
 Note: This hook is intended to be triggered by the fetch spec when the response is set.
 
-1. For each |session| in |sessions|:
+1. For each |session| in [=active BiDi sessions=]:
 
    1. If |session|'s [=network collectors=] is not [=list/empty=]:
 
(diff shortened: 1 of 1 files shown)


juliandescottes:
cc @jgraham @OrKoN

juliandescottes:
There is something slightly odd here where you could be collecting data without having any way to retrieve it because you would not be notified about the request id via a network event. I don't have a good solution that would avoid deadlocks while still ensuring that the client has the necessary events set up to retrieve the request id.

In theory we could add a check when we create the collected data (ie when we clone the response body). But it's not clear what the requirements for the check should be. The only "mandatory" events for a network request are "beforeRequestSent" and "responseStarted", the 3 others depend on the behavior result of the request. 

I feel like it's OK to leave this to clients to setup, and trying to make this fool-proof might make the spec too convoluted?

OrKoN (approved):

PR #996, opened 2025-09-04
wip: refactor the step to allow non-network transports

OrKoN:
I didn't know of a good example of a spec polymorphism so attempted to rewrite as I saw fit.


index.bs
```diff
@@ -33,7 +33,7 @@ spec: RFC6455; urlPrefix: https://datatracker.ietf.org/doc/html/rfc6455
     text: Reading the Client's Opening Handshake; url: section-4.2.1
     text: %x1 denotes a text frame; url: section-5.2
     text: Send a WebSocket Message; url: section-6.1
-    text: A WebSocket Message Has Been Received; url: section-6.2
+    text: a WebSocket Message Has Been Received; url: section-6.2
```
sadym-chromium, line 36: nit: I don't think it is case-sensitive, so no need it this change.

PR #1002, opened 2025-09-10
[Editorial] Explicit result type for commands

Lightning00Blade:
Provides a API that the Command always have a equivalent type with suffix `Result`.
This would allow clients like https://github.com/GoogleChromeLabs/webdriver-bidi-protocol
to create Mapping between Command Response.
It also allows to easily jump to the definitions. 
Currently there are some provided mappings, but they are manually created.
```ts
export interface Commands {
  'bluetooth.handleRequestDevicePrompt': {
    params: BidiBluetooth.Bluetooth.HandleRequestDevicePromptParameters;
    returnType: Bidi.EmptyResult;
  };
 }
  ```

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/1002.html" title="Last updated on Sep 24, 2025, 5:09 PM UTC (cc61894)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/1002/4a1181a...cc61894.html" title="Last updated on Sep 24, 2025, 5:09 PM UTC (cc61894)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -488,13 +488,16 @@ ErrorResponse = {
 }
 
 ResultData = (
+  SessionResult /
+  BrowserResult /
   BrowsingContextResult /
-  EmptyResult /
+  EmulationResult /
   NetworkResult /
   ScriptResult /
-  SessionResult /
   StorageResult /
-  WebExtensionResult
+  InputResult /
+  WebExtensionResult /
+  EmptyResult
 )
 
 EmptyResult = {
@@ -1622,9 +1625,11 @@ SessionCommand = (
 
 <pre class="cddl" data-cddl-module="local-cddl">
 SessionResult = (
-   session.NewResult /
-   session.StatusResult /
-   session.SubscribeResult
+  session.EndResult /
+  session.NewResult /
+  session.StatusResult /
+  session.SubscribeResult /
+  session.UnsubscribeResult
 )
 </pre>
 
@@ -1897,7 +1902,7 @@ This is a [=static command=].
       )
       </pre>
    </dd>
-   <dt>Result Type</dt>
+   <dt>Return Type</dt>
    <dd>
       <pre class="cddl" data-cddl-module="local-cddl">
       session.StatusResult = {
@@ -1951,7 +1956,7 @@ This is a [=static command=].
       }
       </pre>
    </dd>
-   <dt>Result Type</dt>
+   <dt>Return Type</dt>
    <dd>
       <pre class="cddl" data-cddl-module="local-cddl">
       session.NewResult = {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by OrKoN** [comment]
@jgraham @whimboo @sadym-chromium this is a convention that would simplify code generation a bit. Do you have any objections to landing this? (it is only an editorial change)

File: index.bs [hunk 1]
```diff
@@ -488,13 +488,16 @@ ErrorResponse = {
...
+  EmulationResult /
   NetworkResult /
   ScriptResult /
-  SessionResult /
   StorageResult /
-  WebExtensionResult
+  InputResult /
+  WebExtensionResult /
+  EmptyResult
 )
```
juliandescottes: Could we keep this alphabetically sorted?

index.bs
```diff
@@ -3083,12 +3092,17 @@ BrowsingContextCommand = (
...
 <pre class="cddl" data-cddl-module="local-cddl">
 BrowsingContextResult = (
+  browsingContext.ActivateResult /
   browsingContext.CaptureScreenshotResult /
+  browsingContext.CloseResult /
   browsingContext.CreateResult /
   browsingContext.GetTreeResult /
-  browsingContext.LocateNodesResult /
+  browsingContext.HandleUserPromptResult /
+  browsingContext.LocateNodes /
```
```suggestion
  browsingContext.LocateNodesResult /
```

Lightning00Blade: Nice catch!

index.bs
Lightning00Blade: Sure I just used the module order but alphabetically sounds OK to me as well.

index.bs
```diff
@@ -488,13 +488,16 @@ ErrorResponse = {
...
-  EmptyResult /
+  EmulationResult /
   NetworkResult /
   ScriptResult /
-  SessionResult /
   StorageResult /
-  WebExtensionResult
+  InputResult /
+  WebExtensionResult /
+  EmptyResult
```
juliandescottes: Relaying feedback from @whimboo 

Do we still need EmptyResult here?

Lightning00Blade: I think we can safely remove it.

juliandescottes:
@Lightning00Blade thanks for the replies! there's no further comment/concern from peers on the mozilla side, so I'll approve once the comments have been addressed.

OrKoN:
@Lightning00Blade I think after a rebase SetDownloadBehaviorResult is still wrapped in `<code>` instead of a cddl `<pre>` (see failing test).

juliandescottes (approved):
thanks lgtm

PR #1004, opened 2025-09-17
[editorial]: typed "WebDriver BiDi" hooks

sadym-chromium:
* Add types for all external ("WebDriver BiDi ...") hooks.
* Proper refernce to the the |navigation status| struct fields.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3070,8 +3070,8 @@ rather than <code>navigable</code>, and the protocol uses the term
 <code>context</code> to refer to navigables, particularly as a field in command
 and response parameters.
 
-The progress of navigation is communicated using an immutable
-<dfn export>WebDriver BiDi navigation status</dfn> struct, which has the following items:
+The progress of navigation is communicated using an immutable [=struct=]
+<dfn export>WebDriver BiDi navigation status</dfn>, which has the following [=struct/item|items=]:
 
 <dl>
   <dt><dfn export for="WebDriver BiDi navigation status" id="navigation-status-id">id</dfn></dt>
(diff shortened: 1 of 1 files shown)


sadym-chromium:
@OrKoN @jgraham WDYT?

index.bs
```diff
@@ -5550,13 +5559,13 @@ Issue: Remove after HTML spec switched to [=WebDriver BiDi download will begin=]
    <code>timestamp</code> field set to |navigation info|["<code>timestamp</code>"], the
    <code>url</code> field set to |navigation info|["<code>url</code>"] and
    <code>suggestedFilename</code> field set to |navigation status|'s
```
whimboo, line 5561: Is there actually a reason why you are using a mixture of `dict[name]` and `dict's name`? Shouldn't we always use the same approach? This applies to a lot of other places in this patch as well.

index.bs
```diff
@@ -5783,8 +5796,8 @@ failed</dfn> steps given |navigable| and |navigation status|:
 </dl>
 
 <div algorithm>
-The [=remote end event trigger=] is the <dfn export>WebDriver BiDi user prompt
-closed</dfn> steps given |window|, |type|, |accepted| and optional |user text|
+The [=remote end event trigger=] is the <dfn export>WebDriver BiDi user prompt closed</dfn> steps
+given {{Window}} |window|, string |type|, boolean |accepted| and optional text |user text|
```
whimboo, line 5800: If we are going to use the actual type for arguments we still have a lot of other algorithms which don't have them yet. Shall we maybe update them at once?

index.bs
```diff
@@ -5855,8 +5868,8 @@ To <dfn>get navigable's user prompt handler</dfn> given |type| and |navigable|:
 </div>
 
 <div algorithm>
-The [=remote end event trigger=] is the <dfn export>WebDriver BiDi user prompt
-opened</dfn> steps given |window|, |type|, |message|, and optional |default value|
+The [=remote end event trigger=] is the <dfn export>WebDriver BiDi user prompt opened</dfn> steps
+given {{Window}} |window|, string |type|, string |message|, and optional string |default value|
```
whimboo: Whether we use `string` or `text`. The closed variant of this event uses `text`. Given the CDDL definition we probably want to use `text` here as well.

```suggestion
given {{Window}} |window|, string |type|, string |message|, and optional text |default value|
```

index.bs
sadym-chromium: The `|navigation info|` is a `[=/map=]` returned by `[=get the navigation info=]`, so its fields should be accessed as dict properties. The `|navigation status|` is a `[=/struct=]` `[=WebDriver BiDi navigation status=]`, so we can refer to it's items by link.

index.bs
sadym-chromium: I don't think it is critical, and can be done (or not) later. The scope of this change is to provide types only to the externally-exposed hooks.

whimboo: Sure, that sounds fine. We can do that later.

whimboo (approved):
Seems to be fine now. Thanks!

index.bs
whimboo: Ok, thanks.

PR #1011, opened 2025-09-26
Add support for request dataType to network data collection

### LINKED ISSUE #748: Support reading request body
Related to https://github.com/w3c/webdriver-bidi/issues/747 but for network requests: there should be a way to read the body content of the outgoing requests. 

In Puppeteer, the data is exposed via https://pptr.dev/api/puppeteer.httprequest.postdata although there is a more structured interface in CDP postDataEntries https://chromedevtools.github.io/devtools-protocol/tot/Network/#type-Request that splits request data into parts.

It would be useful to support reading request data using WebDriver BiDi.

juliandescottes:
Fixes #748 

- Adds another entry point expected to be called from the fetch spec `clone network request body` (alternatively could be called explicitly from beforeRequestSent for now)
- Extracts a generic algorithm `maybe collect network data` from `maybe collect network response body`
- Update `maybe collect network response body` to call `maybe collect network data`
- Adds a new algorithm `maybe collect network request body` called from the beforeRequestSent steps

Otherwise this mostly reuses the logic added for response body collection. Commands do not need to be udpated AFAICT

***
<a href="https://pr-preview.s3.amazonaws.com/juliandescottes/webdriver-bidi/pull/1011.html" title="Last updated on Sep 29, 2025, 9:30 AM UTC (2eca438)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/1011/bf20c0b...juliandescottes:2eca438.html" title="Last updated on Sep 29, 2025, 9:30 AM UTC (2eca438)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6877,7 +6877,7 @@ To <dfn>match collector for navigable</dfn> given |collector| and |navigable|:
 </div>
 
 <div algorithm>
-To <dfn export>clone network request body</dfn> given |request|:
+To <dfn export>clone network request body</dfn> given [=/request=] |request|:
 
 Note: This hook is intended to be triggered by the fetch spec when the request body has been safely extracted.
 See step 9 of https://fetch.spec.whatwg.org/#concept-fetch
@@ -6998,7 +6998,8 @@ To <dfn>maybe collect network response body</dfn> given |request| and |response|
 </div>
 
 <div algorithm>
-To <dfn>maybe collect network data</dfn> given |request|, |collected data|, |size| and |data type|:
+To <dfn>maybe collect network data</dfn> given [=/request=] |request|,
+[=network data=] |collected data|, js-uint |size| and [=network.DataType=] |data type|:
 
 1. Set |collected data|'s <code>pending</code> to false.
 


index.bs
```diff
@@ -6876,11 +6876,40 @@ To <dfn>match collector for navigable</dfn> given |collector| and |navigable|:
 
 </div>
 
+<div algorithm>
+To <dfn export>clone network request body</dfn> given |request|:
```
sadym-chromium: I'd propose to use typed arguments for externally used hooks: https://github.com/w3c/webdriver-bidi/pull/1004
```suggestion
To <dfn export>clone network request body</dfn> given [=/request=] |request|:
```

index.bs
```diff
@@ -6876,11 +6876,40 @@ To <dfn>match collector for navigable</dfn> given |collector| and |navigable|:
...
+
+Note: This hook is intended to be triggered by the fetch spec when the request body has been safely extracted.
+See step 9 of https://fetch.spec.whatwg.org/#concept-fetch
```
sadym-chromium, line 6883: I don't think this comment is valid.

index.bs
sadym-chromium: oh, despite the comment, it is not used externally

index.bs
```diff
@@ -6925,19 +6954,51 @@ To <dfn>maybe abort network response body collection</dfn> given |request|:
...
+   model such that it uses less size than storing the decoded data, as long as the data
+   returned to clients in getData is identical to the decoded data. The potential
+   tradeoff between storage and performance is up to the implementation.
+
+1. [=Maybe collect network data=] with |request|, |collected data|, |size| and "response".
+
+</div>
+
+<div algorithm>
+To <dfn>maybe collect network data</dfn> given |request|, |collected data|, |size| and |data type|:
```
sadym-chromium: Even though we usually don't do it, we can provide types here for readability:
```suggestion
To <dfn>maybe collect network data</dfn> given [=/request=] |request|, [=network data=] |collected data|, js-uint |size| and [=network.DataType=] |data type|:
```

index.bs
juliandescottes: What do you mean?

index.bs

index.bs
sadym-chromium: The hook `clone network request body` is not called from fetch spec

index.bs
juliandescottes: In theory it should be called from fetch, but yes at the moment it's blocked on https://github.com/whatwg/fetch/pull/1540 

Alternatively I can rewrite the PR to trigger this algorithm from the beforeRequestSent event handler if you prefer?

index.bs
juliandescottes: If that's about the fact the fetch spec doesn't include this, see my other comment about https://github.com/whatwg/fetch/pull/1540

And same proposal: we can trigger this from beforeRequestSent if we prefer?

Otherwise can you explain what is not valid?

juliandescottes:
Thanks for the review! I addressed the suggestions to add types. Let me know if you want to switch to an internal trigger for the algorithm for now. I'll discuss internally to make the fetch PR move.

index.bs
sadym-chromium: This comment is not a blocker (I already approved the PR). 

I cannot see in the [fetch RP](https://github.com/whatwg/fetch/pull/1540) any reference to this hook. Maybe this comment meant to underline that this is a part of the `WebDriver BiDi before request sent` hook, which is invoked by the fetch spec in [fetch RP](https://github.com/whatwg/fetch/pull/1540)?

index.bs
sadym-chromium: Resolving in favor of https://github.com/w3c/webdriver-bidi/pull/1011#discussion_r2387320976

index.bs
juliandescottes: Oh no you're right, I haven't updated James' PR. Not quite sure how to do that but I can give it a try.

And in theory, WebDriver BiDi before request sent is a bit too late, because other consumers might have read the stream already, so we really need to clone it as soon as possible.

jgraham (approved):

PR #1010, opened 2025-09-26
[editorial] extend type references for data collectors

sadym-chromium:
And add note clarifying behavior.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/1010.html" title="Last updated on Sep 26, 2025, 10:25 AM UTC (706eb4d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/1010/915f5e3...706eb4d.html" title="Last updated on Sep 26, 2025, 10:25 AM UTC (706eb4d)">Diff</a>

PR #1013, opened 2025-09-26
[editorial] rename

sadym-chromium:
To avoid confusion, rename some internal struct items and variables. Do not rename command's fields to avoid breaking changes.

* /s/max encoded data size/max encoded item size
* /s/max total data size/max total collected size

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6811,16 +6811,16 @@ A <dfn>network data</dfn> is a [=/struct=] with:
 * [=struct/Item=] named <dfn for="network-data">type</dfn>, which is a <code>[=network.DataType=]</code>.
 
 A <dfn for="network">collector</dfn> is a [=/struct=] with:
-* [=struct/Item=] named <dfn for="network-collector">max encoded data size</dfn>, which is a js-uint;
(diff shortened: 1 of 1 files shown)


sadym-chromium:
@juliandescottes I re-applied the change, as you asked in https://github.com/w3c/webdriver-bidi/pull/1010#discussion_r2381971194

juliandescottes (approved):
My comment on the other PR was that we do not need to revert but thanks for resubmitting to review anyway. In general it's good to have some context as to why we are renaming things. 

I don't really find the new names clearer but I didn't really find the previous ones that great either. If that works better for you, that's fine, I think there are pros and cons with both.

One thing to consider is that now the command arguments no longer match the properties of the data collector. Do you want to change the API? It landed relatively recently so maybe we can have a short backward compat window while we update known consumer. The more we wait the more complicated it will be to update.

sadym-chromium:
> Do you want to change the API? 

It would be challenging for us to update API directly, so we would need to accept both `maxEncodedDataSize` and `maxEncodedItemSize` for some time. I don't think it is required, as we have explicit notes describing the behavior.

PR #1020, opened 2025-10-21
Update "emulation.setLocaleOverride" to also override "Accept-Language" header.

lutien:
In the relation with "network.setExtraHeaders", my approach was that "emulation.setLocaleOverride" always takes precedent and overrides the values which might be set with "network.setExtraHeaders", since clients, when they use "emulation.setLocaleOverride", probably want locale values to be aligned across APIs. But let me know if I missed some use case scenarios.

***
<a href="https://pr-preview.s3.amazonaws.com/lutien/webdriver-bidi/pull/1020.html" title="Last updated on Oct 21, 2025, 9:57 AM UTC (158790a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/1020/18639ef...lutien:158790a.html" title="Last updated on Oct 21, 2025, 9:57 AM UTC (158790a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -9635,10 +9635,12 @@ navigables|:
    navigables. In that case it's unclear in which order to override the headers.
 
 1. [=Update headers=] with |request| and |session|'s
-   [=session/extra headers=]' [=extra headers/default headers=]
+   [=session/extra headers=]' [=extra headers/default headers=].
 
 1. Let |user context headers| be |session|'s [=session/extra headers=]'
    [=extra headers/user context headers=].
+
+1. Let |accept language override| be a null.
 
 1. For |navigable| in |related navigables|:
 
@@ -9646,7 +9648,10 @@ navigables|:
 
   1. If |user context headers| [=map/contains=] |user context|
      then [=update headers=] with |request| and
-     |user context headers|[|user context|]
+     |user context headers|[|user context|].
+
+   1. If [=locale overrides map=] [=map/contains=] |user context|,
+      set |accept language override| to [=locale overrides map=][|user context|].
 
 1. Let |navigable headers| be |session|'s [=session/extra headers=]'
    [=extra headers/navigable headers=].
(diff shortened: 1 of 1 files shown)


sadym-chromium:
> In the relation with "network.setExtraHeaders", my approach was that "emulation.setLocaleOverride" always takes precedent and overrides the values which might be set with "network.setExtraHeaders", since clients, when they use "emulation.setLocaleOverride", probably want locale values to be aligned across APIs. But let me know if I missed some use case scenarios.

I would expect the emulated locale not to be distinguishable from the one selected by the user, but this behavior creates an observable difference in emulated and non-emulated locale behavior.

sadym-chromium:
Also the [Fetch spec](https://fetch.spec.whatwg.org/#concept-fetch) adds `Accept-Language` only if it is not yet set. E.g.:
```
await fetch(..., {headers: { 'Accept-Language': 'SOME_STRING' }})
```
In this case overriding the user's 'Accept-Language' headers can mask some issues, like `fetch` in JS code mistakenly sending a wrong `Accept-Language` header.

I believe the better approach would be to define "an appropriate `Accept-Language` [header value](https://fetch.spec.whatwg.org/#header-value)" in [fetch](https://fetch.spec.whatwg.org/#concept-fetch), step 12.

lutien:
> Also the [Fetch spec](https://fetch.spec.whatwg.org/#concept-fetch) adds `Accept-Language` only if it is not yet set. E.g.:
> 
> ```
> await fetch(..., {headers: { 'Accept-Language': 'SOME_STRING' }})
> ```
> 
> In this case overriding the user's 'Accept-Language' headers can mask some issues, like `fetch` in JS code mistakenly sending a wrong `Accept-Language` header.
> 
> I believe the better approach would be to define "an appropriate `Accept-Language` [header value](https://fetch.spec.whatwg.org/#header-value)" in [fetch](https://fetch.spec.whatwg.org/#concept-fetch), step 12.

Yeah, I think it makes sense. I've created [a PR for the fetch spec](https://github.com/whatwg/fetch/pull/1869). I've reused the hook added for `navigator.language`, and I think nothing has to be done here. So I'm going to close this PR.

@sadym-chromium, could you also create an issue for this feature? (I need for the PR check list)

PR #1023, opened 2025-10-28
Reject with no such user context in setExtraHeaders and addDataCollector

juliandescottes:
Update two commands which incorrectly throw invalid argument error when an invalid user context id is provided.
The wdspec tests for both commands already assume that `no such user context` is thrown, so there is no test to update here.

***
<a href="https://pr-preview.s3.amazonaws.com/juliandescottes/webdriver-bidi/pull/1023.html" title="Last updated on Oct 28, 2025, 9:37 AM UTC (4af22c6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/1023/7fafc35...juliandescottes:4af22c6.html" title="Last updated on Oct 28, 2025, 9:37 AM UTC (4af22c6)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -8759,7 +8759,7 @@ The [=remote end steps=] given |session| and |command parameters| are:
 
        1. Let |user context| be [=get user context=] with |user context id|.
 
-       1. If |user context| is null, return [=error=] with [=error code=] [=invalid argument=].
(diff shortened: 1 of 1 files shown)


juliandescottes:
> Thanks! Do we need an update of the tests for addDataCollector?

The wdspec tests for both commands already assume that no such user context is thrown, so there is no test to update here.

Issue #956, opened 2025-07-10
Apply emulations globally


sadym-chromium:
Currently, emulations can be applied only to a specific browsing or user contexts. Consider extending emulation commands to apply globally if no browsing or user contexts are provided.

The [WebDriver configuration](https://w3c.github.io/webdriver-bidi/#webdriver-configuration) is aimed to help with the migration.

Also it would be great to migrate all other applicable configurations to be [WebDriver configuration](https://w3c.github.io/webdriver-bidi/#webdriver-configuration).

Issue #151, opened 2021-11-29
`browsingContext.create` should be able to support WebDriver classic `New Window` command


bwalderman:
To support future efforts to reformulate WebDriver Classic on top of WebDriver BiDi, it should be possible implement [New Window](https://w3c.github.io/webdriver/#new-window) in terms of [browsingContext.create](https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/133.html#command-browsingContext-create).

There are a couple small gaps in the current spec PR #133

- The classic command uses the implicit [current browsing context](https://w3c.github.io/webdriver/#dfn-current-browsing-context) to determine which OS window the implementation should try to create the new browsing context in. However, BiDi doesn't have a concept of a "current" or "default" browsing context (by design).
- The classic command result value has a `type` that indicates whether the new browsing context was created as a "tab" or a "window". This depends on whether the new context is created in the same OS window as the current browsing context or a different OS window.

We'll likely need to add a parameter to the `browsingContext.create` command that the WebDriver classic spec can use to explicitly set the desired target window. Exactly how this looks will depend on whether we decide to introduce OS-level windows as a concept in WebDriver BiDi.

If we do add a "Window" concept like this, the `browsingContext.create` command could accept an optional `"window"` parameter and use that as a hint of where to create the new browsing context. The reformulated WebDriver classic spec could maintain a handle to the [current browsing context's](https://w3c.github.io/webdriver/#dfn-current-browsing-context) parent "Window" to use for this purpose. Otherwise, if we don't add a "Window" concept, we can still support the needs of WebDriver classic by adding a `"context"` parameter.

PR #1030, opened 2025-11-11
Add "emulation.setScreenSettingsOverride" command.

### LINKED ISSUE #981: Support overriding screen dimensions
To support emulation of mobile devices, the clients should be able to emulate `screen.width/screen.heights` properties. CDP also supports it via [Emulation.setDeviceMetricsOverride](https://chromedevtools.github.io/devtools-protocol/tot/Emulation/#method-setDeviceMetricsOverride). 
The question is here what would be the best interface. I see a few options:
- the viewport settings set via `browsingContext.setViewport` command should automatically apply to screen dimensions;
- `browsingContext.setViewport` command has a new argument/s to set the screen dimensions independently of viewport;
- we have a new command to just set the screen settings.

lutien:
Closes #981.

My approach was to more or less to document the current implementation of Responsive Design mode behaviour of browsers. That's why I think for this specific command it makes more sense to override it on the CSS spec level since it touches exactly the data which we currently override. That's also why I think we have to override not just dimensions, but the coordinates as well (they are currently reset when setting the dimensions). It might be useful in the future to allow in the future to set coordinates as well (?). So that's why I came up with the name  `setScreenAreaOverride` rather than `setScreenDimensionsOverride` or something else.

The draft PR in the CSS spec to illustrate how it might look like there: https://github.com/w3c/csswg-drafts/pull/13091.

Let me know what you think.

***
<a href="https://pr-preview.s3.amazonaws.com/lutien/webdriver-bidi/pull/1030.html" title="Last updated on Nov 20, 2025, 3:44 PM UTC (f77e482)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/1030/5ce7a32...lutien:f77e482.html" title="Last updated on Nov 20, 2025, 3:44 PM UTC (f77e482)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3217,7 +3217,7 @@ an [=struct/item=] named <dfn attribute for="screen settings">width</dfn> which 
 an [=struct/item=] named <dfn attribute for="screen settings">x</dfn> which is an integer,
 an [=struct/item=] named <dfn attribute for="screen settings">y</dfn> which is an integer.
 
-A [=remote end=] has a <dfn>screen settings overrides</dfn> which is [=struct=]
+A [=remote end=] has a <dfn>screen settings overrides</dfn> which is a [=struct=] with
 an [=struct/item=] named <dfn for="screen settings overrides">user context screen settings</dfn>,
 which is a weak map between [=user contexts=] and [=screen settings=],
 and an [=struct/item=] named <dfn for="screen settings overrides">navigable screen settings</dfn>,


sadym-chromium:
With the [\[feature request\] Specification for emulation · Issue #151 · w3c/window-management](https://github.com/w3c/window-management/issues/151) for [Window Management specification](https://w3c.github.io/window-management), I believe eventually we want to emulate screens fully. Can we make the command more generic so that it can be extended with other emulated properties?

lutien:
> With the [[feature request] Specification for emulation · Issue #151 · w3c/window-management](https://github.com/w3c/window-management/issues/151) for [Window Management specification](https://w3c.github.io/window-management), I believe eventually we want to emulate screens fully. Can we make the command more generic so that it can be extended with other emulated properties?

Sure, we can do that. I was just thinking that since we already have orientation override in a separate command and device pixel ration in the `setViewport` command. Maybe we want to have the rest separate? But maybe the other way is true, and we rather want to group it and maybe later move other properties in this command?

jgraham:
I think grouping makes more sense than having many commands that all need to be sent separately. If it makes sense later we can make existing commands options for the new command.

lutien:
Alright, I've renamed the command to `emulation.setScreenSettingsOverride`.

index.bs
```diff
@@ -3210,6 +3211,15 @@ A [=remote end=] has a <dfn>viewport overrides map</dfn> which is a weak map bet
 A [=remote end=] has a <dfn>locale overrides map</dfn> which is a weak map between
 [=navigables=] or [=user contexts=] and string.
 
+A <dfn>screen area</dfn> is a [=struct=] with an [=struct/item=] named
+<dfn attribute for="screen area">height</dfn> which is an integer,
+a [=struct/item=] named <dfn attribute for="screen area">width</dfn> which is an integer,
+a [=struct/item=] named <dfn attribute for="screen area">x</dfn> which is an integer,
+a [=struct/item=] named <dfn attribute for="screen area">y</dfn> which is an integer.
+
+A [=remote end=] has a <dfn>screen area overrides map</dfn> which is a weak map between
```
sadym-chromium: nit: @jgraham used a struct to store the config for [session's extra headers](https://www.w3.org/TR/webdriver-bidi/#session-extra-headers). Consider using this pattern as well.

index.bs
```diff
@@ -6445,6 +6455,116 @@ The [=remote end steps=] with |command parameters| and |session| are:
...
+        params: emulation.SetScreenSettingsOverrideParameters
+      )
+
+      emulation.ScreenArea = {
+        width: js-uint,
+        height: js-uint
+      }
+
+      emulation.SetScreenSettingsOverrideParameters = {
+        screenArea: emulation.ScreenArea / null,
```
sadym-chromium, line 6481: should we add `aviableScreenArea` for completeness? 

```suggestion
        totalScreenArea: emulation.ScreenArea / null,
        availableScreenArea: emulation.ScreenArea / null,
```

index.bs
```diff
@@ -6445,6 +6455,116 @@ The [=remote end steps=] with |command parameters| and |session| are:
...
+   <dd>
+      <pre class="cddl" data-cddl-module="local-cddl">
+      emulation.SetScreenSettingsOverrideResult = EmptyResult
+      </pre>
+   </dd>
+</dl>
+
+<div algorithm>
+
+The <dfn export>WebDriver BiDi emulated screen area</dfn> steps given a given [=/navigable=] |navigable|:
```
```suggestion
The <dfn export>WebDriver BiDi emulated screen area</dfn> steps given [=/navigable=] |navigable|:
```

sadym-chromium: I assume it would be easier to pass [environment settings object](https://html.spec.whatwg.org/multipage/webappapis.html#environment-settings-object) `environment settings` to this algo, instead of defining the algo of finding the related navigables it in the CSSOM spec.

index.bs
```diff
@@ -6445,6 +6455,116 @@ The [=remote end steps=] with |command parameters| and |session| are:
...
+
+The [=remote end steps=] with |command parameters| are:
+
+1. If |command parameters| [=map/contains=] "<code>userContexts</code>"
+   and |command parameters| [=map/contains=] "<code>contexts</code>",
+   return [=error=] with [=error code=] [=invalid argument=].
+
+1. If |command parameters| doesn't [=map/contain=] "<code>userContexts</code>"
+   and |command parameters| doesn't [=map/contain=] "<code>contexts</code>",
+   return [=error=] with [=error code=] [=invalid argument=].
```
sadym-chromium, line 6541: Should not we emulate it globally instead?

index.bs
```diff
@@ -6445,6 +6455,116 @@ The [=remote end steps=] with |command parameters| and |session| are:
...
+
+      1. Otherwise, [=map/set=] [=screen area overrides map=][|user context|] to |emulated screen area|.
+
+      1. [=list/For each=] |top-level traversable| of the list of all
+         [=/top-level traversables=] whose [=associated user context=] is
+         |user context|:
+
+         1. [=list/Append=] |top-level traversable| to |navigables|.
+
+1. For each |navigable| of |navigables|:
```
sadym-chromium: The config per navigable should be stored in the `1. If the <code>contexts</code> field of |command parameters| is present:` case, as otherwise setting `null` to user context would remove per-navigable emulations.

lutien: Ah, right! Fixed ✅

index.bs
lutien: Sure, updated ✅

index.bs
lutien: So that's a good question, from what I could see that's not what either Chrome or Firefox DevTools support at the moment. In both case, the available screen area dimension just set to screen area dimensions. But implementation-wise, we probably could do this. So I guess the question is if that's what the users/clients need, do you know from Puppeteer side if that's something what is requested?

If we do this, we probably have to think about some kind of validation, because the values can not be completely different (the available area should be inside total area)? It's also probably doesn't make much sense to reset them individually?

index.bs
lutien: It actually looks like in the CSSOM spec they do work with navigables (or better say still with browsing context) a lot, see e.g. in https://drafts.csswg.org/cssom-view/#dom-window-moveto step 2. So I used the same pattern in https://github.com/w3c/csswg-drafts/pull/13091.

index.bs
lutien: I think we maybe first should do this for `setViewport` (making possible to set it globally)? I don't think it makes much sense setting only screen settings globally without viewport.

sadym-chromium: There is an issue to enable global config for all the emulations: https://github.com/w3c/webdriver-bidi/issues/956

sadym-chromium:
General thoughts: eventually, to emulate [`window.getScreenDetails().screens`](https://www.w3.org/TR/window-management/#api-screendetails-screens-attribute), we will need to introduce the concept of emulated screen. I wonder how this approach can be extended later on to fully support emulations of multiple screens, and moving browser windows around them.

index.bs
sadym-chromium: Eventually, we will want to emulate the full [window management](https://www.w3.org/TR/window-management), including total and available screen areas. I wonder what is the current behavior of Chromium and Firefox behavior? Does it emulate both of the values to the provided values?

lutien: At the moment, both Chromium and Firefox emulate total and available screen areas to the same provided values, so you can not set them to different values.

index.bs
lutien: Yeah, I guess we should talk about it at the next spec meeting 🙂

lutien:
> General thoughts: eventually, to emulate [`window.getScreenDetails().screens`](https://www.w3.org/TR/window-management/#api-screendetails-screens-attribute), we will need to introduce the concept of emulated screen. I wonder how this approach can be extended later on to fully support emulations of multiple screens, and moving browser windows around them.

I think we could introduce then screen IDs and specify this emulation for a screen id. We will have to decide then if the emulation keeps getting attached to a browsing context/user context no matter on which screen it's, or it will be attached to that screen and if browser window are moved around it uses different settings. 
So I think we can still make it work, but as you say that's something what we will have to do eventually.

sadym-chromium:
> > General thoughts: eventually, to emulate [`window.getScreenDetails().screens`](https://www.w3.org/TR/window-management/#api-screendetails-screens-attribute), we will need to introduce the concept of emulated screen. I wonder how this approach can be extended later on to fully support emulations of multiple screens, and moving browser windows around them.
> 
> I think we could introduce then screen IDs and specify this emulation for a screen id. We will have to decide then if the emulation keeps getting attached to a browsing context/user context no matter on which screen it's, or it will be attached to that screen and if browser window are moved around it uses different settings. So I think we can still make it work, but as you say that's something what we will have to do eventually.

It's not a blocker.

index.bs
sadym-chromium: ok, then we can keep the property name `screenArea`, and extend the `emulation.ScreenArea` later on with `screenArea: emulation.ScreenArea / {type: "something", total: emulation.ScreenArea, available: emulation.ScreenArea} / null,`

index.bs
sadym-chromium: ok, not a blocker

index.bs
```diff
@@ -6445,6 +6458,118 @@ The [=remote end steps=] with |command parameters| and |session| are:
...
+   <dd>
+      <pre class="cddl" data-cddl-module="local-cddl">
+      emulation.SetScreenSettingsOverrideResult = EmptyResult
+      </pre>
+   </dd>
+</dl>
+
+<div algorithm>
+
+The <dfn export>WebDriver BiDi emulated screen area</dfn> steps given [=/navigable=] |navigable|:
```
sadym-chromium: I'd propose to split this into `total` and `available`, even though we are going to return the same values. https://github.com/w3c/csswg-drafts/pull/13091/files#r2534196418

index.bs
```diff
@@ -3210,6 +3211,18 @@ A [=remote end=] has a <dfn>viewport overrides map</dfn> which is a weak map bet
...
 [=navigables=] or [=user contexts=] and string.
 
+A <dfn>screen settings</dfn> is a [=struct=] with an [=struct/item=] named
+<dfn attribute for="screen settings">height</dfn> which is an integer,
+an [=struct/item=] named <dfn attribute for="screen settings">width</dfn> which is an integer,
+an [=struct/item=] named <dfn attribute for="screen settings">x</dfn> which is an integer,
+an [=struct/item=] named <dfn attribute for="screen settings">y</dfn> which is an integer.
+
+A [=remote end=] has a <dfn>screen settings overrides</dfn> which is [=struct=]
+an [=struct/item=] named <dfn for="screen settings overrides">user context screen settings</dfn>,
```
```suggestion
A [=remote end=] has a <dfn>screen settings overrides</dfn> which is a [=struct=] with
an [=struct/item=] named <dfn for="screen settings overrides">user context screen settings</dfn>,
```

juliandescottes (approved):
LGTM, thanks Sasha.

PR #1043, opened 2025-12-10
Bump ubuntu to `24.04`

sadym-chromium:
Python 3.12 is [required](https://github.com/w3c/webdriver-bidi/actions/runs/20072014806/job/57576471890) for bikeshed. Ubunto 24.04 is LTS and has python 3.12 by default.

### DIFF
--- a/.github/workflows/auto-publish.yml
+++ b/.github/workflows/auto-publish.yml
@@ -8,6 +8,9 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
     - uses: actions/checkout@v5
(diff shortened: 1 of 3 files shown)


whimboo:
@sadym-chromium any idea why the deploy task is still failing even though you added 3.12 there as well?

jgraham (approved):

sadym-chromium:
> @sadym-chromium any idea why the deploy task is still failing even though you added 3.12 there as well?

I assume it's because of the python used via `pip`. I bumped ubuntu to 24.04, which has python 3.12.

PR #1046, opened 2025-12-13
feat: add userContext to protocol events

### LINKED ISSUE #1032: All events are scoped by `subscriptionID`
Usually we enable events via `session.subscribe` command with optional `contexts/userContexts`. On client side we are trying to associate the handler (method) which should be invoked as soon as client retrieved event from remote end. It becomes hard on client side to differentiate which event is directed whom.

It would be nice if `event params` contains `subscription[]`, which indicates for whom this event should be dispatched. Like network interception: we see `intercepters[]` property.

jimevans:
Since the spec added the ability to subscribe to events using user context IDs, it becomes more difficult to link a given event handler with the user context its event is initiated from. This change makes the event parameters more ergonomic and able to make that link easier to process.

Fixes #1032.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/1046.html" title="Last updated on Feb 11, 2026, 5:20 PM UTC (bbad9ff)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/1046/78fc518...bbad9ff.html" title="Last updated on Feb 11, 2026, 5:20 PM UTC (bbad9ff)">Diff</a>


sadym-chromium (approved):

lutien:
@jimevans, I think it looks good. But I wanted to check one thing before I approve. I've noticed that `input.fileDialogOpened` was not updated. Was it intentional, or did you just miss it?

jimevans:
@lutien It was purely an oversight. I've fixed it in a subsequent commit. Thanks for catching it.

lutien (approved):
Looks good, thanks!

whimboo:
I’m wondering what the best way is to ship this to users. With `userContext` now being required, this becomes a breaking change, especially since browsers do not yet include it in the event payload. Clients would therefore need to be tolerant for the time being. The same concern applies to tests, which we cannot add immediately without causing a significant amount of breakage.

Do you consider this addition critical enough to justify that disruption, or should we take a more gradual approach? Or am I being overly cautious about landing this change?

jimevans:
@whimboo Since a client already has to be able to handle additional values in command responses and events (both are marked as `Extensible`), existing clients shouldn't see a negative impact. Where things fall down is that a client updates to be strictly compliant with the spec, and the browser implementation hasn't caught up yet. This doesn't seem to be a huge burden on clients, unless they're attempting to force strict spec compliance. As the author of such a client library, I have a strategy for handling this scenario, and I would expect other client authors to do the same.

From the WPT angle, it seems like this is exactly the kind of thing that those tests are intended to enforce, that implementations are compliant with the spec, so I'm fine with WPT test failures (I already have a [PR for those](https://github.com/web-platform-tests/wpt/pull/56744)), especially if they act as a forcing function for implementations to update themselves. But these are just my opinions, and they're not strongly held.

whimboo:
@jimevans in this case we extend the params of the individual events and for those I cannot see that `Extensible` is set. Or do I miss something? Here an example for navigation related events:

https://w3c.github.io/webdriver-bidi/#cddl-type-browsingcontextnavigationinfo

sadym-chromium:
@whimboo if your concern is that with the merge all the implementations become invalid, we can mark CDDL property as an optional for a month, which would give us time to implement without breaking the protocol rolls, and make the field required afterwards. I don't think its that complicated and it should work. WDYT?

whimboo:
> @whimboo if your concern is that with the merge all the implementations become invalid, we can mark CDDL property as an optional for a month, which would give us time to implement without breaking the protocol rolls, and make the field required afterwards. I don't think its that complicated and it should work. WDYT?

I think that would be a compromise, yes. Then browsers supporting events need to be updated first.

CC'ing @lauromoura regarding the Webkit implementation that we don't want to break as well.

lauromoura:
> > @whimboo if your concern is that with the merge all the implementations become invalid, we can mark CDDL property as an optional for a month, which would give us time to implement without breaking the protocol rolls, and make the field required afterwards. I don't think its that complicated and it should work. WDYT?
> 
> I think that would be a compromise, yes. Then browsers supporting events need to be updated first.
> 
> CC'ing @lauromoura regarding the Webkit implementation that we don't want to break as well.

While we still have some open issues regarding proper user context support, the changes related to this PR should be fine. Thanks for the heads-up.

whimboo:
Thanks @lauromoura. In that case lets go ahead and add the `userContexts` as optional for now. @jimevans mind updating the PR accordingly? It would be good to get a new issue filed as well which is about turning them into required fields - something that we can do later this year.

jimevans:
@whimboo I've now made all of the additions of `userContext` properties optional for the moment. I will open an issue to make these mandatory for implementation at a later time.

whimboo (approved):
Thank you @jimevans! Would you mind updating the existing wpt tests to check for a proper user context if it is returned?

whimboo:
On https://github.com/w3c/webdriver-bidi/issues/1071 the `userContext` argument will be made mandatory.

whimboo:
@jimevans when updating the BiDi support sheet I noticed that we do not have the `userContext` field for the `log.entryAdded` event. I assume it's a miss? Also what about the various network, and the `script.message` and `script.realmDestroyed` (here we have no realmInfo) events?

lutien:
> @jimevans when updating the BiDi support sheet I noticed that we do not have the `userContext` field for the `log.entryAdded` event. I assume it's a miss? Also what about the various network, and the `script.message` and `script.realmDestroyed` (here we have no realmInfo) events?

`log.entryAdded` and `script.message` share https://www.w3.org/TR/webdriver-bidi/#cddl-type-scriptsource, which contains `userContext`. Network events share https://www.w3.org/TR/webdriver-bidi/#cddl-type-networkbaseparameters, which also contain `userContext`.  So the only question is about `script.realmDestroyed` which indeed doesn't contain `userContext`, but also has no `context` field, so I think it's also harder to add.

whimboo:
Ah you are right. I missed to check those sub types. So yes, it looks all good in that sense and ignoring `realmDestroyed` for now might be fine even though realms should be destroyed before the context and as such should have a proper reference to both the context and the user context.

whimboo:
> Thank you @jimevans! Would you mind updating the existing wpt tests to check for a proper user context if it is returned?

There is actually https://github.com/web-platform-tests/wpt/pull/56744 for it.

Issue #772, opened 2024-09-13
Support mobile emulation


whimboo:
Both Puppeteer and Playwright support an `isMobile` boolean argument for the `setViewport` command, which is currently missing in WebDriver BiDi. We should add support for this parameter with the following behavior:

- **Default Behavior**: By default `isMobile` is set to `false`, meaning the device is not treated as a mobile device and touch events are not enabled.
  
- **Mobile Mode**: When `isMobile` is set to `true`, the browser will behave as if it's running on a mobile device. This includes rendering the website in mobile mode by taking the `meta viewport` tag and CSS media queries into account. Additionally, touch events are enabled, allowing mobile interactions like swiping and tapping.

Question: We currently support `touch` events in `input.performActions` by default. How should these behave when mobile mode is disabled? Should we consider disallowing their use in this case?

whimboo:
> Question: We currently support `touch` events in `input.performActions` by default. How should these behave when mobile mode is disabled? Should we consider disallowing their use in this case?

There is actually a `hasTouch` argument as well that we do not cover yet and I filed https://github.com/w3c/webdriver-bidi/issues/773 for.

lutien:
@OrKoN, @sadym-chromium, I'm gathering requirements for this feature, and I'm trying to understand now what is actually implemented behind "mobile" argument of "Emulation.setDeviceMetricsOverride" command. [It said in the CDP documentation](https://chromedevtools.github.io/devtools-protocol/tot/Emulation/#method-setDeviceMetricsOverride):

>  Whether to emulate mobile device. This includes viewport meta tag, overlay scrollbars, text autosizing and more.

I think `viewport meta tag` and `overlay scrollbars` is clear, but could expand on `text autosizing`? And also what hides behind `more` ? 🙂

gsnedders:
> [@OrKoN](https://github.com/OrKoN), [@sadym-chromium](https://github.com/sadym-chromium), I'm gathering requirements for this feature, and I'm trying to understand now what is actually implemented behind "mobile" argument of "Emulation.setDeviceMetricsOverride" command. [It said in the CDP documentation](https://chromedevtools.github.io/devtools-protocol/tot/Emulation/#method-setDeviceMetricsOverride):
> 
> > Whether to emulate mobile device. This includes viewport meta tag, overlay scrollbars, text autosizing and more.
> 
[...]

Issue #692, opened 2024-03-22
Clarify the behavior of "browsing.setViewport" command with scrollbar


lutien:
At the moment, [the spec](https://w3c.github.io/webdriver-bidi/#command-browsingContext-setViewport) is not very clear on what is correct behavior when updating the viewport in case there is scrollbar, if we should update the inner window (which means that the area available for the document will be smaller than the requested) or we update the visual viewport, but then we might have problems if certain elements appear or disappear and influence the presence of scrollbar and change the dimensions of the viewport.
One possible option could be to have a flag to force an overlay scrollbar, the way it works on mobile. In that case, the scrollbar will not take up the space of the viewport and make the behavior more predictable.

Note: at the moment, Chroma and Firefox both set the requested dimensions to an inner window (`window.innerWidth` and `window.innerHeight`), which includes the scrollbar.

whimboo:
CC @OrKoN, @sadym-chromium for feedback.

Note that when eg a new element gets added and one or both scrollbars appear a reflow can happen and affect the rendering of content. This can easily be seen with grid layouts.

OrKoN:
As far as I see `window.innerWidth` and `window.innerHeight` reflect the layout viewport dimensions so it sounds like current implementations are correct and match the spec. And the scrollbar might be inside the layout viewport/the scrollable area, thus, affecting the layout.

jgraham:
I think our conclusion here is that the spec and implementations match, but they don't match what people probably want for emulating mobile, which is to force overlay scrollbars, thereby sidestepping this problem. So I think the solution here is probably to have a field in the setViewport command that forces the viewport to use some kind of zero-width ("overlay") scrollbars (obviously with the possibility to return an error if you don't actually support that).

OrKoN:
@jgraham I would not say it is exclusively for mobile device emulation because it is also useful for media query testing on desktop. In that case, I think it is useful to see if the scrollbar would affect the layout. Especially, given that on some platform overlay scrollbars are on the default (I think Mac? probably also browser specific). I agree that we probably need something to configure the scrollbars but I wonder if it should be a capability, or a separate method, or a flag on setViewport as you suggest. Perhaps `scrollbar-width: none` can be used as a workaround until something like this is supported?

OrKoN:
It looks like the following issue is related: https://github.com/w3c/webdriver-bidi/issues/592

lutien:
[...]

PR #1050, opened 2025-12-18
Extend `browsingContext.setViewport` with `scrollbarType`

sadym-chromium:
# Closed in favor of https://github.com/w3c/webdriver-bidi/pull/1064

Addressing yet another part of https://github.com/w3c/webdriver-bidi/issues/772.

Related CSS PR: https://github.com/w3c/csswg-drafts/pull/13240

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/1050.html" title="Last updated on Feb 19, 2026, 2:46 PM UTC (e081a30)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/1050/d5ddb9e...e081a30.html" title="Last updated on Feb 19, 2026, 2:46 PM UTC (e081a30)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4975,7 +4975,7 @@ To <dfn>set scrollbar type</dfn> given [=/navigable=] |navigable| and scrollbar 
    make the |navigable|'s [=active document=] to use <a>overlay scrollbars</a>.
 
 1. Otherwise, run [=implementation-defined=] steps to make the |navigable|'s [=active document=] to
-   use a UA defined default scrollbar type.
+   use an [=implementation-defined=] default scrollbar type.
 
 </div>
 


index.bs
```diff
@@ -5036,6 +5066,17 @@ The [=remote end steps=] with |command parameters| are:
...
+   1. If |command parameters| [=map/contains=] "<code>scrollbarMode</code>":
+
+      1. Let |scrollbarMode| be the value of the <code>scrollbarMode</code> field of
+         |command parameters|.
+
+      1. If |scrollbarMode| is null, remove |navigable| from [=scrollbar mode overrides=]'s
+         [=scrollbar mode overrides/navigable scrollbar modes=].
+
+      1. Otherwise, [=map/set=] [=scrollbar mode overrides=]'s
+         [=scrollbar mode overrides/navigable scrollbar modes=][|navigable|] to |scrollbarMode|.
```
OrKoN: I think there should be steps to make sure that the override applies immediately?

sadym-chromium: I added a note in the CSS spec, analog in html: https://html.spec.whatwg.org/#dom-navigator-online

OrKoN: I think a getter steps are not enough since we do not know when the CSS spec would invoke them and we probably want the scrollbar to apply by the time the WebDriver BiDi command resolves?

index.bs
```diff
@@ -3200,6 +3200,12 @@ A <dfn>viewport configuration</dfn> is a [=struct=] with:
 * [=struct/Item=] named <dfn for="viewport-configuration">viewport</dfn> which is a [=viewport dimensions=] or null;
 * [=struct/Item=] named <dfn for="viewport-configuration">devicePixelRatio</dfn> which is a float or null.
 
+A [=remote end=] has a <dfn>viewport overrides map</dfn> which is a weak map between [=user context|user contexts=] and [=viewport configuration=].
+
+A [=remote end=] has a <dfn>scrollbar mode overrides</dfn> which is a [=struct=] with:
```
whimboo: Why isn't this part of the viewport configuration as defined above?

index.bs
```diff
@@ -4904,13 +4908,16 @@ The <dfn export for=commands>browsingContext.setViewport</dfn> command modifies
...
+        ? scrollbarMode: browsingContext.ScrollbarMode / null,
         ? userContexts: [+browser.UserContext],
       }
 
       browsingContext.Viewport = {
         width: js-uint,
         height: js-uint,
       }
+
+      browsingContext.ScrollbarMode = "classic" / "overlay"
```
whimboo: What about `standard` instead of `classic` to force the scrollbar to take layout space? I would avoid namings that might collide with the WebDriver classic protocol. Also should we have a `none` display mode which would hide the scrollbar?

```suggestion
      browsingContext.ScrollbarMode = "standard" / "overlay" / "none"
```

index.bs
sadym-chromium: Agreed. Changed to `set scrollbar type` method, which I exposed in CDD spec.

index.bs
sadym-chromium: TL;DR; I moved it there. 

It was required if we use the hook "WebDriver BiDi emulated scrollbar type". However, the hook would not allow for the BiDi command to be finished AFTER the scrollbar type is changed. So I changed approach, so that BiDi invokes CSS's `set scrollbar type`.

index.bs
sadym-chromium: [CSS spec](https://drafts.csswg.org/css-overflow-3/#ref-for-classic-scrollbars) defines only ["classic"](https://drafts.csswg.org/css-overflow-3/#classic-scrollbars) and ["overlay"](https://drafts.csswg.org/css-overflow-3/#overlay-scrollbars) scrollbar types.

sadym-chromium: I don't think the "none" scrollbar is required for mobile emulations, and it requires significant CSS spec changes. I'd propose to keep it out of scope of this PR.

whimboo: Ah I might have looked at the wrong spec document then. I agree that this looks fine then and that we should not include `none` - at least not for now.

index.bs
```diff
@@ -272,6 +272,9 @@ spec: HR-TIME; urlPrefix: https://w3c.github.io/hr-time/
 spec: RFC4648; urlPrefix: https://datatracker.ietf.org/doc/html/rfc4648
   type: dfn
     text: Base64 Encode; url: section-4
+spec: CSS-OVERFLOW-3; urlPrefix: https://drafts.csswg.org/css-overflow-3/
+  type: dfn
+    text: set scrollbar type; url: #set-scrollbar-type
```
whimboo: Did you maybe reference the wrong spec here? I cannot find the section for setting the scrollbar type and the anchor isn't there as well.

lutien:
I recall that there were concerns about issues with setting some scrollbar types on certain platforms (like, e.g., it's fine for desktop but not for mobile). So maybe it would be better to have a separate command so it's easier to indicate if some option is not supported?

index.bs
sadym-chromium: That method I introduced in another pr, and it is blocked on this one

https://github.com/w3c/csswg-drafts/pull/13240

sadym-chromium: That method I introduced in another pr, and it is blocked on this one

https://github.com/w3c/csswg-drafts/pull/13240

sadym-chromium:
> I recall that there were concerns about issues with setting some scrollbar types on certain platforms (like, e.g., it's fine for desktop but not for mobile). So maybe it would be better to have a separate command so it's easier to indicate if some option is not supported?

The method allows for setting only a single parameter "scrollbarType", so the client can send 2 commands if they want to.

index.bs
whimboo: Ah, I missed that. Thanks. But shouldn't we also get a review for that PR before landing our own changes so that we know that this logic is getting approved in the CSS spec?

whimboo (commented):
Looks fine to me, but can you please check if this to introduce override will work across all platforms? Like setting `classic` for mobile might not work, or? There is a still [open action item for myself](https://github.com/w3c/webdriver-bidi/issues/692#issuecomment-2880828479) as I just noticed, which I haven't had the time for to check.

sadym-chromium:
> I don't quite see why the change to set scrollbars needs to be in CSS Overflow, it's extremely webdriver-specific. I suggest inlining that here instead and just linking to the definitions of classic and overlay scrollbars.

Moved to the WebDriver BiDi spec.

> I also think that helps with Henrik's feedback; because we know we have a catch all "if you aren't going to be able to do this operation abort" at the start of the setViewport steps it's easier to justify the assumption that the algorithm can be treated as infallible. However I'd still suggest it's better to make it fallible and just know that when we call it _from that command_ it's not going to fail.

I mentioned it in the first step of the "browsingContext.setViewport" command:
> If the implementation is unable to adjust the [layout viewport](https://drafts.csswg.org/cssom-view/#layout-viewport) parameters or **scrollbar type** with the given command parameters for any reason, return [error](https://w3c.github.io/webdriver/#errors) with [error code](https://w3c.github.io/webdriver/#dfn-error-code) [unsupported operation](https://w3c.github.io/webdriver/#dfn-unsupported-operation).

index.bs
```diff
@@ -4955,10 +4963,25 @@ To <dfn>set device pixel ratio override</dfn> given |navigable| and |device pixe
 
 </div>
 
+<div algorithm>
+To <dfn>set scrollbar type</dfn> for [=/navigable=] |navigable| to string or null |scrollbar type|:
```
whimboo: Lets make the wording identically to the other algorithms in use:

```suggestion
To <dfn>set scrollbar type</dfn> given [=/navigable=] |navigable| and scrollbar type  |scrollbar type|:
```

index.bs
```diff
@@ -4955,10 +4963,25 @@ To <dfn>set device pixel ratio override</dfn> given |navigable| and |device pixe
...
+1. Otherwise, if |scrollbar type| is "<code>overlay</code>", run [=implementation-defined=] steps to
+   make the |navigable|'s [=active document=] to use <a>overlay scrollbars</a>.
+
+1. Otherwise, run [=implementation-defined=] steps to make the |navigable|'s [=active document=] to
+   use a UA defined default scrollbar type.
+
+</div>
 
 <div algorithm>
 To <dfn>set viewport</dfn> given given [=/navigable=] |navigable| and
```
whimboo: While you are making changes around these lines you may wanna fix this as well:

```suggestion
To <dfn>set viewport</dfn> given [=/navigable=] |navigable| and
```

sadym-chromium:
@jgraham WDYT?

sadym-chromium:
Deprecated by https://github.com/w3c/webdriver-bidi/pull/1064

PR #1061, opened 2026-01-15
Add io module for streams

jgraham:
This adds the initial primitives for working with readable streams, although they are not yet connected up to any other parts of the specification.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/1061.html" title="Last updated on Apr 16, 2026, 11:09 AM UTC (022d91e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/1061/0e36053...022d91e.html" title="Last updated on Apr 16, 2026, 11:09 AM UTC (022d91e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -586,6 +586,7 @@ CommandData = (
   BrowsingContextCommand //
   EmulationCommand //
   InputCommand //
+  IoCommand //
   NetworkCommand //
   ScriptCommand //
   SessionCommand //
@@ -628,6 +629,7 @@ ResultData = (
   BrowsingContextResult /
   EmulationResult /
   InputResult /
+  IoResult /
   NetworkResult /
   ScriptResult /
   SessionResult /
@@ -802,11 +804,17 @@ with the following additional codes:
   <dt><dfn for=errors export>no such storage partition</dfn>
   <dd>Tried to access data in a non-existent storage partition.
 
+  <dt><dfn for=errors export>no such stream</dfn>
+  <dd>Tried to access an unknown [=Stream=].
+
   <dt><dfn for=errors export>no such user context</dfn>
   <dd>Tried to reference an unknown [=user context=].
 
   <dt><dfn for=errors export>no such web extension</dfn>
   <dd>Tried to reference an unknown web extension.
+
+  <dt><dfn for=errors export>stream read error</dfn>
+  <dd>Unable to read from a [=Stream=].
 
   <dt><dfn for=errors export>unable to close browser</dfn>
   <dd>Tried to close the browser, but failed to do so.
@@ -842,10 +850,12 @@ ErrorCode = "invalid argument" /
             "no such node" /
             "no such request" /
             "no such script" /
+            "no such stream" /
             "no such storage partition" /
             "no such user context" /
             "no such web extension" /
             "session not created" /
+            "stream read error" /
             "unable to capture screen" /
             "unable to close browser" /
             "unable to set cookie" /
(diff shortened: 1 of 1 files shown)


jgraham:
@MattiasBuelens are you the right person to check if this way of integrating with the streams spec is reasonable?

MattiasBuelens:
@jgraham I'll have a look. 🙂

index.bs
```diff
@@ -7185,6 +7195,212 @@ is legacy, user agent might run [=implementation-defined=] steps to respect the
...
+The [=remote end steps=] given |session| and |command parameters| are:
+
+1. Let |stream| be the result of trying to [=get a readable stream=] with
+   |session| and |command parameters|["<code>stream</code>"].
+
+1. If |command parameters| [=map/contains=] "<code>reason</code>, let |reason|
+   be |command parameters|["<code>reason</code>"]. Otherwise let |reason| be
+   undefined.
+
+1. Run [$ReadableStreamCancel$]\(|stream|, |reason|).
```
MattiasBuelens: Use [cancel a readable stream](https://streams.spec.whatwg.org/#readablestream-cancel) instead.

```suggestion
1. [=ReadableStream/Cancel=] |stream| with |reason|.
```

index.bs
```diff
@@ -7185,6 +7195,212 @@ is legacy, user agent might run [=implementation-defined=] steps to respect the
...
+1. If |stream|.\[[state]] is "<code>closed</code>", perform |read request|'s close steps.
+
+1. Otherwise, if |stream|.\[[state]] is "<code>errored</code>", perform |read
+   request|'s error steps given |stream|.\[[storedError]].
+
+1. Otherwise,
+
+    1. Assert: stream.\[[state]] is "readable".
+
+    1. Perform ! |stream|.\[[controller]].\[[PullSteps]](|read request|).
```
MattiasBuelens: This is touching the stream's internals, which is not recommended. You should use [read a chunk from a ReadableStreamDefaultReader reader](https://streams.spec.whatwg.org/#readablestreamdefaultreader-read-a-chunk) instead.

```suggestion
1. Let |reader| be the result of [=ReadableStream/getting a reader=] for |stream|.

1. [=ReadableStreamDefaultReader/Read a chunk=] from |reader| given |read request|.
```

index.bs
```diff
@@ -7185,6 +7195,212 @@ is legacy, user agent might run [=implementation-defined=] steps to respect the
...
+
+1. Assert: |result| is not null.
```
MattiasBuelens, line 7703: This is not correct if `[[PullSteps]]` does not immediately [enqueue](https://streams.spec.whatwg.org/#readablestream-enqueue) a chunk in response to the read request.

Instead, you should probably create a promise and have your read request callbacks resolve/reject it. Have a look at the spec text for [ReadableStreamDefaultReader.read()](https://streams.spec.whatwg.org/#default-reader-read) for some inspiration.

MattiasBuelens (changes requested):
You should use the algorithms defined in [section 9 of the Streams standard](https://streams.spec.whatwg.org/#other-specs) in order to safely interact with streams. Using abstract ops from other parts of the Streams spec is not recommended.

jgraham:
I've updated this with the concept of a specification realm in which we can run scripts without it being visible to content.

I'm not entirely sure this is sound; a lot of the machinery wasn't really designed with this kind of use case in mind.

index.bs
```diff
@@ -7392,6 +7494,218 @@ is legacy, user agent might run [=implementation-defined=] steps to respect the
...
+The [=remote end steps=] given |session| and |command parameters| are:
+
+1. Let |stream| be the result of [=trying=] to [=get a readable stream=] with
+   |session| and |command parameters|["<code>stream</code>"].
+
+1. Let |result| be null.
+
+1. [=Run in the specification realm=] with the following steps:
+
+  1. Let |reader| be ? [$AcquireReadableStreamDefaultReader$](|stream|).
```
MattiasBuelens, line 7661: This [locks the stream](https://streams.spec.whatwg.org/#locking) to this reader. You must [release the lock](https://streams.spec.whatwg.org/#release-a-lock) afterwards, otherwise the stream will be forever locked to a reader that is no longer accessible.

Also, as I've said before, you should really use [the algorithms in this section](https://streams.spec.whatwg.org/#other-specs) instead, such as:
* [get a reader](https://streams.spec.whatwg.org/#readablestream-get-a-reader)
* [read a chunk](https://streams.spec.whatwg.org/#readablestreamdefaultreader-read-a-chunk)
* [release a reader](https://streams.spec.whatwg.org/#readablestreamdefaultreader-release)
* [cancel a stream](https://streams.spec.whatwg.org/#readablestream-cancel)

index.bs
```diff
@@ -7392,6 +7494,218 @@ is legacy, user agent might run [=implementation-defined=] steps to respect the
...
+           <code>done</code> field set to false.
+
+        1. Set |result| to [=success=] with data |response|.
+
+       <dt>When the promise is rejected with <var ignore>reason</var>:
+       <dd>
+         1. Set |result| to [=error=] with [=error code=] [=stream read error=].
+     </dl>
+
+  1. If |react result| is an [=abrupt completion=], return [=error=] with [=error
```
MattiasBuelens, line 7700: *react result* is the result of [react to promise](https://webidl.spec.whatwg.org/#dfn-perform-steps-once-promise-is-settled), which returns a `Promise`. So you won't get an abrupt completion here.

index.bs
MattiasBuelens: This is still not correct, even with the latest changes. The fulfill and reject steps of "react to promise" run *asynchronously*. So this assert step happens *before* any reaction steps have actually ran.

I think what you really want to do is define an *event* for when a read request has been fulfilled or rejected.

MattiasBuelens: Okay, apparently you *can* await promises in this spec? For example, [`script.evaluate`](https://www.w3.org/TR/webdriver-bidi/#command-script-evaluate) has an `awaitPromise` option which uses [`Await()`](https://tc39.es/ecma262/#await).

That might be a better option then.

jgraham: That works because with `evaluate` we're already running script in a content realm. We _could_ try to model our use of streams as if we were running scripts in some relevant realm, but that risks side effects that are visible to content (e.g. events dispatched to content if there are unhandled promises or errors).

PR #1063, opened 2026-01-16
Always restore the client window if it is currently minimized or maximized before setting a different window state

### LINKED ISSUE #1035: clarify minimized-to-fullscreen transition behavior
In [`set client window state`](https://w3c.github.io/webdriver-bidi/#set-the-client-window-state) the following text describes how to fullscreen a document:

> If document’s [visibility state](https://html.spec.whatwg.org/multipage/interaction.html#visibility-state) is "visible", [fullscreen an element](https://fullscreen.spec.whatwg.org/#fullscreen-an-element) with document’s [document element](https://dom.spec.whatwg.org/#ref-for-dom-document-documentelement).

However, this seems to imply that attempting to fullscreen a minimized document should not do anything. It might be worth clarifying this behavior in the specification, though I'm not sure what the desired behavior should be.

whimboo:
Fixed #1035.

This was most likely an oversight. We need to restore the window first if it is minimized before actually switching to fullscreen. This is [in-par with WebDriver classic](https://w3c.github.io/webdriver/#fullscreen-window).

***
<a href="https://pr-preview.s3.amazonaws.com/whimboo/webdriver-bidi/pull/1063.html" title="Last updated on Mar 5, 2026, 7:08 PM UTC (4f8bc14)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/1063/9e65594...whimboo:4f8bc14.html" title="Last updated on Mar 5, 2026, 7:08 PM UTC (4f8bc14)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2489,7 +2489,7 @@ To <dfn>set the client window state</dfn> given |window| and |state|:
 
        Note: This is a no-op for documents in window that are not fullscreen.
 
-1. If |current state| is "<code>maximized</code>", or "<code>minimized</code>":
+1. If |current state| is "<code>maximized</code>" or "<code>minimized</code>":
 
   1. [=Restore the client window=] |window|.
 


index.bs
```diff
@@ -2494,6 +2494,8 @@ To <dfn>set the client window state</dfn> given |window| and |state|:
    <dl>
     <dt>"<code>fullscreen</code>"
     <dd>
+      1. [=Restore the client window=] |window|.
```
whimboo: Maybe we need this every time when we transition out of minimized state. I see issues with minimize -> maximize tests for Firefox as well.

lutien: I don't think we need it for all the transitions, but we probably need it for the maximize window, since i[t's the same in Classic](https://w3c.github.io/webdriver/#maximize-window).

whimboo: Yes, that is the other state change, but given that for normal state we also have to restore the window, we need it for all changes. I'll move the call to `restore the client window` out of the switch and have it similar to fullscreen call that API when needed.

index.bs
```diff
@@ -2614,6 +2489,10 @@ To <dfn>set the client window state</dfn> given |window| and |state|:
 
        Note: This is a no-op for documents in window that are not fullscreen.
 
+1. If |current state| is "<code>maximized</code>", or "<code>minimized</code>":
```
lutien:
```suggestion
1. If |current state| is "<code>maximized</code>" or "<code>minimized</code>":
```

lutien (approved):

PR #1064, opened 2026-01-20
`emulation.setScrollbarTypeOverride`

sadym-chromium:
Addressing yet another part of https://github.com/w3c/webdriver-bidi/issues/772.

As discussed in https://github.com/w3c/webdriver-bidi/pull/1050#issuecomment-3756320242, I moved the emulation to a dedicated command.

As discussed ion https://github.com/w3c/webdriver-bidi/issues/772#issuecomment-3756268948, we are going to have quite a lot of emulation commands with the similar syntax and scopes. So I introduced a concept of `WebDriver Configuration` which is supposed to be a placeholders for all the configs. Moving existing configs there is out of scope. It is supposed to reduce both specification and implementation effort.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/1064.html" title="Last updated on Feb 19, 2026, 12:38 PM UTC (e556e6b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/1064/fdf90f8...e556e6b.html" title="Last updated on Feb 19, 2026, 12:38 PM UTC (e556e6b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -455,10 +455,13 @@ Returns [=WebDriver configuration/unset=] if configuration is not set.
 To <dfn>get WebDriver configuration value</dfn> of [=WebDriver configuration=] |configuration| for
 [=/navigable=] |navigable|:
 
-1. If |configuration|'s [=WebDriver configuration/navigables=] [=map/contains=] |navigable|:
+1. Let |top-level traversable| be |navigable|'s [=navigable/top-level traversable=].
+
+1. If |configuration|'s [=WebDriver configuration/navigables=] [=map/contains=]
+   |top-level traversable|:
 
    1. Let |navigable configuration value| be |configuration|'s
-      [=WebDriver configuration/navigables=][|navigable|].
+      [=WebDriver configuration/navigables=][|top-level traversable|].
 
    1. If |navigable configuration value| is not [=WebDriver configuration/unset=], return
       |navigable configuration value|.


sadym-chromium:
@whimboo WDYT?

index.bs
```diff
@@ -5987,12 +5992,125 @@ EmulationResult = (
...
   emulation.SetTimezoneOverrideResult /
   emulation.SetTouchOverrideResult /
   emulation.SetUserAgentOverrideResult
 )
 </pre>
 
+A <dfn export>stored config</dfn> is a [=struct=] with:
+* [=struct/item=] named <dfn for="stored config">scrollbar type</dfn> which is a string or null initially null.
+
+A [=BiDi session=] has an <dfn for=session>session stored configuration</dfn> which is a [=struct=] with:
```
whimboo: I didn't review that many emulation related changes yet so what is the difference to have the emulation data stored on the session vs the remote end? We seem to vary between those two locations.

The problem I see here is that different sessions cannot set different scrollbar types. So maybe this should be session independent?

index.bs
```diff
@@ -5987,12 +5992,125 @@ EmulationResult = (
...
   emulation.SetScreenOrientationOverrideResult /
   emulation.SetScriptingEnabledResult /
+  emulation.SetScrollbarTypeOverrideResult /
   emulation.SetTimezoneOverrideResult /
   emulation.SetTouchOverrideResult /
   emulation.SetUserAgentOverrideResult
 )
 </pre>
 
+A <dfn export>stored config</dfn> is a [=struct=] with:
```
whimboo: Is this struct scrollbar related? If yes we should reflect that in the name.

index.bs
sadym-chromium: The core difference is in the behavior when the session is ended. If we want the default behavior to be restored after the session initiated emulation to be restored (e.g. extra headers), it should be associated with the session. If not (window position), it should be associated with the remote end. I assume all the emulations are expected to be removed after the session is gone. However IDK if we do this in practice. Probably it would required another look after someone implemented multiple sessions.

index.bs
sadym-chromium: No. As mentioned in the PR description, we are going to have lots of emulation commands, and this structure is expected to be a placeholder for all such configurations (not necessarily emulation related though). This would save us from copy-pasting the same logic over and over again. We can extend it further if later on. I tried to keep it readable, so decided not to pass "algorithm" etc.

index.bs
whimboo: If we have multiple sections we would have a conflict because individual sessions cannot set different scrollbar types. Only one can actually apply at the same time. I feel this is close to the screen orientation override which has its struct on the remote end and not on the session. 

Overall I feel we need a notice that explains when an override is on the remote end and when on the session. Maybe this is something we should add first to get clarity? CC @jgraham

sadym-chromium: > Maybe this is something we should add first to get clarity?

I don't think this should stop all the emulation commands, as the change in the spec would be trivial, and we don't have implementations supporting more then 1 session at the moment

sadym-chromium: @whimboo friendly ping

OrKoN (approved):

index.bs
jgraham: I agree we shouldn't block on this, but I also think there's a misunderstanding about what happens when the session ends. There isn't any implicit cleanup of state set on the session; if you want state to be cleaned up you need to specify explicit cleanup steps in https://w3c.github.io/webdriver-bidi/#cleanup-the-session. In general I don't think the protocol should enforce cleanup of browser state (as opposed to protocol state) because it prevents certain use cases (e.g. consider Firefox devtools which allows setting up device emulation state that persists after devtools itself is closed).

So, in general:
* Only state specific to the protocol should be stored on the session.
* Cleanup of state needs to be specified explicitly where required.
* State that affects the browser itself rather than the protocol should be stored on the remote end, and should generally persist beyond the end of the session, so that applications can clean it up, or not, according to their requirements.

Note that this is a bit different to WebDriver Classic, which baked in the assumption of only one session and frequently made the implicit assumption that the browser lifetime was equal to the session lifetime (because that's how Selenium works).

sadym-chromium: I updated the PR and now store the config in the remote end.

sadym-chromium: @whimboo @jgraham please assert it is what you meant

index.bs
```diff
@@ -5007,6 +5011,8 @@ TODO: Move it as a hook in the html spec instead.
       1. [=Set device pixel ratio override=] with |navigable| and [=viewport overrides map=][|user context|]'s
          [=viewport-configuration/devicePixelRatio=].
 
+1. [=Update scrollbar type override=].
```
whimboo: It's a different command so I assume that this should get its own algorithm similar to viewport and dpr.

index.bs
```diff
@@ -5987,12 +5994,127 @@ EmulationResult = (
...
   emulation.SetTimezoneOverrideResult /
   emulation.SetTouchOverrideResult /
   emulation.SetUserAgentOverrideResult
 )
 </pre>
 
+A <dfn export>stored remote end config</dfn> is a [=struct=] with:
+* [=struct/item=] named <dfn for="stored remote end config">scrollbar type</dfn> which is a string or null initially null.
+
+A [=remote end=] has an <dfn for="remote end">remote end stored configuration</dfn> which is a [=struct=] with:
```
```suggestion
A [=remote end=] has a <dfn for="remote end">remote end configuration</dfn> which is a [=struct=] with:
```

And for other references if needed.

index.bs
```diff
@@ -5987,12 +5994,127 @@ EmulationResult = (
...
   emulation.SetScreenOrientationOverrideResult /
   emulation.SetScriptingEnabledResult /
+  emulation.SetScrollbarTypeOverrideResult /
   emulation.SetTimezoneOverrideResult /
   emulation.SetTouchOverrideResult /
   emulation.SetUserAgentOverrideResult
 )
 </pre>
 
+A <dfn export>stored remote end config</dfn> is a [=struct=] with:
```
```suggestion
A <dfn export>stored configuration</dfn> is a [=struct=] with:
```

or `configuration object`? I would remove `remote end` here in case we will re-use it for session based configuration values.

sadym-chromium: We will need a separate sturture for configurations per session

index.bs
sadym-chromium: Not really. Here is declares, that 
> After creating a document in a new `navigable` and before the "run WebDriver BiDi preload scripts" algorithm is invoked,

the remote end have to 

> Update scrollbar type override

Basically this means that the new navigables should be configured based on the global or user-context saved configuration.

sadym-chromium:
@jgraham could you please take a look? This is a blocker for other emulations, as I want to avoid declaring the config storage over and over again.

sadym-chromium:
@jgraham @whimboo friendly ping

index.bs
```diff
@@ -5987,12 +5994,181 @@ EmulationResult = (
...
   emulation.SetScreenOrientationOverrideResult /
   emulation.SetScriptingEnabledResult /
+  emulation.SetScrollbarTypeOverrideResult /
   emulation.SetTimezoneOverrideResult /
   emulation.SetTouchOverrideResult /
   emulation.SetUserAgentOverrideResult
 )
 </pre>
 
+Note: [=stored remote end configuration=] is an umbrella structure that holds all configurations associated with the
```
jgraham: This clearly shouldn't be defined under emulation, but perhaps in the infrastructure section.

index.bs
```diff
@@ -5987,12 +5994,181 @@ EmulationResult = (
...
+[=remote end=]. Unlike session configurations, these configurations are not reset after the session is closed, and are
```
jgraham: Session configurations are also not (automatically) reset when the browser is closed. It's just that session configuration should only be for properties of the session itself, so they should become irrelevant once the session has ended.

index.bs
```diff
@@ -5987,12 +5994,181 @@ EmulationResult = (
...
+* [=struct/item=] named <dfn for="remote end configuration">scrollbar type</dfn> which is a string or null initially null.
+
+A [=remote end=] has a <dfn for="remote end">stored remote end configuration</dfn> which is a [=struct=] with:
+* [=struct/item=] named <dfn for="stored remote end configuration">global</dfn> which is a [=remote end configuration=], initially an empty [=remote end configuration=];
+* [=struct/item=] named <dfn for="stored remote end configuration">user contexts</dfn> which is a weak map between [=user context|user contexts=] and [=remote end configuration=], initially empty;
+* [=struct/item=] named <dfn for="stored remote end configuration">navigables</dfn> which is a weak map between [=/navigables=] and [=remote end configuration=], initially empty.
+
+<div algorithm>
+
+Note: this algorithm allows for accessing the stored configuration for a given navigable by checking the remote-end's
```
jgraham: This seems like the key property of this configuration, as it answers the question "why would I use this shared configuration object rather than just putting data directly on the remote end?".

index.bs
```diff
@@ -5987,12 +5994,181 @@ EmulationResult = (
...
   emulation.SetUserAgentOverrideResult
 )
 </pre>
 
+Note: [=stored remote end configuration=] is an umbrella structure that holds all configurations associated with the
+[=remote end=]. Unlike session configurations, these configurations are not reset after the session is closed, and are
+shared between sessions.
+
+A <dfn export>remote end configuration</dfn> is a [=struct=] with:
+* [=struct/item=] named <dfn for="remote end configuration">scrollbar type</dfn> which is a string or null initially null.
```
jgraham: So this requires us to centralise the definition of all configuration in a specific place. That feels pretty anti-modular; for example it's impossible for extension specs to use this mechanism.

index.bs
```diff
@@ -5987,12 +5994,181 @@ EmulationResult = (
...
+
+A [=remote end=] has a <dfn for="remote end">stored remote end configuration</dfn> which is a [=struct=] with:
```
jgraham: This feels like the key piece of infrastructure here. It's effectively a new generic container that knows how to store data of some type in a way that it's associated with the session, a user context or a navigable. 

I think instead of having the `remote end configuration` struct that stores all of these, we should just define this type and the associated get/set algorithms. Then when we want to define a property that has these sematics we just say:

`A [=remote end=] has a [=scrollbar type=] which is a [=stored remote end configuration=]`.

(I also don't love the name "stored remote end configuration". I can't imagine what an "unstored" configuration would be, so removing the word "stored" doesn't really change the meaning. I think just "remote end configuration" is a better, but still doesn't quite convey the semantics that it encapsulates configuration that might happen at one of several contexts. It also occurs to me that sometimes this kind of configuration makes sense on a session e.g. it's very close to how we store event subscriptions. So maybe even "remote end" is unnecessary and we should be calling this something like a "WebDriver configuration").

index.bs
```diff
@@ -5987,12 +5994,181 @@ EmulationResult = (
...
+   1. If [=remote end=]'s [=stored remote end configuration=]'s [=stored remote end configuration/user contexts=]
+      [=map/contains=] |user context|:
+
+      1. Set [=remote end=]'s [=stored remote end configuration=]'s
+         [=stored remote end configuration/user contexts=][|user context|]'s |item| to null.
+
+      1. If [=remote end=]'s [=stored remote end configuration=]'s
+         [=stored remote end configuration/user contexts=][|user context|] contains only nulls, remove
+         |user context| from [=remote end=]'s [=stored remote end configuration=]'s
+         [=stored remote end configuration/user contexts=].
```
jgraham: What semantics are you aiming for with the store-null-then-remove setup? It seems to me that in this setup `null` has to have the exact same meaning as "missing" because the case where you set key1, unset key1, and do the cleanup is semantically identical (for key1) to the case where you set key1, set key2, and set key1 to null, not doing the cleanup.

Rather than using null as a sometimes-tombstone value, I'd much rather have an explicit "unset" operation, which removes the value either from the entire configuration or from the given tier of configuration. Further cleanup would then be optional (since it's not observable in the spec). It would also keep `null` available e.g. as a specific "reset" marker if we want to support the case where a user context has a given value but a browsing context has the default value, for a specific kind of data (of we can support that anyway, but need to invent a specific "reset" value to store).

index.bs
sadym-chromium: > why would I use this shared configuration object rather than just putting data directly on the remote end?

Putting configuration to the remote end requires to break the **DRY** principle, and describing the same algorithms over and over again. We already do it in several places, and it causes some inconsistency e.g. in the way we process global configurations: in some commands we don't allow them, in some we do. And it is simply due to the fact we copy-paste the configuration storing logic for each config.

index.bs
sadym-chromium: Agree. Replaced with more flexible mechanism: 
> The \<dfn>scrollbar type override configuration\</dfn> is [=WebDriver configuration=] with
[=WebDriver configuration/type=] string.

index.bs
sadym-chromium: Done and renamed to `WebDriver configuration`. This indeed looks way better.

index.bs
sadym-chromium: Resolving for now

index.bs
```diff
@@ -426,6 +430,157 @@ To <dfn export>resume</dfn> given |name|, |id| and |parameters|:
 
 </div>
 
+The <dfn>WebDriver configuration</dfn> is a structure providing a generic way to store
+configurations per user context, per navigable or globally.
+
+The <dfn>WebDriver configuration unset value</dfn> is a specific value indicating the value for the
```
jgraham: I think it would be easier if we just referred to this as "unset" like `<dfn for="Webdriver configuration">Unset</dfn> is a value indicating that a specific configuration has not be set`.

index.bs
```diff
@@ -426,6 +430,157 @@ To <dfn export>resume</dfn> given |name|, |id| and |parameters|:
...
+* [=struct/item=] named <dfn for="WebDriver configuration struct">global</dfn> which is a [=WebDriver configuration/value=], initially [=WebDriver configuration unset value=];
+* [=struct/item=] named <dfn for="WebDriver configuration struct">user contexts</dfn> which is a weak map between [=user context|user contexts=] and [=WebDriver configuration/value=], initially empty;
+* [=struct/item=] named <dfn for="WebDriver configuration struct">navigables</dfn> which is a weak map between [=/navigables=] and [=WebDriver configuration/value=], initially empty.
+
+Each [=WebDriver configuration=] has an associated <dfn for="WebDriver configuration">values</dfn>
+which is a [=WebDriver configuration struct=].
+
+<div algorithm>
+
+Note: this algorithm allows for accessing the [=WebDriver configuration=] for a given [=/navigable=]
```
```suggestion
Note: this algorithm allows accessing the [=WebDriver configuration=] for a given [=/navigable=]
```

index.bs
```diff
@@ -426,6 +430,157 @@ To <dfn export>resume</dfn> given |name|, |id| and |parameters|:
...
+The <dfn>WebDriver configuration</dfn> is a structure providing a generic way to store
+configurations per user context, per navigable or globally.
+
+The <dfn>WebDriver configuration unset value</dfn> is a specific value indicating the value for the
+specific configuration is unset.
+
+Each [=WebDriver configuration=] has an associated <dfn for="WebDriver configuration">type</dfn>,
+which defines the possible values for the configuration.
+
+Each [=WebDriver configuration=] has a <dfn for="WebDriver configuration">value</dfn>, which is
```
jgraham: I don't think we need to define this? We have a struct with an associated type like (pseudo-code):
```
Configuration<T> {
  global: Unset | T,
  userContext: WeakMap<UserContext, Unset | T>
  navigable: WeakMap<Navigable, Unset | T>
}
```
I guess the idea here is to define `ConfiguationValue<T> = Unset | T`, but it feels like it ends up being a bit of extra indirection.

index.bs
```diff
@@ -426,6 +430,157 @@ To <dfn export>resume</dfn> given |name|, |id| and |parameters|:
...
+   1. If |user context configuration value| is not [=WebDriver configuration unset value=], return
+      |user context configuration value|.
+
+1. Return |configuration values|' [=WebDriver configuration struct/global=].
+
+</div>
+
+<div algorithm>
+
+Note: this is a generic algorithm for storing [=WebDriver configuration struct/global=]
```
jgraham: Personally I'd have a single algorithm like
```
To <dfn>store WebDriver configuration</dfn> given [=WebDriver configuration=] |configuration|, |target| which is a [=navigable=], a [=user context=] or null, and |value| of type |configuration|'s [=WebDriver Configuration/value=]:

  1. If |target| is null, set |configurations|'s  [=WebDriver configuration struct/global=] to |value|.
  1. If |target| is a [=user context=] …
  1. If |target| is a [=navigable=] …
```

index.bs
```diff
@@ -6945,6 +7084,84 @@ The [=remote end steps=] with |command parameters| are:
...
+[=WebDriver configuration/type=] string.
+
+<div algorithm>
+To <dfn>update scrollbar type override</dfn> for [=/navigable=] |navigable|:
+
+1. Let |scrollbar type override| be the result of [=get WebDriver configuration value=] of
+   [=scrollbar type override configuration=] for |navigable|.
+
+1. Assert that |scrollbar type override| is "<code>classic</code>", "<code>overlay</code>" or
+   [=WebDriver configuration unset value=].
```
```suggestion
1. Assert: |scrollbar type override| is "<code>classic</code>", "<code>overlay</code>" or
   [=WebDriver configuration unset value=].
```

index.bs
```diff
@@ -6945,6 +7084,84 @@ The [=remote end steps=] with |command parameters| are:
...
+
+1. If |scrollbar type override| is "<code>classic</code>", run [=implementation-defined=] steps to
+   make the |navigable|'s [=active document=] to use <a>classic scrollbars</a> and return.
+
+1. if |scrollbar type override| is "<code>overlay</code>", run [=implementation-defined=]
+   steps to make the |navigable|'s [=active document=] to use <a>overlay scrollbars</a> and return.
+
+1. Assert the |scrollbar type override| is [=WebDriver configuration unset value=].
```
```suggestion
1. Assert: |scrollbar type override| is [=WebDriver configuration unset value=].
```

index.bs

index.bs
sadym-chromium: updated the notes

index.bs
sadym-chromium: Added a dedicated value indicating the configuration is not set.

index.bs
sadym-chromium: Agree. Added.

index.bs

index.bs

index.bs
sadym-chromium: I want to define `WebDriver configuration` as `Configuration<T>`. If so:
> The \<dfn>scrollbar type override configuration\</dfn> is [=WebDriver configuration=] with
[=WebDriver configuration/type=] string.

Meaning `scrollbar type override configuration = Configuration<T>`. If so, it is unclear how to refer to `T` or `Unset | T`. I don't think keeping them divided creates too much overhead, but make the spec very clear of what exactly it referenced. At least it helped me a lot in writing this spec.

index.bs

sadym-chromium:
@jgraham thanks for the review! I believe I addressed your concerns. Please take another look.

index.bs
```diff
@@ -426,6 +430,135 @@ To <dfn export>resume</dfn> given |name|, |id| and |parameters|:
...
+1. If |configuration values|' [=WebDriver configuration struct/user contexts=] [=map/contains=]
+   |user context|:
+
+   1. Let |user context configuration value| be |configuration values|'
+      [=WebDriver configuration struct/user contexts=][|user context|].
+
+   1. If |user context configuration value| is not [=WebDriver configuration/unset=], return
+      |user context configuration value|.
+
+1. Return |configuration values|' [=WebDriver configuration struct/global=].
```
juliandescottes: Looking at this in scope of #1068 . 

What does this mean for specs which call bidi hooks to get a configuration value?

Typically for CSP bypass / Scripting enabled, we store a boolean, but if no configuration is set we return a default hardcoded value.
(eg [here](https://github.com/w3c/webdriver-bidi/blob/db627ad6e800e4318574df25166e91759941a28a/index.bs#L6899))

Should we keep a simple hook in the BiDi spec which first calls this algorithm and if the return value is `[=WebDriver configuration/unset=]` returns the default value instead? Or should other specs handle `[=WebDriver configuration/unset=]` directly?

index.bs
```diff
@@ -426,6 +430,135 @@ To <dfn export>resume</dfn> given |name|, |id| and |parameters|:
...
+
+</div>
+
+<div algorithm>
+
+Note: this is a generic algorithm for storing [=WebDriver configuration=] per |target|, which can be
+either [=/navigable=], [=user context=], or store it globally if the |target| is null or emitted.
```
juliandescottes: emitted -> omitted?

index.bs
```diff
@@ -6945,6 +7062,84 @@ The [=remote end steps=] with |command parameters| are:
...
+   <dt>Return Type</dt>
+   <dd>
+      <pre class="cddl" data-cddl-module="local-cddl">
+      emulation.SetScrollbarTypeOverrideResult = EmptyResult
+      </pre>
+   </dd>
+</dl>
+
+The <dfn>scrollbar type override configuration</dfn> is [=WebDriver configuration=] with
+[=WebDriver configuration/type=] "<code>classic</code>" / "<code>overlay</code>".
```
juliandescottes: I think this was discussed with @jgraham , but shouldn't we still say where this configuration is stored (remote end vs session?).

index.bs
sadym-chromium: I'm in favor of returning `null` in case of the override is not set, like in https://w3c.github.io/webdriver-bidi/#webdriver-bidi-emulated-max-touch-points or in https://w3c.github.io/webdriver-bidi/#webdriver-bidi-emulated-user-agent. We can also return [=unset=], but that would require updating all the existing hooks when we will align them under this approach.

index.bs

index.bs
sadym-chromium: In this specification version the configuration is stored in remote end (as the configuration is not tight to sessions).

sadym-chromium: We can extend the definition for configurations scoped to sessions on-demand, but it's out of scope of this PR.

jgraham: I think @juliandescottes is correct and you need to be explicit about this e.g. 

```
A [=remote end=] has a <dfn>scrollbar type override configuration</dfn> which is a [=WebDriver configuration=] with [=WebDriver configuration/type=] "<code>classic</code>" / "<code>overlay</code>".
```

As a side note, I'm not sure there's anything else in spec land to support this use of a choice between literals as an enumeration type? I see that it's pretty close to CDDL and one can work out from context what's intended, but I think really you'd want something in infra to point at.

index.bs
juliandescottes: Alright, so that means we keep individual hooks for each entry point. Do we really need to `export` this algorithm then? Maybe it can stay local to this spec?

index.bs
juliandescottes: Storing on remote end sounds good, but then yes it needs to be spelled out explicitly.

index.bs
sadym-chromium: removed

index.bs
sadym-chromium: ok, I introduced "WebDriver configuration scoped to remote end", and we can introduce "WebDriver configuration scoped to session" later on.

index.bs
```diff
@@ -433,6 +433,11 @@ To <dfn export>resume</dfn> given |name|, |id| and |parameters|:
 The <dfn>WebDriver configuration</dfn> is a structure providing a generic way to store
 configurations per user context, per navigable or globally.
 
+The <dfn>WebDriver configuration scoped to remote end</dfn> is a [=WebDriver configuration=] which
+is scoped to [=remote end=].
```
juliandescottes: I don't fully understand the last update. I thought the only thing we needed was the suggestion from James, the scrollbar configuration should be defined as stored on the remote end and that's it. 

Then the individual algorithms already operate on a specific configuration, they don't need to know if it's stored on the session or the remote end?

sadym-chromium: Added a sentence to `values` definition. WDYT?

index.bs
sadym-chromium: And reverted "WebDriver configuration scoped to remote end". Scoped "WebDriver configuration/values" to "remote end" instead. Also changed type of "scrollbar type override configuration" to string.

index.bs
juliandescottes: Trying to understand why we can't apply the suggestion from James.

If I understand correctly "WebDriver configuration" is some kind of abstract type / class / concept, and not really a struct. And I feel like you don't want the "configuration types" (eg scrollbar configuration) to be owned specifically by the remote end or the session. That's why you are proposing to only link the `values` struct to the remote end, and not the whole configuration?

I think we could just say the whole configuration is owned by the remote end. The configuration might not look like a regular struct / data structure but since it has a 1:1 relationship with its values, I really think they're interchangeable here.

Am I missing something, is there another strong reason which prevents from just doing:

> A [=remote end=] has a <dfn>scrollbar type override configuration</dfn> which is a [=WebDriver configuration=] with [=WebDriver configuration/type=] "<code>classic</code>" / "<code>overlay</code>".

?

sadym-chromium: > If I understand correctly "WebDriver configuration" is some kind of abstract type / class / concept, and not really a struct. And I feel like you don't want the "configuration types" (eg scrollbar configuration) to be owned specifically by the remote end or the session. That's why you are proposing to only link the `values` struct to the remote end, and not the whole configuration?

Eventually we can want to associate WebDriver configuration with sessions, and it would be really complicated with this wording. We need to know the difference between "session" and "remote end"-scoped configurations at the setter and getter algorithms. So I want to keep the "WebDriver configuration" abstract.

sadym-chromium: I believe scoping the `values` to `remote end` and eventually to `session` allows for using the terms in the most clear way.

jgraham: I agree with @juliandescottes  here.

We are defining a generic kind of struct to store data in, and some algorithms that work over that struct. Those structs shouldn't care where they're stored; in particular I can't see any reason that the getter or setter algorithms should care who owns the struct.

Given that, the easiest thing is just that each concrete instance of the configuration struct specifies who owns the data i.e. my earlier suggestion.

index.bs
```diff
@@ -426,6 +430,135 @@ To <dfn export>resume</dfn> given |name|, |id| and |parameters|:
 
 </div>
 
+The <dfn>WebDriver configuration</dfn> is a structure providing a generic way to store
+configurations per user context, per navigable or globally.
+
+<dfn for="WebDriver configuration">Unset</dfn> is a value indicating that a specific configuration
+[=WebDriver configuration/value=] has not be set.
```
```suggestion
[=WebDriver configuration/value=] has not been set.
```

index.bs
sadym-chromium: @juliandescottes @jgraham done. Please take a look.

index.bs
```diff
@@ -426,6 +430,135 @@ To <dfn export>resume</dfn> given |name|, |id| and |parameters|:
...
+Each [=WebDriver configuration=] has a <dfn for="WebDriver configuration">value</dfn>, which is
+either [=WebDriver configuration/unset=] or a value of the [=WebDriver configuration/type=].
+
+The <dfn>WebDriver configuration struct</dfn> is a [=struct=] with:
+* [=struct/item=] named <dfn for="WebDriver configuration struct">global</dfn> which is a [=WebDriver configuration/value=], initially [=WebDriver configuration/unset=];
+* [=struct/item=] named <dfn for="WebDriver configuration struct">user contexts</dfn> which is a weak map between [=user context|user contexts=] and [=WebDriver configuration/value=], initially empty;
+* [=struct/item=] named <dfn for="WebDriver configuration struct">navigables</dfn> which is a weak map between [=/navigables=] and [=WebDriver configuration/value=], initially empty.
+
+Each [=WebDriver configuration=] has an associated <dfn for="WebDriver configuration">values</dfn>
+which is a [=WebDriver configuration struct=].
```
jgraham: This still feels like an unnecessary layer of indirection to me.

I really think we could just say something like
```
A <dfn>WebDriver configuration</dfn> is a struct with
* A [=struct/item=] <dfn for="WebDriver configuration">global</dfn> which is a [=WebDriver configuration/value=], initially [=WebDriver configuration/unset=].
* A [=struct/item=] <dfn for="WebDriver configuration">user contexts</dfn> which is a weak map between [=user context|user contexts=] and the [=WebDriver configuration/values=], initially empty.
* [=struct/item=] <dfn for="WebDriver configuration struct">navigables</dfn> which is a weak map between [=/navigables=] and [=WebDriver configuration/values=], initially empty.

A [=WebDriver configuration=] has an <dfn for="WebDriver configuration">associated type</dfn> which is a type.

The <dfn for="WebDriver configuration">value</dfn> for a [=WebDriver configuration=] is either a value whose type is the [=WebDriver configuration/associated type=] for that configuration or [=WebDriver configuration/unset=].
```

But I don't insist on this change.

sadym-chromium: @jgraham I removed it and aligned with your proposal.

index.bs
```diff
@@ -426,6 +430,124 @@ To <dfn export>resume</dfn> given |name|, |id| and |parameters|:
...
+
+<div algorithm>
+
+Note: this algorithm allows accessing the [=WebDriver configuration=] for a given [=/navigable=]
+by checking values in [=WebDriver configuration/navigables=], then in
+[=WebDriver configuration/user contexts=] and finally in [=WebDriver configuration/global=].
+Returns [=WebDriver configuration/unset=] if configuration is not set.
+
+To <dfn>get WebDriver configuration value</dfn> of [=WebDriver configuration=] |configuration| for
+[=/navigable=] |navigable|:
```
juliandescottes, line 456: Since we only store configurations for top-level navigables per the `store WebDriver configuration` steps, should we either:
- retrieve the top-level traversable for the navigable 
- or assert navigable is a top-level traversable
?

(the current usage is fine for the scrollbar configuration, but an assert could make this a bit clearer)

index.bs
```diff
@@ -6962,6 +7068,84 @@ The [=remote end steps=] with |command parameters| are:
...
+
+1. Let |scrollbar type override| be |command parameters|["<code>scrollbarType</code>"].
+
+1. If |scrollbar type override| is null, set |scrollbar type override| to
+   [=WebDriver configuration/unset=].
+
+1. If the implementation does not support setting |scrollbar type override|, then return [=error=]
+   with [=error code=] [=unsupported operation=].
+
+1. Let |affected navigables| be the result of [=store WebDriver configuration=]
```
juliandescottes: (Checked with James) Since `store WebDriver configuration` can return an `invalid argument` error, we should use `[=trying=]` or `[=try=]` here to make sure we return the error. 
```suggestion
1. Let |affected navigables| be the result of [=trying=] to [=store WebDriver configuration=]
```

index.bs
```diff
@@ -6962,6 +7068,84 @@ The [=remote end steps=] with |command parameters| are:
...
+1. Let |scrollbar type override| be the result of [=get WebDriver configuration value=] of
+   [=scrollbar type override configuration=] for |navigable|.
+
+1. Assert: |scrollbar type override| is "<code>classic</code>", "<code>overlay</code>" or
+   [=WebDriver configuration/unset=].
+
+1. If |scrollbar type override| is "<code>classic</code>", run [=implementation-defined=] steps to
+   make the |navigable|'s [=active document=] to use <a>classic scrollbars</a> and return.
+
+1. if |scrollbar type override| is "<code>overlay</code>", run [=implementation-defined=]
```
```suggestion
1. If |scrollbar type override| is "<code>overlay</code>", run [=implementation-defined=]
```

nit: case

index.bs
sadym-chromium: I implemented the "retrieve the top-level traversable for the navigable" approach

sadym-chromium:
@jgraham please take another look

jgraham (approved):

PR #1068, opened 2026-02-05
Add browsingContext.setBypassCSP command

### LINKED ISSUE #1033: Should BiDi expose a command / configuration to bypass CSP
At the moment Chrome seems to bypass CSPs by default (see #1024 ). 

But I can see in CDP there is a dedicated command: https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-setBypassCSP
It seems exposed in puppeteer https://pptr.dev/api/puppeteer.page.setbypasscsp and in playwright https://playwright.dev/docs/api/class-testoptions#test-options-bypass-csp

Are those options similar to the topic discussed in #1024 ? (ie allowing to use "eval", "new Function" etc...) via script evaluation, even if the content page is using strict CSPs.

cc @OrKoN @yury-s

juliandescottes:
Fixes #1033 

This adds a new browsingContext command which will store the desired configuration regarding bypassing CSP either for a set of contexts or user contexts.

The configuration is stored in a map on the remote end.

BiDi exposes a hook to be called by the CSP3 spec in order to check if CSP Bypass is in effect for a given navigable.

***
<a href="https://pr-preview.s3.amazonaws.com/juliandescottes/webdriver-bidi/pull/1068.html" title="Last updated on Feb 20, 2026, 1:48 PM UTC (6155a5a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/1068/bf30ceb...juliandescottes:6155a5a.html" title="Last updated on Feb 20, 2026, 1:48 PM UTC (6155a5a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3361,14 +3361,6 @@ weak map between [=user context|user contexts=] and [=unhandled prompt behavior 
 A [=remote end=] has a <dfn>scripting enabled overrides map</dfn> which is a weak
 map between [=/navigables=] or [=user context|user contexts=] and boolean.
 
-A [=remote end=] has a <dfn>bypass CSP configuration</dfn> which is a [=struct=] with
-an [=struct/item=] named <dfn for="bypass CSP configuration">default bypass CSP configuration</dfn>,
-which is initially null,
-an [=struct/item=] named <dfn for="bypass CSP configuration">user context bypass CSP configuration</dfn>,
-which is a weak map between [=user context|user contexts=] and boolean which is initially empty,
-and an [=struct/item=] named <dfn for="bypass CSP configuration">navigable bypass CSP configuration</dfn>,
-which is a weak map between [=/navigables=] and boolean which is initially empty.
-
 ### Types ### {#module-browsingcontext-types}
 
 #### The browsingContext.BrowsingContext Type #### {#type-browsingContext-Browsingcontext}
@@ -5059,6 +5051,9 @@ Note: When CSP bypass is enabled, all CSP directives are bypassed, including tho
    </dd>
 </dl>
 
+A [=remote end=] has a <dfn>bypass CSP configuration</dfn>, which is
+[=WebDriver configuration=] with [=WebDriver configuration/associated type=] boolean.
+
 <div algorithm>
 
 The <dfn export>WebDriver BiDi CSP is bypassed</dfn> steps given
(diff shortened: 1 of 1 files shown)


juliandescottes:
I am not getting the build error locally with the same version of bikeshed ...

(alright after a rebase on main I could see the same errors finally)

whimboo:
Just one thought for now.. maybe we can name it like CDP `setBypassCSP`? IMO this as well reads better by not having the capital letters in the middle of the name.

index.bs
```diff
@@ -3234,6 +3236,9 @@ weak map between [=user context|user contexts=] and [=unhandled prompt behavior
 A [=remote end=] has a <dfn>scripting enabled overrides map</dfn> which is a weak
 map between [=/navigables=] or [=user context|user contexts=] and boolean.
 
+A [=remote end=] has a <dfn>CSP bypass overrides map</dfn> which is a weak
+map between [=/navigables=] or [=user context|user contexts=] and boolean.
```
sadym-chromium: This would be easier if done on top of https://github.com/w3c/webdriver-bidi/pull/1064

index.bs
```diff
@@ -4887,6 +4892,92 @@ The [=remote end steps=] with |command parameters| are:
...
+   <dt>Command Type</dt>
+   <dd>
+    <pre class="cddl" data-cddl-module="remote-cddl">
+      browsingContext.SetCSPBypass = (
+        method: "browsingContext.setCSPBypass",
+        params: browsingContext.SetCSPBypassParameters
+      )
+
+      browsingContext.SetCSPBypassParameters = {
+        bypass: bool,
```
sadym-chromium: I would prefer to have it `true / null`, indicating either to "force bypass CSP" or to "default". This would allow for avoiding confusion in cases like "setting bypass false for navigable and true for user context".

index.bs
```diff
@@ -4887,6 +4892,92 @@ The [=remote end steps=] with |command parameters| are:
...
+   <dd>
+      <pre class="cddl" data-cddl-module="local-cddl">
+      browsingContext.SetCSPBypassResult = EmptyResult
+      </pre>
+   </dd>
+</dl>
+
+<div algorithm>
+
+The <dfn export>WebDriver BiDi CSP bypass is enabled</dfn> steps given
```
sadym-chromium: NIT: "WebDriver BiDi force bypass CSP"?

index.bs
```diff
@@ -4887,6 +4892,92 @@ The [=remote end steps=] with |command parameters| are:
...
+
+The [=remote end steps=] given |command parameters| are:
+
+1. If |command parameters| [=map/contains=] "<code>userContexts</code>"
+   and |command parameters| [=map/contains=] "<code>contexts</code>",
+   return [=error=] with [=error code=] [=invalid argument=].
+
+1. If |command parameters| doesn't [=map/contain=] "<code>userContexts</code>"
+   and |command parameters| doesn't [=map/contain=] "<code>contexts</code>",
+   return [=error=] with [=error code=] [=invalid argument=].
```
sadym-chromium: Consider allowing for global configurations: https://github.com/w3c/webdriver-bidi/issues/956

index.bs
juliandescottes: Ok to swap CSP bypass -> bypass CSP, not sure about adding `force` ? 

Maybe something like "WebDriver BiDi CSP is bypassed" would be better instead of "WebDriver BiDi (force) bypass CSP is enabled"? Let me know what you think

index.bs
juliandescottes: Sounds good. 

(Alternatively we can align to `setScriptingEnabled` and repurpose as `setCSPEnabled` and accept `false / null`?)

index.bs
juliandescottes: Let's wait until that other PR is approved, I can update then.

index.bs
juliandescottes: Sure.

juliandescottes:
@sadym-chromium FWIW, I just rebased on top of the WebDriver Configuration PR

juliandescottes:
No feedback after 1 month on https://github.com/w3c/webappsec-csp/pull/799 , I am merging this on our side.

Issue #1039, opened 2025-11-20
Optional nullable arguments might be not nullable?


nvborisenko:
I am trying to implement BiDi spec in Selenium/.NET project. Recently I found interesting case.

`setViewport` command: https://w3c.github.io/webdriver-bidi/#command-browsingContext-setViewport

```
? viewport: browsingContext.Viewport / null,
? devicePixelRatio: float / null
```

Here `viewport` argument is optional, meaning we can avoid sending it in `json` over wire. And Selenium implemented it, if it is `default/null`, then don't send.

BUT if client sends `viewport` argument in `json` as `null`, then browser behaves as "reset viewport to default". And here the interpretation is in problematic state: how to not send null if it is null, and at the same time provide a possibility to send null if client really wants.

Most likely introducing new command like **resetViewport** can resolve this situation.

jgraham:
That sounds like the Selenium implementation is wrong. Missing and `null` have different meanings in this case, and it's a pattern we use in multiple places in WebDriver-BiDi.

nvborisenko:
I understand. Thank you for proving that `undefined` and `null` have different meanings. It is pity that spec is TS/JS oriented. I quickly "googled":

> Java, Python, Ruby, Go, C#, many others generally have a single null/nil/None value (or use Option/Optional types to model absence) and do not have a built-in separate “undefined” value.

whimboo:
In our Python client for wpt tests we defined a custom type of `UNDEFINED` to handle the different states:
https://github.com/web-platform-tests/wpt/blob/master/tools/webdriver/webdriver/bidi/modules/browsing_context.py#L238-L239

I assume something like this should work with the other listed languages as well?

nvborisenko:
Thank you @whimboo , again you proved that JS/TS is a priority. Meaning all others should apply workarounds. Like "-0" number in json. I just raised an implementation issues out of JS world, up to you how you deal with it.

whimboo:
So what's the problem with using `dict.TryGetValue()`? It returns a boolean that indicates if the given key is present in the dictionary or not. If present the `out value` argument will then have whether a value or `null`. Note that I'm not using .NET so if the above doesn't work please explain it in more detail how an implementation would look like.

Also given that the above logic might be used more often (eg across the emulation commands) a custom class / type on the client side could potentially be used to simplify the logic there.

nvborisenko:
Summary:

BiDi supports three-states (like `undefined` and `null` in JS):
[...]

PR #1069, opened 2026-02-06
Add commands for screencasting

### LINKED ISSUE #636: Support for an screencast method
Hello. I wonder if it's planning to add a similar method to screencast the frames from a session/test, similar to the CDP method Page.startScreencast.

After get/save the frames, it's possible to process and convert them in a video file. Useful for reporting when run the test in a headless browser and see what happened.

I know that the purpose of the webdriver-bidi protocol is not to make a CDP clone, but maybe adding this feature could be very interesting. I also think that this CDP feature is not very complex to implement in a real project, so maybe  it's worth doing the effort to add it to the set of tool you already have.

CDP example
```js
cdp.on('message', msg => {
  const obj = JSON.parse(msg);
  if(obj.method === 'Page.screencastFrame') {
    ws.send(JSON.stringify({
      id: id++,
      method: 'Page.screencastFrameAck',
      params: { sessionId: obj.params.sessionId },
    }));
    // Do your stuff...
  }
});

cdp.send(JSON.stringify({
  id: id++,
  method: 'Page.startScreencast',
  params: {
    maxWidth: 800,
    maxHeight: 600,
  }
}));
```

https://github.com/w3c/webdriver-bidi/assets/23271049/401008e8-9e17-4eb0-beb3-00168e39c872

_video created with FFMPEG_

Please, let me know if this can be planned in a future and feel free to correct me if this is not the right place to ask or if you need more context or info.

Thanks! :)

lutien:
Closes #636.

Introduce the interface to record a video of the screen of the provided navigable and save it to a file.  

~~The PR for the hook in the Screen Capture specification: https://github.com/w3c/mediacapture-screen-share/pull/328.~~

The PR for the hook in the Viewport Capture specification: https://github.com/w3c/mediacapture-viewport/pull/34.

***
<a href="https://pr-preview.s3.amazonaws.com/lutien/webdriver-bidi/pull/1069.html" title="Last updated on May 4, 2026, 4:17 PM UTC (b8241ec)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/1069/898d896...lutien:b8241ec.html" title="Last updated on May 4, 2026, 4:17 PM UTC (b8241ec)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3258,7 +3258,7 @@ A [=BiDi session=] has a <dfn>screencast recordings map</dfn> which is a [=/map=
 which the keys are [[!RFC9562|UUID]]s, and the values are [=structs=] with
 an [=struct/item=] named <code>mediaRecorder</code>, which is a {{MediaRecorder}},
 an [=struct/item=] named <code>mimeType</code>, which is a string,
-an [=struct/item=] named <code>recordedChunks</code>, which is a list,
+an [=struct/item=] named <code>recordedChunks</code>, which is a [=/list=] of {{BlobEvent/data}},
 an [=struct/item=] named <code>context</code>, which is a navigable.
 
 ### Types ### {#module-browsingcontext-types}
--- a/index.bs
+++ b/index.bs
@@ -3262,11 +3262,11 @@ map between [=/navigables=] or [=user context|user contexts=] and boolean.
 
 A [=BiDi session=] has a <dfn>screencast recordings map</dfn> which is a [=/map=] in
 which the keys are [[!RFC9562|UUID]]s, and the values are [=structs=] with
-an [=struct/item=] named <code>mediaRecorder</code>, which is a {{MediaRecorder}},
-an [=struct/item=] named <code>mimeType</code>, which is a string,
-an [=struct/item=] named <code>path</code>, which is a string,
-an [=struct/item=] named <code>size</code>, which is a number,
-an [=struct/item=] named <code>context</code>, which is a navigable.
+an [=struct/item=] named <dfn attribute for="screencast recordings map">mediaRecorder</dfn>, which is a {{MediaRecorder}},
+an [=struct/item=] named <dfn attribute for="screencast recordings map">mimeType</dfn>, which is a string,
+an [=struct/item=] named <dfn attribute for="screencast recordings map">path</dfn>, which is a string,
+an [=struct/item=] named <dfn attribute for="screencast recordings map">size</dfn>, which is a number,
+an [=struct/item=] named <dfn attribute for="screencast recordings map">context</dfn>, which is a navigable.
 
 ### Types ### {#module-browsingcontext-types}
 
--- a/index.bs
+++ b/index.bs
@@ -329,9 +329,9 @@ spec: MEDIACAPTURE-RECORD; urlPrefix: https://w3c.github.io/mediacapture-record/
     text: is type supported; url: #abstract-opdef-is-type-supported
     text: MediaRecorder; url: #mediarecorder-api
     text: MediaRecorderOptions; url: #mediarecorderoptions-section
-spec: MEDIACAPTURE-SCREEN-SHARE; urlPrefix: https://w3c.github.io/mediacapture-screen-share/
+spec: MEDIACAPTURE-VIEWPORT; urlPrefix: https://w3c.github.io/mediacapture-viewport/
   type: dfn
-    text: capture a browser tab; url: #dfn-capture-a-browser-tab
+    text: capture a browsing context viewport; url: #dfn-capture-a-browsing-context-viewport
 spec: MEDIAQUERIES4; urlPrefix: https://drafts.csswg.org/mediaqueries-4/
   type: dfn
     text: resolution media feature; url: #resolution
@@ -5179,7 +5179,7 @@ The [=remote end steps=] with |command parameters| are:
 1. Let <var>realm</var> be the result of [=trying=] to [=get a realm from a navigable=]
    with |navigable id| and null.
 
-1. Let |promise| be the result of [=trying=] to [=capture a browser tab=] with
+1. Let |promise| be the result of [=trying=] to [=capture a browsing context viewport=] with
    |navigable| and |command parameters|["<code>streamOptions</code>"].
 
 1. Let |media stream| be <a spec=ECMASCRIPT>Await</a>(|promise|).
--- a/index.bs
+++ b/index.bs
@@ -3258,7 +3258,8 @@ A [=BiDi session=] has a <dfn>screencast recordings map</dfn> which is a [=/map=
 which the keys are [[!RFC9562|UUID]]s, and the values are [=structs=] with
 an [=struct/item=] named <code>mediaRecorder</code>, which is a {{MediaRecorder}},
 an [=struct/item=] named <code>mimeType</code>, which is a string,
-an [=struct/item=] named <code>recordedChunks</code>, which is a [=/list=] of {{BlobEvent/data}},
+an [=struct/item=] named <code>path</code>, which is a string,
+an [=struct/item=] named <code>size</code>, which is a number,
 an [=struct/item=] named <code>context</code>, which is a navigable.
 
 ### Types ### {#module-browsingcontext-types}
@@ -5149,7 +5150,8 @@ starts the screencast of a given navigable according to provided options.
    <dd>
     <pre class="cddl" data-cddl-module="local-cddl">
       browsingContext.StartScreencastResult = {
-         screencast: browsingContext.Screencast
+         screencast: browsingContext.Screencast,
+         path: text
       }
 
       browsingContext.Screencast = text
@@ -5184,7 +5186,7 @@ The [=remote end steps=] with |command parameters| are:
 
 1. Let |media stream| be <a spec=ECMASCRIPT>Await</a>(|promise|).
 
-1. Let |recorded chunks| be an empty [=/list=].
+1. Let |path| be an implementation-defined file path where the recording will be stored.
 
 1. Let |media recorder options| be a struct with <code>mimeType</code> |mime type|.
 
@@ -5192,25 +5194,33 @@ The [=remote end steps=] with |command parameters| are:
    with {{MediaRecorder/stream}} |media stream| and
    {{MediaRecorderOptions}} |media recorder options|.
 
+1. Let |recording| be a new [=struct=] with
+   <code>mediaRecorder</code> |media recorder|,
+   <code>mimeType</code> |mime type|,
+   <code>path</code> |path|,
+   <code>size</code> 0,
+   <code>context</code> |navigable|.
+
 1. [=Add an event listener=] to |media recorder| with type {{MediaRecorder/dataavailable}}
    and callback that performs the following steps given |event|:
 
    1. Let |data| be |event|'s {{BlobEvent/data}}.
 
-   1. [=list/Append=] |data| to |recorded chunks|.
+   1. Let |bytes| be |data|'s underlying [=byte sequence=].
 
-1. Call {{MediaRecorder/start()}} on |media recorder|.
+   1. Append |bytes| to the file at |path|.
+
+   1. Set |recording|["<code>size</code>"] to |recording|["<code>size</code>"] + |bytes|'s length.
+
+1. Call {{MediaRecorder/start()}} on |media recorder| with <code>1000</code>.
 
 1. Let |screencast| be the string representation of a [[!RFC9562|UUID]].
 
-1. Set [=screencast recordings map=][|screencast|] to a struct with
-   <code>mediaRecorder</code> |media recorder|,
-   <code>mimeType</code> |mime type|,
-   <code>recordedChunks</code> |recorded chunks|,
-   <code>context</code> |navigable|.
+1. Set [=screencast recordings map=][|screencast|] to |recording|.
 
 1. Return a new [=/map=] matching the <code>browsingContext.StartScreencastResult</code>
-   with the <code>screencast</code> field set to |screencast|.
+   with the <code>screencast</code> field set to |screencast| and <code>path</code> field
+   set to |path|.
 
 </div>
 
@@ -5257,32 +5267,18 @@ The [=remote end steps=] with |command parameters| are:
 
 1. Let |media recorder| be |screencast recording|["<code>mediaRecorder</code>"].
 
-1. Let |recorded chunks| be |screencast recording|["<code>recordedChunks</code>"].
-
-1. Let |navigable id| be the [=navigable id=] for |screencast recording|["<code>navigable</code>"].
-
-1. Let <var>realm</var> be the result of [=trying=] to [=get a realm from a navigable=]
-   with |navigable id| and null.
+1. Let |path| be |screencast recording|["<code>path</code>"].
 
 1. Call {{MediaRecorder/stop()}} on |media recorder|.
 
 1. Wait until |media recorder|'s {{MediaRecorder/stop}} [=fire an event|event fires=].
 
-1. Let |blob options| be a struct with <code>type</code> |screencast recording|["<code>mimeType</code>"].
-
-1. Let |blob| to be a new {{Blob}} in |realm| with {{blobParts}} |recorded chunks| and
-   {{BlobPropertyBag}} |blob options|.
-
-1. Let |path| be an implementation-defined file path where the recording will be stored.
-
-1. Let |bytes| be |blob|'s underlying [=byte sequence=].
-
-1. Write |bytes| to the file system at |path|.
+1. Let |size| be |screencast recording|["<code>size</code>"].
 
 1. [=map/Remove=] |screencast| from [=screencast recordings map=].
 
 1. Return a new [=/map=] matching the <code>browsingContext.StopScreencastResult</code>
-   with the <code>path</code> field set to |path| and <code>size</code> field set to |blob|'s {{Blob/size}}.
+   with the <code>path</code> field set to |path| and <code>size</code> field set to |size|.
 
 </div>
 
--- a/index.bs
+++ b/index.bs
@@ -5237,7 +5237,8 @@ stops the screencast.
    <dd>
     <pre class="cddl" data-cddl-module="local-cddl">
       browsingContext.StopScreencastResult = {
-         path: text
+         path: text,
+         size: js-uint
       }
     </pre>
    </dd>
@@ -5281,7 +5282,7 @@ The [=remote end steps=] with |command parameters| are:
 1. [=map/Remove=] |screencast| from [=screencast recordings map=].
 
 1. Return a new [=/map=] matching the <code>browsingContext.StopScreencastResult</code>
-   with the <code>path</code> field set to |path|.
+   with the <code>path</code> field set to |path| and <code>size</code> field set to |blob|'s {{Blob/size}}.
 
 </div>
 
--- a/index.bs
+++ b/index.bs
@@ -5143,7 +5143,6 @@ starts the screencast of a given navigable according to provided options.
          ? height: js-uint,
          ? aspectRatio: js-uint,
          ? frameRate: js-uint,
-         ? facingMode: text,
          ? resizeMode: text,
          ? sampleRate: js-uint,
          ? sampleSize: js-uint,
@@ -5154,7 +5153,6 @@ starts the screencast of a given navigable according to provided options.
          ? channelCount: js-uint,
          ? deviceId: text,
          ? groupId: text,
-         ? backgroundBlur: bool
       }
 
       </pre>
--- a/index.bs
+++ b/index.bs
@@ -289,12 +289,17 @@ spec: CSSOM-VIEW; urlPrefix: https://drafts.csswg.org/cssom-view/
     text: web-exposed screen area; url: #web-exposed-screen-area
 spec: DOM; urlPrefix: https://dom.spec.whatwg.org/
   type: dfn
+    text: add an event listener; url: #add-an-event-listener
     text: root; url: #concept-tree-root
     text: document element; url: #ref-for-dom-document-documentelement
     text: evaluate; url: #dom-xpathevaluatorbase-evaluate
     text: nodes; url: #concept-node
     text: ORDERED_NODE_SNAPSHOT_TYPE; url: #dom-xpathresult-ordered_node_snapshot_type
     text: snapshotItem; url: #dom-xpathresult-snapshotitem
+spec: FILE-API; urlPrefix: https://w3c.github.io/FileAPI/
+  type: dfn
+    text: Blob; url: #dfn-Blob
+    text: blobParts; url: #dfn-blobParts
 spec: FULLSCREEN; urlPrefix: https://fullscreen.spec.whatwg.org/
   type: dfn
     text: fullscreen an element; url: #fullscreen-an-element
@@ -319,6 +324,14 @@ spec: ACCNAME; urlPrefix:https://www.w3.org/TR/accname-1.2
 spec: CORE-AAM; urlPrefix:https://www.w3.org/TR/core-aam-1.2
   type: dfn
     text: computed role; url: /#roleMappingComputedRole
+spec: MEDIACAPTURE-RECORD; urlPrefix: https://w3c.github.io/mediacapture-record/
+  type: dfn
+    text: is type supported; url: #abstract-opdef-is-type-supported
+    text: MediaRecorder; url: #mediarecorder-api
+    text: MediaRecorderOptions; url: #mediarecorderoptions-section
+spec: MEDIACAPTURE-SCREEN-SHARE; urlPrefix: https://w3c.github.io/mediacapture-screen-share/
+  type: dfn
+    text: capture a browser tab; url: #dfn-capture-a-browser-tab
 spec: MEDIAQUERIES4; urlPrefix: https://drafts.csswg.org/mediaqueries-4/
   type: dfn
     text: resolution media feature; url: #resolution
@@ -670,6 +683,9 @@ with the following additional codes:
 
   <dt><dfn for=errors export>no such request</dfn>
   <dd>Tried to continue an unknown [=/request=].
+
+  <dt><dfn for=errors export>no such screencast</dfn>
+  <dd>Tried to stop an unknown screencast recording.
 
   <dt><dfn for=errors export>no such script</dfn>
   <dd>Tried to remove an unknown [=preload script=].
@@ -3146,6 +3162,8 @@ BrowsingContextCommand = (
   browsingContext.Print //
   browsingContext.Reload //
   browsingContext.SetViewport //
+  browsingContext.StartScreencast //
+  browsingContext.StopScreencast //
   browsingContext.TraverseHistory
 )
 </pre>
@@ -3165,6 +3183,8 @@ BrowsingContextResult = (
   browsingContext.PrintResult /
   browsingContext.ReloadResult /
   browsingContext.SetViewportResult /
+  browsingContext.StartScreencastResult /
+  browsingContext.StopScreencastResult /
   browsingContext.TraverseHistoryResult
 )
 
@@ -3233,6 +3253,13 @@ weak map between [=user context|user contexts=] and [=unhandled prompt behavior 
 
 A [=remote end=] has a <dfn>scripting enabled overrides map</dfn> which is a weak
 map between [=/navigables=] or [=user context|user contexts=] and boolean.
+
+A [=BiDi session=] has a <dfn>screencast recordings map</dfn> which is a [=/map=] in
+which the keys are [[!RFC9562|UUID]]s, and the values are [=structs=] with
+an [=struct/item=] named <code>mediaRecorder</code>, which is a {{MediaRecorder}},
+an [=struct/item=] named <code>mimeType</code>, which is a string,
+an [=struct/item=] named <code>recordedChunks</code>, which is a list,
+an [=struct/item=] named <code>context</code>, which is a navigable.
 
 ### Types ### {#module-browsingcontext-types}
 
@@ -5083,6 +5110,179 @@ The [=remote end steps=] with |command parameters| are:
          1. [=Set device pixel ratio override=] with |navigable| and |device pixel ratio|.
 
 1. Return [=success=] with data null.
+
+</div>
+
+#### The browsingContext.startScreencast Command ####  {#command-browsingContext-startScreencast}
+
+The <dfn export for=commands>browsingContext.startScreencast</dfn> command
+starts the screencast of a given navigable according to provided options.
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+      <pre class="cddl" data-cddl-module="remote-cddl">
+      browsingContext.StartScreencast = (
+        method: "browsingContext.startScreencast",
+        params: browsingContext.StartScreencastParameters
+      )
+
+      browsingContext.StartScreencastParameters = {
+        context: browsingContext.BrowsingContext,
+        mimeType: text,
+        ? streamOptions: browsingContext.MediaStreamOptions,
+      }
+
+      browsingContext.MediaStreamOptions = {
+         ? video: bool / browsingContext.MediaTrackConstraints .default true;
+         ? audio: bool / browsingContext.MediaTrackConstraints .default false;
+      }
+
+      browsingContext.MediaTrackConstraints = {
+         ? width: js-uint,
+         ? height: js-uint,
+         ? aspectRatio: js-uint,
+         ? frameRate: js-uint,
+         ? facingMode: text,
+         ? resizeMode: text,
+         ? sampleRate: js-uint,
+         ? sampleSize: js-uint,
+         ? echoCancellation: bool / text,
+         ? autoGainControl: bool,
+         ? noiseSuppression: bool,
+         ? latency: js-uint,
+         ? channelCount: js-uint,
+         ? deviceId: text,
+         ? groupId: text,
+         ? backgroundBlur: bool
+      }
+
+      </pre>
+   </dd>
+   <dt>Return Type</dt>
+   <dd>
+    <pre class="cddl" data-cddl-module="local-cddl">
+      browsingContext.StartScreencastResult = {
+         screencast: browsingContext.Screencast
+      }
+
+      browsingContext.Screencast = text
+    </pre>
+   </dd>
+</dl>
+
+<div algorithm="remote end steps for browsingContext.startScreencast">
+The [=remote end steps=] with |command parameters| are:
+
+1. Let |navigable| be the result of [=trying=] to [=get a navigable=]
+   with |command parameters|["<code>context</code>"].
+
+1. If |navigable| is not a [=/top-level traversable=], return [=error=] with
+   [=error code=] [=invalid argument=].
+
+1. Let |mime type| to be |command parameters|["<code>mimeType</code>"].
+
+1. If [=is type supported=] with |mime type| returns false, return [=error=] with
+   [=error code=] [=invalid argument=].
+
+1. If the implementation is unable to record a screencast of |navigable| for any
+   reason then return [=error=] with [=error code=] [=unsupported operation=].
+
+1. Let |media stream| be the result of [=trying=] to [=capture a browser tab=] with
+   |navigable| and |command parameters|["<code>streamOptions</code>"].
+
+1. Let |recorded chunks| be an empty [=/list=].
+
+1. Let |media recorder options| be a struct with <code>mimeType</code> |mime type|.
+
+1. Let |media recorder| be a new {{MediaRecorder}} with {{MediaRecorder/stream}} |media stream| and
+   {{MediaRecorderOptions}} |media recorder options|.
+
+1. [=Add an event listener=] to |media recorder| with type {{MediaRecorder/dataavailable}}
+   and callback that performs the following steps given |event|:
+
+   1. Let |data| be |event|'s {{BlobEvent/data}}.
+
+   1. [=list/Append=] |data| to |recorded chunks|.
+
+1. Call {{MediaRecorder/start()}} on |media recorder|.
+
+1. Let |screencast| be the string representation of a [[!RFC9562|UUID]].
+
+1. Set [=screencast recordings map=][|screencast|] to a struct with
+   <code>mediaRecorder</code> |media recorder|,
+   <code>mimeType</code> |mime type|,
+   <code>recordedChunks</code> |recorded chunks|,
+   <code>context</code> |navigable|.
+
+1. Return a new [=/map=] matching the <code>browsingContext.StartScreencastResult</code>
+   with the <code>screencast</code> field set to |screencast|.
+
+</div>
+
+#### The browsingContext.stopScreencast Command ####  {#command-browsingContext-stopScreencast}
+
+The <dfn export for=commands>browsingContext.stopScreencast</dfn> command
+stops the screencast.
+
+<dl>
+   <dt>Command Type</dt>
+   <dd>
+      <pre class="cddl" data-cddl-module="remote-cddl">
+      browsingContext.StopScreencast = (
+        method: "browsingContext.stopScreencast",
+        params: browsingContext.StopScreencastParameters
+      )
+
+      browsingContext.StopScreencastParameters = {
+        screencast: browsingContext.Screencast
+      }
+      </pre>
+   </dd>
+   <dt>Return Type</dt>
+   <dd>
+    <pre class="cddl" data-cddl-module="local-cddl">
+      browsingContext.StopScreencastResult = {
+         path: text
+      }
+    </pre>
+   </dd>
+</dl>
+
+<div algorithm="remote end steps for browsingContext.stopScreencast">
+The [=remote end steps=] with |command parameters| are:
+
+1. Let |screencast| be the value of the "<code>screencast</code>" field in |command
+   parameters|.
+
+1. If [=screencast recordings map=] does not <a for=map>contain</a> |screencast|, return
+   [=error=] with [=error code=] [=no such screencast=].
+
+1. Let |screencast recording| be [=screencast recordings map=][|screencast|].
+
+1. Let |media recorder| be |screencast recording|["<code>mediaRecorder</code>"].
+
+1. Let |recorded chunks| be |screencast recording|["<code>recordedChunks</code>"].
+
+1. Call {{MediaRecorder/stop()}} on |media recorder|.
+
+1. Wait until |media recorder|'s {{MediaRecorder/stop}} [=fire an event|event fires=].
+
+1. Let |blob options| be a struct with <code>type</code> |screencast recording|["<code>mimeType</code>"].
+
+1. Let |blob| to be a new {{Blob}} with {{blobParts}} |recorded chunks| and
+   {{BlobPropertyBag}} |blob options|.
+
+1. Let |path| be an implementation-defined file path where the recording will be stored.
+
+1. Let |bytes| be |blob|'s underlying [=byte sequence=].
+
+1. Write |bytes| to the file system at |path|.
+
+1. [=map/Remove=] |screencast| from [=screencast recordings map=].
+
+1. Return a new [=/map=] matching the <code>browsingContext.StopScreencastResult</code>
+   with the <code>path</code> field set to |path|.
 
 </div>
 
--- a/index.bs
+++ b/index.bs
@@ -5188,8 +5188,10 @@ The [=remote end steps=] with |command parameters| are:
 1. If the implementation is unable to record a screencast of |navigable| for any
    reason then return [=error=] with [=error code=] [=unsupported operation=].
 
-1. Let |media stream| be the result of [=trying=] to [=capture a browser tab=] with
+1. Let |promise| be the result of [=trying=] to [=capture a browser tab=] with
    |navigable| and |command parameters|["<code>streamOptions</code>"].
+
+1. Let |media stream| be <a spec=ECMASCRIPT>Await</a>(|promise|).
 
 1. Let |recorded chunks| be an empty [=/list=].
 
--- a/index.bs
+++ b/index.bs
@@ -3261,12 +3261,10 @@ A [=remote end=] has a <dfn>scripting enabled overrides map</dfn> which is a wea
 map between [=/navigables=] or [=user context|user contexts=] and boolean.
 
 A [=BiDi session=] has a <dfn>screencast recordings map</dfn> which is a [=/map=] in
-which the keys are [[!RFC9562|UUID]]s, and the values are [=structs=] with
-an [=struct/item=] named <dfn attribute for="screencast recordings map">mediaRecorder</dfn>, which is a {{MediaRecorder}},
-an [=struct/item=] named <dfn attribute for="screencast recordings map">mimeType</dfn>, which is a string,
-an [=struct/item=] named <dfn attribute for="screencast recordings map">path</dfn>, which is a string,
-an [=struct/item=] named <dfn attribute for="screencast recordings map">size</dfn>, which is a number,
-an [=struct/item=] named <dfn attribute for="screencast recordings map">context</dfn>, which is a navigable.
+which the keys are [[!RFC9562|UUID]]s, and the values are <dfn>screencast recording</dfn>, which is a [=struct=] with
+an [=struct/item=] named <dfn for="screencast recording">mediaRecorder</dfn>, which is a {{MediaRecorder}},
+an [=struct/item=] named <dfn for="screencast recording">path</dfn>, which is a string,
+an [=struct/item=] named <dfn for="screencast recording">size</dfn>, which is a number.
 
 ### Types ### {#module-browsingcontext-types}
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -5086,6 +5113,179 @@ The [=remote end steps=] with |command parameters| are:
...
+        mimeType: text,
+        ? streamOptions: browsingContext.MediaStreamOptions,
+      }
+
+      browsingContext.MediaStreamOptions = {
+         ? video: bool / browsingContext.MediaTrackConstraints .default true;
+         ? audio: bool / browsingContext.MediaTrackConstraints .default false;
+      }
+
+      browsingContext.MediaTrackConstraints = {
```
jgraham, line 5146: Do all of these options actually make sense for capturing a tab? Some of them seem more relevant to capturing video e.g. from a camera.

index.bs
```diff
@@ -5086,6 +5113,179 @@ The [=remote end steps=] with |command parameters| are:
...
+
+1. Let |mime type| to be |command parameters|["<code>mimeType</code>"].
+
+1. If [=is type supported=] with |mime type| returns false, return [=error=] with
+   [=error code=] [=invalid argument=].
+
+1. If the implementation is unable to record a screencast of |navigable| for any
+   reason then return [=error=] with [=error code=] [=unsupported operation=].
+
+1. Let |media stream| be the result of [=trying=] to [=capture a browser tab=] with
```
jgraham: IIUC `[=capture a browser tab=]` is returning a promise, and we later pass that to `{{MediaRecorder}}` as if it's a resolved stream.

index.bs
```diff
@@ -5086,6 +5113,179 @@ The [=remote end steps=] with |command parameters| are:
...
+   |navigable| and |command parameters|["<code>streamOptions</code>"].
+
+1. Let |recorded chunks| be an empty [=/list=].
+
+1. Let |media recorder options| be a struct with <code>mimeType</code> |mime type|.
+
+1. Let |media recorder| be a new {{MediaRecorder}} with {{MediaRecorder/stream}} |media stream| and
```
jgraham: This feels somewhat sketchy in that it is invoking a lot of JS machinery without being clear about which realm it's operating it.

nvborisenko:
Command versus events... From client side perspective it doesn't make sense to just `browsingContext.StartScreencast`, in any case a client needs to provide a handler for events. Just a proposal to think about: what if it would be regular event pattern.. I subscribe (actually starting screencast on remote end), and handle events data.

So instead of (pseudo programming language):
```
var scr = browsingContext.startScreencast(new Options { frameRate: 5 })
browsingContext.on('screencastDataReceived', e => doSomething())
scr.stopScreencast()
```

We can do the following:
```
var scr = browsingContext.on('screencastDataReceived', new Options { frameRate: 5 }, e => doSomething())
scr.unsubscribe()
```

For a client `startScreencast` method seems just as enabler for event's data, why not to try to reuse events modelling in this case? I understand it requires to review existing model, and allow some additional arguments when subscribing. And it is good! We already have a possibility to provide `userContext` or `browsingContext` when subscribing. I can imagine the following example where it can be useful: `network.beforeRequestSent` event is raised for all requests.. would be nice if I can listen to only "**/*.png" requests.

index.bs
lutien: Right, added the "await" logic

index.bs
lutien: Right, I've removed `facingMode` and `backgroundBlur`, which are indeed camera specific. 
There is a bunch of audio-specific options, and I guess that's still a question to still discuss if we want to support audio recording at all, but that's why I kept them for now.
Let me know if you see anything else.

index.bs
lutien: Ok, I've added the logic to create `MediaRecorder` and `Blob` in the navigable's realm. But since I couldn't really find a similar case in the BiDi spec, I had to improvise with the wording :) So feel free to correct me

lutien:
@nvborisenko, the current idea for the screencast API is to start with commands to record the video and save it to a file (sorry, it was probably not very clear, I've updated the PR description now). And for this interface, we need start and stop commands. The next step would be to introduce the streaming option. In that case we probably will just pass a different `type` to the start command to indicate that we want to get a stream, then the command will return some handle that can be used to read the stream (see #1061). So we actually do not currently plan to introduce the event-based interface, but even if we decide to do this in the future, I guess I would see it as another `type` for `startScreencast` command. You can check the minutes from the last group discussion here: https://github.com/w3c/webdriver-bidi/issues/636#issuecomment-3751265528.

index.bs
```diff
@@ -3234,6 +3254,13 @@ weak map between [=user context|user contexts=] and [=unhandled prompt behavior
 A [=remote end=] has a <dfn>scripting enabled overrides map</dfn> which is a weak
 map between [=/navigables=] or [=user context|user contexts=] and boolean.
 
+A [=BiDi session=] has a <dfn>screencast recordings map</dfn> which is a [=/map=] in
+which the keys are [[!RFC9562|UUID]]s, and the values are [=structs=] with
+an [=struct/item=] named <code>mediaRecorder</code>, which is a {{MediaRecorder}},
+an [=struct/item=] named <code>mimeType</code>, which is a string,
+an [=struct/item=] named <code>recordedChunks</code>, which is a list,
```
```suggestion
an [=struct/item=] named <code>recordedChunks</code>, which is a [=/list=] of {{BlobEvent/data}},
```

index.bs
```diff
@@ -5086,6 +5113,190 @@ The [=remote end steps=] with |command parameters| are:
...
+         ? resizeMode: text,
+         ? sampleRate: js-uint,
+         ? sampleSize: js-uint,
+         ? echoCancellation: bool / text,
+         ? autoGainControl: bool,
+         ? noiseSuppression: bool,
+         ? latency: js-uint,
+         ? channelCount: js-uint,
+         ? deviceId: text,
+         ? groupId: text,
```
sadym-chromium: what are params `deviceId` and `groupId` for? Do we really support `echoCancellation`?

index.bs
```diff
@@ -5086,6 +5113,190 @@ The [=remote end steps=] with |command parameters| are:
...
+
+      browsingContext.StartScreencastParameters = {
+        context: browsingContext.BrowsingContext,
+        mimeType: text,
+        ? streamOptions: browsingContext.MediaStreamOptions,
+      }
+
+      browsingContext.MediaStreamOptions = {
+         ? video: bool / browsingContext.MediaTrackConstraints .default true;
+         ? audio: bool / browsingContext.MediaTrackConstraints .default false;
```
sadym-chromium: I don't think the whole `browsingContext.MediaTrackConstraints` makes sense in case of audio.

I'd propose to restrict both video and audio to `true/false` and extend it on demand.

index.bs
```diff
@@ -319,6 +324,14 @@ spec: ACCNAME; urlPrefix:https://www.w3.org/TR/accname-1.2
...
   type: dfn
     text: computed role; url: /#roleMappingComputedRole
+spec: MEDIACAPTURE-RECORD; urlPrefix: https://w3c.github.io/mediacapture-record/
+  type: dfn
+    text: is type supported; url: #abstract-opdef-is-type-supported
+    text: MediaRecorder; url: #mediarecorder-api
+    text: MediaRecorderOptions; url: #mediarecorderoptions-section
+spec: MEDIACAPTURE-SCREEN-SHARE; urlPrefix: https://w3c.github.io/mediacapture-screen-share/
+  type: dfn
+    text: capture a browser tab; url: #dfn-capture-a-browser-tab
```
sadym-chromium: The link does not work.

lutien: Yes, the hook is not merged yet. Here is the PR for it: https://github.com/w3c/mediacapture-screen-share/pull/328.

index.bs
lutien: > I don't think the whole `browsingContext.MediaTrackConstraints` makes sense in case of audio.

Yeah, that's just the approach that is used for `getDisplayMedia` (see https://w3c.github.io/mediacapture-screen-share/#dom-displaymediastreamoptions), but I agree that we can start smaller and expand. I would only suggest adding already `width` and `height` constraints for the video, since we know that Playwright requires them. The PR is updated.

index.bs
lutien: I've removed them as discussed in the other comment about constraints.

index.bs
```diff
@@ -5086,6 +5113,178 @@ The [=remote end steps=] with |command parameters| are:
...
+
+1. Let |navigable id| be |command parameters|["<code>context</code>"].
+
+1. Let |navigable| be the result of [=trying=] to [=get a navigable=]
+   with |navigable id|.
+
+1. If |navigable| is not a [=/top-level traversable=], return [=error=] with
+   [=error code=] [=invalid argument=].
+
+1. Let |mime type| to be |command parameters|["<code>mimeType</code>"].
```
```suggestion
1. Let |mime type| be |command parameters|["<code>mimeType</code>"].
```

index.bs
```diff
@@ -5086,6 +5113,178 @@ The [=remote end steps=] with |command parameters| are:
...
+
+1. Let |path| be an implementation-defined file path where the recording will be stored.
+
+1. Let |bytes| be |blob|'s underlying [=byte sequence=].
+
+1. Write |bytes| to the file system at |path|.
+
+1. [=map/Remove=] |screencast| from [=screencast recordings map=].
+
+1. Return a new [=/map=] matching the <code>browsingContext.StopScreencastResult</code>
```
juliandescottes, line 5305: Could be interesting to return the size (bytes' length?) as well?

juliandescottes (commented):
I can only provide limited feedback. From my point of view it sounds reasonable.

index.bs
```diff
@@ -5086,6 +5113,178 @@ The [=remote end steps=] with |command parameters| are:
...
+1. If [=is type supported=] with |mime type| returns false, return [=error=] with
+   [=error code=] [=invalid argument=].
+
+1. If the implementation is unable to record a screencast of |navigable| for any
+   reason then return [=error=] with [=error code=] [=unsupported operation=].
+
+1. Let <var>realm</var> be the result of [=trying=] to [=get a realm from a navigable=]
+   with |navigable id| and null.
+
+1. Let |promise| be the result of [=trying=] to [=capture a browser tab=] with
```
OrKoN: is this expected to trigger any permission dialogs or screen view selection dialogs? I think it should not but just want to double check (could not follow the link).

index.bs
```diff
@@ -5086,6 +5113,178 @@ The [=remote end steps=] with |command parameters| are:
...
+1. Let |blob options| be a struct with <code>type</code> |screencast recording|["<code>mimeType</code>"].
+
+1. Let |blob| to be a new {{Blob}} in |realm| with {{blobParts}} |recorded chunks| and
+   {{BlobPropertyBag}} |blob options|.
+
+1. Let |path| be an implementation-defined file path where the recording will be stored.
+
+1. Let |bytes| be |blob|'s underlying [=byte sequence=].
+
+1. Write |bytes| to the file system at |path|.
```
OrKoN: would implementations be allowed to write to file during screencast recording? it does not seem like we should require writing only in the stop command?

index.bs
lutien: Here is the link to the PR: https://github.com/w3c/mediacapture-screen-share/pull/328.
Yes, the idea is that we should bypass permission and screen view selection dialogs.

lutien:
For the visibility, [in the PR for the `getDisplayMedia` hook](https://github.com/w3c/mediacapture-screen-share/pull/328#discussion_r2828687208), there is a proposal to move the hook to https://www.w3.org/TR/mediacapture-viewport/#capturing-viewport-media, which is basically a subset of `getDisplayMedia` to work only with top-level browsing context, so it should be easier to integrate. That sounds reasonable to me, so I'm going to move it there.

index.bs
```diff
@@ -5086,6 +5113,178 @@ The [=remote end steps=] with |command parameters| are:
...
+      )
+
+      browsingContext.StartScreencastParameters = {
+        context: browsingContext.BrowsingContext,
+        mimeType: text,
+        ? streamOptions: browsingContext.MediaStreamOptions,
+      }
+
+      browsingContext.MediaStreamOptions = {
+         ? video: bool / browsingContext.MediaTrackConstraints .default true;
```
nvborisenko: It is hard for strongly typed languages. `boolean` or `object`? I propose to avoid it early. Kind of https://github.com/w3c/webdriver-bidi/issues/1039, but in this case it is even harder.

whimboo:
>  which is basically a subset of `getDisplayMedia` to work only with top-level browsing context, so it should be easier to integrate. That sounds reasonable to me, so I'm going to move it there.

Is there a reason why we only want to allow recording of a screencast for a top-level browsing context? We do not have this restriction for capturing a screenshot and I can see use cases when consumers only want to record a frame.

lutien:
> Is there a reason why we only want to allow recording of a screencast for a top-level browsing context? We do not have this restriction for capturing a screenshot and I can see use cases when consumers only want to record a frame.

There is no support for recording a video of a frame in `getDisplayMedia` and, as far as I know, there is no other WebRTC API currently to do that. I think CDP also only supports that (maybe @OrKoN or @sadym-chromium can correct me if I'm wrong). In my opinion it's fine as a start, and we can add it in the future if/when the support on the WebRTC side will be added.

sadym-chromium:
> Is there a reason why we only want to allow recording of a screencast for a top-level browsing context? We do not have this restriction for capturing a screenshot and I can see use cases when consumers only want to record a frame.

It is unlikely that Chromium will support nested browsing contexts' screencasts. I would make a decision based on the following:
1. If there are other vendors who are going to support "nested browsing contexts' screencasts" **AND** there is a requirement from the clients for that, then add it into the spec.
2. Otherwise leave the main spec tight for the top-level browsing contexts. If a single vendor wants to implement it, they can extend the command with vendor-specific parameter (like `moz:nestedBrowsingContext` or something) to implement it.

index.bs
lutien: Sounds good, I've added the size property

index.bs
lutien: Yes, I think it makes sense. I've updated the PR to write to the file during the recording, and `startScreecast` will now also return the file path.

index.bs
```diff
@@ -5086,6 +5120,174 @@ The [=remote end steps=] with |command parameters| are:
...
+
+   1. Let |data| be |event|'s {{BlobEvent/data}}.
+
+   1. Let |bytes| be |data|'s underlying [=byte sequence=].
+
+   1. Append |bytes| to the file at |path|.
+
+   1. Set |recording|["<code>size</code>"] to |recording|["<code>size</code>"] + |bytes|'s length.
+
+1. Call {{MediaRecorder/start()}} on |media recorder| with <code>1000</code>.
```
lutien: [MediaRecorder.start() requires a timeslice](https://developer.mozilla.org/en-US/docs/Web/API/MediaRecorder/start#timeslice) to provide the data regularly; otherwise, it will only do it when the `stop` or `requestData` commands are called. `1000ms` seems to be the common value, but we can, of course, make it configurable. When we call `stop` in `stopScreencast` the data will be available quicker if it's required.

index.bs
OrKoN: Ack, thanks

index.bs
lutien: This is the type structure from https://w3c.github.io/mediacapture-main/#dom-mediastreamconstraints, and we use this spec as a base for these APis. I understand that it doesn't look very elegant, but I'm not sure why it's difficult to implement for strongly typed languages. We have the same type definitions for, e.g., `getDisplayMedia` in our C++ code. For #1039 I can see that some languages don't have the mechanism to distinguish between `null` and `undefined`, but as far as I know, usually the languages have some way to specify the union type.

lutien:
All comments should be addressed. The new link for the external hook PR is https://github.com/w3c/mediacapture-viewport/pull/34 (also updated in the PR description).

index.bs
```diff
@@ -1693,6 +1709,12 @@ To <dfn>cleanup the session</dfn> given |session|:
    1. For each |collected data| in [=collected network data=], [=remove collector from data=]
       with |collected data| and |collector id|.
 
+1. For each |screencast recording| in |session|'s [=screencast recordings map=]:
+
+   1. Let |media recorder| be |screencast recording|["<code>mediaRecorder</code>"].
```
sadym-chromium, line 1708: Please use typed reference here and in other places:

```suggestion
   1. Let |media recorder| be |screencast recording|'s [=screencast recording/mediaRecorder=].
```

index.bs
```diff
@@ -3234,6 +3260,14 @@ weak map between [=user context|user contexts=] and [=unhandled prompt behavior
 A [=remote end=] has a <dfn>scripting enabled overrides map</dfn> which is a weak
 map between [=/navigables=] or [=user context|user contexts=] and boolean.
 
+A [=BiDi session=] has a <dfn>screencast recordings map</dfn> which is a [=/map=] in
+which the keys are [[!RFC9562|UUID]]s, and the values are [=structs=] with
+an [=struct/item=] named <code>mediaRecorder</code>, which is a {{MediaRecorder}},
```
sadym-chromium: Let's use typed definition to ease reference finding:
```suggestion
A [=BiDi session=] has a <dfn>screencast recordings map</dfn> which is a [=/map=] in
which the keys are [[!RFC9562|UUID]]s, and the values are <dfn>screencast recording</dfn>, which is a [=struct=] with
an [=struct/item=] named <dfn for="screencast recording">mediaRecorder</dfn>, which is a {{MediaRecorder}},
```

index.bs
```diff
@@ -5086,6 +5120,174 @@ The [=remote end steps=] with |command parameters| are:
...
+
+1. Let |path| be an implementation-defined file path where the recording will be stored.
+
+1. Let |media recorder options| be a struct with <code>mimeType</code> |mime type|.
+
+1. Let |media recorder| be a new {{MediaRecorder}} in |realm|
+   with {{MediaRecorder/stream}} |media stream| and
+   {{MediaRecorderOptions}} |media recorder options|.
+
+1. Let |recording| be a new [=struct=] with
```
```suggestion
1. Let |recording| be a new [=screencast recording=] with
```

OrKoN:
Does Firefox support `navigator.mediaDevices.getViewportMedia`?

lutien:
> Does Firefox support `navigator.mediaDevices.getViewportMedia`?

It doesn't. But since `getViewportMedia` is basically a subset of `getDisplayMedia`, my plan implementation-wise is still to use parts of `getDisplayMedia`. And refactor later on, when `getViewportMedia` is implemented.

index.bs
lutien: It should be fixed now.

index.bs
lutien: It should be addressed now.

index.bs
lutien: And this should be fixed as well.

index.bs
```diff
@@ -5086,6 +5118,172 @@ The [=remote end steps=] with |command parameters| are:
...
+1. Let |path| be an implementation-defined file path where the recording will be stored.
+
+1. Let |media recorder options| be a struct with <code>mimeType</code> |mime type|.
+
+1. Let |media recorder| be a new {{MediaRecorder}} in |realm|
+   with {{MediaRecorder/stream}} |media stream| and
+   {{MediaRecorderOptions}} |media recorder options|.
+
+1. Let |recording| be a new [=screencast recording=] with
+   <code>mediaRecorder</code> |media recorder|,
```
sadym-chromium: Please use the proper linking in other places as well
```suggestion
   [=screencast recording/mediaRecorder=] |media recorder|,
```

lutien: Ah, right, sorry. Hopefully with the last commit, I found all the remaining usage of the "screencast recording" items.

lutien:
Henrik had a question about error handling of writing to the file (e.g., no space on the disk). 
With the update to write to the file while recording, it's harder to handle since the writing process is happening asynchronously from the actual command call. I guess we could add an event, e.g., `browsingContext.screenscastFailed` which would include an error message. Does it sound like something that might be helpful for the clients? Alternatively, the `stopRecording` command could still return an error when an ID of the failed recording is provided. It could also be both. Please let me know what you think.

index.bs
lutien: Added the command parameter, which defaults to `1000ms`.

lutien:
Ok, I've updated `browsingContext.stopScreencast` to throw an error if it happens while writing to a file. And also now we stop recording if there was an error.

index.bs
```diff
@@ -5086,6 +5120,182 @@ The [=remote end steps=] with |command parameters| are:
...
+1. If [=is type supported=] with |mime type| returns false, return [=error=] with
+   [=error code=] [=invalid argument=].
+
+1. If the implementation is unable to record a screencast of |navigable| for any
+   reason then return [=error=] with [=error code=] [=unsupported operation=].
+
+1. Let <var>realm</var> be the result of [=trying=] to [=get a realm from a navigable=]
+   with |navigable id| and null.
+
+1. Let |promise| be the result of [=trying=] to [=capture a browsing context viewport=] with
```
jgraham, line 5194: This still feels a bit sketchy to me; the "capture a browsing context viewport" algorithm creates a promise without any JS running and without specifying which realm the promise is in. 

We are also passing in an Infra Map as `options`, whereas the algorithm expects a `WebIDL `dictionary`. I _think_ this is OK because WebIDL suggests that dictonary instances are Maps, but I'm not 100% sure.

index.bs
```diff
@@ -5086,6 +5120,182 @@ The [=remote end steps=] with |command parameters| are:
...
+   |navigable| and |command parameters|["<code>streamOptions</code>"].
+
+1. Let |media stream| be <a spec=ECMASCRIPT>Await</a>(|promise|).
+
+1. Let |path| be an implementation-defined file path where the recording will be stored.
+
+1. Let |media recorder options| be a [=struct=] with an [=struct/item=] named <code>mimeType</code>
```
jgraham: Doesn't this need to be a Map, not a struct, if it's taking the place of a Web IDL Dictionary type?

index.bs
```diff
@@ -5086,6 +5120,182 @@ The [=remote end steps=] with |command parameters| are:
...
+1. If the implementation is unable to record a screencast of |navigable| for any
+   reason then return [=error=] with [=error code=] [=unsupported operation=].
+
+1. Let <var>realm</var> be the result of [=trying=] to [=get a realm from a navigable=]
+   with |navigable id| and null.
+
+1. Let |promise| be the result of [=trying=] to [=capture a browsing context viewport=] with
+   |navigable| and |command parameters|["<code>streamOptions</code>"].
+
+1. Let |media stream| be <a spec=ECMASCRIPT>Await</a>(|promise|).
```
jgraham: I don't think you can just use `await` here. You probably want to use https://webidl.spec.whatwg.org/#dfn-perform-steps-once-promise-is-settled Also note https://webidl.spec.whatwg.org/#mark-a-promise-as-handled which you likely want to avoid getting content-visible errors if the promise is rejected.

index.bs
```diff
@@ -5086,6 +5120,182 @@ The [=remote end steps=] with |command parameters| are:
...
+1. Let |media recorder| be a new {{MediaRecorder}} in |realm|
+   with {{MediaRecorder/stream}} |media stream| and
+   {{MediaRecorderOptions}} |media recorder options|.
+
+1. Let |recording| be a new [=screencast recording=] with
+   [=screencast recording/mediaRecorder=] |media recorder|,
+   [=screencast recording/path=] |path|,
+   [=screencast recording/size=] 0.
+
+1. [=Add an event listener=] to |media recorder| with type {{MediaRecorder/dataavailable}}
```
jgraham: [Add an event listener](https://dom.spec.whatwg.org/#add-an-event-listener) doesn't allow a set of steps as the callback argument. 

I think we can do something more hacky here, like:

```
Whenever the implementation is going to [=fire a blob event=] named [=dataavailable=] at |media recorder|with |blob|, run the following steps:
```

It's not ideal (because it has COMEFROM semantics), but I think it's better than trying to use the DOM API from the spec.

index.bs
```diff
@@ -5086,6 +5120,182 @@ The [=remote end steps=] with |command parameters| are:
...
+   and callback that performs the following steps given |event|:
+
+   1. Let |data| be |event|'s {{BlobEvent/data}}.
+
+   1. Let |bytes| be |data|'s underlying [=byte sequence=].
+
+   1. If appending |bytes| to the file at |path| fails:
```
```suggestion
   1. Append |bytes| to the file at |path|. If this fails:
```

index.bs
```diff
@@ -5086,6 +5120,182 @@ The [=remote end steps=] with |command parameters| are:
...
+
+      1. Set |recording|'s [=screencast recording/writeError=] to an
+         implementation-defined string describing the write failure.
+
+      1. Call {{MediaRecorder/stop()}} on |media recorder|.
+
+   1. Otherwise, set |recording|'s [=screencast recording/size=] to |recording|'s [=screencast recording/size=] + |bytes|'s length.
+
+1. Call {{MediaRecorder/start()}} on |media recorder| with |command parameters|["<code>timeslice</code>"].
```
jgraham: So, every time we do Call, it might need to actually be https://tc39.es/ecma262/#sec-call

Also, I think that currently any exceptions thrown by the JS leak into the content realm, assuming that's where the code is run. We coul maybe try creating a new sandbox realm for this to run in, to avoid that. I don't think anything really depends on which realm we're in; the Media Capture spec doesn't seem to use it.

index.bs
```diff
@@ -5086,6 +5120,182 @@ The [=remote end steps=] with |command parameters| are:
 
 </div>
 
+#### The browsingContext.startScreencast Command ####  {#command-browsingContext-startScreencast}
+
+The <dfn export for=commands>browsingContext.startScreencast</dfn> command
+starts the screencast of a given navigable according to provided options.
```
jgraham: "according to the provided options" is true for all commands. In this case it would be more useful to call out that it's to a file on disk.

Also I notice that we always leak the file and depend on the client to clean up. That's probably fine, but there are cases where it could be impossible (e.g. if the browser had r/w access to the filesystem but the client ro access). It's likely worth a note.

jgraham (changes requested):

index.bs
lutien: I am not sure if we can fix something in the BiDi spec, but I guess I could update https://github.com/w3c/mediacapture-viewport/pull/34 to create a promise in the realm of the passed browsing context. What do you think?

index.bs
lutien: Fixed ✅

index.bs
lutien: Updated ✅

index.bs
lutien: Ah, ok, we used `await` for the script evaluation, that's why I thought I could just use it here. I've updated this algorithm to use `react to promise` steps. But I'm not sure what you meant about the promise rejection, I've  now added the step to throw an error when the promise is rejected. Do I still have to mark the promise as handled?

index.bs
lutien: Ok, I've updated to use `Call` and created a sandbox for the MediaRecorder.

index.bs
lutien: I've updated the description and added the note.

index.bs
jan-ivar: It's not just the promise but the returned `MediaStream` object and its embedded `MediaStreamTrack` objects for video and audio we need to consider. All four are JS objects given no realm to live in.

From [handle an incoming message](https://pr-preview.s3.amazonaws.com/lutien/webdriver-bidi/pull/1069.html#handle-an-incoming-message) it seems _remote end steps_ all run [in parallel](https://html.spec.whatwg.org/multipage/infrastructure.html#in-parallel):

Image

So there's no JS set up here yet AFAIU. How are you planning on using these objects? From JS or c++?

index.bs
```diff
@@ -5086,6 +5121,189 @@ The [=remote end steps=] with |command parameters| are:
...
+
+      1. Let |path| be an implementation-defined file path where the recording will be stored.
+
+      1. Let |media recorder options| be a new [=/map=] with the <code>mimeType</code> field
+         set to |mime type|.
+
+      1. Let |screencast| be the string representation of a [[!RFC9562|UUID]].
+
+      1. Let |realm| be result of [=trying=] to [=get or create a sandbox realm=] given
+         |screencast| and |navigable|.
```
jan-ivar: Is this the destination realm for the promise, its stream, and tracks? If so, perhaps we can pass it into the capture algorithm earlier?

index.bs
```diff
@@ -5086,6 +5121,189 @@ The [=remote end steps=] with |command parameters| are:
...
+         with {{MediaRecorder/stream}} |media stream| and
+         {{MediaRecorderOptions}} |media recorder options|.
+
+      1. Let |recording| be a new [=screencast recording=] with
+         [=screencast recording/mediaRecorder=] |media recorder|,
+         [=screencast recording/path=] |path|,
+         [=screencast recording/size=] 0.
+
+      1. Whenever the implementation is going to [=fire a blob event=] named {{MediaRecorder/dataavailable}}
+         at |media recorder| with |blob|, run the following steps:
```
jan-ivar, line 5222: Is there a precedent for this approach in some other command?

I'm not super familiar with WebDriver-BiDi and sandbox realms, but it's surprising to me to see JS-facing APIs used in this way in parallel. I'd normally associate this with data races. These JS objects are being created on the navigable being captured? What thread do their constructors run on?

Is this the long-term plan?
- If yes, passing in the realm to the capture algorithm earlier might work
- If no, might adding automation steps to https://w3c.github.io/mediacapture-record  be another approach to ultimately try to pass around non-JS objects here?

jan-ivar (commented):
Thanks for working on this!

I have some questions to aid review in https://github.com/w3c/mediacapture-viewport/pull/34.

index.bs
jgraham: > Is there a precedent for this approach in some other command?

I think the answer is "not really".

Historically WebDriver has managed to call into algorithms that are written entirely in terms of abstract spec objects. In browser terms this is roughly equivalent to native code (i.e. C++ or similar).

However as we add more modern platform features we're more frequently running into cases where the spec itself assumes that it's being called from executing script, and is written in terms of operations on JS objects. That obviously makes sense if the only entry point is via scripting interfaces defined in WebIDL, but a WebDriver endpoint is not that.

A current idea I have is to create a agent/environment settings object/realm that's defined in the WebDriver spec and is only used for running spec-internal code, somewhat similar to parent process JS in Firefox. I asked about this idea on [matrix](https://matrix.to/#/!AGetWbsMpFPdSgUrbs:matrix.org/$zDK1t596MAKbL0nNVDs2r-VxstfZ8gaSG1WlOSzBofY?via=matrix.org&via=mozilla.org&via=igalia.com), but so far no one gave any feedback on whether that was a silly idea…

index.bs
lutien: Ok, so I've added "Prepare to run script" to set up JS in the specification environment (something like the Firefox parent process).

index.bs
```diff
@@ -5086,6 +5121,196 @@ The [=remote end steps=] with |command parameters| are:
...
+
+1. Let |mime type| be |command parameters|["<code>mimeType</code>"].
+
+1. If [=is type supported=] with |mime type| returns false, return [=error=] with
+   [=error code=] [=invalid argument=].
+
+1. If the implementation is unable to record a screencast of |navigable| for any
+   reason then return [=error=] with [=error code=] [=unsupported operation=].
+
+1. Let |environment settings| be the [=environment settings object=] representing
```
jgraham, line 5187: This needs to be better defined, but for the purposes of the PR an `ISSUE` would be fine.

index.bs
lutien: So the plan is to have two options for this command:
- save the MediaStream into the file (what is this PR about now);
- stream the MediaStream over the websocket to the client (after we introduce the general streaming support with #1061).

So I think we still need to get out of me `capture a browsing context viewport` at least the MediaStream. I guess we could try to resolve the promise on the `capture a browsing context viewport` side, but I'm not sure if it would make it easier.

index.bs
```diff
@@ -5086,6 +5121,196 @@ The [=remote end steps=] with |command parameters| are:
...
+   1. If |promise| was fulfilled with value |media stream|, then:
+
+      1. Let |path| be an implementation-defined file path where the recording will be stored.
+
+      1. Let |media recorder options| be a new [=/map=] with the <code>mimeType</code> field
+         set to |mime type|.
+
+      1. Let |screencast| be the string representation of a [[!RFC9562|UUID]].
+
+      1. Let |realm| be result of [=trying=] to [=get or create a sandbox realm=] given
```
jgraham: If we're going to use the concept of an execution script environment earlier in the algorithm, I think we can use the realm for that environment here.

lutien: Makes sense, updated

index.bs
lutien: Added the issue.

index.bs
```diff
@@ -5086,6 +5121,197 @@ The [=remote end steps=] with |command parameters| are:
...
+
+1. If the implementation is unable to record a screencast of |navigable| for any
+   reason then return [=error=] with [=error code=] [=unsupported operation=].
+
+1. Let |environment settings| be the [=environment settings object=] representing
+   a specification execution environment.
+
+   Issue: The specification execution environment has to be better defined.
+
+1. [=Prepare to run script=] with |environment settings|.
```
OrKoN, line 5192: Can we specify it without requiring the implementations to run scripts?

index.bs
```diff
@@ -5086,6 +5121,197 @@ The [=remote end steps=] with |command parameters| are:
...
+      1. Let |screencast| be the string representation of a [[!RFC9562|UUID]].
+
+      1. Let |realm| be |environment settings|' [=realm execution context=]'s Realm component.
+
+      1. Let |media recorder| be a new {{MediaRecorder}} in |realm|
+         with {{MediaRecorder/stream}} |media stream| and
+         {{MediaRecorderOptions}} |media recorder options|.
+
+      1. Let |recording| be a new [=screencast recording=] with
+         [=screencast recording/mediaRecorder=] |media recorder|,
```
OrKoN, line 5217: Can we save objects like |media recorder| across navigations? I thought media recorder would be bound to a specific realm?

index.bs
jgraham: The problem we have is that we want to call existing platform APIs that assume in their specification that they're being called from a WebIDL interface which can only be invoked when running script. So they do things like use promises, whose sematics are undefined outside of the context of JS execution. If we don't use that model we have to reimplement the entire API with different semantics

My plan is to have a "specification agent" which allows us to run script in a way that's invisible to the content process, and is basically only a specification formalism. Implementations will be free to not actually run script as long as the observable behaviour is correct.

index.bs
```diff
@@ -295,6 +295,10 @@ spec: DOM; urlPrefix: https://dom.spec.whatwg.org/
     text: nodes; url: #concept-node
     text: ORDERED_NODE_SNAPSHOT_TYPE; url: #dom-xpathresult-ordered_node_snapshot_type
     text: snapshotItem; url: #dom-xpathresult-snapshotitem
+spec: FILE-API; urlPrefix: https://w3c.github.io/FileAPI/
+  type: dfn
+    text: Blob; url: #dfn-Blob
+    text: blobParts; url: #dfn-blobParts
```
tidoust: The PR does not seem to make use of these.

```suggestion
```

(Also, "Blob" should be readily available as `{{Blob}}`)

index.bs
```diff
@@ -319,6 +324,15 @@ spec: ACCNAME; urlPrefix:https://www.w3.org/TR/accname-1.2
 spec: CORE-AAM; urlPrefix:https://www.w3.org/TR/core-aam-1.2
   type: dfn
     text: computed role; url: /#roleMappingComputedRole
+spec: MEDIACAPTURE-RECORD; urlPrefix: https://w3c.github.io/mediacapture-record/
+  type: dfn
+    text: is type supported; url: #abstract-opdef-is-type-supported
```
tidoust: Spec already exports that term.

```suggestion
```

index.bs
```diff
@@ -319,6 +324,15 @@ spec: ACCNAME; urlPrefix:https://www.w3.org/TR/accname-1.2
...
+    text: fire a blob event; url: #fire-a-blob-event
+    text: MediaRecorder; url: #mediarecorder-api
+    text: MediaRecorderOptions; url: #mediarecorderoptions-section
```
tidoust: Spec already exports these terms, and the PR (correctly) references them as IDL term through the `{{...}}` shorthand, which this declaration does not expose (it creates "term" definitions for them).

```suggestion
```

index.bs
```diff
@@ -308,7 +312,8 @@ spec: SELECTORS4; urlPrefix: https://drafts.csswg.org/selectors-4/
 spec: WEB-IDL; urlPrefix: https://webidl.spec.whatwg.org/
   type: dfn
     text: DOMException; url: #idl-DOMException
-    text: SyntaxError; url:#syntaxerror
+    text: SyntaxError; url: #syntaxerror
+    text: react; url: #dfn-perform-steps-once-promise-is-settled
```
tidoust: WebIDL already exports "react" (although it may be better to reference it through `[=promise/react=]`)

```suggestion
```

index.bs
```diff
@@ -5086,6 +5121,197 @@ The [=remote end steps=] with |command parameters| are:
...
+
+      browsingContext.StartScreencastParameters = {
+        context: browsingContext.BrowsingContext,
+        mimeType: text,
+        ? streamOptions: browsingContext.MediaStreamOptions,
+        ? timeslice: js-uint .default 1000
+      }
+
+      browsingContext.MediaStreamOptions = {
+         ? video: bool / browsingContext.MediaTrackConstraints .default true;
```
OrKoN: are there any output format for the video/audio that is expected by the proposed spec?

index.bs
lutien: hmm... it looks like this export was still required, see https://github.com/w3c/webdriver-bidi/actions/runs/24392829522/job/71243049434?pr=1069.

tidoust: Oh, that's because it's defined as an `abstract-op` definition, and Bikeshed [uses a different shorthand syntax](https://speced.github.io/bikeshed/#autolink-wrappers) for abstract operations. It should work with `[$is type supported$]`.

I got confused because ReSpec does not use a specific shorthand syntax for abstract operations...

lutien: Ah, alright, I've fixed it to use the correct syntax

index.bs
lutien: There is a [mimeType](https://pr-preview.s3.amazonaws.com/lutien/webdriver-bidi/pull/1069.html#cddl-type-browsingcontextstartscreencastparameters) parameter that is passed to `MediaRecorder` to define the output format.

OrKoN: Right, but basically no specific format that has to be supported is required? Do we expect the mime type to also define the audio codec that is used? (IIUC there could be various audio codes within same video formats)

lutien: > Right, but basically no specific format that has to be supported is required? 

Ah, you mean that we don't have specific formats listed here that have to be supported?
My idea here was that it should be all formats that are supported by the `MediaRecorder` implementation. That's why there is [step 5](https://pr-preview.s3.amazonaws.com/lutien/webdriver-bidi/pull/1069.html#commands-browsingcontextstartscreencast) to validate if it's supported by a browser. But I guess that's something we could discuss.
 
> Do we expect the mime type to also define the audio codec that is used? (IIUC there could be various audio codes within same video formats)

Yes, that's the idea (e.g., `video/webm;codecs=vp9,opus`).

OrKoN: Do you know if there is a matrix of what formats/codecs are currently supported by the web-facing API by web browsers?

lutien: No, sorry 😞 I found only some testing pages, e.g., https://www.sitelint.com/lab/media-recorder-supported-mime-type/, but you have to run them in each browser to see the support data.

index.bs
```diff
@@ -5283,6 +5113,197 @@ The [=remote end steps=] with |command parameters| are:
...
+
+1. Let |navigable| be the result of [=trying=] to [=get a navigable=]
+   with |navigable id|.
+
+1. If |navigable| is not a [=/top-level traversable=], return [=error=] with
+   [=error code=] [=invalid argument=].
+
+1. Let |mime type| be |command parameters|["<code>mimeType</code>"].
+
+1. If [$is type supported$] with |mime type| returns false, return [=error=] with
```
OrKoN: we are likely to support only few formats in the automation and not everything that is supported in the web API. I would suggest we drop this check in favour of the next unsupported-operation check. WebAPI-supported things can be checked by evaluating `MediaSource.isTypeSupported` via scripts for users who want to test what is available in the web API. WDYT?

index.bs
OrKoN: Thanks, that was very helpful!

index.bs
lutien: Yeah, I think that should be fine. I'll update the spec

index.bs
OrKoN: @jgraham @lutien do you think we can rewrite this spec proposal without using promises / IDL / requiring a JS execution context? it looks like in the media viewport the relevant portion of the spec is https://w3c.github.io/mediacapture-viewport/#dom-mediadevices-getviewportmedia (step 10.3)? And then it just becomes a [MediaStream from IDL](https://www.w3.org/TR/mediacapture-streams/#dom-mediastream)? Perhaps we could link to the constraints defined by the spec and eventually make a WebDriver BiDi stream out of it re-using the stream spec algorithms.

```
The provided media MUST include precisely one video track, which MUST be a live-capture of the [browser](https://www.w3.org/TR/screen-capture/#dfn-browser) [display surface](https://www.w3.org/TR/screen-capture/#dfn-display-surface) of the [relevant global object](https://html.spec.whatwg.org/multipage/webappapis.html#concept-relevant-global)'s [associated Document](https://html.spec.whatwg.org/multipage/nav-history-apis.html#concept-document-window)'s [top-level browsing context](https://html.spec.whatwg.org/multipage/document-sequences.html#top-level-browsing-context)'s [viewport](https://html.spec.whatwg.org/multipage/#viewport).

The provided media MUST include at most one audio track, which, if provided, MUST be the combined audio produced by the sum of documents that consist of the [relevant global object](https://html.spec.whatwg.org/multipage/webappapis.html#concept-relevant-global)'s [associated Document](https://html.spec.whatwg.org/multipage/nav-history-apis.html#concept-document-window)'s [top-level browsing context](https://html.spec.whatwg.org/multipage/document-sequences.html#top-level-browsing-context)'s [active document](https://html.spec.whatwg.org/multipage/document-sequences.html#nav-document), and all [active documents](https://html.spec.whatwg.org/multipage/document-sequences.html#nav-document) in nested [browsing context](https://html.spec.whatwg.org/multipage/document-sequences.html#browsing-context)s of the [relevant global object](https://html.spec.whatwg.org/multipage/webappapis.html#concept-relevant-global)'s [associated Document](https://html.spec.whatwg.org/multipage/nav-history-apis.html#concept-document-window)'s [top-level browsing context](https://html.spec.whatwg.org/multipage/document-sequences.html#top-level-browsing-context). This audio track MUST NOT be included if audio was not specified in requestedMediaTypes, or if it was specified as false.

The source of a [MediaStreamTrack](https://www.w3.org/TR/mediacapture-streams/#dom-mediastreamtrack) MUST NOT change.

If the result of the request is "[granted](https://www.w3.org/TR/permissions/#dom-permissionstate-granted)", then for each device that is sourcing the provided media, using a stable and private id for the device, deviceId, set [[devicesLiveMap]][deviceId] to true, if it isn’t already true, and set the [[devicesAccessibleMap]][deviceId] to true, if it isn’t already true.

The user agent MUST NOT store a "[granted](https://www.w3.org/TR/permissions/#dom-permissionstate-granted)" permission entry.
```

index.bs
```diff
@@ -5283,6 +5113,197 @@ The [=remote end steps=] with |command parameters| are:
...
+        ? timeslice: js-uint .default 1000
+      }
+
+      browsingContext.MediaStreamOptions = {
+         ? video: bool / browsingContext.MediaTrackConstraints .default true;
+         ? audio: bool .default false;
+      }
+
+      browsingContext.MediaTrackConstraints = {
+         ? width: js-uint,
```
OrKoN, line 5147: how are width / height parameters used by the underlying spec? Is it something to scale the output to or is it mix/max constraint for the stream selection? I do not seem to find mentions of it (same for `timeslice`).

index.bs
```diff
@@ -5283,6 +5113,197 @@ The [=remote end steps=] with |command parameters| are:
...
+      browsingContext.StartScreencastParameters = {
+        context: browsingContext.BrowsingContext,
+        mimeType: text,
+        ? streamOptions: browsingContext.MediaStreamOptions,
+        ? timeslice: js-uint .default 1000
+      }
+
+      browsingContext.MediaStreamOptions = {
+         ? video: bool / browsingContext.MediaTrackConstraints .default true;
+         ? audio: bool .default false;
```
OrKoN, line 5143: do we expect audio: true, video: false to work? the underlying spec seems to require us to reject that.

> If options.video is false, return a promise [rejected](https://webidl.spec.whatwg.org/#reject) with a newly [created](https://webidl.spec.whatwg.org/#dfn-create-exception) [TypeError](https://webidl.spec.whatwg.org/#exceptiondef-typeerror).

lutien: Yeah, I've just also noticed that viewport capturing with video is not allowed. I've updated the types.

index.bs
lutien: `width` and `height` are supposed to be used for scaling. You can see more info here: https://www.w3.org/TR/screen-capture/#constrainable-properties. `timeslice` is used as a parameter for `MediaRecorder.start(timeslice)` (spec: https://w3c.github.io/mediacapture-record/#dom-mediarecorder-start)

index.bs
lutien: Updated.

index.bs
lutien: It looks like `MediaStream` is supposed to live in some JS execution context as well (see comment here: https://github.com/w3c/webdriver-bidi/pull/1069#discussion_r2943054864). I've created a draft PR with an attempt to sketch something without requiring JS: https://github.com/w3c/webdriver-bidi/pull/1113. I've used some working from https://w3c.github.io/mediacapture-viewport/, which we could potentially share, but mostly it's just abstract describing, since I couldn't really find the way to share more with other specs.

OrKoN: Thanks! I guess if we had https://github.com/w3c/webdriver-bidi/pull/1061#discussion_r3094302072 specified without JS execution realms we could just use WebDriver BiDi's own stream instead of MediaStream?

lutien: From what I've understood (tagging @jgraham to correct me/clarify), it still might be an issue for working with the streams, because the stream algorithms are really determined that they are running JS.

index.bs
```diff
@@ -5283,6 +5113,194 @@ The [=remote end steps=] with |command parameters| are:
...
+      browsingContext.StartScreencast = (
+        method: "browsingContext.startScreencast",
+        params: browsingContext.StartScreencastParameters
+      )
+
+      browsingContext.StartScreencastParameters = {
+        context: browsingContext.BrowsingContext,
+        mimeType: text,
+        ? streamOptions: browsingContext.MediaStreamOptions,
+        ? timeslice: js-uint .default 1000
```
OrKoN: to confirm: the timeslice only controls the size of the chunks but does not affect the framerate? Do we need to expose timeslice in this case? (we just write to file right and with the stream API the client would probably determine the size of the chunks).

lutien: Yes, that's right. It's only required to define when to write to a file (if we want to write while recording), and I agree we don't really have to expose it.

OrKoN: Since we do not have the dedicated event for this, I think clients would have difficulty to extract chunks of the right timeslice duration from the file. I would suggest we remove timeslice and leave it to be implementation -defined and instead we add a framerate that would be very useful for clients?

index.bs
```diff
@@ -5283,6 +5113,194 @@ The [=remote end steps=] with |command parameters| are:
...
+   <dd>
+      <pre class="cddl" data-cddl-module="remote-cddl">
+      browsingContext.StartScreencast = (
+        method: "browsingContext.startScreencast",
+        params: browsingContext.StartScreencastParameters
+      )
+
+      browsingContext.StartScreencastParameters = {
+        context: browsingContext.BrowsingContext,
+        mimeType: text,
```
OrKoN: wdyt to make mimeType optional to mean implementation-defined default mime type?

index.bs
lutien: Sounds good. I've updated the PR to make `timeslice` an implementation-defined and added `frameRate` to the video constraints.

index.bs
lutien: I think that should be fine. Updated the PR ✅

index.bs
OrKoN: Do we have a conclusion here? I think it would be nice to avoid JS for specifying this and I think maybe we can specify our own stream behavior that also does not rely on running JS?

lutien: So to unblock things, we decided to focus on https://github.com/w3c/webdriver-bidi/pull/1113, but only for saving screencasting to the file for now. And then come back to the streaming option after we're more certain about the generic streaming API. We were planning to talk about it on Wednesday, but we can discuss it now if you have any feedback already.

OrKoN: That sounds good to me, should I review https://github.com/w3c/webdriver-bidi/pull/1113/changes?

lutien: Sure (I thought it would be good to review internally first, to save you the trouble. But if you have time, I guess there is no point in delaying 🙂 ).

Issue #1053, opened 2025-12-22
Allow forcing viewport meta tag respect


sadym-chromium:
Partly addresses https://github.com/w3c/webdriver-bidi/issues/772.
Related Specification: [CSS Viewport Module Level 1](https://www.w3.org/TR/css-viewport-1/)

### Description

Currently, WebDriver BiDi lacks a mechanism to mandate that the browser engine respect the `<meta name="viewport">` tag.

On desktop platforms, engines traditionally ignore viewport metadata and use a fixed-width layout logic. There is currently no standardized, protocol-level way to force an engine to transition into a layout mode that honors these document instructions.

### Problems & Gaps

1. **Inaccurate Mobile Emulation:** Users can resize a window to a mobile width (e.g., 390px), but without the engine parsing the meta tag, it continues to use a desktop rendering path (often a 980px fixed width). This results in "shrunk" content rather than a true mobile reflow.
2. **Inability to Test Scaling Constraints:** Document instructions such as `initial-scale`, `minimum-scale`, and `user-scalable` are ignored by desktop-based automation sessions. This prevents the verification of accessibility and UX requirements related to zooming.
3. **Decoupling of Layout and Visual Viewports:** Automated tests cannot currently simulate scenarios where the **Layout Viewport** (dictated by the meta tag, e.g., `width=600`) differs from the **Visual Viewport** (the window size, e.g., 390px), which is standard behavior on mobile devices.

### Scenarios
Apart of the mobile emulation scenario, the following scenarios are impossible at the moment.

#### Verifying Responsive Reflow

A developer wants to verify that a website correctly reflows when a viewport meta tag specifies `width=device-width`. Currently, setting a narrow viewport in a desktop browser fails to trigger the mobile layout path, rendering a narrow "slice" of a desktop-width page rather than reflowing the content.

#### Custom Layout Viewports
A site uses a specific width for its mobile view (e.g., `<meta name="viewport" content="width=600">`) while viewed on a 390px device. Currently, the browser ignores the `600px` instruction and sets the layout width to the window width (390px), failing to reproduce the horizontal scrolling behavior that a real device would exhibit.

### Proposed Solution
We propose introducing a command `browsingContext.setViewportMetaEnabled`. This command serves as a signal to the User Agent to transition from its default desktop rendering logic to the mobile-specific processing model defined in the CSS Viewport specification.

whimboo:
See as well issue #934.

sadym-chromium:
[...]

PR #1073, opened 2026-02-12
`emulation.setViewportMetaOverride`

sadym-chromium:
Addressing #1053.

Introduce a new command `emulation.setViewportMetaOverride` allowing for force-using viewport `<meta>` element. 

Defined the logic in the "8. Patches to Other Specifications" until the related CSS PR https://github.com/w3c/csswg-drafts/pull/13548 is merged.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/1073.html" title="Last updated on Jul 28, 2026, 1:34 PM UTC (56ee296)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/1073/513a880...56ee296.html" title="Last updated on Jul 28, 2026, 1:34 PM UTC (56ee296)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6971,8 +6971,11 @@ The [=remote end steps=] given |command parameters| are:
 1. Let |affected navigables| be the result of [=store WebDriver configuration=]
    [=viewport meta override configuration=] |viewport meta override| for |command parameters|.
 
-   Note: no need to update the |affected navigables|, as |viewport meta override| takes affect
-   only after the next navigation.
+1. For each |navigable| of |affected navigables|:
+
+   1. Let |document| be |navigable|'s [=active document=].
+
+   1. Run the [=evaluate media queries and report changes=] steps for |document|.
 
 1. Return [=success=] with data null.
 
--- a/index.bs
+++ b/index.bs
@@ -6971,11 +6971,8 @@ The [=remote end steps=] given |command parameters| are:
 1. Let |affected navigables| be the result of [=store WebDriver configuration=]
    [=viewport meta override configuration=] |viewport meta override| for |command parameters|.
 
-1. For each |navigable| of |affected navigables|:
-
-   1. Let |document| be |navigable|'s [=active document=].
-
-   1. Run the [=evaluate media queries and report changes=] steps for |document|.
+1. For each |navigable| of |affected navigables|, run [=evaluate media queries and report changes=]
+   for [=/document=] currently loaded in a specified |navigable|.
 
 1. Return [=success=] with data null.
 
--- a/index.bs
+++ b/index.bs
@@ -6079,7 +6079,8 @@ EmulationCommand = (
   emulation.SetScrollbarTypeOverride //
   emulation.SetTimezoneOverride //
   emulation.SetTouchOverride //
-  emulation.SetUserAgentOverride
+  emulation.SetUserAgentOverride //
+  emulation.SetViewportMetaOverride
 )
 
 </pre>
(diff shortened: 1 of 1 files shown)


OrKoN (approved):

sadym-chromium:
@lutien WDYT?

index.bs
```diff
@@ -6974,6 +6976,74 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+1. Let |viewport meta override| be |command parameters|["<code>viewportMeta</code>"].
+
+1. If |viewport meta override| is null, set |viewport meta override| to
+   [=WebDriver configuration/unset=].
+
+1. Let |affected navigables| be the result of [=store WebDriver configuration=]
+   [=viewport meta override configuration=] |viewport meta override| for |command parameters|.
+
+   Note: no need to update the |affected navigables|, as |viewport meta override| takes affect
+   only after the next navigation.
```
lutien: I think it's clear that we have to trigger reflow somehow for the viewport meta tag to take effect. And I guess navigation is the only way to do it now (?). But I'm not sure if it's fine to just delay it to the next navigation. So tagging @jgraham and @juliandescottes to get their opinions.

juliandescottes: I'm not sure if it's only about a reflow, or if handling the viewport meta is only something that happens at navigation time. If it only requires a reflow, it sounds like we could try to apply this dynamically and maybe the spec should reflect that.

If not, if it's really something that can only be applied after navigation, it might be good to expose this to a user via an additional parameter that allows to reload impacted contexts? Otherwise if you want to set the emulation and make sure all impacted contexts are applying it you'd need to set the emulation, then get all the contexts, and then send individual reload commands.

sadym-chromium: Agree. I updated the algorithm to trigger "evaluate media queries and report changes" which should be sufficient to re-calculate the viewport changes.

index.bs
```diff
@@ -6909,6 +6911,74 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+
+1. Let |viewport meta override| be |command parameters|["<code>viewportMeta</code>"].
+
+1. If |viewport meta override| is null, set |viewport meta override| to
+   [=WebDriver configuration/unset=].
+
+1. Let |affected navigables| be the result of [=store WebDriver configuration=]
+   [=viewport meta override configuration=] |viewport meta override| for |command parameters|.
+
+1. For each |navigable| of |affected navigables|, run [=evaluate media queries and report changes=]
```
sadym-chromium, line 7430: I believe this is sufficient to re-calculate the viewport meta, as according to the spec, the UA MUST use it after it is enabled.

index.bs
```diff
@@ -6909,6 +6911,74 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+<div algorithm="remote end steps for emulation.setViewportMetaOverride">
+
+The [=remote end steps=] given |command parameters| are:
+
+1. Let |viewport meta override| be |command parameters|["<code>viewportMeta</code>"].
+
+1. If |viewport meta override| is null, set |viewport meta override| to
+   [=WebDriver configuration/unset=].
+
+1. Let |affected navigables| be the result of [=store WebDriver configuration=]
```
juliandescottes: This should use trying to
```suggestion
1. Let |affected navigables| be the result of [=trying=] to [=store WebDriver configuration=]
```

index.bs
```diff
@@ -6909,6 +6911,74 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+A [=remote end=] has a <dfn>viewport meta override configuration</dfn>, which is
+[=WebDriver configuration=] with [=WebDriver configuration/associated type=] boolean.
+
+<div algorithm>
+
+The <dfn export>WebDriver BiDi viewport meta state</dfn> steps given
+[=/Document=] |document| are:
+
+1. Let |navigable| be |document|'s [=node navigable=].
+
```
juliandescottes, line 7408: I guess we don't really care about retrieving the top level navigable here, since the viewport meta is only relevant for top level documents? We could still do it for consistency though.

juliandescottes (approved):
LGTM, one question + one thing to fix.

Issue #750, opened 2024-07-16
Support for testing various media features


OrKoN:
Puppeteer allows testing various [media features](https://developer.mozilla.org/en-US/docs/Web/CSS/@media) on a specific browsing context. Apart from the viewport media queries the following use cases are common:

1) dark vs light mode
2) reduced motion
3) color-gamut

jelly:
We would also like to be able to change color-scheme in BiDi as we currently do via [CDP](https://chromedevtools.github.io/devtools-protocol/tot/Emulation/#method-setEmulatedMedia). This allows us to easily test dark/light mode but it only works in Chromium based browsers right now as [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1549434) chose to not implement it.

whimboo:
> There is an open PR for forced colors at [#905](https://github.com/w3c/webdriver-bidi/pull/905).

This was actually implemented as [`emulation.setForcedColorsModeThemeOverride`](https://w3c.github.io/webdriver-bidi/#command-emulation-setForcedColorsModeThemeOverride), which allows you to set a theme (light, dark).

@OrKoN so points 1 and 4 could be striked out from your initial list. Should we add contrast emulation as well?

sadym-chromium:
> There is an open PR for forced colors at [#905](https://github.com/w3c/webdriver-bidi/pull/905).

These are 2 different features: 

* `forced-colors` **(Accessibility / High Contrast)**
This feature indicates whether the user agent has restricted the color palette to a limited, user-defined set.  It is intended for accessibility (High Contrast Mode).
  * CSS Specification: [CSS Color Adjustment Module Level 1 - Forced Colors](https://www.w3.org/TR/css-color-adjust-1/#forced-colors-properties) 
  * Behavior: When active, the browser overrides the author's specified colors (background, text, links) and replaces them with system colors. 

* `prefers-color-scheme` **(User Preference / Dark Mode)**
This feature is used to detect if the user has requested the system use a light or dark color theme.
  * CSS Specification: [Media Queries Level 4 - Prefers Color Scheme](https://www.w3.org/TR/mediaqueries-4/#prefers-color-scheme) 
  * Behavior: It does not override author styles. Instead, it allows developers to adapt their UI using @media (prefers-color-scheme: dark) to provide a theme that matches the OS setting.

sadym-chromium:
### Proposed Solution
[...]

PR #1076, opened 2026-02-16
`emulation.setMediaFeaturesOverride`

sadym-chromium:
Addressing #750.

Allow for overriding media features. 

Related specs: 
* https://www.w3.org/TR/mediaqueries-4
* https://drafts.csswg.org/mediaqueries-5

I added a CSS spec patch until the CSS PR https://github.com/w3c/csswg-drafts/pull/13549 is merged.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/1076.html" title="Last updated on Jul 28, 2026, 1:19 PM UTC (442ae9c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/1076/a7b8b67...442ae9c.html" title="Last updated on Jul 28, 2026, 1:19 PM UTC (442ae9c)">Diff</a>


index.bs
```diff
@@ -6475,7 +6479,101 @@ The [=remote end steps=] with |command parameters| are:
...
+        params: emulation.SetMediaFeaturesOverrideParameters
+      )
+
+      emulation.SetMediaFeaturesOverrideParameters = {
+        features: emulation.MediaFeatures / null,
+        ? contexts: [+browsingContext.BrowsingContext],
+        ? userContexts: [+browser.UserContext],
+      }
+
+      emulation.MediaFeatures = {
```
OrKoN: I am not a fan of this API as it requires changing WebDriver BiDi spec whenever there is a new media feature available in the underlying browser.

OrKoN (approved):

index.bs
OrKoN: Since media query features generally support unknown values, e.g., in CSS rules or in the matchMedia API, I think it might be beneficial to just pass a list of media features to the browser without re-listing all possible feature names in the WebDriver BiDi's CDDL. WDYT @jgraham ?

index.bs
```diff
@@ -14989,6 +15087,27 @@ Insert the following steps at the start of the [=determine the device pixel rati
 
 1. If [=device pixel ratio overrides=] [=map/contains=] <var ignore>window</var>'s [=window/navigable=], return [=device pixel ratio overrides=][<var ignore>window</var>'s [=window/navigable=]].
 
+### Evaluating Media Queries ### {#patchs-evaluating-media-queries}
+
+Issue: Remove after https://github.com/w3c/csswg-drafts/pull/13549 is merged.
+
+The [<dfn export>evaluate media feature</dfn> algorithm is modified to support WebDriver BiDi
```
lutien:
```suggestion
The <dfn export>evaluate media feature</dfn> algorithm is modified to support WebDriver BiDi
```

index.bs
```diff
@@ -6475,7 +6479,101 @@ The [=remote end steps=] with |command parameters| are:
...
+      )
+
+      emulation.SetMediaFeaturesOverrideParameters = {
+        features: emulation.MediaFeatures / null,
+        ? contexts: [+browsingContext.BrowsingContext],
+        ? userContexts: [+browser.UserContext],
+      }
+
+      emulation.MediaFeatures = {
+        ? "color": js-uint,
```
lutien: Hmm... so as far as I can see, there is no way to reset one media feature override? I think we should support it.

sadym-chromium: > Hmm... so as far as I can see, there is no way to reset one media feature override?

~correct. Allowing for reseting media features one-by-one would require for non-trivial logic of getting actual emulation configuration, as all the overrides (per navigable, user context and global) would have to be merged.~

> I think we should support it.

~I don't think so, but it can be done as a follow-up, as it would be an incremental non-breaking change.~

**UPD:** @lutien I added a way to set and remove media feature emulation one-by-one or remove them all together.

index.bs
```diff
@@ -6475,7 +6479,132 @@ The [=remote end steps=] with |command parameters| are:
...
+        features: emulation.MediaFeatures / null,
+        ? contexts: [+browsingContext.BrowsingContext],
+        ? userContexts: [+browser.UserContext],
+      }
+
+      emulation.MediaFeatures = {
+        ? "color": js-uint / null,
+        ? "color-index": js-uint / null,
+        ? "monochrome": js-uint / null,
+        ? "scan": "interlace" / "progressive" / null,
```
lutien: I couldn't find the information about the `scan` feature. Is it something Chrome-specific?

index.bs
```diff
@@ -6475,7 +6479,132 @@ The [=remote end steps=] with |command parameters| are:
...
+        ? "pointer": "none" / "coarse" / "fine" / null,
+        ? "hover": "none" / "hover" / null,
+        ? "any-pointer": "none" / "coarse" / "fine" / null,
+        ? "any-hover": "none" / "hover" / null,
+        ? "prefers-color-scheme": "light" / "dark" / null,
+        ? "prefers-contrast": "no-preference" / "more" / "less" / "custom" / null,
+        ? "prefers-reduced-motion": "no-preference" / "reduce" / null,
+        ? "prefers-reduced-transparency": "no-preference" / "reduce" / null,
+        ? "prefers-reduced-data": "no-preference" / "reduce" / null,
+        ? "display-mode": "fullscreen" / "standalone" / "minimal-ui" / "browser" / "picture-in-picture" / null,
```
lutien: Looks like there is also [window-controls-overlay](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@media/display-mode#window-controls-overlay). Could you also sort the options alphabetically here? It's then easier to make sure we have everything listed.

index.bs
```diff
@@ -6475,7 +6479,132 @@ The [=remote end steps=] with |command parameters| are:
...
+        ? "monochrome": js-uint / null,
+        ? "scan": "interlace" / "progressive" / null,
+        ? "grid": 0 / 1 / null,
+        ? "update": "none" / "slow" / "fast" / null,
+        ? "overflow-block": "none" / "scroll" / "paged" / null,
+        ? "overflow-inline": "none" / "scroll" / null,
+        ? "color-gamut": "srgb" / "p3" / "rec2020" / null,
+        ? "pointer": "none" / "coarse" / "fine" / null,
+        ? "hover": "none" / "hover" / null,
+        ? "any-pointer": "none" / "coarse" / "fine" / null,
```
lutien: I think it would be better to sort features alphabetically to verify if we have/miss some features.

index.bs
sadym-chromium: https://www.w3.org/TR/mediaqueries-4/#scan

index.bs
sadym-chromium: `window-controls-overlay` is not part of the specification https://www.w3.org/TR/mediaqueries-4/#media-descriptor-table

lutien: Ah, ok, I shouldn't have trusted MDN.

sadym-chromium: I updated the features with https://drafts.csswg.org/mediaqueries-5 as well

index.bs

index.bs
```diff
@@ -6475,7 +6479,142 @@ The [=remote end steps=] with |command parameters| are:
...
+        ? "display-mode": "fullscreen" / "standalone" / "minimal-ui" / "browser" / "picture-in-picture" / null,
+        ? "dynamic-range": "standard" / "high" / null,
+        ? "environment-blending": "opaque" / "additive" / "subtractive" / null,
+        ? "forced-colors": "none" / "active" / null,
+        ? "grid": 0 / 1 / null,
+        ? "horizontal-viewport-segments": js-uint / null,
+        ? "hover": "none" / "hover" / null,
+        ? "inverted-colors": "none" / "inverted" / null,
+        ? "monochrome": js-uint / null,
+        ? "nav-controls": "back" / null,
```
lutien: It looks like this one should also have the `none` option: https://drafts.csswg.org/mediaqueries-5/#descdef-media-nav-controls

lutien (approved):
Thanks! I think it looks fine now

index.bs
sadym-chromium: @lutien WDYT about this proposal?

lutien: I don't think we can ensure interoperability between the browsers if we don't specify the supported media features here, and we have to write wpt tests anyway for every feature.

index.bs
```diff
@@ -327,6 +327,8 @@ spec: MEDIAQUERIES4; urlPrefix: https://drafts.csswg.org/mediaqueries-4/
   type: dfn
     text: resolution media feature; url: #resolution
     text: media type; url: #media-type
+    text: media feature; url: #media-features
+    text: mf-name; url: #typedef-mf-name
```
tidoust: The spec already exports these terms although, for `<mf-name>`, the correct way to reference the term is going to be with the `<<mf-name>>` shorthand syntax.

```suggestion
```

(FWIW, the spec also exports "media type", but it's been introduced before this PR)

index.bs
```diff
@@ -14989,6 +15128,27 @@ Insert the following steps at the start of the [=determine the device pixel rati
...
+
+Issue: Remove after https://github.com/w3c/csswg-drafts/pull/13549 is merged.
+
+The <dfn export>evaluate media feature</dfn> algorithm is modified to support WebDriver BiDi
+overrides. Replace the step "The result is the result of evaluating the specified media feature" to:
+
+<div algorithm="evaluate media feature">
+To get the result of evaluating a [=media feature=] |media feature| in the given [=/Document=] |document|:
+
+1. Let |media feature name| be the |media feature|'s [=mf-name=].
```
tidoust, line 15386:
```suggestion
1. Let |media feature name| be the |media feature|'s <<mf-name>>.
```

Note: this will render as `<mf-name>`, but that matches how the CSS spec defines it.

lutien:
I thought we agreed to have a list of supported media features as possible parameters. Looks like the PR got updated after being accepted to just accept any string for a media feature name and value.

sadym-chromium:
ouch. My bad, rebased in the wrong way.

sadym-chromium:
> I thought we agreed to have a list of supported media features as possible parameters. Looks like the PR got updated after being accepted to just accept any string for a media feature name and value.

https://github.com/w3c/webdriver-bidi/pull/1149

PR #1087, opened 2026-02-26
Do not allow for forcing classic scrollbars

sadym-chromium:
In some cases it is non-trivial and meaningless to try to emulate classic scrollbar. E.g. on mobile device, while forcing overlay scrollbar instead of the classic one is very common use-case. This PR restricts the emulation to only force using overlay scrollbars.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/1087.html" title="Last updated on Feb 26, 2026, 2:18 PM UTC (1651933)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/1087/4883e50...1651933.html" title="Last updated on Feb 26, 2026, 2:18 PM UTC (1651933)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -274,7 +274,6 @@ spec: RFC4648; urlPrefix: https://datatracker.ietf.org/doc/html/rfc4648
     text: Base64 Encode; url: section-4
 spec: CSS-OVERFLOW-3; urlPrefix: https://drafts.csswg.org/css-overflow-3/
   type: dfn
-    text: classic scrollbars; url: #classic-scrollbars
     text: overlay scrollbars; url: #overlay-scrollbars
 spec: CSS-VALUES-3; urlPrefix: https://drafts.csswg.org/css-values-3/
   type: dfn
@@ -7021,7 +7020,7 @@ scrollbar type on the given top-level traversables, user contexts or globally.
       )
 
       emulation.SetScrollbarTypeOverrideParameters = {
-        scrollbarType: "classic" / "overlay" / null,
+        scrollbarType: "default" / "overlay" / null,
         ? contexts: [+browsingContext.BrowsingContext],
         ? userContexts: [+browser.UserContext],
       }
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -7021,7 +7020,7 @@ scrollbar type on the given top-level traversables, user contexts or globally.
       )
 
       emulation.SetScrollbarTypeOverrideParameters = {
-        scrollbarType: "classic" / "overlay" / null,
+        scrollbarType: "default" / "overlay" / null,
```
OrKoN, line 7023: this does not look editorial

gsnedders (changes requested):
So on macOS, where overlay scrollbars are (more-or-less) the default, we now have `emulation.SetScrollbarTypeOverrideParameters.scrollbarType` with the values: "default" (overlay), "overlay", or null (the default, so overlay).

This fundamentally does not work for WPT where large numbers of tests rely on classic scrollbars.

This is far from the only case with emulation where we're going to run into "the endpoint node doesn't support this", and in almost every case we should just be returning an error where it isn't supported.

Endpoints which only support "overlay" scrollbars should return an error when you try and set it; endpoints which only support "classic" scrollbars should return an error when you try and set it.

index.bs
sadym-chromium: agree. Removed `editorial` mark

sadym-chromium:
> So on macOS, where overlay scrollbars are (more-or-less) the default, we now have `emulation.SetScrollbarTypeOverrideParameters.scrollbarType` with the values: "default" (overlay), "overlay", or null (the default, so overlay).
> 
> This fundamentally does not work for WPT where large numbers of tests rely on classic scrollbars.
> 
> This is far from the only case with emulation where we're going to run into "the endpoint node doesn't support this", and in almost every case we should just be returning an error where it isn't supported.
> 
> Endpoints which only support "overlay" scrollbars should return an error when you try and set it; endpoints which only support "classic" scrollbars should return an error when you try and set it.

Hi @gsnedders! Thanks for your comment. I see the motivation, but I'm not sure if I understood your proposal. Which of the following options seems reasonable for you?
1. (current) `classic` / `overlay` / null
2. (proposed) `default` / `overlay` / null
3. `default` / `classic` / `overlay` / null

sadym-chromium:
@gsnedders could you please clarify your concern?

whimboo:
@gsnedders could you please take a look? Thanks.

whimboo:
https://github.com/w3c/webdriver-bidi/issues/692 is the related issue.

PR #713, opened 2024-05-21
Update `browsingContext.locateNodes`

sadym-chromium:
Do not decrease max depth when traversing document or document fragment

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/713.html" title="Last updated on May 21, 2024, 6:53 AM UTC (d6ff4d6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/713/34104f2...d6ff4d6.html" title="Last updated on May 21, 2024, 6:53 AM UTC (d6ff4d6)">Diff</a>

PR #1090, opened 2026-03-03
browsingContext.locateNodes should use the navigable's active document as default if no start nodes are specified

whimboo:
Compared to WebDriver classic using the "browsingContext.locateNodes" command for BiDi to retrieve the HTML (documentElement) element of a page doesn't return any value. This is because [the WebDriver BiDi spec defaults to the documentElement](https://w3c.github.io/webdriver-bidi/#cddl-value-browsingcontextlocatenodes-method-browsingcontextlocatenodes:~:text=If%20start%20nodes%20parameter%20is%20null%2C%20append%20the%20document%20element) already in case no context node is specified.

To fix that we should default to the navigable's active document instead.

***
<a href="https://pr-preview.s3.amazonaws.com/whimboo/webdriver-bidi/pull/1090.html" title="Last updated on Mar 3, 2026, 8:46 AM UTC (ef65889)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/1090/e3c596d...whimboo:ef65889.html" title="Last updated on Mar 3, 2026, 8:46 AM UTC (ef65889)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4631,8 +4631,8 @@ The [=remote end steps=] with |session| and |command parameters| are:
 
 1. Let |context nodes| be an empty [=/list=].
 
-1. If |start nodes parameter| is null, [=list/append=] the [=document element=]
-   of |navigable|'s [=active document=] to |context nodes|. Otherwise, for each
+1. If |start nodes parameter| is null, [=list/append=] the |navigable|'s
+   [=active document=] to |context nodes|. Otherwise, for each
    |serialized start node| in |start nodes parameter|:
 
    1. Let |start node| be the result of [=trying=] to [=deserialize shared reference=] given


whimboo:
I'm going to add tests via https://bugzilla.mozilla.org/show_bug.cgi?id=2020578.

OrKoN:
I think it might be overlapping with this change https://github.com/w3c/webdriver-bidi/pull/713/changes that should already account for the case of searching for documentElement?

whimboo:
> I think it might be overlapping with this change https://github.com/w3c/webdriver-bidi/pull/713/changes that should already account for the case of searching for documentElement?

These changes are only part of the `innerText` locator type and as such not affecting the `css` locator` which is broken and not able to find the root element. This would be similar to calling `document.documentElement.querySelector(":root")` which returns `undefined`.

PR #1094, opened 2026-03-08
rename contexts to contextInfoList in GetTreeResult

### LINKED ISSUE #82: Confusing naming. `context: string` and `contexts: object`
`context` is used both as an object: `contexts: [*BrowsingContextInfo]` implying `context` is a `BrowsingContextInfo`, and as an ID string: `context: BrowsingContext` specified as an ID string. This twofold interpretation makes ыщьу confusion. 

The response to the `browsingContext.getTree` command should look like:
```
...
    "contexts": [{
        "context": "3BD893F38972478DB095B393E8C3AFE7",
        "parent": null,
        "url": "about:blank",
        "type": "page"
      }]
...
```

While reading, I tend to misinterpret `BrowsingContextInfo` as a `context`, because of the property name `contexts`. 

We should stick to the consistent naming and either renaming string `context` to `contextID` or array `contexts` to `contextInfos`.

ojuschugh1:
This PR Fixes https://github.com/w3c/webdriver-bidi/issues/82

***
<a href="https://pr-preview.s3.amazonaws.com/ojuschugh1/webdriver-bidi/pull/1094.html" title="Last updated on Mar 9, 2026, 5:47 PM UTC (cca48cd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/1094/43236c6...ojuschugh1:cca48cd.html" title="Last updated on Mar 9, 2026, 5:47 PM UTC (cca48cd)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4248,7 +4248,7 @@ or all top-level contexts when no parent is provided.
    <dd>
     <pre class="cddl" data-cddl-module="local-cddl">
         browsingContext.GetTreeResult = {
-          contexts: browsingContext.InfoList
+          contextInfoList: browsingContext.InfoList
         }
     </pre>
    </dd>
@@ -4279,7 +4279,7 @@ The [=remote end steps=] with <var ignore>session</var> and |command parameters|
   1. Append |info| to |navigables infos|
 
 1. Let |body| be a [=/map=] matching the <code>browsingContext.GetTreeResult</code>
-   production, with the <code>contexts</code> field set to |navigables infos|.
+   production, with the <code>contextInfoList</code> field set to |navigables infos|.
 
 1. Return [=success=] with data |body|.
 


whimboo:
Thank you for the PR! It's sad that the filed issue got under the wire in all the last years. Nowadays changing it is quite a breaking change and I don't think that this is something that we can do - or is worth doing. Note that all supported browsers would have to make the change, and each existent client needs to be adapted + we have to make it backward compatible.

Instead what we might be able to do is changing internal names - means types which are internally resolved and do not cause a change in the payload of the returned value. But it wont work here.

@jgraham what do you think? I would propose to not make any change like that.

jgraham:
Yes, agreed. For better or worse I don't think we can change anything here.

PR #1105, opened 2026-03-21
feat: add accuracyMode to emulation.setGeolocationOverride

alvinjiooo:
This pull request adds an optional `accuracyMode` parameter to the `emulation.setGeolocationOverride` command.

This aligns WebDriver BiDi with the upcoming changes to the [W3C Geolocation API](https://github.com/w3c/geolocation/pull/195), which introduces support for approximate geolocation via the `accuracyMode` attribute on the `GeolocationPosition` interface.

***
<a href="https://pr-preview.s3.amazonaws.com/alvinjiooo/webdriver-bidi/pull/1105.html" title="Last updated on Apr 13, 2026, 7:27 AM UTC (b9a46f8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/1105/0e36053...alvinjiooo:b9a46f8.html" title="Last updated on Apr 13, 2026, 7:27 AM UTC (b9a46f8)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6288,6 +6288,7 @@ The <dfn export for=commands>emulation.setGeolocationOverride</dfn> command modi
          latitude: -90.0..90.0,
          longitude: -180.0..180.0,
          ? accuracy: (float .ge 0.0) .default 1.0,
+         ? accuracyMode: "approximate" / "precise",
          ? altitude: float / null .default null,
          ? altitudeAccuracy: (float .ge 0.0) / null .default null,
          ? heading: (0.0...360.0) / null .default null,
@@ -6307,14 +6308,7 @@ The <dfn export for=commands>emulation.setGeolocationOverride</dfn> command modi
    </dd>
 </dl>
 
-A <dfn>geolocation override</dfn> is a [=struct=] with:
-* [=struct/item=] named <dfn attribute for="geolocation override">latitude</dfn> which is a float;
-* [=struct/item=] named <dfn attribute for="geolocation override">longitude</dfn> which is a float;
-* [=struct/item=] named <dfn attribute for="geolocation override">accuracy</dfn> which is a float;
-* [=struct/item=] named <dfn attribute for="geolocation override">altitude</dfn> which is a float or null;
-* [=struct/item=] named <dfn attribute for="geolocation override">altitudeAccuracy</dfn> which is a float or null;
-* [=struct/item=] named <dfn attribute for="geolocation override">heading</dfn> which is a float or null;
-* [=struct/item=] named <dfn attribute for="geolocation override">speed</dfn> which is a float or null.
+A <dfn>geolocation override</dfn> is a [=/map=].
 
 A [=remote end=] has a <dfn>geolocation override configuration</dfn>, which is
 [=WebDriver configuration=] with [=WebDriver configuration/associated type=] [=geolocation override=].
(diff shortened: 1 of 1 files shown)


alvinjiooo:
Hi @OrKoN,
May I have your help to review this PR? 
Thanks!
Alvin

index.bs
```diff
@@ -6280,6 +6280,7 @@ The <dfn export for=commands>emulation.setGeolocationOverride</dfn> command modi
           (coordinates: emulation.GeolocationCoordinates / null) //
           (error: emulation.GeolocationPositionError)
         ),
+        ? accuracyMode: "precise" / "approximate",
```
whimboo: Does that also apply to errors? If not we should add this field to `GeolocationCoordinates`.
```suggestion
        ? accuracyMode: "approximate", "precise",
```

index.bs
```diff
@@ -6359,6 +6361,15 @@ The [=remote end steps=] with |command parameters| are:
...
    [=WebDriver configuration/unset=].
 
+1. If |emulated position data| is not [=WebDriver configuration/unset=]:
+
+   1. Let |accuracy mode| be "<code>precise</code>".
+
+   1. If |command parameters| [=map/contains=] "<code>accuracyMode</code>", set |accuracy mode| to
+      |command parameters|["<code>accuracyMode</code>"].
+
+   1. [=map/set|Set=] |emulated position data|["<code>accuracyMode</code>"] to |accuracy mode|.
```
whimboo: I don't think that we need all the new additions at all. Per CDDL spec we allow only the two values, which we then pass-through. And if the field is not specified shouldn't the [`set emulated position data`](https://www.w3.org/TR/geolocation/#dfn-set-emulated-position-data) algorithm from the Geolocation spec handle the default value?

whimboo (changes requested):

index.bs
alvinjiooo: Regarding the placement of `accuracyMode`: While we could move it into `GeolocationCoordinates`, it is intended to describe the operating mode of the browser's geolocation service (based on the website's request and the user's permission state) rather than a property of the coordinates themselves. In the [underlying Geolocation spec PR #195](https://github.com/w3c/geolocation/pull/195), `accuracyMode` is defined on the `GeolocationPosition` interface at the same level as `coords` (`GeolocationCoordinates`). 

And yes, `accuracyMode` should be paired only with `coordinates`,  I wonder would it work if I move it to the group with `coordinates` like:

```cddl
  (
    (
      coordinates: emulation.GeolocationCoordinates / null,
      ? accuracyMode: "precise" / "approximate"
    ) //
  ),
}
```

Will this break existing tooling and tests which already using setGeolocationOverride?

index.bs
alvinjiooo: I have this covered in https://github.com/w3c/webdriver-bidi/pull/1105#discussion_r2992266577

index.bs
alvinjiooo: I actually update the PR further for `emulated position data` to either `emulation.GeolocationCoordinates` or `GeolocationPositionError`. I think it probably will be better to pass  `emulated position data` as a map with {coordinates:..., accuracyMode?: ... } or {error: {code: ...}}, so the Geolocation side can decide what to created at their side. I also have Geolocation [PR#195](https://pr-preview.s3.amazonaws.com/alvinjiooo/geolocation/pull/195.html#acquire-a-position) (at step 5.5 of Acquire a position algorithm) updated to hook this up. 
PTAL at latest version and thanks!

whimboo: > Will this break existing tooling and tests which already using setGeolocationOverride?

This should work fine given that this is a new optional parameter. Older versions of browsers will just ignore it as long as the handling of the parameter is not supported.

whimboo: Hm, I don't see the proposed changes from your second last comment yet. `accuracyMode` should be part of `emulation.GeolocationCoordinates` given that if `coordinates` is `null` we won't need this parameter's value.

alvinjiooo: Hi @whimboo, 
Sorry for the confusion, my original plan was to add `accuracyMode` outside of `GeolocationCoordinates`, but after discussing with https://github.com/w3c/geolocation/pull/195 reviewer we think it seems also feasible to add new `accuracyMode` into `GeolocationCoordinates`. So I updated the CDDL and `remote end steps` for the new proposal (add new `accuracyMode` into `GeolocationCoordinates`). Thanks!

index.bs
alvinjiooo: These steps has been updated. Let's review the latest change and I am resolving this one.

index.bs
alvinjiooo: Hi @whimboo, 
Do you get a chance to review the latest PR? 
Thanks!

index.bs
```diff
@@ -6292,6 +6292,7 @@ The <dfn export for=commands>emulation.setGeolocationOverride</dfn> command modi
          ? altitudeAccuracy: (float .ge 0.0) / null .default null,
          ? heading: (0.0...360.0) / null .default null,
          ? speed: (float .ge 0.0) / null .default null,
+         ? accuracyMode: "approximate" / "precise",
```
```suggestion
? accuracy: (float .ge 0.0) .default 1.0,
? accuracyMode: "approximate" / "precise",
? altitude: float / null .default null,
? altitudeAccuracy: (float .ge 0.0) / null .default null,
? heading: (0.0...360.0) / null .default null,
? speed: (float .ge 0.0) / null .default null,
```

index.bs
```diff
@@ -6307,14 +6308,7 @@ The <dfn export for=commands>emulation.setGeolocationOverride</dfn> command modi
...
 
-A <dfn>geolocation override</dfn> is a [=struct=] with:
-* [=struct/item=] named <dfn attribute for="geolocation override">latitude</dfn> which is a float;
-* [=struct/item=] named <dfn attribute for="geolocation override">longitude</dfn> which is a float;
-* [=struct/item=] named <dfn attribute for="geolocation override">accuracy</dfn> which is a float;
-* [=struct/item=] named <dfn attribute for="geolocation override">altitude</dfn> which is a float or null;
-* [=struct/item=] named <dfn attribute for="geolocation override">altitudeAccuracy</dfn> which is a float or null;
-* [=struct/item=] named <dfn attribute for="geolocation override">heading</dfn> which is a float or null;
-* [=struct/item=] named <dfn attribute for="geolocation override">speed</dfn> which is a float or null.
+A <dfn>geolocation override</dfn> is a [=/map=].
```
whimboo, line 6311: Why did you change all these lines? It's for the internal data structure. You need to get the accuracy mode added. Maybe this was caused by a merge conflict? The same applies to the changes below.

whimboo (changes requested):
Sorry, it looks like that I missed the emails from a week ago given that there was a public holiday and I was out as well for some days.

I think that one of our latest changes (https://github.com/w3c/webdriver-bidi/pull/1086) caused merge conflicts for you.

index.bs
alvinjiooo: Hi @whimboo,
Thanks for the review! The reason for changing the geolocation override from a struct to a map was intend to fix an existing type inconsistency in the spec regarding how we handle the error parameter. 
In the original code, geolocation override was defined as a strict struct containing only coordinate floats (latitude, longitude, etc.). However, in the algorithm steps for setGeolocationOverride, when an "error" is provided, the spec does this:

  > Let emulated position data be a map matching GeolocationPositionError production...

This created an inconsistency where the geolocation override configuration (which is typed as a geolocation override) was being forced to store a map representing a GeolocationPositionError, instead of the expected struct of coordinates. I believe changing internal definition of geolocation override to be a map, containing either GeolocationCoordinates or GeolocationPositionError, should resolve this conflict. So Geolocation side can check what does geolocation override contain and decide what to create. Please let me know what do you think about this. 

Thanks!

index.bs
alvinjiooo: Done

PR #1106, opened 2026-03-25
Use the URLPattern specification instead of custom URL pattern parsing

### LINKED ISSUE #919: Feature Request: Support UrlPattern for network interception
https://urlpattern.spec.whatwg.org

### Motivation
Currently to filter out network requests to be intercepted it happens on client side, meaning a client "enables" interception of all requests. It would be nice a client can specify "most interesting" requests to be intercepted.

whimboo:
Fixes #919.

This change aligns our implementation for parsing and matching URL patterns with the [URLPattern specification](https://urlpattern.spec.whatwg.org/), ensuring consistent and spec-compliant behavior across matching operations. It reduces maintenance overhead by removing custom parsing code from the WebDriver BiDi specification, and leverages the native URLPattern implementation for improved correctness and clarity.

Existing functionality is preserved, but edge cases may now follow the semantics defined by the specification.

Follow-up work will extend support for additional [`URLPatternInit`](https://urlpattern.spec.whatwg.org/#dictdef-urlpatterninit) fields that we do not support yet, and as well for handling the `baseURL`, and parse options like `ignoreCase`.

***
<a href="https://pr-preview.s3.amazonaws.com/whimboo/webdriver-bidi/pull/1106.html" title="Last updated on Jun 16, 2026, 1:27 PM UTC (ebf35c8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/1106/fcd2302...whimboo:ebf35c8.html" title="Last updated on Jun 16, 2026, 1:27 PM UTC (ebf35c8)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7841,7 +7841,7 @@ To <dfn>get the network intercepts</dfn> given |session|, |event|, |request|, an
 
     1. For each |url pattern| in |url patterns|:
 
-      1. If [=match URL pattern=] with |url pattern| and |url|:
+      1. If [=URL pattern/match|URL pattern match=] with |url pattern| and |url|:
 
         1. [=list/Append=] |intercept id| to |intercepts|.
 
@@ -8859,13 +8859,12 @@ network.UrlPattern = (
 
 network.UrlPatternPattern = {
     type: "pattern",
-    ?protocol: text,
-    ?hostname: text,
-    ?port: text,
-    ?pathname: text,
-    ?search: text,
+    ? protocol: text,
+    ? hostname: text,
+    ? port: text,
+    ? pathname: text,
+    ? search: text,
 }
-
 
 network.UrlPatternString = {
     type: "string",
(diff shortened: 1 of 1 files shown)


whimboo:
@juliandescottes could you maybe do the initial review?

index.bs
```diff
@@ -8875,296 +8879,50 @@ network.UrlPatternString = {
...
-
-   1. Let |url query| be |url|'s [=url/query=].
-
-   1. If |url query| is null, set |url query| to the empty string.
-
-   1. If |url pattern|'s search is not equal to |url query|, return false.
-
-1. Return true.
+1. Return the result of running [=URL pattern match=] given |url pattern| and
+   |url string|.
```
juliandescottes: Since we only use this in one spot, maybe we can get rid of the algorithm and invoke URL pattern match directly?

index.bs
```diff
@@ -8875,296 +8879,50 @@ network.UrlPatternString = {
...
+     |init|["<code>search</code>"] to |pattern|["<code>search</code>"].
 
-        1. Return [=error=] with [=error code=] [=invalid argument=].
-
-    1. Append |search| to |pattern url|.
-
-  1. Otherwise:
-
-    1. Set |has search| to false.
+  1. Let |url pattern| be the result of [=creating a URLPattern=] given |init|.
```
juliandescottes: `create a URLPattern` can throw a TypeError, so we should probably use `trying` here?

Then we need to handle the error case, probably by still sending an invalid argument error?

juliandescottes (changes requested):
Looks like a great simplification. Overall I think this is fine, but we need to handle the error case.

index.bs
whimboo: Absolutely. This separate algorithm is no longer needed. I'll combine.

index.bs
whimboo: We use `trying` when calling algorithms of our own spec given that in all these cases we set a `data` field for the return value. This is not the case when calling algorithms of external specifications. Search for `is failure` in the BiDi spec to see other usages of this style. So I think that this is correct.

juliandescottes: Ah yes, I missed the bullet point below. Looks fine thanks.

juliandescottes (approved):

whimboo:
@OrKoN could you please review the changes as well? Thanks!

whimboo:
> Follow-up work will extend support for additional [`URLPatternInit`](https://urlpattern.spec.whatwg.org/#dictdef-urlpatterninit) fields that we do not support yet, and as well for handling the `baseURL`, and parse options like `ignoreCase`. This change is intended as a strict 1:1 replacement of the existing functionality without introducing new behavior.

I filed https://github.com/w3c/webdriver-bidi/issues/1109 for this follow-up work.

OrKoN:
@whimboo do you already have an implementation? I wonder what the effect on existing WPT tests will be

index.bs
```diff
@@ -8875,296 +8879,42 @@ network.UrlPatternString = {
...
-  1. Let |pattern url| be the empty string.
-
-  1. If |pattern| [=map/contains=] "<code>protocol</code>":
-
-    1. If |pattern|["<code>protocol</code>"] is the empty string, return [=error=]
-       with [=error code=] [=invalid argument=].
-
-    1. Let |protocol| be the result of [=trying=] to [=unescape
-       URL Pattern=] with |pattern|["<code>protocol</code>"].
+  1. Let |init| be a new [=URLPatternInit=].
```
tidoust: It would seem better to use the IDL shorthand syntax here:

```suggestion
  1. Let |init| be a new {{URLPatternInit}}.
```

index.bs
```diff
@@ -337,6 +337,11 @@ spec: STREAMS; urlPrefix: https://streams.spec.whatwg.org/
 spec: TOUCH-EVENTS; urlPrefix: https://www.w3.org/community/reports/touchevents/CG-FINAL-touch-events-20240704/
   type: dfn
     text: expose legacy touch event APIs; url: #conditionally-exposing-legacy-touch-event-apis
+spec: URLPATTERN; urlPrefix: https://urlpattern.spec.whatwg.org/
+  type: dfn
+    text: creating a URLPattern; url: #url-pattern-create
+    text: URL pattern match; url: #url-pattern-match
+    text: URLPatternInit; url: #dictdef-urlpatterninit
```
tidoust: Shouldn't be needed if you use the IDL shorthand syntax (see other comment)


```suggestion
```

index.bs
```diff
@@ -337,6 +337,11 @@ spec: STREAMS; urlPrefix: https://streams.spec.whatwg.org/
 spec: TOUCH-EVENTS; urlPrefix: https://www.w3.org/community/reports/touchevents/CG-FINAL-touch-events-20240704/
   type: dfn
     text: expose legacy touch event APIs; url: #conditionally-exposing-legacy-touch-event-apis
+spec: URLPATTERN; urlPrefix: https://urlpattern.spec.whatwg.org/
+  type: dfn
+    text: creating a URLPattern; url: #url-pattern-create
+    text: URL pattern match; url: #url-pattern-match
```
tidoust: I note that these terms can be directly referenced through `[=URL pattern/create=]` and `[=URL pattern/map=]`, which would avoid the need to re-create local definitions for them with more brittle fragments.

Bikeshed would render the above references as "create" and "match". To label them directly, you may use something like `[=URL pattern/match|URL pattern match=]`.

whimboo:
> @whimboo do you already have an implementation? I wonder what the effect on existing WPT tests will be

@OrKoN, I haven't had when you were asking. But I spend some time today in updating our interception related code to check what will change. In general I noticed that several values are handled more permissively compared to our current implementation. The main differences are:

* Wildcards (*) now work consistently across all fields and act as the default when a field is not specified.
* When using a pattern init object, fields like search treat most special characters as literals (e.g. search: "search#", which gets encoded (e.g. search%23). This can lead to mismatches with URLs such as http://example.com/?search#, since URL parsing interprets # as a delimiter. The ? character will still cause a parser error.
* An empty pathname currently matches all schemes in our implementation, but with URLPattern, this is no longer the case. For special schemes, / must be used explicitly to achieve the same behavior.
* Invalid cases like https://example.com:port/ as a string pattern will now work because example.com:port is taken as value for the hostname only.

All the relevant test changes you can find in: https://bugzilla.mozilla.org/attachment.cgi?id=9568221

As given by Julian we expected breaking changes and from our side it looks ok. What do you think?

index.bs
whimboo: Oh wow, I didn't know that this works. Thanks for letting me know. I'll update all the different usages.

index.bs
whimboo: It would duplicate the description for `URL pattern/create` because it is used twice but that is probably fine. Thanks.

OrKoN:
I added a basic implementation here https://github.com/GoogleChromeLabs/chromium-bidi/pull/4138 and there quite a few WPT tests that are failing, I wonder if the list of failed tests matches what you have?

whimboo:
> I added a basic implementation here [GoogleChromeLabs/chromium-bidi#4138](https://github.com/GoogleChromeLabs/chromium-bidi/pull/4138) and there quite a few WPT tests that are failing, I wonder if the list of failed tests matches what you have?

There is actually only a single test which seems to perma fail and some others that fail intermittently. Here a job which shows 3 failing tests:

https://github.com/GoogleChromeLabs/chromium-bidi/actions/runs/24882696358/job/72854858369?pr=4138

```
FAILED tests/network/test_add_intercept.py::test_add_intercept_type_string_one_valid_and_one_invalid - AssertionError: Regex pattern did not match.
 Regex: '{\'error\': \'invalid argument\', \'message\': "Failed to construct \'URL\': Invalid URL \'foo\'"}'
 Input: '{\'error\': \'invalid argument\', \'message\': "Invalid URL pattern: Failed to construct \'URLPattern\': Relative constructor string \'foo\' must have a base URL passed as the second argument."}'
FAILED tests/network/test_add_intercept.py::test_add_intercept_type_pattern_protocol_empty_invalid - Failed: DID NOT RAISE <class 'Exception'>
FAILED tests/network/test_add_intercept.py::test_add_intercept_type_pattern_hostname_invalid[abc:com] - Failed: DID NOT RAISE <class 'Exception'>
= 3 failed, 136 passed, 1 skipped, 1 xfailed, 1 warning, 12 rerun in 85.89s (0:01:25) =
```

It looks like that I did not update this particular test file as part of my patch on https://phabricator.services.mozilla.com/D293101. I should take a look. Thanks for pointing out.

whimboo:
> ```
> FAILED tests/network/test_add_intercept.py::test_add_intercept_type_string_one_valid_and_one_invalid - AssertionError: Regex pattern did not match.
>  Regex: '{\'error\': \'invalid argument\', \'message\': "Failed to construct \'URL\': Invalid URL \'foo\'"}'
>  Input: '{\'error\': \'invalid argument\', \'message\': "Invalid URL pattern: Failed to construct \'URLPattern\': Relative constructor string \'foo\' must have a base URL passed as the second argument."}'
> FAILED tests/network/test_add_intercept.py::test_add_intercept_type_pattern_protocol_empty_invalid - Failed: DID NOT RAISE <class 'Exception'>
> FAILED tests/network/test_add_intercept.py::test_add_intercept_type_pattern_hostname_invalid[abc:com] - Failed: DID NOT RAISE <class 'Exception'>
> = 3 failed, 136 passed, 1 skipped, 1 xfailed, 1 warning, 12 rerun in 85.89s (0:01:25) =
> ```

@OrKoN these failing tests are not part of the [public wdspec tests](https://wpt.fyi/results/webdriver/tests/bidi/network/add_intercept/add_intercept.py?label=master&label=experimental&product=firefox&product=firefox_android&product=chrome&product=chrome_android&product=safari&view=subtest).

But nevertheless here a quick investigation:

* **test_add_intercept_type_string_one_valid_and_one_invalid** fails because its a relative URL and the Base URL is missing. Note that we have a follow-up issue for that as filed as https://github.com/w3c/webdriver-bidi/issues/1109.
* **test_add_intercept_type_pattern_protocol_empty_invalid** fails because it assumes that URLPattern should cause an error if "" for the protocol is specified. Note that `new URLPattern({protocol: "", hostname: "www.example.com", port: "80" })` doesn't raise in Chrome as well.
* **test_add_intercept_type_pattern_hostname_invalid[abc:com]** unexpectedly passes for you when it should fail. For us it fails correctly so I assume there is a bug in your implementation? I'm going to add this case to my patch which updates the tests.

With all that I assume it should not be blocked to get the changes landed?

whimboo:
@OrKoN over on https://phabricator.services.mozilla.com/D293101 I added a couple more tests to cover the additional features that we now allow to use by clients which URLPattern offers for us. It should apply to the upstream wpt repo (when you ignore the prefix in the path) so that you can run these tests with your implementation as well.

Please let me know how it looks like. Thanks!

whimboo:
One thing to keep in mind, and likely the most visible change for users, is that characters with special meaning in `URLPattern` syntax (`+`, `(`, `)`, `{`, `}`, `:`) must be escaped with `\` when used literally in string patterns. This will most commonly affect `+` characters in query strings, such as those resulting from URL-encoded spaces.

whimboo:
I had to push some smaller updates based on our internal review and that I saw that no note was added describing the new behavior + issue for the follow-up work.

@juliandescottes please review again. Thanks

index.bs
```diff
@@ -9157,312 +8859,70 @@ network.UrlPattern = (
...
+defined by [[!URLPattern]].
 
-</div>
+Note: The component fields of <code>network.UrlPatternPattern</code> and the
+<code>pattern</code> field of <code>network.UrlPatternString</code> are parsed
+using [[!URLPattern]] syntax. Characters such as <code>*</code>,
+<code>+</code>, <code>?</code>, <code>:</code>, and
+<code>{</code>...<code>}</code> have special meaning as wildcards, modifiers,
+named groups, and grouping constructs respectively. To match these characters
+literally, escape them with a backslash (e.g. <code>\\+</code>).
```
juliandescottes: Do we need one or two backslashes to escape the character? The example makes the description a bit confusing.

index.bs
```diff
@@ -9157,312 +8859,70 @@ network.UrlPattern = (
...
-
-        1. Return [=error=] with [=error code=] [=invalid argument=].
-
-    1. Append |search| to |pattern url|.
-
-  1. Otherwise:
-
-    1. Set |has search| to false.
+  1. Let |url pattern| be the result of
+     [=URL pattern/create|creating a URLPattern=] given |init|. If this
```
juliandescottes: Looking at the URLPattern spec, I feel like we should properly mention the two other arguments to the create algorithm: 

`given |init|, null and and an empty [=/map=].`

?

index.bs
```diff
@@ -9157,312 +8859,70 @@ network.UrlPattern = (
...
-     1. Otherwise:
-
-       1. If |url|'s [=url/scheme=] [=is special=] and |url|'s [=url/scheme=]'s
-          [=default port=] is not null, and |url|'s [=url/port=] is null or is
-          equal to [=url/scheme=]'s [=default port=]:
-
-         1. The empty string.
-
-       1. Otherwise, if |url|'s [=url/port=] is not null:
+  1. Let |url pattern| be the result of [=URL pattern/create|creating a URLPattern=]
```
juliandescottes, line 8921: Same suggestion about the arguments passed to `URL pattern/create`, I think all 3 are mandatory.

juliandescottes (approved):

index.bs
whimboo: Sorry, just one backslash for sure. And I think that we can remove the example.

index.bs
whimboo: Indeed. Good catch. I'll fix it.

whimboo:
@jgraham can you please review as well? Thanks

index.bs
whimboo: FYI handling the error as thrown by the URLPattern constructor cannot be done via `is failure` but requires `if this throws an exception ...`. I've corrected that in my last uploaded revision.

index.bs
```diff
@@ -9157,312 +8859,71 @@ network.UrlPattern = (
...
-specification. These can be escaped by prefixing them with a U+005C (\) character.
-
-<div algorithm>
-To <dfn>unescape URL pattern</dfn> given |pattern|
-
-1. Let |forbidden characters| be the [=/set=] of codepoints «U+0028 ((), U+0029
-   ()), U+002A (*), U+007B ({), U+007D (})»
-
-1. Let |result| be the empty string.
+Issue(1109): Add support for additional {{URLPatternInit}} fields (e.g.
```
jgraham, line 8876: Although this is "cheap" in that we're just passing in extra options, I'm not quite convinced that they all make sense for WebDriver use cases.

index.bs
```diff
@@ -9157,312 +8859,71 @@ network.UrlPattern = (
...
-    1. If |scheme| is not "<code>file</code>", append "<code>placeholder</code>" to
-       |pattern url|.
-
-    1. Set |has hostname| to false.
-
-  1. If |pattern| [=map/contains=] "<code>port</code>":
-
-    1. If |pattern|["<code>port</code>"] is the empty string, return [=error=]
-       with [=error code=] [=invalid argument=].
+  1. Let |init| be a new {{URLPatternInit}}.
```
jgraham, line 8897: Technically you can't really do this; you're trying to create a JS object outside a realm. The URLPattern specification accepts an infra Map object as URLPatternInit in cases where it's not running in a JS realm. I think you want to use the entry point at https://urlpattern.spec.whatwg.org/#build-a-url-pattern-from-an-infra-value.

jgraham (changes requested):

index.bs
```diff
@@ -8139,7 +7841,7 @@ To <dfn>get the network intercepts</dfn> given |session|, |event|, |request|, an
 
     1. For each |url pattern| in |url patterns|:
 
-      1. If [=match URL pattern=] with |url pattern| and |url|:
+      1. If [=URL pattern/match|URL pattern match=] with |url pattern| and |url|:
```
OrKoN: does this return a boolean? I think it returns null if there is no match.

index.bs
```diff
@@ -9157,312 +8859,71 @@ network.UrlPattern = (
...
-
-        1. Return [=error=] with [=error code=] [=invalid argument=].
-
-    1. Append |search| to |pattern url|.
-
-  1. Otherwise:
-
-    1. Set |has search| to false.
+  1. Let |url pattern| be the result of
+     [=URL pattern/create|creating a URLPattern=] given |init|, null, and an
```
OrKoN: > Section [Using URL patterns in other specifications](https://urlpattern.spec.whatwg.org/#other-specs) of `[[!URLPattern]]` explicitly advises:
> *"3. **Be clear about regexp groups**. Some APIs may benefit from only allowing URL patterns which do not [=URL pattern/has regexp groups|have regexp groups=], for example, because user agents are likely to implement them in a different thread or process from those executing author script, and because of security or performance concerns, a JavaScript engine would not ordinarily run there. If so, this should be clearly documented (with reference to [=URL pattern/has regexp groups=]) and the operation should report an error as soon as possible..."*

> By delegating directly to `[=URL pattern/create=]` without checking `[=URL pattern/has regexp groups=]`, `network.addIntercept` now requires all WebDriver BiDi remote ends to evaluate arbitrary ECMAScript regular expression groups (`v`-mode `RegExpBuiltinExec`, e.g. `pathname: "/api/:id(\\d+)"`) in their network interception layer. If user agents implement network interception in a network process or non-JS layer (e.g. Rust/C++ URLPattern matcher without an embedded JS RegExp engine), supporting arbitrary custom JS regexps may be undesirable or require rejecting patterns where `[=URL pattern/has regexp groups=]` is `true` with `[=invalid argument=]`. This design decision should be explicitly confirmed and documented.

Do we want to support regexp groups?

PR #1113, opened 2026-04-23
Alternaitve proposal for screencasting commands without JS primitives

lutien:
An attempt to specify screencasting algorithms without JS primitives to not require running JS.

Only the last commit is different from #1069.

***
<a href="https://pr-preview.s3.amazonaws.com/lutien/webdriver-bidi/pull/1113.html" title="Last updated on May 22, 2026, 8:44 AM UTC (18c3740)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/1113/e9b871c...lutien:18c3740.html" title="Last updated on May 22, 2026, 8:44 AM UTC (18c3740)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3258,7 +3258,7 @@ A [=BiDi session=] has a <dfn>screencast recordings map</dfn> which is a [=/map=
 which the keys are [[!RFC9562|UUID]]s, and the values are [=structs=] with
 an [=struct/item=] named <code>mediaRecorder</code>, which is a {{MediaRecorder}},
 an [=struct/item=] named <code>mimeType</code>, which is a string,
-an [=struct/item=] named <code>recordedChunks</code>, which is a list,
+an [=struct/item=] named <code>recordedChunks</code>, which is a [=/list=] of {{BlobEvent/data}},
 an [=struct/item=] named <code>context</code>, which is a navigable.
 
 ### Types ### {#module-browsingcontext-types}
--- a/index.bs
+++ b/index.bs
@@ -3262,11 +3262,11 @@ map between [=/navigables=] or [=user context|user contexts=] and boolean.
 
 A [=BiDi session=] has a <dfn>screencast recordings map</dfn> which is a [=/map=] in
 which the keys are [[!RFC9562|UUID]]s, and the values are [=structs=] with
-an [=struct/item=] named <code>mediaRecorder</code>, which is a {{MediaRecorder}},
-an [=struct/item=] named <code>mimeType</code>, which is a string,
-an [=struct/item=] named <code>path</code>, which is a string,
-an [=struct/item=] named <code>size</code>, which is a number,
-an [=struct/item=] named <code>context</code>, which is a navigable.
+an [=struct/item=] named <dfn attribute for="screencast recordings map">mediaRecorder</dfn>, which is a {{MediaRecorder}},
+an [=struct/item=] named <dfn attribute for="screencast recordings map">mimeType</dfn>, which is a string,
+an [=struct/item=] named <dfn attribute for="screencast recordings map">path</dfn>, which is a string,
+an [=struct/item=] named <dfn attribute for="screencast recordings map">size</dfn>, which is a number,
+an [=struct/item=] named <dfn attribute for="screencast recordings map">context</dfn>, which is a navigable.
 
 ### Types ### {#module-browsingcontext-types}
 
--- a/index.bs
+++ b/index.bs
@@ -329,9 +329,9 @@ spec: MEDIACAPTURE-RECORD; urlPrefix: https://w3c.github.io/mediacapture-record/
     text: is type supported; url: #abstract-opdef-is-type-supported
     text: MediaRecorder; url: #mediarecorder-api
     text: MediaRecorderOptions; url: #mediarecorderoptions-section
-spec: MEDIACAPTURE-SCREEN-SHARE; urlPrefix: https://w3c.github.io/mediacapture-screen-share/
+spec: MEDIACAPTURE-VIEWPORT; urlPrefix: https://w3c.github.io/mediacapture-viewport/
   type: dfn
-    text: capture a browser tab; url: #dfn-capture-a-browser-tab
+    text: capture a browsing context viewport; url: #dfn-capture-a-browsing-context-viewport
 spec: MEDIAQUERIES4; urlPrefix: https://drafts.csswg.org/mediaqueries-4/
   type: dfn
     text: resolution media feature; url: #resolution
@@ -5179,7 +5179,7 @@ The [=remote end steps=] with |command parameters| are:
 1. Let <var>realm</var> be the result of [=trying=] to [=get a realm from a navigable=]
    with |navigable id| and null.
 
-1. Let |promise| be the result of [=trying=] to [=capture a browser tab=] with
+1. Let |promise| be the result of [=trying=] to [=capture a browsing context viewport=] with
    |navigable| and |command parameters|["<code>streamOptions</code>"].
 
 1. Let |media stream| be <a spec=ECMASCRIPT>Await</a>(|promise|).
--- a/index.bs
+++ b/index.bs
@@ -3258,7 +3258,8 @@ A [=BiDi session=] has a <dfn>screencast recordings map</dfn> which is a [=/map=
 which the keys are [[!RFC9562|UUID]]s, and the values are [=structs=] with
 an [=struct/item=] named <code>mediaRecorder</code>, which is a {{MediaRecorder}},
 an [=struct/item=] named <code>mimeType</code>, which is a string,
-an [=struct/item=] named <code>recordedChunks</code>, which is a [=/list=] of {{BlobEvent/data}},
+an [=struct/item=] named <code>path</code>, which is a string,
+an [=struct/item=] named <code>size</code>, which is a number,
 an [=struct/item=] named <code>context</code>, which is a navigable.
 
 ### Types ### {#module-browsingcontext-types}
@@ -5149,7 +5150,8 @@ starts the screencast of a given navigable according to provided options.
    <dd>
     <pre class="cddl" data-cddl-module="local-cddl">
       browsingContext.StartScreencastResult = {
-         screencast: browsingContext.Screencast
+         screencast: browsingContext.Screencast,
+         path: text
       }
 
       browsingContext.Screencast = text
@@ -5184,7 +5186,7 @@ The [=remote end steps=] with |command parameters| are:
 
 1. Let |media stream| be <a spec=ECMASCRIPT>Await</a>(|promise|).
 
-1. Let |recorded chunks| be an empty [=/list=].
+1. Let |path| be an implementation-defined file path where the recording will be stored.
 
 1. Let |media recorder options| be a struct with <code>mimeType</code> |mime type|.
 
@@ -5192,25 +5194,33 @@ The [=remote end steps=] with |command parameters| are:
    with {{MediaRecorder/stream}} |media stream| and
    {{MediaRecorderOptions}} |media recorder options|.
 
+1. Let |recording| be a new [=struct=] with
+   <code>mediaRecorder</code> |media recorder|,
+   <code>mimeType</code> |mime type|,
+   <code>path</code> |path|,
+   <code>size</code> 0,
+   <code>context</code> |navigable|.
+
 1. [=Add an event listener=] to |media recorder| with type {{MediaRecorder/dataavailable}}
    and callback that performs the following steps given |event|:
 
    1. Let |data| be |event|'s {{BlobEvent/data}}.
 
-   1. [=list/Append=] |data| to |recorded chunks|.
+   1. Let |bytes| be |data|'s underlying [=byte sequence=].
 
-1. Call {{MediaRecorder/start()}} on |media recorder|.
+   1. Append |bytes| to the file at |path|.
+
+   1. Set |recording|["<code>size</code>"] to |recording|["<code>size</code>"] + |bytes|'s length.
+
+1. Call {{MediaRecorder/start()}} on |media recorder| with <code>1000</code>.
 
 1. Let |screencast| be the string representation of a [[!RFC9562|UUID]].
 
-1. Set [=screencast recordings map=][|screencast|] to a struct with
-   <code>mediaRecorder</code> |media recorder|,
-   <code>mimeType</code> |mime type|,
-   <code>recordedChunks</code> |recorded chunks|,
-   <code>context</code> |navigable|.
+1. Set [=screencast recordings map=][|screencast|] to |recording|.
 
 1. Return a new [=/map=] matching the <code>browsingContext.StartScreencastResult</code>
-   with the <code>screencast</code> field set to |screencast|.
+   with the <code>screencast</code> field set to |screencast| and <code>path</code> field
+   set to |path|.
 
 </div>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -5283,6 +5161,195 @@ The [=remote end steps=] with |command parameters| are:
...
+1. If the implementation is unable to record a screencast of |navigable| for any
+   reason then return [=error=] with [=error code=] [=unsupported operation=].
+
+1. Let |stream options| be |command parameters|["<code>streamOptions</code>"].
+
+1. Let |stream| be a new [=screencast stream=] for |navigable|, constructed as follows:
+
+   1. Create a |video track| which must be a live-capture of the [=/browser=] [=display surface=]
+      of the [=relevant global object=]'s [=associated `Document`=]'s |navigable|'s
+      <a spec=css2>viewport</a>.
```
lutien, line 5453: That and the audio track description are basically a copy from https://w3c.github.io/mediacapture-viewport/#dom-mediadevices-getviewportmedia. So we can probably share it.

index.bs
```diff
@@ -5283,6 +5161,207 @@ The [=remote end steps=] with |command parameters| are:
...
+      )
+
+      browsingContext.StartScreencastParameters = {
+        context: browsingContext.BrowsingContext,
+        ? mimeType: text,
+        ? streamOptions: browsingContext.MediaStreamOptions,
+      }
+
+      browsingContext.MediaStreamOptions = {
+         ? video: true / browsingContext.MediaTrackConstraints;
```
OrKoN:
```suggestion
         ? video: browsingContext.MediaTrackConstraints;
```

do we need true?

index.bs
```diff
@@ -5283,6 +5161,207 @@ The [=remote end steps=] with |command parameters| are:
...
+
+   1. Create a |video track| which must be a live-capture of the [=/browser=] [=display surface=]
+      of the [=relevant global object=]'s [=associated `Document`=]'s |navigable|'s
+      <a spec=css2>viewport</a>.
+
+   1. If |stream options|["<code>video</code>"] is a map:
+
+      1. If |stream options|["<code>video</code>"] [=map/contains=] <code>width</code>:
+
+         1. The [=screencast stream/video track=] must satisfy its <code>width</code>,
```
OrKoN: what does satisfy mean? should we scale the output to match the one of the width / height constraints (whatever fits better without changing the aspect ratio)? or does it mean that we reject the screencast if width/height does not match the viewport size.

index.bs
```diff
@@ -3362,6 +3253,62 @@ weak map between [=user context|user contexts=] and [=unhandled prompt behavior
 A [=remote end=] has a <dfn>scripting enabled overrides map</dfn> which is a weak
 map between [=/navigables=] or [=user context|user contexts=] and boolean.
 
+A <dfn>screencast stream</dfn> is an abstract stream of the viewport of a
+[=/top-level traversable=], consisting of a <dfn for="screencast stream">video track</dfn>
+containing the rendered visual output of the top-level document's viewport,
+and optionally an <dfn for="screencast stream">audio track</dfn> containing the audio output of the viewport.
```
OrKoN:
```suggestion
and optionally an <dfn for="screencast stream">audio track</dfn> containing the audio output of the top-level traversable.
```

Audio output of the viewport does not sound accurate to me. Is it based on some media spec?

index.bs
```diff
@@ -3362,6 +3253,62 @@ weak map between [=user context|user contexts=] and [=unhandled prompt behavior
 A [=remote end=] has a <dfn>scripting enabled overrides map</dfn> which is a weak
 map between [=/navigables=] or [=user context|user contexts=] and boolean.
 
+A <dfn>screencast stream</dfn> is an abstract stream of the viewport of a
+[=/top-level traversable=], consisting of a <dfn for="screencast stream">video track</dfn>
+containing the rendered visual output of the top-level document's viewport,
```
OrKoN:
```suggestion
containing the rendered visual output of the [=/top-level traversable=]'s document's viewport,
```

I am not sure top-level document is an established definition?

index.bs
```diff
@@ -3362,6 +3253,62 @@ weak map between [=user context|user contexts=] and [=unhandled prompt behavior
...
+an [=struct/item=] named <dfn for="screencast recording">size</dfn>, which is a number,
+an [=struct/item=] named <dfn for="screencast recording">writeError</dfn>, which is a string or null.
+
+<div algorithm>
+To <dfn>start a screencast recording</dfn> given a [=screencast recording=] |recording|, |mime type| and |timeslice|:
+
+1. The implementation must begin encoding |recording|'s
+   [=screencast recording/stream=] using |mime type|, producing successive chunks of encoded data as
+   [=byte sequences=]. A new chunk must be produced at least every |timeslice|
+   milliseconds while |recording|'s [=screencast recording/state=] is
```
jgraham: I don't like this MUST-level requirement on timing. If the system is under load, or in a sleep state, or something similar, it may be impossible to produce data on the required schedule. I think we should rephrase this as a target / soft limit.

index.bs
```diff
@@ -3362,6 +3253,62 @@ weak map between [=user context|user contexts=] and [=unhandled prompt behavior
...
+an [=struct/item=] named <dfn for="screencast recording">path</dfn>, which is a string,
+an [=struct/item=] named <dfn for="screencast recording">state</dfn>, which is one of
+"<code>recording</code>", "<code>stopping</code>", "<code>stopped</code>",
+an [=struct/item=] named <dfn for="screencast recording">size</dfn>, which is a number,
+an [=struct/item=] named <dfn for="screencast recording">writeError</dfn>, which is a string or null.
+
+<div algorithm>
+To <dfn>start a screencast recording</dfn> given a [=screencast recording=] |recording|, |mime type| and |timeslice|:
+
+1. The implementation must begin encoding |recording|'s
```
jgraham: You generally don't need "the implementation must"; it's easier to say "Begin encoding […]".

Also, this algorithm is supposed to run in parallel, so you should use the explicit machinery for that from HTML, sort of like:

```
1. Any necessary setup steps that should happen synchronously (create the file with appropriate headers?)

1. Run the following steps [=in parallel=]:

  1. Stuff that happens continuously in the background
```

index.bs
```diff
@@ -3362,6 +3253,62 @@ weak map between [=user context|user contexts=] and [=unhandled prompt behavior
...
+
+   1. Append |bytes| to the file at |recording|'s [=screencast recording/path=].
+      If this fails:
+
+      1. Set |recording|'s [=screencast recording/writeError=] to an
+         implementation-defined string describing the write failure.
+
+      1. [=Stop a screencast recording=] given |recording|.
+
+   1. Otherwise, set |recording|'s [=screencast recording/size=] to
```
jgraham: Should we make these updates explicitly atomic? It's unclear if appending some bytes can work in the above step; if that's the case then the size here would be too long.

index.bs
```diff
@@ -3362,6 +3253,62 @@ weak map between [=user context|user contexts=] and [=unhandled prompt behavior
...
+</div>
+
+<div algorithm>
+To <dfn>stop a screencast recording</dfn> given a [=screencast recording=] |recording|:
+
+1. If |recording|'s [=screencast recording/state=] is "<code>recording</code>":
+
+   1. Set |recording|'s [=screencast recording/state=] to "<code>stopping</code>".
+
+   1. The implementation must stop producing new chunks for |recording|, flush any
```
jgraham: Similarly to the above the use of "must" here is unnecessary and suggests that the algorithm is misphrased.

Also, I notice that all the callers of this algorithm act as if it's asynchronous, but then wait on the state change that indicates it's complete. We should probably just make this algorithm explicitly synchronous and encapsulate the asynchronicity.

My suggestion would be that we try to put all the writing behaviour in one place, and have it check the state, so that it can decide if it's writing the final chunk or not. Then this algorithm is just:

```
1. If  |recording|'s [=screencast recording/state=] is not "<code>recording</code>" then return.

1. Set |recording|'s [=screencast recording/state=] to "<code>stopping</code>".

1. Wait until |recording|'s [=screencast recording/state=] is "<code>stopped</code>".
```

index.bs
jgraham: (I have the exact same question :) )

index.bs
```diff
@@ -5283,6 +5161,207 @@ The [=remote end steps=] with |command parameters| are:
...
+
+         1. The [=screencast stream/video track=] must satisfy its <code>frameRate</code>,
+            otherwise if the implementation cannot satisfy these constraints,
+            return [=error=] with [=error code=] [=unknown error=].
+
+   1. [=map/Set=] |video track| to |stream|[[=screencast stream/video track=]].
+
+   1. If |stream options|["<code>audio</code>"] is true:
+
+      1. Create an |audio track| which must be the combined audio produced by the sum of documents
```
```suggestion
      1. Create an |audio track| which contains the combined audio produced by the sum of documents
```

index.bs
```diff
@@ -5283,6 +5161,207 @@ The [=remote end steps=] with |command parameters| are:
...
+         that consist of the [=relevant global object=]'s [=associated `Document`=]'s
+         |navigable|'s [=navigable/active document=], and all [=navigable/active documents=]
+         in nested [=/browsing context=]s of the [=relevant global object=]'s [=associated `Document`=]'s |navigable|.
+
+      1. [=map/Set=] |audio track| to |stream|[[=screencast stream/audio track=]].
+
+   1. If the implementation is unable to produce |stream|, return [=error=] with
+      [=error code=] [=unknown error=].
```
jgraham, line 5487: Why "unknown error"?

index.bs
```diff
@@ -5283,6 +5161,207 @@ The [=remote end steps=] with |command parameters| are:
...
+1. Let |recording| be a new [=screencast recording=] with
+   [=screencast recording/stream=] |stream|,
+   [=screencast recording/path=] |path|,
+   [=screencast recording/state=] "<code>recording</code>",
+   [=screencast recording/size=] 0,
+   [=screencast recording/writeError=] null.
+
+1. Set |session|'s [=screencast recordings map=][|screencast|] to |recording|.
+
+1. Let |timeslice| be an implementation-defined value.
```
jgraham: I don't understand the point of timeslice if it's implementation defined. Was the idea that the user would be in control? If that's no longer the case I think we should just remove it as a parameter and just allow the implementation to write whenever (e.g. when there's "enough" data to fill some implementation-defined buffer, or at some time limit, whichever is sooner).

index.bs
```diff
@@ -5283,6 +5161,207 @@ The [=remote end steps=] with |command parameters| are:
...
+
+1. [=Stop a screencast recording=] given |screencast recording|.
+
+1. Wait until |screencast recording|'s [=screencast recording/state=] is
+   "<code>stopped</code>".
+
+1. [=map/Remove=] |screencast| from |session|'s [=screencast recordings map=].
+
+1. If |screencast recording|'s [=screencast recording/writeError=] is not null,
+   return [=error=] with [=error code=] [=unknown error=].
```
jgraham: It seems possible that we get a partially written file that's still useful. I wonder if we should just return the size and an error flag as part of the normal result (I guess returning the path is fine, but we should already know that from when we started the recording so it's not necessary).

jgraham (changes requested):

index.bs
lutien: Yes, right, it looks like it should be rather the output of the top-level traversable’s document. Updated.

index.bs
lutien: Yeah, sorry, untangling the media constraints algorithm happened to be not so straightforward. 😕 
Ok, so if we still try to follow it, the idea seemed to be if we just specify the number for width and height, it just has to be about best effort for user agents to achieve the specified constraints. We can also add an option of specifying an object with `min` and `max` or `exact` to enforce more specific dimensions. But I'm not sure if it's required as the first step. 
I've used now the wording from MDN since it basically summarizes the final result of applying different parts of the media constraints algorithm in the scope of one property (e.g., [for height](https://developer.mozilla.org/en-US/docs/Web/API/MediaTrackConstraints/height)). So let me know if it's better now.

index.bs
lutien: Ok, I've rephrased and "inlined" `timeslice`

index.bs
lutien: Updated ✅

index.bs
lutien: Yeah, that looks simpler 🙂 Updated ✅

index.bs
lutien: hmm... I went back and forth about it, but I think it's probably still a better fit here. My idea was to handle a failure in the video and audio stream creation. Would you expect `an unsupported operation` here or something else?

index.bs
lutien: Yeah, initially it was a parameter for `MediaRecorder` which requires it to define when to produce a new chunk. But yes, since Alex prefers it to be an implementation-defined, it's not really needed anymore. I've removed it.

index.bs
lutien: Makes sense to me. I've updated to just make the error part of the response in case it's present. 

@OrKoN, does it look reasonable to you?

index.bs
lutien: I think I'm not sure if I fully understand what you're suggesting. Maybe the wording is confusing. 
The idea was exactly to only update the recording size in case appending the bytes didn't fail. If it fails, we only `stop a screencast recording`.

jgraham: Er, I meant to say "too short". The problem is the case where the append operation partially succeeds. Since writing to a file isn't really defined, it's not clear that it's on its own an atomic operation.

We could also just avoid explicitly storing the size and at the end say that we return the size of the file in bytes.

lutien: Ah, alright, that makes sense 🙂 Updated

index.bs
OrKoN: I think using the width / height only as a max constraint on producing a video stream of the original aspect ratio sounds sufficient to me. I also got info that some codecs prefer the width / height to be divisible by 2 so I think we should not be exact here.

index.bs
OrKoN: I think we would need timeslice once we have streaming and it should work with the stream interface, so that a chunk of a stream equals at least a timeslice of the screencast. For file it would not really matter because we do not expect the clients to listen for file updates and read buffers as they go for streaming. Removing it for now SGTM but I think we would need to account for it when designing streams. In the Web API for this, the timeslice defines how often is the event that a blob available is emitted and how many video seconds the blob contains and I think the blob needs to be playable for playback.

index.bs
OrKoN: I would prefer not to return the size. Returning an error code sgtm.

lutien: Ok, I've updated to only return `size` in the success case.

OrKoN: Why do we need to return size? The client should be able to access the file to get the size, if the client has no access to the file, the size does not matter anyway?

lutien: Ah, you meant that you would prefer to not return the `size` field at all.
I think the idea was that it might be interesting for the clients to know and we can calculate it since we're processing it, but of course it was a nice to have thing and the clients can absolutely calculate it themselves.
I've removed it now.

lutien:
@OrKoN, @jgraham, any more feedback?

OrKoN:
@lutien I see several comment threads being open, could youplease mark the ones that have been resolved?

lutien:
> @lutien I see several comment threads being open, could youplease mark the ones that have been resolved?

done

OrKoN (approved):
LGTM thanks!

jgraham (approved):
I suspect we'll want to tighten some of this up over time, but I think it's good enough to land :)

lutien:
Thanks for the review! I'm going to merge it now!

index.bs
```diff
@@ -5302,6 +5373,199 @@ The [=remote end steps=] with |command parameters| are:
...
+      <pre class="cddl" data-cddl-module="remote-cddl">
+      browsingContext.StartScreencast = (
+        method: "browsingContext.startScreencast",
+        params: browsingContext.StartScreencastParameters
+      )
+
+      browsingContext.StartScreencastParameters = {
+        context: browsingContext.BrowsingContext,
+        ? mimeType: text,
+        ? streamOptions: browsingContext.MediaStreamOptions,
```
nvborisenko, line 5398: Perhaps flatten for simplicity?

```
context:
? mimeType:
? video:
? audio:
```

lutien: I think that would be fine. Feel free to provide a PR for the proposed change.

Issue #1071, opened 2026-02-11
Make userContext properties mandatory for events


jimevans:
When #1046 lands, it adds `userContext` properties to the payload of events where browsing contexts are returned for the events. It makes the addition of the `userContext` ID optional. Making them optional is a compromise for clients to adapt to the presence of these properties in events and to give implementations time to implement the inclusion of this information into the event payloads. Ideally, these properties should be mandatory.

juliandescottes:
As noted in #1122 we will also have to update the browsingContext.CreateResult payload, which also marks the userContext as optional even though the algorithm guarantees to retrieve a valid id.

PR #1122, opened 2026-05-19
fix: make userContext mandatory in browsingContext.CreateResult

juliandescottes:
The algorithm for browsingContext.create reads as follows:

> Let body be a [map](https://infra.spec.whatwg.org/#ordered-map) matching the browsingContext.CreateResult production, with the context field set to traversable’s [navigable id](https://w3c.github.io/webdriver-bidi/#navigable-id) and the userContext property set to the [user context id](https://w3c.github.io/webdriver-bidi/#user-context-user-context-id) of traversable’s [associated user context](https://w3c.github.io/webdriver-bidi/#associated-user-context).

Nothing here can lead userContext to be missing so I don't see why the property should be optional?

***
<a href="https://pr-preview.s3.amazonaws.com/juliandescottes/webdriver-bidi/pull/1122.html" title="Last updated on May 19, 2026, 12:45 PM UTC (c424f7d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/1122/585e4f1...juliandescottes:c424f7d.html" title="Last updated on May 19, 2026, 12:45 PM UTC (c424f7d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4144,7 +4144,7 @@ The <dfn export for=commands>browsingContext.create</dfn> command creates a new
     <pre class="cddl" data-cddl-module="local-cddl">
         browsingContext.CreateResult = {
           context: browsingContext.BrowsingContext,
-          ? userContext: browser.UserContext
+          userContext: browser.UserContext
         }
     </pre>
    </dd>


juliandescottes:
Ok I see that this seems to be another follow up of #1046 which not only added userContext to events, but also to the return value of browsingContext.create. 

I will close in favor of #1071, I guess we can make it mandatory at the same time as the rest (in the future we should add a note in the spec when we make fields temporarily optional, that would save time ...)

PR #1135, opened 2026-07-14
Add io module for streams with handling streams in the associated document's realm

lutien:
An alternative approach to #1061 to handle streams in the associated document's realm + adding the serialization of ReadableStreams as RemoteValue to demonstrate how the whole flow would look.

***
<a href="https://pr-preview.s3.amazonaws.com/lutien/webdriver-bidi/pull/1135.html" title="Last updated on Sep 10, 2026, 1:45 PM UTC (c5e536d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/1135/d636c85...lutien:c5e536d.html" title="Last updated on Sep 10, 2026, 1:45 PM UTC (c5e536d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7918,7 +7918,9 @@ The <dfn export for=commands>io.ReadStream</dfn> command reads the next chunk of
 [=readable stream=]. If provided, the <code>timeout</code> parameter specifies the
 maximum number of milliseconds to wait for the next chunk. If the timeout
 expires, the command returns an [=error=] with [=error code=] [=timeout=]
-without removing the stream.
+without removing the stream. Only one invocation for the same stream can be
+active at a time. Additional concurrent invocations return an [=error=] with
+[=error code=] [=stream read error=].
 
 <dl>
   <dt>Command Type</dt>
@@ -7955,6 +7957,11 @@ The [=remote end steps=] given |session| and |command parameters| are:
 1. Let |stream settings| be the result of [=trying=] to [=get readable stream settings=] with
    |session| and |stream id|.
 
+1. If |stream settings|[[=stream settings/read command active=]] is true,
+   return [=error=] with [=error code=] [=stream read error=].
+
+1. Set |stream settings|[[=stream settings/read command active=]] to true.
+
 1. Let |realm id| be |stream settings|[[=stream settings/realm=]].
 
 1. If |realm id| is not null:
@@ -7963,16 +7970,6 @@ The [=remote end steps=] given |session| and |command parameters| are:
 
    1. Let |environment settings| be the [=environment settings object=] whose
       [=realm execution context=]'s Realm component is |realm|.
-
-   1. Let |stream map| be |session|'s [=readable stream settings map=].
-
-   1. Wait until |stream settings|[[=stream settings/read command active=]] is
-      false or |stream map| no longer [=map/contains=] |stream id|.
-
-   1. If |stream map| no longer [=map/contains=] |stream id|, return [=error=]
-      with [=error code=] [=no such stream=].
-
-   1. Set |stream settings|[[=stream settings/read command active=]] to true.
 
    1. Let |stream| be |stream settings|[[=stream settings/stream=]].
 
--- a/index.bs
+++ b/index.bs
@@ -591,6 +591,7 @@ CommandData = (
   BrowsingContextCommand //
   EmulationCommand //
   InputCommand //
+  IoCommand //
   NetworkCommand //
   ScriptCommand //
   SessionCommand //
@@ -633,6 +634,7 @@ ResultData = (
   BrowsingContextResult /
   EmulationResult /
   InputResult /
+  IoResult /
   NetworkResult /
   ScriptResult /
   SessionResult /
@@ -813,11 +815,17 @@ with the following additional codes:
   <dt><dfn for=errors export>no such storage partition</dfn>
   <dd>Tried to access data in a non-existent storage partition.
 
+  <dt><dfn for=errors export>no such stream</dfn>
+  <dd>Tried to access an unknown [=Stream=].
+
   <dt><dfn for=errors export>no such user context</dfn>
   <dd>Tried to reference an unknown [=user context=].
 
   <dt><dfn for=errors export>no such web extension</dfn>
   <dd>Tried to reference an unknown web extension.
+
+  <dt><dfn for=errors export>stream read error</dfn>
+  <dd>Unable to read from a [=Stream=].
 
   <dt><dfn for=errors export>unable to close browser</dfn>
   <dd>Tried to close the browser, but failed to do so.
@@ -854,10 +862,12 @@ ErrorCode = "invalid argument" /
             "no such request" /
             "no such screencast" /
             "no such script" /
+            "no such stream" /
             "no such storage partition" /
             "no such user context" /
             "no such web extension" /
             "session not created" /
+            "stream read error" /
             "unable to capture screen" /
             "unable to close browser" /
             "unable to set cookie" /
@@ -7687,6 +7697,212 @@ is legacy, user agent might run [=implementation-defined=] steps to respect the
    to |maxTouchPoints|.
 
 1. Return [=success=] with data null.
+
+</div>
+
+## The io Module ## {#module-io}
+
+The <dfn export for=modules>io</dfn> module contains commands and events
+relating to IO operations.
+
+### Definition ### {#module-io-definition}
+
+{^remote end definition^}
+
+<pre class="cddl" data-cddl-module="remote-cddl">
+
+IoCommand = (
+  io.Cancel //
+  io.ReadStream //
+)
+
+</pre>
+
+{^local end definition^}
+
+<pre class="cddl" data-cddl-module="local-cddl">
+
+IoResult = (
+   io.ReadStreamResult
+)
+</pre>
+
+A [=BiDi Session=] has a <dfn>readable stream map</dfn> which is a [=/map=] between
+a <code>io.Stream</code> and a [=readable stream=].
+
+<div class="algorithm">
+
+To <dfn>get a readable stream</dfn> given <var ignore>session</var> and |stream|:
+
+1. Let |stream map| be |session|'s [=readable stream map=].
+
+1. If |stream map| [=map/contains=] |stream|, return [=success=] with data
+   |stream map|[|stream|].
+
+1. Return [=error=] with [=error code=] [=no such stream=].
+
+</div>
+
+### Types ### {#module-io-types}
+
+#### The io.ReadableStream Type #### {#type-io-ReadableStream}
+
+<pre class="cddl" data-cddl-module="local-cddl">
+io.ReadableStream = {
+  stream: io.Stream,
+}
+</pre>
+
+The <code>io.ReadableStream</code> type represents a [=readable stream=].
+
+#### The io.ReadData Type #### {#type-io-ReadData}
+
+<pre class="cddl" data-cddl-module="local-cddl">
+io.ReadData = {
+  value: network.BytesValue,
+  done: bool
+}
+</pre>
+
+The <code>io.ReadData</code> type represents the result of reading a value
+from a [=readable stream=].
+
+
+#### The io.Stream Type #### {#type-io-Stream}
+
+<pre class="cddl" data-cddl-module="local-cddl">
+io.Stream = text
+</pre>
+
+The <code>io.Stream</code> type represents a handle to an IO [=stream=].
+
+### Commands ### {#module-io-commands}
+
+#### The io.cancel Command ####  {#command-io-cancel}
+
+The <dfn export for=commands>io.Cancel</dfn> command cancels a [=readable stream=].
+
+<dl>
+  <dt>Command Type</dt>
+  <dd>
+    <pre class="cddl" data-cddl-module="remote-cddl">
+      io.Cancel = (
+        method: "io.cancel",
+        params: io.CancelParameters
+      )
+
+      io.CancelParameters = {
+        stream: io.Stream,
+        ? reason: text,
+      }
+      </pre>
+   </dd>
+   <dt>Return Type</dt>
+   <dd>
+    <pre class="cddl" data-cddl-module="local-cddl">
+      io.CancelResult = EmptyResult
+    </pre>
+  </dd>
+</dl>
+
+<div algorithm="remote end steps for io.Cancel">
+The [=remote end steps=] given |session| and |command parameters| are:
+
+1. Let |stream| be the result of trying to [=get a readable stream=] with
+   |session| and |command parameters|["<code>stream</code>"].
+
+1. If |command parameters| [=map/contains=] "<code>reason</code>, let |reason|
+   be |command parameters|["<code>reason</code>"]. Otherwise let |reason| be
+   undefined.
+
+1. Run [$ReadableStreamCancel$]\(|stream|, |reason|).
+
+</div>
+
+### The io.readStream Command ### {#command-io.readStream}
+
+The <dfn export for=commands>io.ReadStream</dfn> command reads the next chunk of a
+[=readable stream=].
+
+<dl>
+  <dt>Command Type</dt>
+  <dd>
+    <pre class="cddl" data-cddl-module="remote-cddl">
+      io.ReadStream = (
+       method: "io.readStream",
+       params: io.ReadStreamParameters
+      )
+
+      io.ReadStreamParameters = {
+        stream: io.Stream,
+      }
+    </pre>
+   </dd>
+   <dt>Return Type</dt>
+   <dd>
+    <pre class="cddl" data-cddl-module="local-cddl">
+      io.ReadStreamResult = io.ReadData
+    </pre>
+   </dd>
+</dl>
+
+
+<div algorithm="remote end steps for io.readStream">
+The [=remote end steps=] given |session| and |command parameters| are:
+
+1. Let |stream| be the result of [=trying=] to [=get a readable stream=] with
+   |session| and |command parameters|["<code>stream</code>"].
+
+1. Let |result| be null.
+
+1. Let |read request| be a new [=read request=] with the following items:
+  <dl>
+    <dt>[=read request/chunk steps=] given |chunk|:
+    <dd>
+      1. Assert: |result| is null.
+
+      1. Let |response| be a new `io.ReadResult` with the <code>value</code>
+         field set to [=serialize protocol bytes=] given |chunk| and the
+         <code>done</code> field set to false.
+
+      1. Let |result| be [=success=] with data |response|.
+    </dd>
+
+    <dt>[=read request/close steps=]:
+    <dd>
+      1. Assert: |result| is null.
+
+      1. Let |response| be a new `io.ReadResult` with the <code>value</code>
+         field set to <code>""</code> and the <code>done</code> field set to
+         true.
+
+      1. Let |result| be [=success=] with data |response|.
+    </dd>
+
+    <dt>[=read request/error steps=] given <var ignore>e</var>:
+    <dd>
+      1. Assert: |result| is null.
+
+      1. Let |result| be [=error=] with [=error code=] [=stream read error=].
+    </dd>
+  </dl>
+
+1. Set |stream|.\[[disturbed]] to true.
+
+1. If |stream|.\[[state]] is "<code>closed</code>", perform |read request|'s close steps.
+
+1. Otherwise, if |stream|.\[[state]] is "<code>errored</code>", perform |read
+   request|'s error steps given |stream|.\[[storedError]].
+
+1. Otherwise,
+
+    1. Assert: stream.\[[state]] is "readable".
+
+    1. Perform ! |stream|.\[[controller]].\[[PullSteps]](|read request|).
+
+1. Assert: |result| is not null.
+
+1. Return |result|.
 
 </div>
 
--- a/index.bs
+++ b/index.bs
@@ -7781,7 +7781,7 @@ from a [=readable stream=].
 
 #### The io.Stream Type #### {#type-io-Stream}
 
-<pre class="cddl" data-cddl-module="local-cddl">
+<pre class="cddl" data-cddl-module="remote-cddl,local-cddl">
 io.Stream = text
 </pre>
 
@@ -12360,7 +12360,8 @@ script.RemoteValue = (
   script.NodeListRemoteValue /
   script.HTMLCollectionRemoteValue /
   script.NodeRemoteValue /
-  script.WindowProxyRemoteValue
+  script.WindowProxyRemoteValue /
+  script.ReadableStreamRemoteValue
 )
 
 script.ListRemoteValue = [*script.RemoteValue];
@@ -12510,6 +12511,13 @@ script.WindowProxyRemoteValue = {
 
 script.WindowProxyProperties = {
   context: browsingContext.BrowsingContext
+}
+
+script.ReadableStreamRemoteValue = {
+  type: "readable-stream",
+  stream: io.Stream,
+  ? handle: script.Handle,
+  ? internalId: script.InternalId,
 }
 </pre>
 
@@ -12883,6 +12891,31 @@ an |ownership type|, a |serialization internal map|, a |realm| and a |session|:
        production in the {^local end definition^}, with the <code>handle</code>
        property set to |handle id| if it's not null, or omitted otherwise, and
        the <code>value</code> property set to |serialized|.
+
+    <dt>|value| is a [=platform object=] that implements {{ReadableStream}}
+    <dd>
+    1. Let |stream realm| be |value|'s [=relevant settings object=]'s
+       [=realm execution context=]'s Realm component.
+
+    1. Let |navigable| be the result of [=get the navigable=] with |stream realm|.
+
+    1. If |navigable| is null, let |navigable id| be null. Otherwise, let
+       |navigable id| be the [=navigable id=] for |navigable|.
+
+    1. Let |stream id| be the string representation of a [[!RFC9562|UUID]].
+
+    1. Let |stream settings| be a [=stream settings=] struct with the
+       [=stream settings/stream=] item set to |value| and the
+       [=stream settings/navigable=] item set to |navigable id|.
+
+    1. Set |session|'s [=readable stream settings map=][|stream id|] to
+       |stream settings|.
+
+    1. Let |remote value| be a [=/map=] matching the
+       <code>script.ReadableStreamRemoteValue</code> production in the
+       {^local end definition^}, with the <code>stream</code> property set to
+       |stream id|, and the <code>handle</code> property set to |handle id| if
+       it's not null, or omitted otherwise.
 
     <dt>|value| is a [=platform object=]
     <dd>1. Let |remote value| be a [=/map=] matching the <code>script.ObjectRemoteValue</code>
--- a/index.bs
+++ b/index.bs
@@ -7799,16 +7799,6 @@ To <dfn>cancel a readable stream</dfn> given |realm|, |stream| and |reason|:
 
 ### Types ### {#module-io-types}
 
-#### The io.ReadableStream Type #### {#type-io-ReadableStream}
-
-<pre class="cddl" data-cddl-module="local-cddl">
-io.ReadableStream = {
-  stream: io.Stream,
-}
-</pre>
-
-The <code>io.ReadableStream</code> type represents a [=readable stream=].
-
 #### The io.ReadData Type #### {#type-io-ReadData}
 
 <pre class="cddl" data-cddl-module="local-cddl">
@@ -7964,11 +7954,9 @@ The [=remote end steps=] given |session| and |command parameters| are:
         <dd>Reject |promise| with |e|.
      </dl>
 
-   1. Let |read request| be the result of [=ReadableStreamDefaultReader/read a chunk=] from |reader| given |read request|.
+   1. [=ReadableStreamDefaultReader/Read a chunk=] from |reader| given |read request|.
 
-   1. If |read result|.\[[Type]] is <code>normal</code>:
-
-      1. Set |read result| to <a spec=ECMASCRIPT>Await</a>(|read result|.\[[Value]]).
+   1. Let |read result| be <a spec=ECMASCRIPT>Await</a>(|promise|).
 
    1. [=Clean up after running script=] with |environment settings|.
 
@@ -12971,6 +12959,9 @@ an |ownership type|, a |serialization internal map|, a |realm| and a |session|:
        |stream id|, and the <code>handle</code> property set to |handle id| if
        it's not null, or omitted otherwise.
 
+   1. [=Set internal ids if needed=] with |serialization internal map|,
+      |remote value| and |value|.
+
     <dt>|value| is a [=platform object=]
     <dd>1. Let |remote value| be a [=/map=] matching the <code>script.ObjectRemoteValue</code>
            production in the {^local end definition^}, with the <code>handle</code>
@@ -14156,8 +14147,9 @@ Define the following [=unloading document cleanup steps=] with |document|:
    1. For each |stream id| → |stream settings| in |active session|'s
       [=readable stream settings map=]:
 
-      1. If |stream settings|[[=stream settings/realm=]] equals |realm id|,
-         [=remove a readable stream=] given |active session| and |stream id|.
+      1. If |stream settings|[[=stream settings/realm=]] equals |realm id|:
+
+         1. [=Remove a readable stream=] given |active session| and |stream id|.
 
 1. Let |params| be a [=/map=] matching the
    <code>script.RealmDestroyedParameters</code> production, with the
--- a/index.bs
+++ b/index.bs
@@ -816,7 +816,7 @@ with the following additional codes:
   <dd>Tried to access data in a non-existent storage partition.
 
   <dt><dfn for=errors export>no such stream</dfn>
-  <dd>Tried to access an unknown [=Stream=].
+  <dd>Tried to access an unknown [=/stream=].
 
   <dt><dfn for=errors export>no such user context</dfn>
   <dd>Tried to reference an unknown [=user context=].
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -7690,6 +7704,255 @@ is legacy, user agent might run [=implementation-defined=] steps to respect the
...
+
+A <dfn lt="stream|IO stream" export>stream</dfn> is an abstract source of
+data that can be read from or written to.
+
+A [=BiDi Session=] has a <dfn>readable stream settings map</dfn> which is a [=/map=] between
+a <code>io.Stream</code> and [=stream settings=].
+
+<dfn>Stream settings</dfn> is a [=struct=] with
+an [=struct/item=] named <dfn for="stream settings">stream</dfn>, which is a [=readable stream=],
+an [=struct/item=] named <dfn for="stream settings">navigable</dfn>, which is a <code>browsingContext.BrowsingContext</code> or null.
```
jgraham: Should this be a navigable or directly be a realm (e.g. so that we support streams coming from workers)?

index.bs
```diff
@@ -7690,6 +7704,255 @@ is legacy, user agent might run [=implementation-defined=] steps to respect the
...
+
+IoResult = (
+   io.ReadStreamResult
+)
+</pre>
+
+A <dfn lt="stream|IO stream" export>stream</dfn> is an abstract source of
+data that can be read from or written to.
+
+A [=BiDi Session=] has a <dfn>readable stream settings map</dfn> which is a [=/map=] between
```
jgraham, line 7752: As written this would seem to possibly keep navigables (or realms) alive as long as there's a reference to one of their streams. That feels likely to cause memory leaks.

index.bs
```diff
@@ -7690,6 +7704,255 @@ is legacy, user agent might run [=implementation-defined=] steps to respect the
...
+     1. [=Prepare to run script=] with |environment settings|.
+
+     1. Let |promise| be the result of [=ReadableStream/cancel=] |stream| with
+        |reason|.
+
+     1. [=React=] to |promise|:
+
+      1. If |promise| was rejected, then:
+
+         1. [=Clean up after running script=] with |environment settings|.
```
jgraham: I think this needs to be outside the reaction; if the promise isn't already fulfilled the steps end up being called in a future microtask checkpoint when we're no longer top of the execution stack.

index.bs
```diff
@@ -7690,6 +7704,255 @@ is legacy, user agent might run [=implementation-defined=] steps to respect the
...
+
+   1. Let |realm| be the result of [=trying=] to [=get a realm from a navigable=]
+      with |navigable id| and null.
+
+   1. Let |environment settings| be the [=environment settings object=] whose
+      [=realm execution context=]'s Realm component is |realm|.
+
+   1. [=Prepare to run script=] with |environment settings|.
+
+   1. Let |reader| be the result of [=ReadableStream/getting a reader=] for |stream|.
```
jgraham: This can fail, we need to handle the failure.

index.bs
```diff
@@ -7690,6 +7704,255 @@ is legacy, user agent might run [=implementation-defined=] steps to respect the
...
+
+   1. Let |read result| be the result of calling {{ReadableStreamDefaultReader/read()}} for |reader|.
```
jgraham: I think you want to use https://streams.spec.whatwg.org/#readablestreamdefaultreader-read-a-chunk instead, which then affects the logic for the rest of this section.

index.bs
```diff
@@ -11449,6 +11712,13 @@ To <dfn>deserialize local value</dfn> given |local protocol value|, |realm| and
 1. If |local protocol value| matches the <code>script.ChannelValue</code> production,
    return [=create a channel=] with |session|, |realm| and |local protocol value|.
 
+1. If |local protocol value| matches the <code>script.ReadableStreamRemoteValue</code> production, then:
```
jgraham: It feels a bit weird to have a special case for streams here. I think you only get to this point if you send something with `{type: "readable-stream", stream: "uuid"}` but no `handle` property. I don't think we need to support that; if it's a script-accessible stream it should have a handle. If we do support it we at least need to check that the stream belongs to the current realm.

index.bs
```diff
@@ -12621,6 +12899,31 @@ an |ownership type|, a |serialization internal map|, a |realm| and a |session|:
        property set to |handle id| if it's not null, or omitted otherwise, and
        the <code>value</code> property set to |serialized|.
 
+    <dt>|value| is a [=platform object=] that implements {{ReadableStream}}
```
jgraham, line 12942: I'm not sure if we also want to support the case where you try to return a reader to the stream rather than the stream itself.

jgraham (changes requested):

index.bs
lutien: Yeah, I was thinking about it too. Ok, I've updated the PR to use a realm.

index.bs
lutien: Ok, I've added the cleanup steps when the realms are gone.

index.bs
lutien: Ok, I moved it up.

index.bs
lutien: I've added `trying`, is it enough, or do you want to send some specific error if it fails?

index.bs
lutien: Ah, alright, I've updated it.

index.bs
lutien: Ok, I've removed it for now.

index.bs
lutien: Ok, yeah, I'm not sure either. I guess that's a good question for the spec meeting. Let's keep it open for now.

lutien:
Hey, @MattiasBuelens, I've picked up the work on the streaming API from James. So when you have time, could you also review from your side?

lutien:
> Hey, @MattiasBuelens, I've picked up the work on the streaming API from James. So when you have time, could you also review from your side?

Hey @MattiasBuelens, not sure if you have seen my previous message. Would you have time to take a look? Or should we try to find someone else?

index.bs
```diff
@@ -7690,6 +7704,264 @@ is legacy, user agent might run [=implementation-defined=] steps to respect the
...
+        <dd>Resolve |promise| with «[ "value" → |chunk|, "done" → false ]».
+        <dt>[=read request/close steps=]
+        <dd>Resolve promise with «[ "value" → undefined, "done" → true ]».
+        <dt>[=read request/error steps=], given |e|
+        <dd>Reject |promise| with |e|.
+     </dl>
+
+   1. Let |read request| be the result of [=ReadableStreamDefaultReader/read a chunk=] from |reader| given |read request|.
+
+   1. If |read result|.\[[Type]] is <code>normal</code>.:
```
MattiasBuelens: Small typo:
```suggestion
   1. If |read result|.\[[Type]] is <code>normal</code>:
```

index.bs
```diff
@@ -13826,6 +14136,13 @@ worker=] algorithm:
 
 1. Let |realm id| be the [=realm id=] for |realm|.
 
+1. For each |active session| in [=active sessions=]:
+
+   1. For each |stream setting| in [=active sessions=]'s [=stream settings=]:
```
MattiasBuelens: Wrong reference.
```suggestion
   1. For each |stream setting| in [=active sessions=]'s [=readable stream settings map=]:
```

Also, is there some way to remove these entries earlier? I'm worried that we might be holding on to them for too long and cause memory leaks.

index.bs
```diff
@@ -13801,6 +14104,13 @@ Define the following [=unloading document cleanup steps=] with |document|:
...
```
MattiasBuelens: Wrong reference.

```suggestion
   1. For each |stream setting| in [=active sessions=]'s [=readable stream settings map=]:
```

MattiasBuelens (commented):
Apologies for the delay. I did a quick scan through the PR.

index.bs
lutien: Ah, right, I've added the cleanup logic after we're done reading from the stream, in case there's an error and when canceling the stream.

index.bs
```diff
@@ -7690,6 +7719,296 @@ is legacy, user agent might run [=implementation-defined=] steps to respect the
...
+}
+</pre>
+
+The <code>io.ReadableStream</code> type represents a [=readable stream=].
+
+#### The io.ReadData Type #### {#type-io-ReadData}
+
+<pre class="cddl" data-cddl-module="local-cddl">
+io.ReadData = {
+  value: network.BytesValue / null,
```
nvborisenko: `network.BytesValue` is `String` or `Base64`. I interpret it as `String` for real string content like html/json, `Base64` for bytes like png/bin.

If I read the first chunk of data, image it is `String`. The second chunk might be `Base64` (bytes?). Not clear how to concatenate at client side.

Please correct me, seems streaming should always work over `Base64`. If it is true, then for client it would be very convenient to always get `network.Base64Value`.

nvborisenko: Seems the same situation as with `browsingContext.CaptureScreenshotResult`, it is always `text` (base64 encoded).

If we really stream bytes, would be better to define it as `text` (base64 encoded).

index.bs
```diff
@@ -7690,6 +7719,296 @@ is legacy, user agent might run [=implementation-defined=] steps to respect the
...
+#### The io.readStream Command #### {#command-io.readStream}
+
+The <dfn export for=commands>io.ReadStream</dfn> command reads the next chunk of a
+[=readable stream=].
+
+<dl>
+  <dt>Command Type</dt>
+  <dd>
+    <pre class="cddl" data-cddl-module="remote-cddl">
+      io.ReadStream = (
```
nvborisenko, line 7903: Client has no control how much data it wants to read. Sorry, I am not network engineer.

My concern is following:
 1. remote end got 1GB of data immediately
 2. client sends `readStream` command
 3. remote end responds this 1GB to client via websocket (framing?)
 4. client reads ALL websocket frames into RAM (just to understand json message to parse later)

So at client side 1GB is _always_ materialized.

How to fix?.. Hm, guessing, as a client, I want to fetch chunks with 10KB (doesn't matter, just small rational enough) of data and write it to disk.
Or.. at remote end don't send 1GB as websocket message, send small rational chunks by default, also works for me (client).

nvborisenko: One idea came to me remote end sends truly binary data via websocket (not json). But disregard it: 1. it adds complexity 2. it blocks connection for concurrent responses

MattiasBuelens: Web sockets themselves don't yet have a way to stream individual messages. There was some thought about this when @ricea was designing `WebSocketStream` in https://github.com/whatwg/websockets/pull/48 but it ended up [as a non-goal for the first version of that API](https://github.com/ricea/websocketstream-explainer/blob/master/README.md#non-goals-in-the-first-version).

lutien: @nvborisenko, regarding the chunk size, it's actually defined on a stream side. So for the case of a `ReadableStream` being returned from a script (that is the only use case for now in this PR), you can specify the size of a chunk that has to be returned (see https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/ReadableStream#sizechunk). But yes, for network bodies and screencasting streaming, we probably will need some settings for the chunk sizes, since then streams are created on the protocol side.

index.bs
lutien: yeah, I don't think that it should be the case that you have different serializing strategies for different chunks of the same stream. I think we still want to have both `base64` and `string` types, but maybe we should have some mechanic to ensure it's consistent. I'll take a look.

lutien: Alright, we discussed it internally and agreed that it's indeed easier if we just always return the base64 encoded string. I've updated the PR.

lutien:
@OrKoN or @sadym-chromium, could you take a look as well? I think we're getting close here.

index.bs
```diff
@@ -7690,6 +7719,296 @@ is legacy, user agent might run [=implementation-defined=] steps to respect the
...
+
+### Definition ### {#module-io-definition}
+
+{^remote end definition^}
+
+<pre class="cddl" data-cddl-module="remote-cddl">
+
+IoCommand = (
+  io.Cancel //
+  io.ReadStream //
```
sadym-chromium, line 7735: nit: consider using `Read` omiting `Stream` as more consistent with other methods, instead of redundant `ReadStream`.

sadym-chromium:
> @OrKoN or @sadym-chromium, could you take a look as well? I think we're getting close here.

I'm on it

index.bs
sadym-chromium: Note: for the ecmascript's `ReadableStream` I believe it is acceptable not to have chunkSize parameter. However I believe we will need it later when in video streaming.

sadym-chromium (approved):

index.bs
lutien: I've actually thought about it and renamed the `io.cancel` command to `io.cancelStream`. I think that makes more sense because since it's an I/O module, we might have methods to work with other types of data in this module.

lutien:
@MattiasBuelens, thanks for the review! I've updated the PR according to your comments. Does it mean that overall it looks fine, or do you still want to take a closer look?

index.bs
```diff
@@ -7690,6 +7719,296 @@ is legacy, user agent might run [=implementation-defined=] steps to respect the
...
+     <dl>
+        <dt>[=read request/chunk steps=], given |chunk|
+        <dd>Resolve |promise| with «[ "value" → |chunk|, "done" → false ]».
+        <dt>[=read request/close steps=]
+        <dd>Resolve promise with «[ "value" → undefined, "done" → true ]».
+        <dt>[=read request/error steps=], given |e|
+        <dd>Reject |promise| with |e|.
+     </dl>
+
+   1. Let |read request| be the result of [=ReadableStreamDefaultReader/read a chunk=] from |reader| given |read request|.
```
OrKoN: Is it overwriting the previous |read request|? Could we introduce a new variable to avoid shadowing?

index.bs
```diff
@@ -7690,6 +7719,296 @@ is legacy, user agent might run [=implementation-defined=] steps to respect the
...
+   1. Let |realm| be the result of [=trying=] to [=get a realm=] with |realm id|.
+
+   1. Let |environment settings| be the [=environment settings object=] whose
+      [=realm execution context=]'s Realm component is |realm|.
+
+   1. [=Prepare to run script=] with |environment settings|.
+
+   1. Let |reader| be the result of [=trying=] to [=ReadableStream/getting a reader=] for |stream|.
+
+   1. Let |promise| be [=a new promise=].
```
OrKoN, line 7945: Is |promise| used (e.g., awaited) anywhere?

index.bs
```diff
@@ -12621,6 +12948,29 @@ an |ownership type|, a |serialization internal map|, a |realm| and a |session|:
...
+    1. Let |stream id| be the string representation of a [[!RFC9562|UUID]].
+
+    1. Let |stream settings| be a [=stream settings=] struct with the
+       [=stream settings/stream=] item set to |value| and the
+       [=stream settings/realm=] item set to |realm id|.
+
+    1. Set |session|'s [=readable stream settings map=][|stream id|] to
+       |stream settings|.
+
+    1. Let |remote value| be a [=/map=] matching the
```
OrKoN: should internalId be set here?

index.bs
```diff
@@ -13801,6 +14151,14 @@ Define the following [=unloading document cleanup steps=] with |document|:
 
 1. Let |realm id| be the [=realm id=] for |realm|.
 
+1. For each |active session| in [=active sessions=]:
+
+   1. For each |stream id| → |stream settings| in |active session|'s
+      [=readable stream settings map=]:
+
+      1. If |stream settings|[[=stream settings/realm=]] equals |realm id|,
```
OrKoN: This algorithm and the next one seems to mutate the map while iterating it. I am not sure anymore if the Infra standard tolerates that but perhaps it's worth splitting into 2 steps?

index.bs
```diff
@@ -7690,6 +7719,296 @@ is legacy, user agent might run [=implementation-defined=] steps to respect the
...
+
+1. [=Clean up after running script=] with |environment settings|.
+
+1. Return |promise|.
+
+</div>
+
+### Types ### {#module-io-types}
+
+#### The io.ReadableStream Type #### {#type-io-ReadableStream}
```
OrKoN: Is io.ReadableStream  type used anywhere?

OrKoN (commented):
I wonder if instead of tying JS streams to automation streams automatically, we should introduce an indirection: e.g., `io.createStreamFromHandle(objectHandle)` that would retain the stream for automation purposes and generate a streamId? Otherwise, I wonder if we should only retain streams in readable stream settings map if resultOwnership: "root" is specified, or tie it to the handle lifecycle.

jgraham:
> I wonder if instead of tying JS streams to automation streams automatically, we should introduce an indirection: e.g., io.createStreamFromHandle(objectHandle) that would retain the stream for automation purposes and generate a streamId? Otherwise, I wonder if we should only retain streams in readable stream settings map if resultOwnership: "root" is specified, or tie it to the handle lifecycle.

Can you explain the underlying concern here? Is it that people might be surprised that returning a stream leaks the stream (and therefore its owner) unless you close it or the realm is destroyed?

In practice I think that returning a stream without intending to use it is unlikely, but I agree that the need to call `cancelStream`, or navigate, is a bit different to JS primitive types. On the other hand it seems rather similar to the way that we handle Node types.

OrKoN:
> > I wonder if instead of tying JS streams to automation streams automatically, we should introduce an indirection: e.g., io.createStreamFromHandle(objectHandle) that would retain the stream for automation purposes and generate a streamId? Otherwise, I wonder if we should only retain streams in readable stream settings map if resultOwnership: "root" is specified, or tie it to the handle lifecycle.
> 
> Can you explain the underlying concern here? Is it that people might be surprised that returning a stream leaks the stream (and therefore its owner) unless you close it or the realm is destroyed?
> 
> In practice I think that returning a stream without intending to use it is unlikely, but I agree that the need to call `cancelStream`, or navigate, is a bit different to JS primitive types. On the other hand it seems rather similar to the way that we handle Node types.

There could be use cases where you just hold a reference to a stream in a client to pass it to another script evaluation without ever consuming the stream via automation. I think it would be nice to be explicit about the ownership and lifetime of a stream similar to how we have result ownership for handles (which stream handling does not integrate with) and streams might retain a lot of data and underlying infrastructure if not consumed/disposed. How about this: 1) only retain the stream if result ownership is root 2) dispose the stream of the handler is disowned? 

I think there could be also an issue that a stream can have a single reader and can be read once and so this might have effects on the page if the automation starts reading the stream. But I guess that would be fine? An alternative approach would be in some way to clone the stream for automation (upon explicit command), but I think it might be an overkill.

jgraham:
> How about this: 1) only retain the stream if result ownership is root 2) dispose the stream of the handler is disowned?

That seems reasonable to me; it is basically treating streams as JS values rather than as nodes.

> An alternative approach would be in some way to clone the stream for automation (upon explicit command), but I think it might be an overkill.

Yeah, you can tee the stream on the JS side. Auto-teeing seems like it would be surprising.

index.bs
lutien: `Read chunk` doesn't actually return anything, so we don't need to assign it to `|read request|`.

index.bs
lutien: Yeah, that was wrong, updated.

index.bs
lutien: right, added

index.bs
lutien: Sure, I've split it into 2 steps.

index.bs
lutien: Yeah, not currently, I've removed it for now.

lutien:
> > > I wonder if instead of tying JS streams to automation streams automatically, we should introduce an indirection: e.g., io.createStreamFromHandle(objectHandle) that would retain the stream for automation purposes and generate a streamId? Otherwise, I wonder if we should only retain streams in readable stream settings map if resultOwnership: "root" is specified, or tie it to the handle lifecycle.
> > 
> > 
> > Can you explain the underlying concern here? Is it that people might be surprised that returning a stream leaks the stream (and therefore its owner) unless you close it or the realm is destroyed?
> > In practice I think that returning a stream without intending to use it is unlikely, but I agree that the need to call `cancelStream`, or navigate, is a bit different to JS primitive types. On the other hand it seems rather similar to the way that we handle Node types.
> 
> There could be use cases where you just hold a reference to a stream in a client to pass it to another script evaluation without ever consuming the stream via automation. I think it would be nice to be explicit about the ownership and lifetime of a stream similar to how we have result ownership for handles (which stream handling does not integrate with) and streams might retain a lot of data and underlying infrastructure if not consumed/disposed. How about this: 1) only retain the stream if result ownership is root 2) dispose the stream of the handler is disowned?
> 
> I think there could be also an issue that a stream can have a single reader and can be read once and so this might have effects on the page if the automation starts reading the stream. But I guess that would be fine? An alternative approach would be in some way to clone the stream for automation (upon explicit command), but I think it might be an overkill.

Alright, I've updated to only save the stream when result ownership is root and added the steps to cancel the stream when disowning its handle.

index.bs
```diff
@@ -7690,6 +7719,285 @@ is legacy, user agent might run [=implementation-defined=] steps to respect the
...
+  <dt>Command Type</dt>
+  <dd>
+    <pre class="cddl" data-cddl-module="remote-cddl">
+      io.ReadStream = (
+       method: "io.readStream",
+       params: io.ReadStreamParameters
+      )
+
+      io.ReadStreamParameters = {
+        stream: io.Stream,
```
OrKoN, line 7909: would not we need a timeout parameter for the read stream call to support the cases of the stream being slowly populated? This timeout could not be client-side because if the client stops listening for the response, the command they sent might still cause a read from the stream if new data arrives. That means the client would not be able to recover from such situation.

OrKoN:
A meta question: does the proposed API offer anything new compared to just using Web APIs?

```js
import puppeteer from 'puppeteer';

await using browser = await puppeteer.launch({
  protocol: 'webDriverBiDi',
});
const page = await browser.newPage();

await page.evaluate(() => {
  window.createStream = () => {
    const encoder = new TextEncoder();
    return new ReadableStream({
      start(controller) {
        controller.enqueue(encoder.encode('chunk 1, '));
        controller.enqueue(encoder.encode('chunk 2, '));
        controller.enqueue(encoder.encode('chunk 3'));
        controller.close();
      },
    });
  };
});

await using stream = await page.evaluateHandle(() => window.createStream());
await using reader = await page.evaluateHandle((s) => s.getReader(), stream);

while (true) {
  const { done, base64 } = await page.evaluate(async (r) => {
    const { done, value } = await r.read();
    return done ? { done: true } : { done: false, base64: value.toBase64() };
  }, reader);

  if (done) break;
  console.log(Buffer.from(base64, 'base64').toString());
}
```

It seems to be fairly easy to use without any new protocol support.

jgraham:
> A meta question: does the proposed API offer anything new compared to just using Web APIs?

The idea in the medium term is to expose streams that are not typically accessible to content scripts (e.g. network request data).

I agree that on its own the ability to access streams returned from scripts doesn't add a lot that you can't implement in user space, but _not_ allowing such streams to be accessed via the API would seem like a strange and inconsistent choice, and considering that use case has helped us make the API design better vs just landing something totally abstract.

OrKoN:
In this case, I would prefer to keep the streaming API shape and specify how to handle non-content streams instead so that we focus on things that are not possible to implement with existing Web features. To me it's not 100% clear that the current API is ideal for non-content streams: I would have expected that we have some sort of server-side timeout to reliably detect stuck streams and allowing for a recovery. Do you think this would not be needed?

lutien:
> In this case, I would prefer to keep the streaming API shape and specify how to handle non-content streams instead so that we focus on things that are not possible to implement with existing Web features. To me it's not 100% clear that the current API is ideal for non-content streams: I would have expected that we have some sort of server-side timeout to reliably detect stuck streams and allowing for a recovery. Do you think this would not be needed?

I think the timeout argument makes sense. I've updated the PR to add it.

index.bs
```diff
@@ -7957,12 +7973,18 @@ The [=remote end steps=] given |session| and |command parameters| are:
 
    1. [=ReadableStreamDefaultReader/Read a chunk=] from |reader| given |read request|.
 
-   1. Let |read result| be <a spec=ECMASCRIPT>Await</a>(|promise|).
+   1. Wait until |promise| is resolved, or |timer|'s [=timer/timeout fired flag=]
```
OrKoN: The issue with the Web API here is that the promise cannot be cancelled so the read from stream might still happen, even if the client timed out, it means the client would miss a chunk. WDYT? Do you see any other way we can ensure recovery from timeouts?

index.bs
```diff
@@ -7957,12 +7973,18 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+      is set, whichever occurs first.
 
    1. [=Clean up after running script=] with |environment settings|.
 
    1. [=ReadableStreamDefaultReader/Release=] |reader|.
 
+   1. If |promise| is still pending and |timer|'s [=timer/timeout fired flag=]
```
jgraham, line 8038: This doesn't seem safe. If the timeout flag is set the read could still succeed later and the client will miss the data. I think you could make it work by stashing the promise somewhere and re-using the existing promise in subsequent calls if it exists (but I didn't think about it in lots of detail, so there might be a better way).

jgraham (changes requested):

index.bs
jgraham: Oh, this is also what @OrKoN just said :)

index.bs
lutien: If the command times out, we go to the `Release reader` step, which, to my understanding of https://streams.spec.whatwg.org/#ref-for-default-reader-release-lock%E2%91%A1, should reject the pending promise for reading and keep the chunk queued for the next reader. (I've fixed the error order, but that's not really related to your question)

index.bs
lutien: Replied above :) I think it should be fine?

index.bs
OrKoN: I think that works! Thanks!

index.bs
jgraham: Oh I see. So it works because we're acquiring a new reader for each read and releasing it whenever we return. But that has the observable consequence that the stream is unlocked between reads, so you have the possibility that something else with a handle to the stream can race you for data, whereas if we held the reader until the stream was closed then it would be first-to-create-a-reader-wins. 

I also think that acquiring a single reader would be the logical implementation strategy; having to create a new one each time could be non-trivial overhead (but ofc we're also going across the network so maybe it doesn't matter that much?), and is hard to tell apart from the given approach except in edge cases, which feels like a potential source of bugs.

How sure are we that we're OK with this approach?

jgraham: (I think it also means that we couldn't easily extend the API to allow passing a reader itself to the client vs a stream)

OrKoN: I think with the reader would be retained (with lock) it's not clear how to deal with timeouts again?

jgraham: Well you could do what I suggested and keep the promise; one the next call if there's an existing promise you re-use that rather than creating a new one. Or I suppose one could release the stream in the case of a timeout, but that seems potentially surprising if you don't do it in other cases.

lutien: Yeah, I think retaining the same reader makes more sense, and it looks like storing the promise should do the job for the timeout scenario. Updated the PR ✅

index.bs
```diff
@@ -7690,6 +7724,364 @@ is legacy, user agent might run [=implementation-defined=] steps to respect the
...
+           <dt>[=read request/close steps=]
+           <dd>Resolve promise with «[ "value" → undefined, "done" → true ]».
+           <dt>[=read request/error steps=], given |e|
+           <dd>Reject |promise| with |e|.
+        </dl>
+
+      1. [=ReadableStreamDefaultReader/Read a chunk=] from |reader| given
+         |read request|.
+
+   1. Wait until |promise| is resolved, or |timer|'s [=timer/timeout fired flag=]
```
OrKoN: should this wait until promise is settled instead of resolved?

index.bs
```diff
@@ -7690,6 +7724,364 @@ is legacy, user agent might run [=implementation-defined=] steps to respect the
...
+1. If |realm id| is not null:
+
+   1. Let |realm| be the result of [=trying=] to [=get a realm=] with |realm id|.
+
+   1. Let |environment settings| be the [=environment settings object=] whose
+      [=realm execution context=]'s Realm component is |realm|.
+
+   1. Let |stream map| be |session|'s [=readable stream settings map=].
+
+   1. Wait until |stream settings|[[=stream settings/read command active=]] is
```
OrKoN: should we disallow concurrent reads by returning an error early? currently it appears that multiple read commands can be sent and then they will all resume at once on the previous read resolves.

index.bs
```diff
@@ -7690,6 +7724,364 @@ is legacy, user agent might run [=implementation-defined=] steps to respect the
...
+
+      1. Return [=error=] with [=error code=] [=timeout=].
+
+   1. Let |read result| be <a spec=ECMASCRIPT>Await</a>(|promise|).
+
+   1. Set |stream settings|[[=stream settings/read promise=]] to null.
+
+   1. If |read result|.\[[Type]] is <code>throw</code>:
+
+      1. [=ReadableStreamDefaultReader/Release=] |reader|.
```
OrKoN: I think this might double release if a pending read was cancelled by the a cancel command

index.bs
```diff
@@ -7690,6 +7724,364 @@ is legacy, user agent might run [=implementation-defined=] steps to respect the
...
+   1. If |promise| is still pending and |timer|'s [=timer/timeout fired flag=]
+      is set:
+
+      1. [=Clean up after running script=] with |environment settings|.
+
+      1. Set |stream settings|[[=stream settings/read command active=]] to false.
+
+      1. Return [=error=] with [=error code=] [=timeout=].
+
+   1. Let |read result| be <a spec=ECMASCRIPT>Await</a>(|promise|).
```
OrKoN, line 8045: just to confirm: this does not actually throw but always returns the result?

index.bs
```diff
@@ -7690,6 +7724,364 @@ is legacy, user agent might run [=implementation-defined=] steps to respect the
...
+      false or |stream map| no longer [=map/contains=] |stream id|.
+
+   1. If |stream map| no longer [=map/contains=] |stream id|, return [=error=]
+      with [=error code=] [=no such stream=].
+
+   1. Set |stream settings|[[=stream settings/read command active=]] to true.
+
+   1. Let |stream| be |stream settings|[[=stream settings/stream=]].
+
+   1. [=Prepare to run script=] with |environment settings|.
```
OrKoN, line 7981: should we call `Clean up after running script` right after we scheduled a read? Could the current version affect the event loop negatively?

index.bs
```diff
@@ -7690,6 +7724,364 @@ is legacy, user agent might run [=implementation-defined=] steps to respect the
...
+
+      1. Let |response| be a new <code>io.ReadData</code> with the <code>value</code>
+         field set to null and the <code>done</code> field set to true.
+
+   1. Otherwise:
+
+      1. Let |chunk| be |read value|["<code>value</code>"].
+
+      1. Let |response| be a new <code>io.ReadData</code> with the <code>value</code>
+         field set to [=forgiving-base64 encode=] |chunk| and the <code>done</code> field
```
OrKoN: I think the stream could yield non binary chunks? i.e., strings or objects? how do we ensure that we handle only byte sequences here?

index.bs
lutien: Yes, that's right, updated.

index.bs
lutien: Sure, updated to return earlier

index.bs
lutien: Yeah, I guess it's possible. I've added the check to make sure that a reader has not been released already.

index.bs
lutien: Yes, it will not throw here

index.bs
lutien: We could, but then we will have to run "prepare to run script" again for awaiting on a promise.

index.bs
lutien: Yeah, I guess it's possible. Ok, I've added the check for it, and we will reject the promise in this case.

jgraham: I guess a question is whether we want to just handle those in the same way we would for any other kind of JS value.

index.bs
OrKoN: I think the current version might not be correct w.r.t. the event loop and I am not sure that we need it again for awaiting the promise since it's should be in the microtask queue for the script we ran to create the promise?

lutien: Ok, I think you're probably right and we don't need it for await. But I think we still need to release a reader? But maybe it still looks better? I've updated to clean up earlier but also prepare a script for a reader release

index.bs
lutien: Hmm... you mean as RemoveValue or something else?

OrKoN: I guess for JS streams we just might want to invoke serialization algorithm here? i.e., https://www.w3.org/TR/webdriver-bidi/#serialize-as-a-remote-value. Which would mean not returning a ByteValue but returning a RemoteValue and possibly changing the serialization algorithm to optionally serialize binary data as base64? I think this might another argument for not doing JS-backed streams and focus on WebDriver BiDi streams that would be dealing with binary data for sure.

jgraham: I think it's sort of an argument for the opposite, in that focusing on the general case has found a bug that we missed due to thinking too much about the network data case.

OrKoN: My point is that there seems to be no clear use case for having this API apply to JS-supplied streams (except for maybe saving a few lines of glue code), as it is already sufficiently covered by existing script evaluation primitives. In fact, one can argue that a userland solution would be always simpler as the consumer likely knows what sort of stream they are dealing with and do not need to use alternative mechanism for serialization just because the resulting object happened to be a stream. A general case solution does not seem to be required for streams not backed by JS. Do we have a client library that would definitely (and not hypothecally) benefit from this? I think like the API shape is great here and we could just re-use it for streams produced by the WebDriver BiDi implementation.

jgraham: But yes, I agree that we'd need some modifications to the serialize-as-a-remote-value algorithm that would allow Uint8Array values to be optionally serialized to base64. I guess the main difficulty is that in the stream case a lot of the values are temporaries, so providing a handle to them doesn't make much sense. But making it possible to serialize a UInt8Array as base 64 seems independently useful.

lutien:
Alright, I've removed the exposure of streams as a remote value, and I've added the `type` field to `io.ReadData`. For now I've only kept the serialization of bytes, and I've added the note that we will extend for other types using the remote value interface. @jgraham and @OrKoN, could you take another look? Maybe we could get this part merged first, then work on the specific consumers as follow-up work?

lutien:
@sadym-chromium, maybe you want to take a look at the update instead of Alex?

Issue #791, opened 2024-09-26
Update CDDL generation script to also read external definitions


whimboo:
We have external definitions of WebDriver BiDi command and events like this on-exhaustive list shows:

https://w3c.github.io/webdriver-bidi/#external-specifications

Right now we do not yet read their CDDL defintions which means even if implemented in browsers clients will not know about in case they rely on auto-generated code.

We should probably update our cddl generation script to check external documents as well.

CC @jgraham @gsnedders

gsnedders:
We should be careful to not accidentally change the definition of the [Remote End Definition](https://w3c.github.io/webdriver-bidi/#remote-end-definition) (or Local End Definition) if/when we do this; various things are matched against it, and if we explicitly include all external specifications in the definition, does that not oblige them to be implemented?

whimboo:
The inclusion of any external spec would require an agreement by the Browser Testing and Tools group for sure. Further it would apply to any other CDDL changes afterward. 

If we don't agree on that what else could we do to provide a CDDL export that also includes all the extension commands and events?

tidoust:
I haven't had time to dig into it yet but I plan to work on adding CDDL extracts to Webref data in a not so distant future, see https://github.com/w3c/webref/issues/1353.

There are also interesting considerations on the notion of CDDL modules in a [recent IETF thread](https://mailarchive.ietf.org/arch/msg/tools-discuss/-oa5-sKE6OdRPx63f9KaQJHmVlE/) (initially focused on ABNF extracts), which could perhaps be used to create a consolidated view.

tidoust:
Webref now has [CDDL extracts](https://github.com/w3c/webref/tree/curated/ed/cddl) for individual specifications (for WebDriver BiDi related ones, one extract for the remote end definition, one for the local end definition, and one that contains all definitions). Extracts in the `curated` branch are guaranteed to be valid CDDL.

What's missing is creating a merged view of WebDriver BiDi related CDDL extracts from the list in [External specifications](https://w3c.github.io/webdriver-bidi/#external-specifications).

It would be easy to merge relevant CDDL extracts from Webref automatically, starting from the WebDriver BiDi one, but something seems missing: the extension modules currently do not link back to the core CDDL constructs defined in WebDriver BiDi. For example, taking the example of [Web Bluetooth](https://webbluetoothcg.github.io/web-bluetooth/#automated-testing), I would expect that spec to define something like:

```cddl
CommandData //= BluetoothCommand
[...]

PR #1140, opened 2026-07-24
Add a CDDL extension point for webExtension.install parameters

titusfortner:
Adds a per-command CDDL extension point so vendor-specific command parameters can be declared with their own types in the remote end definition. Addresses the parameter-map case of w3c/webdriver-bidi#274, and gives the auto-generated clients in w3c/webdriver-bidi#791 something typed to generate from.

### Context

- `webExtension.InstallParameters` is a closed CDDL map, so a remote end that accepts extra install parameters cannot describe them in the remote end definition.
- Concretely, Firefox ships `moz:permanent` and `moz:allowPrivateBrowsing` on `webExtension.install`. Mozilla is maintaining a machine-readable `Extensions.cddl` for these ([bug 2057588](https://bugzilla.mozilla.org/show_bug.cgi?id=2057588)); this PR adds the base-side extension point that vendor file plugs into.

### Change

- Adds a `webExtension.InstallParametersExtension` group, referenced from `InstallParameters`, whose value is `Extensible`. A vendor or extension spec declares its specific fields by extending this group with a group choice (`//=`), e.g. `webExtension.InstallParametersExtension //= (? "moz:permanent": bool)`.
- Documents the convention generally: an extensible parameter map exposes a named `…Extension` group that vendors extend, with vendor-prefixed field names.

### Notes

- **Forward-compatible and additive.** The group's value is `Extensible` (`* text => any`), so the base still accepts unknown keys exactly as the envelope types do; adding it changes nothing about what `InstallParameters` validates today.
- **Typed, not just tolerated.** Unlike a bare `Extensible` in the map, the named group gives vendor CDDL a stable target to attach *typed* fields to. A code generator can then emit real typed parameters (e.g. a `boolean permanent` argument) instead of an untyped string→any map — which is the difference between a usable and an unusable API in statically-typed client bindings.
- **Scoped to command parameter maps.** These are dispatched by `method`, so they avoid the structural-typing ambiguity behind the broader discussion in w3c/webdriver-bidi#274.
- **Named group rather than an RFC 8610 socket.** A `$$…` group socket would be the idiomatic construct, but the spec toolchain (Bikeshed / cddlparser) has no socket support — it treats every `//=` as extending a prior `=` definition. A named group whose value is `Extensible` is the equivalent the toolchain accepts, and reuses the existing `Extensible` production.

***
<a href="https://pr-preview.s3.amazonaws.com/titusfortner/webdriver-bidi/pull/1140.html" title="Last updated on Aug 6, 2026, 5:38 PM UTC (9ae210c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/1140/a7b8b67...titusfortner:9ae210c.html" title="Last updated on Aug 6, 2026, 5:38 PM UTC (9ae210c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -730,13 +730,10 @@ Authors of external specifications are encouraged to to add new modules rather t
 existing ones. Where it is desired to extend an existing module, it is preferred to integrate the
 extension directly into the specification containing the original module definition.
 
-Some [=command parameters=] types include a dedicated [[!RFC8610|CDDL]]
-extension group, named after the parameters type with an "<code>Extension</code>"
-suffix and having the value <code>Extensible</code>. An implementation may add
-vendor-specific fields to such a type by extending that group with a group
-choice ("<code>//=</code>"). To avoid conflicts with other implementations, each
-added field name must begin with a prefix that is unique to the implementation,
-followed by a U+003A (:) character.
+An implementation may add vendor-specific fields to the [=command parameters=]
+of a [=command=]. To avoid conflicts with other implementations, each such field
+name must begin with a prefix that is unique to the implementation, followed by
+a U+003A (:) character.
 
 ## Commands ## {#commands}
 
@@ -15110,10 +15107,7 @@ The <dfn export for=commands>webExtension.install</dfn> command installs a web e
 
       webExtension.InstallParameters = {
          extensionData: webExtension.ExtensionData,
-         webExtension.InstallParametersExtension,
       }
-
-      webExtension.InstallParametersExtension = ( Extensible )
 
       webExtension.ExtensionData = (
          webExtension.ExtensionArchivePath /
--- a/index.bs
+++ b/index.bs
@@ -730,6 +730,16 @@ Authors of external specifications are encouraged to to add new modules rather t
 existing ones. Where it is desired to extend an existing module, it is preferred to integrate the
 extension directly into the specification containing the original module definition.
 
+An implementation may extend a [=command=]'s [=command parameters=] with
+vendor-specific fields. For this to be expressible in the {^remote end
+definition^}, a parameter map that permits such extension exposes a
+[[!RFC8610|CDDL]] group socket whose name is the parameters type name
+followed by "<code>.extension</code>", prefixed with "<code>$$</code>", and
+whose default is the empty group. An [=extension module=] adds fields by
+augmenting that socket. As with the [=module names=] of [=extension modules=],
+each field name added in this way must begin with a prefix that is unique for a
+given implementation, followed by a U+003A (:) character.
+
 ## Commands ## {#commands}
 
 A <dfn export>command</dfn> is an asynchronous operation, requested by
@@ -15102,7 +15112,10 @@ The <dfn export for=commands>webExtension.install</dfn> command installs a web e
 
       webExtension.InstallParameters = {
          extensionData: webExtension.ExtensionData,
+         $$webExtension.InstallParameters.extension,
       }
+
+      $$webExtension.InstallParameters.extension //= ()
 
       webExtension.ExtensionData = (
          webExtension.ExtensionArchivePath /
--- a/index.bs
+++ b/index.bs
@@ -730,15 +730,13 @@ Authors of external specifications are encouraged to to add new modules rather t
 existing ones. Where it is desired to extend an existing module, it is preferred to integrate the
 extension directly into the specification containing the original module definition.
 
-An implementation may extend a [=command=]'s [=command parameters=] with
-vendor-specific fields. For this to be expressible in the {^remote end
-definition^}, a parameter map that permits such extension includes a dedicated
-[[!RFC8610|CDDL]] group, named after the parameters type with an
-"<code>Extension</code>" suffix, whose value is <code>Extensible</code>. As
-with the [=module names=] of [=extension modules=], each such field name must
-begin with a prefix that is unique for a given implementation, followed by a
-U+003A (:) character. An [=extension module=] may declare the specific fields it
-adds by extending this group with a group choice ("<code>//=</code>").
+Some [=command parameters=] types include a dedicated [[!RFC8610|CDDL]]
+extension group, named after the parameters type with an "<code>Extension</code>"
+suffix and having the value <code>Extensible</code>. An implementation may add
+vendor-specific fields to such a type by extending that group with a group
+choice ("<code>//=</code>"). To avoid conflicts with other implementations, each
+added field name must begin with a prefix that is unique to the implementation,
+followed by a U+003A (:) character.
 
 ## Commands ## {#commands}
 
--- a/index.bs
+++ b/index.bs
@@ -732,11 +732,11 @@ extension directly into the specification containing the original module definit
 
 An implementation may extend a [=command=]'s [=command parameters=] with
 vendor-specific fields. For this to be expressible in the {^remote end
-definition^}, a parameter map that permits such extension exposes a
-[[!RFC8610|CDDL]] group socket whose name is the parameters type name
-followed by "<code>.extension</code>", prefixed with "<code>$$</code>", and
-whose default is the empty group. An [=extension module=] adds fields by
-augmenting that socket. As with the [=module names=] of [=extension modules=],
+definition^}, a parameter map that permits such extension includes a dedicated
+[[!RFC8610|CDDL]] group, named after the parameters type with an
+"<code>Extension</code>" suffix, that is empty by default. An [=extension
+module=] adds fields by extending that group with a group choice
+("<code>//=</code>"). As with the [=module names=] of [=extension modules=],
 each field name added in this way must begin with a prefix that is unique for a
 given implementation, followed by a U+003A (:) character.
 
@@ -15112,10 +15112,10 @@ The <dfn export for=commands>webExtension.install</dfn> command installs a web e
 
       webExtension.InstallParameters = {
          extensionData: webExtension.ExtensionData,
-         $$webExtension.InstallParameters.extension,
+         webExtension.InstallParametersExtension,
       }
 
-      $$webExtension.InstallParameters.extension //= ()
+      webExtension.InstallParametersExtension = ()
 
       webExtension.ExtensionData = (
          webExtension.ExtensionArchivePath /
--- a/index.bs
+++ b/index.bs
@@ -734,11 +734,11 @@ An implementation may extend a [=command=]'s [=command parameters=] with
 vendor-specific fields. For this to be expressible in the {^remote end
 definition^}, a parameter map that permits such extension includes a dedicated
 [[!RFC8610|CDDL]] group, named after the parameters type with an
-"<code>Extension</code>" suffix, that is empty by default. An [=extension
-module=] adds fields by extending that group with a group choice
-("<code>//=</code>"). As with the [=module names=] of [=extension modules=],
-each field name added in this way must begin with a prefix that is unique for a
-given implementation, followed by a U+003A (:) character.
+"<code>Extension</code>" suffix, whose value is <code>Extensible</code>. As
+with the [=module names=] of [=extension modules=], each such field name must
+begin with a prefix that is unique for a given implementation, followed by a
+U+003A (:) character. An [=extension module=] may declare the specific fields it
+adds by extending this group with a group choice ("<code>//=</code>").
 
 ## Commands ## {#commands}
 
@@ -15115,7 +15115,7 @@ The <dfn export for=commands>webExtension.install</dfn> command installs a web e
          webExtension.InstallParametersExtension,
       }
 
-      webExtension.InstallParametersExtension = ()
+      webExtension.InstallParametersExtension = ( Extensible )
 
       webExtension.ExtensionData = (
          webExtension.ExtensionArchivePath /


lutien:
@tidoust, do you know if that's an ok way to extend CDDL like this, or should we just set it as `Extencible`?

tidoust:
> @tidoust, do you know if that's an ok way to extend CDDL like this, or should we just set it as `Extencible`?

Hmm, my CDDL parser seems to accept `webExtension.InstallParametersExtension = ()` but that's a bug: the [CDDL grammar](https://www.rfc-editor.org/info/rfc8610/#appendix-B) requires a `type` in the parentheses and I don't see any way to produce an empty one (at a minimum, a type is a `value` or an `id`, neither of which can be empty).

I believe the definition should rather be:

```cddl
webExtension.InstallParametersExtension = ( Extensible )
```

lutien:
> > @tidoust, do you know if that's an ok way to extend CDDL like this, or should we just set it as `Extencible`?
> 
> Hmm, my CDDL parser seems to accept `webExtension.InstallParametersExtension = ()` but that's a bug: the [CDDL grammar](https://www.rfc-editor.org/info/rfc8610/#appendix-B) requires a `type` in the parentheses and I don't see any way to produce an empty one (at a minimum, a type is a `value` or an `id`, neither of which can be empty).
> 
> I believe the definition should rather be:
> 
> ```
> webExtension.InstallParametersExtension = ( Extensible )
> ```

Ah, right. I think that makes sense

titusfortner:
You're right, the cddl npm also accepted it as empty so it has the same bug as yours. I've updated the code as you suggested.

Setting it as Extensible to support vendor extensions (like `storage.PartitionKey` explicitly calls out) would work, but my code generator could only produce an untyped map from it. 

Mozilla agreed to maintain a vendored cddl file for me to reference with their addons, so a named Extension group here would be preferred. It still uses Extensible as its value, so forward compatibility is unchanged, but it gives my generator a typed target to attach their fields to.

It will be the difference between supporting in my code:
```java
installParameters.setMozPermanent(true)
```
and
```java
installParameters.setExtension("moz:permanent", true)
```

lutien (approved):
That looks fine to me, but I would like one other vendor's representative to review before we merge to make sure we are all fine with the general approach. @OrKoN or @sadym-chromium, could you take a look as well?

index.bs
```diff
@@ -730,6 +730,16 @@ Authors of external specifications are encouraged to to add new modules rather t
 existing ones. Where it is desired to extend an existing module, it is preferred to integrate the
 extension directly into the specification containing the original module definition.
 
+An implementation may extend a [=command=]'s [=command parameters=] with
+vendor-specific fields. For this to be expressible in the {^remote end
+definition^}, a parameter map that permits such extension includes a dedicated
+[[!RFC8610|CDDL]] group, named after the parameters type with an
+"<code>Extension</code>" suffix, whose value is <code>Extensible</code>. As
```
OrKoN: Thanks for the PR. Is there a reason to require certain authoring style for CDDL types? I think it's sufficient to say that anything that is `Extensible` can be extended. Additionally, I think a while ago we established that CDDL naturally does not prevent extensions so `Extensible` might not be necessary at all. I support documenting that vendor prefixes should be used for extension attributes.

index.bs
```diff
@@ -730,6 +730,16 @@ Authors of external specifications are encouraged to to add new modules rather t
...
+with the [=module names=] of [=extension modules=], each such field name must
+begin with a prefix that is unique for a given implementation, followed by a
+U+003A (:) character. An [=extension module=] may declare the specific fields it
```
OrKoN: IIRC extension modules have to extend entire domains and cannot extend specific commands or parameters. It sounds like the last sentence in this paragraph is not necessary then?

titusfortner:
Bare `Extensible` tolerates extra fields but gives nowhere to define what those vendor fields actually are; the named group is a defined attachment point where a vendor's typed fields live in a merged CDDL. My intention with this PR is to explicitly define how vendors should create their own cddl files that I can parse.

My original implementation used a group socket, which is the CDDL-native mechanism to do this, but Bikeshed’s CDDL markup currently does not support sockets. We could update `bikeshed/cddl.py` to support sockets, but that's more than I want to do here. `...Extension` might be awkward, but it defines the attachment-point behavior with the current tooling.

Alternately, if there's another way for vendors to agree on a standard convention for extending a cddl file (full definition, just the vendor parameters, use `=` or use `//=`) that would work. Also, I'm not sure how much vendor-specific functionality to expect, so maybe this is a lot of work for 2 extra parameters.

Also agreed that "extension module" is the wrong term here; this is about extension fields on an existing command's parameters. Vendor-specific fields should use a vendor prefix. I've updated the PR.

OrKoN:
Thanks for the explanation. I am not sure if it's really worth rewriting every CDDL command following certain conventions to allow extensibility by implementations (other specs is a different question but we agreed that those should extend on the module level) and I am not sure if all vendors have to all extend the spec CDDL files in the same way and re-publish the final interface as CDDL. Could vendors who choose to do so patch the CDDL (AST-based to be less brittle) from the spec and the publish the implementation-specific CDDL? It sounds doable to me and I think this would save us the work of naming every possible extension point one by one in this spec. 

To summarize: I think documenting the naming conventions for implementation-defined field that this PR adds is still worth landing but I would exclude the part requiring named extension points and `Extension` suffixes. WDYT?

titusfortner:
That makes sense. I've updated the PR to just the field-naming convention, and dropped the CDDL change entirely since the map does not need to be marked extensible for vendors to patch their own.

jimevans:
Here's my confusion. `Command` is marked with `Extensible`. Individual `CommandData` objects are not so marked. To me, that says I can do:
```json
{
  "id": 1,
  "method": "module.method",
  "params": {
    "namedParam": "parameterValue"
  },
  "foo:someExtensionProperty": "extensionPropertyValue"
}
```

but could **not** do:
```json
{
  "id": 1,
  "method": "module.method",
  "params": {
    "namedParam": "parameterValue"
    "foo:someExtensionProperty": "extensionPropertyValue"
  }
}
```

I'm sure I'm misinterpreting that, because the former makes less sense than the latter in the context of what a vendor would want to extend. Given the current CDDL, are both constructs legal? If the latter construct is legal, is that explicitly called out somewhere, or is CDDL just that lax where anything goes? This seems like a potential area for misinterpretation.

OrKoN:
I think we still need to mark types to include `Extensible` when needed, we just do not need to enforce a specific way to structure CDDL for interoperability with downstream parsers.

index.bs
```diff
@@ -730,6 +730,11 @@ Authors of external specifications are encouraged to to add new modules rather t
 existing ones. Where it is desired to extend an existing module, it is preferred to integrate the
 extension directly into the specification containing the original module definition.
 
+An implementation may add vendor-specific fields to the [=command parameters=]
+of a [=command=]. To avoid conflicts with other implementations, each such field
```
OrKoN, line 734:
```suggestion
An implementation may add vendor-specific fields to <code>Extensible</code> types. To avoid conflicts with other implementations, each such field
```

jgraham:
I sort of vaguely remember that when we looked at this before it seemed like a conformant CDDL implementation should ignore any unknown fields, so `Extensible` was more of a hint that this is a point where the spec expects to allow extensions (and of course we can try to impose a post-processing step that guarantees no other extensions, but in theory that makes it harder to implement on top of a conforming CDDL implementation; I don't know if real world implementations actually allow arbitrary extra fields or not).

If we do adopt this, I think Jim is basically right and we've put `Extensible` in the wrong layers; we want the command parameters, event parameters, etc. to be extensible rather than the wrapper objects.

nvborisenko:
I interpret spec that it allows extensible at command envelop level and its parameters:

```
Command = {
  id: js-uint,
  CommandData,
  Extensible,
}
```

And parameters, I thought all command parameters are inherited from `EmptyParams`:
```
EmptyParams = {
   Extensible
}
```

Note: Code blocks are taken from https://w3c.github.io/webdriver-bidi/

jgraham:
> I interpret spec that it allows extensible at command envelop level

This is true, but not what people want for most use cases. It's only really useful if you're trying to add something which applies to all payloads like some concept of sub-sessions where the transport layer doesn't correspond 1:1 with a specific session.

> And parameters, I thought all command parameters are inherited from EmptyParams

This is not true. For example 

```
browser.CreateUserContextParameters = {
  ? acceptInsecureCerts: bool,
  ? unhandledPromptBehavior: session.UserPromptHandler
}
```

This currently doesn't contain `EmptyParams` or `Extensible` anywhere in the definition.

titusfortner:
After review, I think I may have been asking for something that isn't necessary, which is likely why it wasn't already here. If a conformant consumer ignores unknowns, then the goal isn't ensuring a record allows everything, but allowing the consumer to maintain the collection of sources that describe the knowns (as already needed for bluetooth, digital credentials, permissions, etc). If so, then we don't need to add Extensible markers everywhere a vendor may want to provide custom fields. 

I think the change in this PR now should suffice, and I'll work with Mozilla on the shape of https://phabricator.services.mozilla.com/D314395.

index.bs
whimboo: @titusfortner I second here that we should make this change as proposed by @OrKoN but with a small additional change: `… fields to any map that includes <code>Extensible</code>` because `Extensible` is not a type but a group.

whimboo (changes requested):
It's not only commands that could be extended with vendor specific parameters, but we can have as well vendor commands and even modules. Adding such a paragraph we probably should include it as well.

Please note that for `webExtension.InstallParameters` we are not allowed to have our vendor specific `moz:permanent` parameter because the CDDL doesn't allow the command for additional parameters.

titusfortner:
Let me summarize things as I see them.

What a remote end needs to implement is already spread across multiple specs, so absence from this one isn't the limiting condition for compliance. My intent for this PR is to define how to support vendor-maintained fragments so Selenium can explicitly validate against them.

If we aren't using sockets (which I recognize is more trouble than it is worth), then vendor fragments cannot be directly merged with the other CDDL files and still validate. Selenium's generator will need to parse them separately, and that's fine. This is the case regardless of whether things are marked Extensible, so from my perspective it's a separate concern.

The issue I think needs to be addressed is whether the suggested wording is intended to require every unspecified field on a map marked Extensible to have a vendor prefix. If so, that would let us validate payloads, which is what I find valuable, but I don't think it's feasible. `script.RemoteObjectReference` relies on unprefixed extra keys by design, but those keys can't be distinguished from "vendor-specific fields". That's why I'm proposing the naming requirement be limited to vendor fragments explicitly defined elsewhere.

So, I've updated the PR text to reference defining vendor fragments and the required naming conventions for such commands, events, and fields (modules are already covered by the previous paragraph), without referencing Extensibility.

whimboo:
CC'ing @hbenl who is actually working on a merge tool for Firefox. He has already some prototype running and could give helpful feedback.

hbenl:
See also https://bugzilla.mozilla.org/show_bug.cgi?id=2057588#c6 - that's how we intend to manage Firefox vendor extensions (which follow the naming requirement specified in this PR) and shows how we were able to autogenerate Typescript types that include our extensions.

sadym-chromium (approved):

titusfortner:
Thanks @hbenl, yes, that works well for Selenium. I recently pulled in the `Debugging`, `Fields` and `Profiler` fragments you created to generate our code in [SeleniumHQ/selenium#18060](https://github.com/SeleniumHQ/selenium/pull/18060). (we'll update the code to grab the latest version of what you publish instead of including it in our code directly once it goes live). 

We're parsing the `<Type>Extension` convention, attaching each group to its spec type but keeping the `moz:` fields in a separate vendor section rather than folding them into the shared types. That gives us typed `moz:` parameters and the `moz:debugging` and `moz:profiler` modules without needing to mark things in this spec as `Extensible`.

hbenl:
@titusfortner Note that the latest version of the Firefox patch also has a `Commands.cddl` containing a vendor extension command to an existing module (`webExtension.moz:listExtensions`).

titusfortner:
@hbenl thanks, already in progress: https://github.com/SeleniumHQ/selenium/pull/18071
Once you get these published we'll figure out how to dynamically get the list

index.bs
```diff
@@ -730,6 +730,15 @@ Authors of external specifications are encouraged to to add new modules rather t
 existing ones. Where it is desired to extend an existing module, it is preferred to integrate the
 extension directly into the specification containing the original module definition.
 
+An implementation may augment the {^remote end definition^} and {^local end
+definition^} with its own definition fragments that add vendor-specific fields
+to existing maps, or vendor-specific [=commands=] and [=events=] to existing
+modules. To avoid conflicts with other implementations and with future versions
+of this specification, each such field name must be of the form
+<code>[prefix]:[name]</code> and each such [=command name=] or [=event name=]
+must be of the form <code>[module name].[prefix]:[name]</code>, where
```
whimboo, line 739: So what we actually never discussed is how the naming rules should actually be defined for commands and events. Here we claim that any vendor specific command and event should actually be part of the public module. But we could also have a vendor module with no prefixed command/event name like `[prefix]:[module].[name]`. 

I think that we should make sure that we all agree, which one of those two we should use.

PR #1143, opened 2026-08-04
Update "known subscription ids" when the list of subscribed contexts is empty after a context is destroyed.

lutien:
***
<a href="https://pr-preview.s3.amazonaws.com/lutien/webdriver-bidi/pull/1143.html" title="Last updated on Aug 4, 2026, 4:19 PM UTC (99a23f8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/1143/6a2835c...lutien:99a23f8.html" title="Last updated on Aug 4, 2026, 4:19 PM UTC (99a23f8)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5795,6 +5795,8 @@ the <dfn export>WebDriver BiDi navigable destroyed</dfn> steps given [=/navigabl
 
   1. Let |subscriptions to remove| be a [=/set=].
 
+  1. Let |subscription ids to remove| be a [=/set=].
+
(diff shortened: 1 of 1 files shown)


jgraham (approved):

lutien:
Looking at it again, I think it actually was intentional to avoid throwing an error if the client tries to unsubscribe using the subscription id, for the subscription that was cleaned up when the browsing context was destroyed.

PR #1144, opened 2026-08-06
fix: make userContext mandatory in event payloads

### LINKED ISSUE #1071: Make userContext properties mandatory for events

AutomatedTester:
#1046 added `userContext` to the payloads of events that carry a browsing context, and to `browsingContext.CreateResult`, but marked it optional as a transition measure (#1071). The navigable, and therefore its associated user context, is always known in these algorithms, so the field is always set. Make it mandatory.

The download events splice in `browsingContext.BaseNavigationInfo` but copy fields out of the navigation info one at a time, so the second commit copies `userContext` across as well.

`browsingContext.CreateParameters.userContext` stays optional, since it is a command input and omitting it is meaningful. `script.Source.userContext` stays optional because its `context` field is optional too: a script related event can occur in a realm with no associated navigable.

This also covers the `browsingContext.CreateResult` change from #1122, which was closed in favour of doing it here.

Tests: web-platform-tests/wpt#61830

Fixes #1071

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3739,7 +3739,7 @@ browsingContext.BaseNavigationInfo = (
   navigation: browsingContext.Navigation / null,
   timestamp: js-uint,
   url: text,
-  ? userContext: browser.UserContext,
+  userContext: browser.UserContext,
 )
 
 browsingContext.NavigationInfo = {
@@ -4239,7 +4239,7 @@ The <dfn export for=commands>browsingContext.create</dfn> command creates a new
     <pre class="cddl" data-cddl-module="local-cddl">
         browsingContext.CreateResult = {
           context: browsingContext.BrowsingContext,
-          ? userContext: browser.UserContext
+          userContext: browser.UserContext
         }
     </pre>
    </dd>
@@ -5907,7 +5907,7 @@ given [=/navigable=] |navigable| and [=WebDriver BiDi navigation status|navigati
           context: browsingContext.BrowsingContext,
           timestamp: js-uint,
           url: text,
-          ? userContext: browser.UserContext
+          userContext: browser.UserContext
         }
       </pre>
    </dd>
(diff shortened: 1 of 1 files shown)


AutomatedTester:
Tests: https://github.com/web-platform-tests/wpt/pull/61830

whimboo:
> Note, unrelated to this change and possibly worth its own issue: the `browsingContext.downloadEnd` steps also never set the `status` field the two productions require, and they assert on `|navigation info|["status"]`, which is not a field of `browsingContext.NavigationInfo`.

@AutomatedTester would you mind filing an issue for that or if doable create a PR? Thanks!

index.bs
```diff
@@ -8471,14 +8474,17 @@ network.BaseParameters = (
...
-    ? userContext: browser.UserContext / null,
+    userContext: browser.UserContext / null,
     ? intercepts: [+network.Intercept]
 )
 </pre>
 
 The <code>network.BaseParameters</code> type is an abstract type representing
 the data that's common to all network events.
 
+Note: the <code>userContext</code> field is null if and only if the
```
whimboo: `context` can be `null` as well. So maybe we should include it here as well, similar to what you already did below for `script.source`?

whimboo (changes requested):
Overall this looks fine, but please trim the commit message for this PR as well.

index.bs
AutomatedTester: Reworded the note to cover both fields, mirroring the `script.Source` one:

> Note: the `context` and `userContext` fields are either both non-null, when the request is associated with a navigable, or both null.

`process a network event` sets `navigable id` and `user context id` from the same navigable in the same step, so they are always both null or both set.

AutomatedTester:
Trimmed both commit messages and the PR description, and filed #1158 for the `browsingContext.downloadEnd` `status` field.

Force-pushed: the only content change from the previous revision is the reworded `network.BaseParameters` note.

whimboo (approved):
Looks good to me. But we currently cannot land this until Chrome has actually support for the `userContext` field in events.

@sadym-chromium maybe you could give us some details when you may be able to implement this field?

rodeskike-art (changes requested):
Request Changes

Issue #1134, opened 2026-07-13
Alternative screencast API using screenshots


hbenl:
CDP offers an [experimental screencasting API](https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-startScreencast) that sends screenshots to the client whenever the page was rendered.
Playwright uses this API (and corresponding APIs with the other non-BiDi protocols) for its screencasting and it recently started [exposing the screenshots](https://playwright.dev/docs/release-notes#-screencast) directly to the tests (using the `onFrame` callback).
In contrast, WebDriver BiDi's `browsingContext.startScreencast` command creates a video file in the browser and saves it to disk.
So the current implementation of screencasting with Playwright/BiDi uses `browsingContext.captureScreenshot` instead, requesting 25 screenshots per second. This has some obvious downsides:
- the client doesn't know when the page is rendered, so it always requests 25 screenshots per second whereas the CDP API stops sending screenshots when the page is static
- `browsingContext.captureScreenshot` doesn't have an option to scale the screenshots down to a maximum width/height

Playwright/BiDi could switch to `browsingContext.startScreencast` once it supports streaming (#959), addressing both downsides:
- the network overhead of the video stream should be minimal when the page is static (and network overhead should generally be much lower for a video stream compared to a series of screenshots)
- `browsingContext.startScreencast` can scale the video down

However, to support the `onFrame` callback of Playwright's screencast API (and similar APIs), it would have to be able to extract screenshots from the video stream, adding significant complexity to the implementation, so we should consider adding an alternative screencast API similar to that of CDP.

PR #1145, opened 2026-08-07
Add "imageSize" argument to "browsingContext.captureScreenshot" command.

lutien:
Covers the resizing item in #1134.

***
<a href="https://pr-preview.s3.amazonaws.com/lutien/webdriver-bidi/pull/1145.html" title="Last updated on Aug 11, 2026, 2:14 PM UTC (6b9be43)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/1145/6a2835c...lutien:6b9be43.html" title="Last updated on Aug 11, 2026, 2:14 PM UTC (6b9be43)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -132,6 +132,8 @@ spec: CONSOLE; urlPrefix: https://console.spec.whatwg.org
 spec: ECMASCRIPT-I18N; urlPrefix: https://tc39.es/ecma402/
   type: dfn
     text: DefaultLocale; url: #sec-defaultlocale
+    text: finite; url: #finite
+    text: integral Number; url: #integral-number
     text: IsStructurallyValidLanguageTag; url: #sec-isstructurallyvalidlanguagetag
 spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
   type: dfn
@@ -3983,6 +3985,19 @@ To <dfn>rectangle intersection</dfn> given |rect1| and |rect2|
 </div>
 
 <div algorithm>
+To <dfn>round a number</dfn> given |number|:
+
+1. If |number| is not [=finite=] or |number| is an [=integral Number=], return |number|.
+
+1. If |number| < 0.5𝔽 and |number| > +0𝔽, return +0𝔽.
+
+1. If |number| < -0𝔽 and n ≥ -0.5𝔽, return -0𝔽.
+
+1. Return the [=integral Number=] closest to |number|, preferring the Number closer to +∞ in the case of a tie.
+
+</div>
+
+<div algorithm>
 To <dfn>render document to a canvas</dfn> given |document|, |rect| and |image size|:
 
 1. Let |ratio| be [=determine the device pixel ratio=] given |document|'s
@@ -4012,9 +4027,13 @@ To <dfn>render document to a canvas</dfn> given |document|, |rect| and |image si
       height| is less than |scale|, set |scale| to |max height| divided by
       |framebuffer height|.
 
-1. Let |paint width| be the maximum of 1 and |framebuffer width| multiplied by |scale|, rounded to the nearest integer.
+1. Let |scale width| be [=round a number=] given |framebuffer width| multiplied by |scale| rounded to the nearest integer.
 
-1. Let |paint height| be the maximum of 1 and |framebuffer height| multiplied by |scale|, rounded to the nearest integer.
+1. Let |scale height| be [=round a number=] given |framebuffer height| multiplied by |scale| rounded to the nearest integer.
+
+1. Let |paint width| be the maximum of 1 and |scale width|.
+
+1. Let |paint height| be the maximum of 1 and |scale height|.
 
 1. Let |canvas| be a new {{HTMLCanvasElement}} with {{HTMLCanvasElement/width}} |paint
    width| and {{HTMLCanvasElement/height}} |paint height|.
@@ -4036,15 +4055,24 @@ To <dfn>render document to a canvas</dfn> given |document|, |rect| and |image si
    points (0, 0), (|framebuffer width|, 0), (|framebuffer width|,
    |framebuffer height|), and (0, |framebuffer height|).
 
-1. Invoke |canvas context|'s {{CanvasTransform/scale()}} method with
+1. Run the steps corresponding to the {{CanvasTransform/scale()}} method on |canvas context| with
    <var ignore>x</var> set to |paint width| divided by |framebuffer width|, and
    <var ignore>y</var> set to |paint height| divided by |framebuffer height|.
 
+   Note: Running these steps doesn't cause any script to execute.
+
 1. Complete implementation specific steps equivalent to painting the region of
    |document|'s framebuffer specified by |source rectangle| on the region of
-   |canvas context|'s [=output bitmap=] specified by |destination rectangle|,
-   after applying |canvas context|'s [=current transformation matrix=] to
-   |destination rectangle|.
+   |canvas context|'s [=output bitmap=] specified by |destination rectangle|.
+
+   When the dimensions of |destination rectangle| are smaller than |source rectangle|,
+   then the value painted at a point in the |destination rectangle| is computed
+   by filtering the data from |document|'s framebuffer. The user agent may use
+   any filtering algorithm (for example bilinear interpolation or nearest-neighbor).
+   When the filtering algorithm requires a pixel value from outside the framebuffer,
+   it must instead use the value from the nearest edge pixel. (That is, the filter uses "clamp-to-edge" behavior.)
+   When the filtering algorithm requires a pixel value from outside the |source rectangle|
+   but inside the framebuffer, then the value from the framebuffer must be used.
 
 1. Return |canvas|.
 
--- a/index.bs
+++ b/index.bs
@@ -226,6 +226,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
     text: create a classic script; url: webappapis.html#creating-a-classic-script
     text: create a new browsing context; url: browsers.html#creating-a-new-browsing-context
     text: create a new top-level traversable; url: document-sequences.html#creating-a-new-top-level-traversable
+    text: current transformation matrix; url: canvas.html#current-transformation-matrix
     text: default script fetch options; url: webappapis.html#default-script-fetch-options
     text: default view; url: nav-history-apis.html#dom-document-defaultview
     text: descendant navigables; url: document-sequences.html#descendant-navigables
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3975,18 +3981,60 @@ To <dfn>rectangle intersection</dfn> given |rect1| and |rect2|
...
+        <dd>
+        1. Set |paint width| to |requested width|.
+
+        1. Set |paint height| to |requested height|.
 
-1. Let |paint height| be |rect|'s [=height dimension=] multiplied by |ratio|,
-   rounded to the nearest integer, so it matches the height of |rect| in device
-   pixels.
+        Note: In this case the aspect ratio of |rect| is not preserved unless it
+        happens to match the aspect ratio of the requested size.
```
hbenl: This doesn't match what CDP's [startScreencast](https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-startScreencast) does (and what Playwright expects): it allows you to specify `maxWidth`/`maxHeight` and will preserve the aspect ratio.

index.bs
```diff
@@ -3998,8 +4046,8 @@ To <dfn>render document to a canvas</dfn> given |document| and |rect|:
 
 1. Complete implementation specific steps equivalent to drawing the region of
    the framebuffer representing the region of |document| covered by |rect| to
-   |canvas context|, such that each pixel in the framebuffer corresponds to a pixel in
-   |canvas context| with (|rect|'s [=x coordinate=], |rect|'s [=y coordinate=]) in
+   |canvas context|, scaling it as needed to cover |canvas context|, with
```
jgraham: I think we should write this as render-then-resize rather than resize-then-render even though it's a bigger change.

index.bs
lutien: Ok, I've updated to use maximum values.

index.bs
lutien: I've updated to use canvas `scale` and then the wording from https://html.spec.whatwg.org/multipage/canvas.html#dom-context-2d-drawimage. It still then says that `This specification does not define the precise algorithm to use when scaling an image down, or when scaling an image up when the [imageSmoothingEnabled](https://html.spec.whatwg.org/multipage/canvas.html#dom-context-2d-imagesmoothingenabled) attribute is set to true.`, so I guess it still might be vague.

juliandescottes (approved):
That looks good to me, thanks Sasha.

OrKoN:
I would leave the review of this to @sadym-chromium

index.bs
```diff
@@ -3975,18 +3983,38 @@ To <dfn>rectangle intersection</dfn> given |rect1| and |rect2|
...
-1. Let |paint height| be |rect|'s [=height dimension=] multiplied by |ratio|,
-   rounded to the nearest integer, so it matches the height of |rect| in device
-   pixels.
+   1. If |max height| is not null and |max height| divided by |framebuffer
+      height| is less than |scale|, set |scale| to |max height| divided by
+      |framebuffer height|.
+
+1. Let |paint width| be |framebuffer width| multiplied by |scale|, rounded to the nearest integer.
+
+1. Let |paint height| be |framebuffer height| multiplied by |scale|, rounded to the nearest integer.
```
sadym-chromium: NIT: IDK if we want to handle empty images, or want to add something like this:
```suggestion
1. Let |paint width| be the maximum of 1 and |framebuffer width| multiplied by |scale|, rounded to the nearest integer.

1. Let |paint height| be the maximum of 1 and |framebuffer height| multiplied by |scale|, rounded to the nearest integer.
```

Alternatively we can throw "invalid argument"

index.bs
```diff
@@ -3871,13 +3873,19 @@ Base64-encoded string.
...
       }
 
       browsingContext.ImageFormat = {
          type: text,
          ? quality: 0.0..1.0,
       }
 
+      browsingContext.ImageSize = {
+         ? maxWidth: js-uint .gt 0,
+         ? maxHeight: js-uint .gt 0,
```
```suggestion
         ? maxWidth: (js-uint .gt 0),
         ? maxHeight: (js-uint .gt 0),
```

sadym-chromium: Also we can restrict the image size to at least 1 pixel:

```suggestion
         ? maxWidth: (js-uint .gt 1),
         ? maxHeight: (js-uint .gt 1),
```

sadym-chromium (approved):
LGTM with a couple of nits

index.bs
```diff
@@ -3975,18 +3983,38 @@ To <dfn>rectangle intersection</dfn> given |rect1| and |rect2|
...
+      width|.
 
-1. Let |paint height| be |rect|'s [=height dimension=] multiplied by |ratio|,
-   rounded to the nearest integer, so it matches the height of |rect| in device
-   pixels.
+   1. If |max height| is not null and |max height| divided by |framebuffer
+      height| is less than |scale|, set |scale| to |max height| divided by
+      |framebuffer height|.
+
+1. Let |paint width| be the maximum of 1 and |framebuffer width| multiplied by |scale|, rounded to the nearest integer.
```
jgraham: Nit: the order of operations here is arguably ambiguous; I assume it's supposed to be `max(1, w_fb * scale)` rather than `max(1, w_fb) * scale`, but maybe one could define `|scale width|` like `Let |scale width| be |framebuffer width| multiplied by |scale| rounded to the nearest integer`.

But also note that `rounded to the nearest integer` is ambiguous if `|scale| * |framebuffer width|` is exactly `x.5`. https://tc39.es/ecma262/#sec-math.round has a more precise algorithm if we want always-round-up behaviour.

index.bs
```diff
@@ -3996,16 +4024,29 @@ To <dfn>render document to a canvas</dfn> given |document| and |rect|:
...
+   (|source x|, |source y|), (|source x| + |framebuffer width|, |source y|),
+   (|source x| + |framebuffer width|, |source y| + |framebuffer height|), and
+   (|source x|, |source y| + |framebuffer height|).
+
+1. Let |destination rectangle| be the rectangle whose corners are the four
+   points (0, 0), (|framebuffer width|, 0), (|framebuffer width|,
+   |framebuffer height|), and (0, |framebuffer height|).
 
-1. Return [=canvas=].
+1. Invoke |canvas context|'s {{CanvasTransform/scale()}} method with
```
jgraham: I wonder if we should wordsmith this in a way that avoids the implication that you actually need to run JS at this point e.g. "Run the steps corresponding to the {{CanvasTransform/scale()}} method on |canvas context|` with a note that this does not actually cause any script to execute.

index.bs
```diff
@@ -3996,16 +4024,29 @@ To <dfn>render document to a canvas</dfn> given |document| and |rect|:
...
+   <var ignore>x</var> set to |paint width| divided by |framebuffer width|, and
```
jgraham: To be clear: these factors are equal to `|scale|` except in the case where the scaling would produce zero-width output in one dimension, in which case they might be different to each other, and bigger than `|scale|`? That makes sense, but I wonder if there should be a note on why you can't just pass in `|scale|` directly.

index.bs
```diff
@@ -3996,16 +4024,29 @@ To <dfn>render document to a canvas</dfn> given |document| and |rect|:
...
+   <var ignore>y</var> set to |paint height| divided by |framebuffer height|.
+
+1. Complete implementation specific steps equivalent to painting the region of
```
jgraham, line 4057: Sorry, I'm having difficulty following this, and may be missing something. So we have `|source rectangle|` which is a region on the framebuffer we want to paint, in display coordinates. We also have `|destination rectangle|` which is a translation of `|source rectangle|` to the origin which in the unscaled case represents the place where we should paint the input pixels onto the canvas. But in the scaled case `|destination rectangle|` is bigger than the actual canvas we're painting to, and we instead want to paint to a rectangle covering the area `(0,0)` to (`|paint width|`, `|paint height|`). And by construction scaling the destination rectangle by the transformation matrix will produce a rectangle of that size.

So I guess that's what the text is trying to do? And it does seem to match the first part of step 6 in https://html.spec.whatwg.org/#dom-context-2d-drawimage.

But nothing here actually tells me what to paint at each pixel. Like to give a silly example, if the input rectangle is 3x1 with a red pixel, a green pixel, and a blue pixel, and the output rectangle is 2x1, what should the output colours be?

Later in https://html.spec.whatwg.org/#dom-context-2d-drawimage it says:

> If the original image data is a bitmap image, then the value painted at a point in the destination rectangle is computed by filtering the original image data. The user agent may use any filtering algorithm (for example bilinear interpolation or nearest-neighbor). When the filtering algorithm requires a pixel value from outside the original image data, it must instead use the value from the nearest edge pixel. (That is, the filter uses 'clamp-to-edge' behavior.) When the filtering algorithm requires a pixel value from outside the source rectangle but inside the original image data, then the value from the original image data must be used.

So maybe that's closer to the actual text we need here i.e. we can remove the part about setting the scale matrix, directly construct the destination rectangle at the size we need, and say something like:


"""When the dimensions of |destination rect| are smaller than |source rect|, then the value painted at a point in the destination rectangle is computed by filtering the data from |document|'s framebuffer. The user agent may use any filtering algorithm (for example bilinear interpolation or nearest-neighbor). When the filtering algorithm requires a pixel value from outside the framebuffer, it must instead use the value from the nearest edge pixel. (That is, the filter uses 'clamp-to-edge' behavior.) When the filtering algorithm requires a pixel value from outside the |source rect| but inside the framebuffer, then the value from the framebuffer must be used."""

jgraham (changes requested):

index.bs
lutien: Alright, I've updated it to split out `|scale width|` and add the rounding similar to JS Math.round.

index.bs
lutien: Sure, updated.

index.bs
lutien: yeah, exactly. I've added the note.

index.bs
lutien: Ah, ok, yeah, I've tried to base it on drawImage but missed the part about mapping pixels. Updated to use the suggested wording.

index.bs
jgraham: I think you can just delete this step now? We don't refer to the transformation matrix of the canvas in any later step.

lutien: ah, right, removed.

index.bs
```diff
@@ -226,6 +228,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
     text: create a classic script; url: webappapis.html#creating-a-classic-script
     text: create a new browsing context; url: browsers.html#creating-a-new-browsing-context
     text: create a new top-level traversable; url: document-sequences.html#creating-a-new-top-level-traversable
+    text: current transformation matrix; url: canvas.html#current-transformation-matrix
```
jgraham: I guess you can remove this as well?

jgraham (approved):
Thanks! One tiny clean up, but otherwise it looks good to me.

index.bs
jgraham:
```suggestion
```

lutien:
Thanks, everyone, for the reviews! I'm going to merge it now.

whimboo:
On our side we have [Bug 2069002](https://bugzilla.mozilla.org/show_bug.cgi?id=2069002) now where the necessary tests will be added.

PR #1148, opened 2026-08-14
Emit script.realmDestroyed when a worker event loop is destroyed

### LINKED ISSUE #1147: script.realmDestroyed is never emitted when a worker event loop is destroyed
### Summary

The [`script.realmDestroyed`](https://w3c.github.io/webdriver-bidi/#event-script-realmDestroyed) remote end event trigger has two branches. The *unloading document cleanup steps* branch emits correctly (twice — once in the worklet loop, once for the document's own realm). The **"Whenever a worker event loop is destroyed"** branch does not emit at all.

It runs:

1. `environment settings` ← the settings object whose responsible event loop this is
2. `related navigables` ← *get related navigables*
3. `realm` → `realm id`
4. build `params`
5. build `body`

…and then the algorithm ends. There is no *set of sessions for which an event is enabled* loop and no *emit an event* step.

### Why this looks like an editorial omission rather than a design choice

- **`related navigables` is computed and never read.** Its only possible consumer is the missing emit loop.
- **The sibling branch in the same algorithm emits.** The document branch ends with exactly the two steps that are absent here.
- **`script.realmCreated` is symmetric and does emit.** Its trigger fires on *set up a window / worker / worklet environment settings object* in a single algorithm and ends with the emit loop — so realm creation is announced for all three global types while destruction is announced for only two.

### Impact

A client that maintains a `realm → owning context` map (the only way to attribute `script.message` and `log.entryAdded` from a worker, since `script.Source`'s `context` is optional and worker realm infos carry no `context`) has no teardown signal for worker realms and leaks every one it ever saw.

This is not limited to explicit `terminate()`: when a document unloads, its dedicated workers are terminated too, so the same branch is responsible for ordinary navigation teardown.

### Both implementations already emit it

Tested against Chrome 151.0.7922.138 (chromedriver 151.0.7922.138) and Firefox, driving raw BiDi over a WebSocket. In each case a dedicated worker's realm id was captured from `script.realmCreated`, then `worker.terminate()` was called:

**Chrome**

```
dedicated worker's realm, from realmCreated: "-7399823684229122840.-9127975584400262148"
after worker.terminate(): 1 realmDestroyed [{"realm":"-7399823684229122840.-9127975584400262148"}]
  -> names exactly that worker realm: YES
```

**Firefox**

```
dedicated worker's realm, from realmCreated: "d2dd9baf-a55a-4b88-aa08-a70249dd9af5"
after worker.terminate(): 1 realmDestroyed [{"realm":"d2dd9baf-a55a-4b88-aa08-a70249dd9af5"}]
  -> names exactly that worker realm: YES
```

So two independent implementations emit an event the spec never instructs them to, naming precisely the terminated worker's realm. Adding the missing step codifies shipped behaviour and requires no implementation change; a WPT test for it should pass on both engines immediately.

### Proposed fix

Append to the worker event loop branch, mirroring the document branch verbatim:

```
1. For each |session| in the [=set of sessions for which an event is enabled=]
   given "<code>script.realmDestroyed</code>" and |related navigables|:

  1. [=Emit an event=] with |session| and |body|.
```

### Out of scope

`script.RealmDestroyedParameters` carries only `realm` — no `context`, no `owners` — so once the event fires, the client's own map is the only way to attribute it, and there is no recovery for a worker whose `realmCreated` was missed. That is a design change rather than a correction, so it is better handled separately.

Worth noting for anyone testing this: neither Chrome nor Firefox currently announces a `shared-worker` realm via `script.realmCreated` at all, even with live `SharedWorker` handles in two tabs. So the destroy path stays unverifiable for shared workers on either engine regardless of this fix.

AutomatedTester:
Fixes #1147

The worker event loop branch of the [`script.realmDestroyed`](https://w3c.github.io/webdriver-bidi/#event-script-realmDestroyed) trigger built `params` and `body` but was missing the step that emits the event, so no event was sent when a worker event loop was destroyed. This adds it, copied from the document branch of the same algorithm.

Chrome and Firefox both already emit this event, naming the terminated worker's realm.

Also fixes two `field set of` typos in the same event definition, which should read `field set to`.

***
<a href="https://pr-preview.s3.amazonaws.com/AutomatedTester/webdriver-bidi/pull/1148.html" title="Last updated on Aug 18, 2026, 7:05 PM UTC (8c7d28d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/1148/a8d59e7...AutomatedTester:8c7d28d.html" title="Last updated on Aug 18, 2026, 7:05 PM UTC (8c7d28d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -13942,7 +13942,7 @@ Define the following [=unloading document cleanup steps=] with |document|:
   1. Let |realm id| be the [=realm id=] for |realm|.
 
   1. Let |params| be a [=/map=] matching the <code>script.RealmDestroyedParameters</code>
-     production, with the <code>realm</code> field set of |realm id|.
+     production, with the <code>realm</code> field set to |realm id|.
 
   1. Let |body| be a [=/map=] matching the <code>script.RealmDestroyed</code>
      production, with the <code>params</code> field set to |params|.
@@ -13987,10 +13987,15 @@ worker=] algorithm:
 1. Let |realm id| be the [=realm id=] for |realm|.
 
 1. Let |params| be a [=/map=] matching the <code>script.RealmDestroyedParameters</code>
-   production, with the <code>realm</code> field set of |realm id|.
+   production, with the <code>realm</code> field set to |realm id|.
 
 1. Let |body| be a [=/map=] matching the <code>script.RealmDestroyed</code>
    production, with the <code>params</code> field set to |params|.
+
+1. For each |session| in the [=set of sessions for which an event is enabled=]
+   given "<code>script.realmDestroyed</code>" and |related navigables|:
+
+  1. [=Emit an event=] with |session| and |body|.
 
 </div>
 


jgraham (changes requested):
I think this change is correct, but we can't land it with such an obviously AI generated commit message. The actual issue here is that we were literally just missing the line in the algorithm that emitted the events. That's all the message needs to say.

AutomatedTester:
Rewritten and force-pushed — the commit message now just says the emit step was missing from the algorithm. The diff is unchanged. I trimmed the PR description down too.

jgraham (approved):

Issue #832, opened 2024-12-13
Need an event/data point to know when a frame finished loading the about:blank document


OrKoN:
Initial about:blank loads are not navigations but clients often need to know if a frame is finishing loading the about:blank document. It is relevant for distinguishing regular frames from lazy loading frames that are paused and are not loading.

cc @sadym-chromium

OrKoN:
(not sure if there is a duplicate spec issue for this)

OrKoN:
> Is this request similar to the [`Page.ready` event](https://github.com/microsoft/playwright/blob/b0cec5b3518b8fa5d107068ea39612bcc1341b53/browser_patches/firefox/juggler/protocol/PageHandler.js#L88) that Playwright is using via Juggler with Firefox?

yes, I think it is. I think you mentioned this on some other issue but I could not find it.

hbenl:
> It is relevant for distinguishing regular frames from lazy loading frames that are paused and are not loading.

Another important use case is pages opened by `window.open()` or `window.open(url)`: it's impossible for the client to tell from the `contextCreated` event if the page will navigate to a URL or if it will stay at `about:blank` (in which case no more events will be sent after `contextCreated`).

whimboo:
> Another important use case is pages opened by `window.open()` or `window.open(url)`: it's impossible for the client to tell from the `contextCreated` event if the page will navigate to a URL or if it will stay at `about:blank` (in which case no more events will be sent after `contextCreated`).

Therefor see https://github.com/w3c/webdriver-bidi/issues/766 as well.

PR #1151, opened 2026-08-24
Pass a new "hasPlannedNavigation" flag to the "browsingContext.contextCreated" event.

lutien:
Covers part of #832.
Introduce a new `hasPlannedNavigation` flag to the "browsingContext.contextCreated" event that should indicate if a browsing context will navigate from the `about:blank` page. For lazy iframes, it looks like the BiDi event happens too early, so we cannot identify if an iframe is in the viewport. So we probably will need an event to know when they're rendered in the viewport and another property for a `browsingContext.contextCreated` to indicate that it's a lazy iframe that will have to wait for another event.
The HTML spec part is https://github.com/whatwg/html/pull/12833.

***
<a href="https://pr-preview.s3.amazonaws.com/lutien/webdriver-bidi/pull/1151.html" title="Last updated on Sep 11, 2026, 1:48 PM UTC (ac74ddc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/1151/f685b93...lutien:ac74ddc.html" title="Last updated on Sep 11, 2026, 1:48 PM UTC (ac74ddc)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5691,30 +5691,41 @@ are:
       <pre class="cddl" data-cddl-module="local-cddl">
         browsingContext.ContextCreated = (
          method: "browsingContext.contextCreated",
-         params: browsingContext.Info
+         params: browsingContext.ContextCreatedParameters
         )
+
+        browsingContext.ContextCreatedParameters = {
+          ~browsingContext.Info,
+          hasImmediateNavigation: bool,
+        }
       </pre>
    </dd>
 </dl>
 
 <div algorithm>
 
-To <dfn>Recursively emit context created events</dfn> given |session| and |navigable|:
+To <dfn>Recursively emit context created events</dfn> given |session|, |navigable|,
+and |has immediate navigation|:
 
-1. [=Emit a context created event=] with |session| and |navigable|.
+1. [=Emit a context created event=] with |session|, |navigable|, and
+   |has immediate navigation|.
 
 1. For each child navigable, |child|, of |navigable|:
 
-  1. [=Recursively emit context created events=] given |session| and |child|.
+  1. [=Recursively emit context created events=] given |session|, |child|, and
+     |has immediate navigation|.
 
 </div>
 
 <div algorithm>
 
-To <dfn>Emit a context created event</dfn> given |session| and |navigable|:
+To <dfn>Emit a context created event</dfn> given |session|, |navigable|, and
+|has immediate navigation|:
 
 1. Let |params| be the result of [=get the navigable info=] given
    |navigable|, 0, and true.
+
+1. Set |params|["<code>hasImmediateNavigation</code>"] to |has immediate navigation|.
 
 1. Let |body| be a [=/map=] matching the
    <code>browsingContext.ContextCreated</code> production, with the
@@ -5729,7 +5740,7 @@ To <dfn>Emit a context created event</dfn> given |session| and |navigable|:
 
 The [=remote end event trigger=] is
 the <dfn export>WebDriver BiDi navigable created</dfn> steps given
-[=/navigable=] |navigable| and [=/navigable=] |opener navigable|:
+[=/navigable=] |navigable|, |has immediate navigation|, and [=/navigable=] |opener navigable|:
 
 1. Set |navigable|'s [=original opener=] to |opener navigable|,
    if |opener navigable| is provided.
@@ -5743,7 +5754,8 @@ the <dfn export>WebDriver BiDi navigable created</dfn> steps given
 1. For each |session| in the [=set of sessions for which an event is enabled=]
    given "<code>browsingContext.contextCreated</code>" and |related navigables|:
 
-  1. [=Emit a context created event=] given |session| and |navigable|.
+  1. [=Emit a context created event=] given |session|, |navigable|, and
+     |has immediate navigation|.
 
 </div>
 
@@ -5754,7 +5766,7 @@ The [=remote end subscribe steps=], with [=subscribe priority=] 1, given
 
 1. For each |navigable| in |navigables|:
 
-  1. [=Recursively emit context created events=] given |session| and |navigable|.
+  1. [=Recursively emit context created events=] given |session|, |navigable|, and false.
 
 </div>
 


index.bs
```diff
@@ -5691,31 +5691,42 @@ are:
       <pre class="cddl" data-cddl-module="local-cddl">
         browsingContext.ContextCreated = (
          method: "browsingContext.contextCreated",
-         params: browsingContext.Info
+         params: browsingContext.ContextCreatedParameters
         )
+
+        browsingContext.ContextCreatedParameters = {
+          ~browsingContext.Info,
+          hasStartedNavigating: bool,
```
jgraham: I feel like `hasImmediateNavigation` or `willNavigate` might be more obvious names?

lutien: I think we still want to indicate that navigation should start now, so `willNavigate` can be ambiguous for lazy iframes, but `hasImmediateNavigation` sounds good to me. Updated.

OrKoN (approved):

whimboo:
@lutien before finally reviewing this PR one question regarding the HTML spec related work. That other PR is still in draft mode. When do you think it can be requested review for it?

lutien:
> @lutien before finally reviewing this PR one question regarding the HTML spec related work. That other PR is still in draft mode. When do you think it can be requested review for it?

When we all agree here that this is the right approach for us. 🙂

whimboo:
The name of the property looks fine to me, if that is what you need from me at the moment. I assume that every browser will be able to guess that an immediate navigation will follow or not. If not we wouldn't be able to make this property mandatory.

lutien:
Ok, after trying to convince everyone that we have to cover lazy iframes in this property, thanks to Henrik's comment, I realized that the `browsingContext.contextCreated` event happens too early to know if an iframe is in a viewport. So I think the new proposed behavior is that the property is `true` for lazy iframes if they will navigate at some point. That's why I renamed the property to rather be `hasPlannedNavigation`. 
To fully cover the lazy iframe behavior, I think we will need to have another property to indicate that they're lazy and maybe have another event to indicate when they're in the viewport (or just wait for the load events/start of a navigation). But I would handle it as a follow-up. 

The PR, the title, and the description of this and the HTML spec PR are updated and ready to be reviewed.

OrKoN:
@sadym-chromium PTAL

index.bs
```diff
@@ -5688,31 +5691,42 @@ are:
...
+        }
       </pre>
    </dd>
 </dl>
 
 <div algorithm>
 
-To <dfn>Recursively emit context created events</dfn> given |session| and |navigable|:
+To <dfn>Recursively emit context created events</dfn> given |session|, |navigable|,
+and |has planned navigation|:
```
sadym-chromium: I don't think we ever invoke `Recursively emit context created events` with `|has planned navigation|: true`. do I miss something?

index.bs
```diff
@@ -5726,7 +5740,7 @@ To <dfn>Emit a context created event</dfn> given |session| and |navigable|:
 
 The [=remote end event trigger=] is
 the <dfn export>WebDriver BiDi navigable created</dfn> steps given
-[=/navigable=] |navigable| and [=/navigable=] |opener navigable|:
+[=/navigable=] |navigable|, |has planned navigation|, and [=/navigable=] |opener navigable|:
```
```suggestion
[=/navigable=] |navigable|, [=boolean=] |has planned navigation|, and
optional [=/navigable=] |opener navigable| (default: null):
```

index.bs
```diff
@@ -5688,31 +5691,42 @@ are:
       <pre class="cddl" data-cddl-module="local-cddl">
         browsingContext.ContextCreated = (
          method: "browsingContext.contextCreated",
-         params: browsingContext.Info
+         params: browsingContext.ContextCreatedParameters
         )
+
+        browsingContext.ContextCreatedParameters = {
+          ~browsingContext.Info,
```
sadym-chromium: I don't think we have a precedent of using `~`. Do we need it?

index.bs
lutien: No, you're right, I haven't noticed that we don't have any other usage for `Recursively emit context created events`. I've updated it to not pass `|has planned navigation|` in it.

index.bs
lutien: It looks like it's correct cddl syntax for unwrapping: https://www.rfc-editor.org/info/rfc8610/#section-3.7. I guess alternatively we could just specify all the fields from `browsingContext.Info` but I thought it's a nice simplification.

sadym-chromium: we can extract a group `browsingContext.BaseInfo` and reuse it:
```
browsingContext.BaseInfo = (
  url: text,
  userContext: browser.UserContext,
  parent: browsingContext.BrowsingContext / null,
  children: [* browsingContext.Info] / null,
)

browsingContext.Info = {
  browsingContext.BaseInfo,
}

browsingContext.ContextCreatedParameters = {
  browsingContext.BaseInfo,
  hasPlannedNavigation: bool,
}
```

lutien: Yeah, I think that should work as well, updated ✅

sadym-chromium: great, thanks!

sadym-chromium (approved):

index.bs
```diff
@@ -5688,8 +5695,13 @@ are:
       <pre class="cddl" data-cddl-module="local-cddl">
         browsingContext.ContextCreated = (
          method: "browsingContext.contextCreated",
-         params: browsingContext.Info
+         params: browsingContext.ContextCreatedParameters
         )
+
+        browsingContext.ContextCreatedParameters = {
+         browsingContext.BaseInfo,
+         hasPlannedNavigation: bool,
```
whimboo, line 5703: Just to make sure: is this a field for which we can always provide a value, or are there situations where the value would be unknown and the field should therefore be omitted?

lutien: We can always provide the value

whimboo: Ok, thank you. So it means for whatever happens with lazy loaded iframes in the future we will probably have a separate field here?

index.bs
```diff
@@ -5751,7 +5769,7 @@ The [=remote end subscribe steps=], with [=subscribe priority=] 1, given
 
 1. For each |navigable| in |navigables|:
 
-  1. [=Recursively emit context created events=] given |session| and |navigable|.
+  1. [=Recursively emit context created events=] given |session|, |navigable|, and false.
```
whimboo: This adds `false` as third argument which the definition doesn't define. Looks like we automatically use `false` so that we can just remove the extra argument here?

index.bs
whimboo, line 5703: Can you please at least add a one-liner description what this field stands for? We have it usually for any other field where it's not immediately clear what it's used for.

index.bs
```diff
@@ -5726,7 +5742,8 @@ To <dfn>Emit a context created event</dfn> given |session| and |navigable|:
 
 The [=remote end event trigger=] is
 the <dfn export>WebDriver BiDi navigable created</dfn> steps given
-[=/navigable=] |navigable| and [=/navigable=] |opener navigable|:
+[=/navigable=] |navigable|, [=boolean=] |has planned navigation|, and
+optional [=/navigable=] |opener navigable| (default: null):
 
 1. Set |navigable|'s [=original opener=] to |opener navigable|,
    if |opener navigable| is provided.
```
whimboo: I assume that we can remove the `is provided` part here given that we have a declared default (null), which means that a value will always be present.

whimboo (changes requested):

index.bs
lutien: yes

lutien: ok, added

index.bs
lutien: yep, updated

index.bs
lutien: yes, I've missed to clean it up here

index.bs
```diff
@@ -5708,11 +5725,14 @@ To <dfn>Recursively emit context created events</dfn> given |session| and |navig
...
-To <dfn>Emit a context created event</dfn> given |session| and |navigable|:
+To <dfn>Emit a context created event</dfn> given |session|, |navigable|, and
+|has planned navigation|:
 
 1. Let |params| be the result of [=get the navigable info=] given
    |navigable|, 0, and true.
 
+1. Set |params|["<code>hasPlannedNavigation</code>"] to |has planned navigation|.
+
 1. Let |body| be a [=/map=] matching the
```
whimboo: Question. For network events we have an extra assertion here to check that the object really matches the target production. Should we do that here as well given that we use a shared group as basis now?

index.bs
```diff
@@ -5726,10 +5746,10 @@ To <dfn>Emit a context created event</dfn> given |session| and |navigable|:
 
 The [=remote end event trigger=] is
 the <dfn export>WebDriver BiDi navigable created</dfn> steps given
-[=/navigable=] |navigable| and [=/navigable=] |opener navigable|:
+[=/navigable=] |navigable|, [=boolean=] |has planned navigation|, and
```
whimboo: Note that the boolean declaration autolinks to the ECMAscript boolean value and not to the infra boolean as in all the other places.

whimboo (changes requested):
Some more issues that I noticed after cross-checking with the WhatWG HTML spec changes. I noticed that there is currently a mismatch of the arguments for the `create a new top-level traversable` algorithm. Even through it might change based on its review we should at least try to be in sync with our landed patch.

index.bs
lutien: I am not sure why we do it for the network events, to be honest, if we build the event payload in the same spec, but ok, I've added it here as well.

index.bs
lutien: So I guess we then should just use the plan `boolean` word? At least that's how it is in the other places.

index.bs
```diff
@@ -4354,8 +4285,8 @@ The [=remote end steps=] with |command parameters| are:
 
   <!-- This is based on step 5 of https://w3c.github.io/webdriver/#new-window,
        but without using the "current browsing context" concept. -->
-  1. Let |traversable| be the result of trying to [=/create a new top-level traversable=] steps with null and empty string,
-     and setting the [=associated user context=] for the newly created [=/top-level traversable=] to |user context|.
+  1. Let |traversable| be the result of trying to [=/create a new top-level traversable=] steps with null, empty string,
+     null and false, and setting the [=associated user context=] for the newly created [=/top-level traversable=] to |user context|.
```
```suggestion
     null, and false, and setting the [=associated user context=] for the newly created [=/top-level traversable=] to |user context|.
```

lutien:
> Some more issues that I noticed after cross-checking with the WhatWG HTML spec changes. I noticed that there is currently a mismatch of the arguments for the `create a new top-level traversable` algorithm. Even through it might change based on its review we should at least try to be in sync with our landed patch.

ok, I've updated the call of `create a new top-level traversable` algorithm

index.bs
whimboo: Maybe @jgraham or @juliandescottes can explain. If we don't need it we can remove them all in a follow-up review. I only noticed this discrepancy.

whimboo (approved):
Thanks. Looks good to me now.

Issue #868, opened 2025-01-26
`script.NodeRemoteValue` without any attribute?


nvborisenko:
I am reading the following definition for `NodeRemoteValue`:

```
script.NodeRemoteValue = {
  type: "node",
  ? sharedId: script.SharedId,
  ? handle: script.Handle,
  ? internalId: script.InternalId,
  ? value: script.NodeProperties,
}
```

Hmm, everything is optional? Then what does define it as concrete value? Seems `sharedId`...

OrKoN:
What attributes will be set is specified in the serialization steps https://w3c.github.io/webdriver-bidi/#serialize-as-a-remote-value and depends on parameters like the ownership.

whimboo:
There is nothing that needs an update in the BiDi specification. So lets close.

PR #1156, opened 2026-08-27
Require sharedId on browsingContext.locateNodes result nodes

titusfortner:
`LocateNodesResult` types its nodes as `script.NodeRemoteValue`, where `sharedId` is optional. The command's algorithm guarantees more:

- result ownership is `"none"`, so `handle` is never created
- every located node is in a document with a navigable, so serialization always produces a shared id

So a conforming-by-the-CDDL response can carry no reference of any kind, which also strands `startNodes` — `script.SharedReference` requires `sharedId`, and a previous `locateNodes` result is its only source for a freshly located node. Safari does this today:

```ruby
forms = browsing_context.locate_nodes(
  context: driver.window_handle,
  locator: BrowsingContext::CssLocator.new(value: 'form'),
  max_node_count: 1
)
Script::SharedReference.new(shared_id: forms.nodes.first.shared_id)
# => ArgumentError: Selenium::WebDriver::BiDi::Protocol::Script::SharedReference#shared_id is required
#                   ./rb/lib/selenium/webdriver/bidi/serialization/record.rb:115:in `validate_values'
```

A client generated from the CDDL parses the browser response as valid; the error comes from trying to use it, which results in a confusing error.

Additionally, WPT already asserts `sharedId` on every `locate_nodes` result node, so this change makes the type say what the tests and conforming implementations already enforce.

Changes:

- `LocateNodesResult` now uses `script.SharedNodeRemoteValue`: `script.NodeRemoteValue` with `sharedId` required, named after `script.SharedReference`
- return `no such node` for a start node whose node navigable is null (e.g. adopted into a `<template>`'s inert content document) — it still deserializes, and the non-CSS strategies traverse its subtree, yielding nodes that would serialize without a shared id
- assert `sharedId` at serialization, following the `network.authRequired` assert

Notes:

- narrower than #868: only the result type of one command whose serialization inputs are fully pinned down

***
<a href="https://pr-preview.s3.amazonaws.com/titusfortner/webdriver-bidi/pull/1156.html" title="Last updated on Sep 8, 2026, 2:44 PM UTC (4563d92)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/1156/a7b8b67...titusfortner:4563d92.html" title="Last updated on Sep 8, 2026, 2:44 PM UTC (4563d92)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4472,7 +4472,7 @@ list of all nodes matching the specified locator.
    <dd>
     <pre class="cddl" data-cddl-module="local-cddl">
         browsingContext.LocateNodesResult = {
-            nodes: [ * script.NodeRemoteValue ]
+            nodes: [ * script.SharedNodeRemoteValue ]
         }
     </pre>
    </dd>
@@ -4734,6 +4734,14 @@ The [=remote end steps=] with |session| and |command parameters| are:
    1. Let |start node| be the result of [=trying=] to [=deserialize shared reference=] given
       |serialized start node|, |realm| and |session|.
 
+   1. If |start node|'s [=/node navigable=] is null, return [=error=] with
+      [=error code=] [=no such node=].
+
+      Note: This happens when the node was adopted into a document that is not
+      associated with a navigable, such as the associated inert template
+      document of a <{template}> element. Nodes located from such a start node
+      could not be serialized with a shared id.
+
    1. [=list/Append=] |start node| to |context nodes|.
 
 1. Assert [=list/size=] of |context nodes| is greater than 0.
@@ -4824,6 +4832,8 @@ The [=remote end steps=] with |session| and |command parameters| are:
    1. Let |serialized node| be the result of [=serialize as a remote value=] with
       |result node|, |serialization options|, |result ownership|, a new [=/map=]
       as serialization internal map, |realm| and |session|.
+
+   1. Assert: |serialized node| [=map/contains=] "<code>sharedId</code>".
 
    1. [=list/Append=] |serialized node| to |serialized nodes|.
 
@@ -12247,6 +12257,21 @@ script.WindowProxyRemoteValue = {
 
 script.WindowProxyProperties = {
   context: browsingContext.BrowsingContext
+}
+</pre>
+
+A <code>script.SharedNodeRemoteValue</code> is a <code>script.NodeRemoteValue</code>
+that is known to carry a shared id. It is used for results of commands that
+serialize nodes in a context where [=get shared id for a node=] cannot return
+null.
+
+<pre class="cddl" data-cddl-module="local-cddl">
+script.SharedNodeRemoteValue = {
+  type: "node",
+  sharedId: script.SharedId,
+  ? handle: script.Handle,
+  ? internalId: script.InternalId,
+  ? value: script.NodeProperties,
 }
 </pre>
 
--- a/index.bs
+++ b/index.bs
@@ -4739,7 +4739,7 @@ The [=remote end steps=] with |session| and |command parameters| are:
 
       Note: This happens when the node was adopted into a document that is not
       associated with a navigable, such as the associated inert template
-      document of a <{template}> element. Nodes located from such a start node
+      document of a <code>template</code> element. Nodes located from such a start node
       could not be serialized with a shared id.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -12432,6 +12260,21 @@ script.WindowProxyProperties = {
...
+serialize nodes in a context where [=get shared id for a node=] cannot return
+null.
+
+<pre class="cddl" data-cddl-module="local-cddl">
+script.SharedNodeRemoteValue = {
+  type: "node",
+  sharedId: script.SharedId,
+  ? handle: script.Handle,
+  ? internalId: script.InternalId,
+  ? value: script.NodeProperties,
```
whimboo, line 12274: The problem here is that with the usage of `script.NodeProperties` we have again a usage of the script's `NodeRemoteValue` for any `children` and `shadowRoot` nodes, which then doesn't have a required `sharedId` for any entry.

Personally I would factor the shared fields into groups (exactly what `BaseRealmInfo` already does in the spec), then build both variants plus a shared `NodeProperties` variant so it works also recursively.

index.bs
```diff
@@ -4733,6 +4734,14 @@ The [=remote end steps=] with |session| and |command parameters| are:
    1. Let |start node| be the result of [=trying=] to [=deserialize shared reference=] given
       |serialized start node|, |realm| and |session|.
 
+   1. If |start node|'s [=/node navigable=] is null, return [=error=] with
```
whimboo, line 4737: Could you add a wpt test for such a particular case? I don't think that we have one yet.

index.bs
```diff
@@ -4733,6 +4734,14 @@ The [=remote end steps=] with |session| and |command parameters| are:
...
+      [=error code=] [=no such node=].
+
+      Note: This happens when the node was adopted into a document that is not
+      associated with a navigable, such as the associated inert template
+      document of a <{template}> element. Nodes located from such a start node
```
whimboo, line 4742: We don't use autolink references anywhere else in this document. Maybe you can use `[=template=]` instead?

index.bs
```diff
@@ -12432,6 +12260,21 @@ script.WindowProxyProperties = {
 }
 </pre>
 
+A <code>script.SharedNodeRemoteValue</code> is a <code>script.NodeRemoteValue</code>
+that is known to carry a shared id. It is used for results of commands that
+serialize nodes in a context where [=get shared id for a node=] cannot return
+null.
+
+<pre class="cddl" data-cddl-module="local-cddl">
+script.SharedNodeRemoteValue = {
```
whimboo, line 12269: Lets place this definition closely to `NodeProperties` so that it makes it easier to update the blocks in case an entry changes, gets added, or removed.

index.bs
```diff
@@ -4824,6 +4833,8 @@ The [=remote end steps=] with |session| and |command parameters| are:
       |result node|, |serialization options|, |result ownership|, a new [=/map=]
       as serialization internal map, |realm| and |session|.
 
+   1. Assert: |serialized node| [=map/contains=] "<code>sharedId</code>".
```
whimboo, line 4836: This assert only runs for the top-level node but not for any children or shadowRoot's.

whimboo (changes requested):
Thank you @titusfortner. The description makes sense and we should potentially fix that issue for clients. I've a couple of inline comments but beside those it would be good to as well get some feedback from @jgraham.

index.bs
titusfortner: Done in 00b3db0

index.bs
titusfortner: Done in 00b3db0

index.bs
titusfortner: Changed to Assert: |serialized node| matches the script.SharedNodeRemoteValue production (same form as the network event asserts), which covers nested children and shadow roots.

index.bs
titusfortner: I couldn't get [=template=] to work, so I switched to plain <code>template</code>, matching how input.setFiles references the input element in prose.

index.bs
titusfortner: https://github.com/web-platform-tests/wpt/pull/62528

PR #1161, opened 2026-09-09
Round dimensions only when downscaling in the "browsingContext.captureScreenshot" command.

lutien:
During the implementation, I've noticed that using the same rounding mechanic when we apply device pixel ratio might cause issues. So I think we should do it when we downscale.

***
<a href="https://pr-preview.s3.amazonaws.com/lutien/webdriver-bidi/pull/1161.html" title="Last updated on Sep 9, 2026, 3:31 PM UTC (e53537f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/1161/dc8cf11...lutien:e53537f.html" title="Last updated on Sep 9, 2026, 3:31 PM UTC (e53537f)">Diff</a>


sadym-chromium (approved):

index.bs
```diff
@@ -4023,13 +4023,23 @@ To <dfn>render document to a canvas</dfn> given |document|, |rect| and |image si
       height| is less than |scale|, set |scale| to |max height| divided by
```
whimboo, line 4023: I just noticed that we set scale here again. It could override an already set scale based on the width above which would be different if we don't have a 1:1 aspect ratio. Maybe we need to conditionally update and pick the smallest value?

lutien: But that's what we do here already. We only update the scale if the new scale is less than the previously calculated one.

whimboo: Not sure why I didn't notice that. Please drop it.

whimboo (approved):
I think that looks fine to me as well. Thank you Sasha!

PR #1163, opened 2026-09-11
Add a "destinationFolder" parameter to the "browsingContext.startScreencast" command.

lutien:
We've got some requests to make it possible to configure the location of a screencast file.

***
<a href="https://pr-preview.s3.amazonaws.com/lutien/webdriver-bidi/pull/1163.html" title="Last updated on Sep 11, 2026, 2:46 PM UTC (56a515a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/1163/dc8cf11...lutien:56a515a.html" title="Last updated on Sep 11, 2026, 2:46 PM UTC (56a515a)">Diff</a>


index.bs
```diff
@@ -5552,8 +5553,16 @@ The [=remote end steps=] given |session| and |command parameters| are:
    1. If the implementation is unable to produce |stream|, return [=error=] with
       [=error code=] [=unknown error=].
 
-1. Let |path| be an implementation-defined file path where the recording will
-   be stored.
+1. If |command parameters| [=map/contains=] the <code>destinationFolder</code> field:
+
+   1. Let |destination folder| be
+      |command parameters|["<code>destinationFolder</code>"].
```
sadym-chromium: should we fail command if the destination folder is not a valid path or unavailable for whatever reason?

sadym-chromium (commented):
API LGTM

index.bs
lutien: We don't really have an existing mechanic in the spec for such validation (e.g., we don't have it for `setDownloadBehavior`), but I think it, of course, would be good. So I've improvised something.

sadym-chromium: HTML defines it:

> If an error occurs (for example, a network error, not enough storage, an unavailable destination folder);
> then invoke [WebDriver BiDi download end](https://w3c.github.io/webdriver-bidi/#webdriver-bidi-download-end) with ... [status](https://w3c.github.io/webdriver-bidi/#navigation-status-status) is "[canceled](https://w3c.github.io/webdriver-bidi/#navigation-status-canceled)".

sadym-chromium: https://html.spec.whatwg.org/#handle-as-a-download, step 5.2

sadym-chromium: my concern is that the user records something only to know the path is incorrect after they done all the actions they wanted to record. I believe we discussed it already at some point, but cannot find related minutes

lutien: Ah, alright, I've checked, of course, only the BiDi spec, but it's good to know that it's covered.
But yes, if it maybe was not clear, I've updated this PR to validate the path for the screencast file. 🙂

index.bs
```diff
@@ -5552,8 +5576,7 @@ The [=remote end steps=] given |session| and |command parameters| are:
    1. If the implementation is unable to produce |stream|, return [=error=] with
       [=error code=] [=unknown error=].
 
-1. Let |path| be an implementation-defined file path where the recording will
-   be stored.
+1. If |command parameters| [=map/contains=] the <code>destinationFolder</code> field:
```
```suggestion
```

index.bs
```diff
@@ -5501,6 +5502,29 @@ The [=remote end steps=] given |session| and |command parameters| are:
 1. If |navigable| is not a [=/top-level traversable=], return [=error=] with
    [=error code=] [=invalid argument=].
 
+1. If |command parameters| [=map/contains=] the <code>destinationFolder</code> field:
```
sadym-chromium, line 5505: I believe checking `|destination folder entry|` is enough before we have a stream. Creating a file beofre the stream is ready can leave orphaned empty file if there were some problems in the streaming.

```suggestion
1. Let |destination folder entry| be null.

1. If |command parameters| [=map/contains=] the <code>destinationFolder</code> field:
```

index.bs
```diff
@@ -5501,6 +5502,29 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+
+   1. Let |destination folder| be |command parameters|["<code>destinationFolder</code>"].
+
+   1. Let |destination folder locator| be a [=directory locator=] with
```
```suggestion
   1. Set |destination folder entry| to [=locate an entry=] given
```

index.bs
```diff
@@ -5501,6 +5502,29 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+      [=invalid argument=].
+
+   1. Let |path| be an implementation-defined file path for a new file in
+      |destination folder entry| where the recording will be stored.
+
+   1. Perform implementation-defined steps to create an empty file at |path|. If
+      they fail, return [=error=] with [=error code=] [=unknown error=].
+
+1. Otherwise, let |path| be an implementation-defined file path where the
+   recording will be stored.
```
sadym-chromium, line 5520:
```suggestion
```

index.bs
```suggestion
1. If |destination folder entry| is not null:

   1. Let |path| be an implementation-defined file path for a new file in
     |destination folder entry| where the recording will be stored.

1. Otherwise, let |path| be an implementation-defined file path where the
   recording will be stored.

1. Perform implementation-defined steps to create an empty file at |path|. If
   they fail, return [=error=] with [=error code=] [=unknown error=].
```

sadym-chromium (approved):

index.bs
whimboo: Is that suggestion still wanted? Looks like it hasn't been addressed yet.

index.bs
whimboo: Not sure what this suggestion actually is. Do I miss something?

lutien: I've updated it all in one comment below, and Maksim already approved

index.bs
lutien: Nope, I resolved now all the unneeded branches

sadym-chromium: affirmative. @lutien addressed my concerns in another way, so the comment is obsolete and resolved.

index.bs
```diff
@@ -5511,6 +5502,23 @@ The [=remote end steps=] given |session| and |command parameters| are:
...
+   1. Let |destination folder| be |command parameters|["<code>destinationFolder</code>"].
+
+   1. Let |destination folder entry| be [=locate an entry=] given
+      [=file system locator/path=] |destination folder| and
+      [=file system locator/root=] corresponding to the root of the file system.
+
+   1. Let |path| be an implementation-defined file path for a new file in
+      |destination folder entry| where the recording will be stored.
+
+   1. If |destination folder entry| is null, return [=error=] with [=error code=]
```
whimboo, line 5513: Hm, the step before is already using the destination folder entry even through it could be `null`. We should reverse the order.

whimboo (changes requested):

index.bs
lutien: Updated

whimboo (approved):
Looks good. I assume you are going to add tests as well when implementing the field for Firefox.

